PR #1, opened 2018-12-30
Premiere checks just return true now. 

bfayers:
Most checks utilise "isRegistered()" method, that returns true. Some check the isPaid variable directly so I've set that to true also.

### DIFF
--- a/MediaBrowserTv/app/src/main/java/tv/emby/embyatv/TvApp.java
+++ b/MediaBrowserTv/app/src/main/java/tv/emby/embyatv/TvApp.java
@@ -84,7 +84,7 @@ public class TvApp extends Application implements ActivityCompat.OnRequestPermis
 
     private String lastDeletedItemId = "";
 
-    private boolean isPaid = false;
+    private boolean isPaid = true;
     private RegistrationInfo registrationInfo;
 
     private Calendar lastPlayback = Calendar.getInstance();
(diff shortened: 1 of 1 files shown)


MediaBrowserTv/app/src/main/java/tv/emby/embyatv/TvApp.java
```diff
@@ -417,18 +417,23 @@ public void setLastUserInteraction(long lastUserInteraction) {
     }
 
     public boolean checkPaidCache() {
+        /*
```
JustAMan: I think it's better to remove the old code (that is never going to be used) altogether, commented out code is hard to catch in diff.

MediaBrowserTv/app/src/main/java/tv/emby/embyatv/TvApp.java
```diff
@@ -417,18 +417,23 @@ public void setLastUserInteraction(long lastUserInteraction) {
...
+        return true;
     }
 
     public boolean isPaid() {
-        return isPaid;
+        return true;
     }
 
     public void setPaid(boolean isPaid) {
+        /*
```
JustAMan: Same

MediaBrowserTv/app/src/main/java/tv/emby/embyatv/TvApp.java
```diff
@@ -444,7 +449,10 @@ public boolean isValid() {
     }
 
     public boolean isRegistered() {
+        /*
```
JustAMan: Same

MediaBrowserTv/app/src/main/java/tv/emby/embyatv/TvApp.java
bfayers: Resolved via c3c7b95

MediaBrowserTv/app/src/main/java/tv/emby/embyatv/TvApp.java
bfayers: Resolved via c3c7b95

MediaBrowserTv/app/src/main/java/tv/emby/embyatv/TvApp.java
bfayers: Resolved via c3c7b95

PR #3, opened 2019-01-02
replace all tv.emby.atv with org.jellyfin.androidtv

bfayers:
I have built and tested the app still functions as it did before this change. (it does)


bfayers:
Replaced with #5

PR #16, opened 2019-01-11
Remove support for Connect

thornbill:
Removes support for Connect and related strings

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/Utils.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/Utils.java
@@ -1601,7 +1601,6 @@ public class Utils {
                 logger.Debug("No server available...");
                 Utils.showToast(activity, "No MB Servers available...");
                 break;
-            case ConnectSignIn:
             case ServerSignIn:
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/util/Utils.java
```diff
@@ -1635,6 +1628,7 @@ public void onResponse(ArrayList<ServerInfo> serverResponse) {
                     }
                 });
                 break;
```
JustAMan: How is it possible to get in this state if you're removing connect login?

thornbill: That state is set by the Java API client when no servers are available. That logic should probably be removed at some point.

https://github.com/jellyfin/jellyfin-apiclient-java/blob/master/android/src/main/java/mediabrowser/apiinteraction/connectionmanager/ConnectionManager.java#L133

thornbill:
I just tested automatic discovery and confirmed it still works as expected.

app/src/main/java/org/jellyfin/androidtv/util/Utils.java
JustAMan: Thanks!

PR #31, opened 2019-01-30
Minor build fix

grafixeyehero:
fix Failure [INSTALL_FAILED_NO_MATCHING_ABIS: Failed to extract native libraries, res=-113]

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -22,6 +22,14 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
     }
+	splits {
+        abi {
+            enable true
+            reset()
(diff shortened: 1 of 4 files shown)


app/build.gradle
```diff
@@ -21,6 +21,14 @@ android {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
+    }
+	splits {
```
dkanada: I would switch to spaces here, looks like a rogue tab jumped in.

dkanada: I would switch to spaces here, looks like a rogue tab jumped in.

grafixeyehero: this error happening on andorid emulator to install the apk

PR #37, opened 2019-02-08
Restore gradle wrapper

thornbill:
Restores the gradle wrapper for use in build scripts.


joshuaboniface:
Works perfectly and allows building with the facilities in #39~

PR #38, opened 2019-02-08
General Cleanup

thornbill:
* Removes unused resources
* Optimizes and removes unused imports


app/src/main/res/values-cs/strings.xml
```diff
@@ -1,45 +1,19 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools" tools:ignore="MissingTranslation">
     <string name="app_name">Jellyfin</string>
-    <string name="current_queue">Aktuální fronta</string>
```
dkanada: Is there a chance we are going to use any of these strings again?

thornbill: I don’t know. I prefer to not leave unused things around. It’s always available in git history if needed.

PR #39, opened 2019-02-09
Import Docker build infrastructure from jellyfin-android

joshuaboniface:
Imports the Docker-based APK build facilities from [jellyfin-android](https://github.com/jellyfin/jellyfin-android). Combined with the `gradlew` configuration from #37 generates APKs in a similar format and set of release types (production, unminified, and debug).


docker-build.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Handle the release type
+case ${RELEASE} in
+    'production')
+        RELEASE_SUFFIX=""
+        NODE_ENV="production"
+        RFLAG="assembleRelease"
+    ;;
+    'unminified')
+        RELEASE_SUFFIX="unminified_"
```
dkanada: We don't use node so this section won't do anything, need to find a new way to specify debug as well.

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# Docker run environment
+ENV SOURCE_DIR=/repo
+ENV ARTIFACT_DIR=/dist
+ENV ANDROID_DIR=/usr/lib/android-sdk
+
+# Prepare Debian build environment
+RUN echo "deb http://deb.debian.org/debian stretch-backports main" >> /etc/apt/sources.list.d/backports.list \
+ && apt-get update \
+ && apt-get install -y mmv wget unzip git \
+ && apt-get install -t stretch-backports -y npm android-sdk openjdk-8-jdk
```
thornbill: `npm` is not needed here.

docker-build.sh
joshuaboniface: The `assembleDebug` does that, or do you mean specifying unminified?

Dockerfile
joshuaboniface: Good catch, no sense wasting the space.

docker-build.sh
dkanada: specifying unminified

joshuaboniface: As discussed on Riot, might be OK to just not have this as this app seems to not have as many issues with minification. Worth research for the future though.

.gitignore
```diff
@@ -4,4 +4,4 @@ local.properties
 
 /.idea
 /.gradle
-build
+gradle
```
thornbill: I don’t think this should be added back. Merge conflict maybe?

joshuaboniface: Ah yes I mis-corrected that merge conflict, I'll remove this too.

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN echo "deb http://deb.debian.org/debian stretch-backports main" >> /etc/apt/sources.list.d/backports.list \
+ && apt-get update \
+ && apt-get install -y mmv wget unzip git \
+ && apt-get install -t stretch-backports -y android-sdk openjdk-8-jdk
+
+# Install Android repository tools
+RUN rm -rf ${ANDROID_DIR}/tools \
+ && wget http://dl-ssl.google.com/android/repository/tools_r25.2.3-linux.zip -O tools.zip \
+ && unzip tools.zip -d ${ANDROID_DIR}/ \
+ && rm -f tools.zip
```
dkanada: I still think we should be able to remove this somehow but that can be for another day.

joshuaboniface: As long as we need to use sdkmanager, we need it :(

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# Install SDK tools
+# `yes` is required to accept licenses:
+# https://stackoverflow.com/questions/38096225/automatically-accept-all-sdk-licences/42125740#42125740
+# There will be a bit of delay (approx 30 seconds)
```
LeoVerto: I know I'm repeating myself but this should be something along the lines of "up to a couple minutes".

joshuaboniface: Agreed, I updated the wording.

PR #44, opened 2019-02-13
Create single build flavor

thornbill:
Emby was maintaining a private fork of ExoMedia that used Amazon's fork of ExoPlayer (refs: https://github.com/brianwernick/ExoMedia/issues/161). This seems to be the only real reason for having an Amazon specific flavor of the build.

While it seems like it would be trivial to fork ExoMedia and use the Amazon fork of ExoPlayer, I would like to avoid this overhead unless we encounter playback issues on Fire TV devices.


cvium:
While I agree with the idea, I'm just not certain it's wise. Just skimming over their patches it looks like they did a bunch to make Optical Output and Dolby work eg. https://github.com/amzn/exoplayer-amazon-port/blob/09f605d6ab230202e3f49bba74511cd375e545c1/library/core/src/main/java/com/google/android/exoplayer2/audio/MediaCodecAudioRenderer.java#L434-L442 and https://developer.amazon.com/docs/fire-tv/dolby-integration-guidelines.html#exoplayer. Won't this be a problem for us as well?

thornbill:
I agree. It looks like we will need a fork of ExoMedia that includes the ExoPlayer patches.

PR #63, opened 2019-03-06
Refactor Utils class

thornbill:
Refactor Utils class into separate classes based on functionality to improve maintainability. Also changes some method naming to use Java conventions and makes some general improvements to the readability of functions.


thornbill:
I’ve been running a build with this included for several days and haven’t noticed any regressions.

PR #64, opened 2019-03-06
Fix linting errors

thornbill:
Fixes lint errors when building.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -34,10 +34,6 @@ android {
             versionName = version + "a"
         }
     }
-
-    lintOptions {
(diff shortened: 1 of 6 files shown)


dkanada:
Lint doesn't like the basic TextView and ImageView?

thornbill:
Not when you are using appcompat...
> In order to support features such as tinting, the appcompat library will automatically load special appcompat replacements for the builtin widgets. However, this does not work for your own custom views.
>
> Instead of extending the android.widget classes directly, you should instead extend one of the delegate classes in `android.support.v7.widget.AppCompat`.

PR #69, opened 2019-03-09
 Build improvements for wrapping infrastructure

joshuaboniface:
**Changes**
Makes some minor modifications to better support the wrapping `jellyfin-build` infrastructure. Specifically:
1. Put the output in `../bin` rather than `../`, which is cleaner, and easier to grab from.
1. Add a `build.yaml` configuration listing the available package types. `build` within the repo doesn't use this (yet).


joshuaboniface:
@thornbill Right now its just a placeholder and doesn't do anything, so we can settle on a version when its ready.

@dkanada It matches all the other build systems, avoiding polluting the repo with artifacts. I prefer that method.

Issue #34, opened 2019-02-06
Google Android TV using incorrect icon


dinki:
The icon being used for JF on Android TV appears to be a banner when it should be just the fin/triangle icon.  Here's what it looks like on my Xiaomi Mibox 3:

![image](https://user-images.githubusercontent.com/2521542/52314568-dd78b300-2978-11e9-8f8f-f520dde17418.png)

and this is how Emby's icon looks:
![image](https://user-images.githubusercontent.com/2521542/52314556-c4700200-2978-11e9-8585-86cb7e97e314.png)

anthonylavado:
@thornbill remember to just tag us over at UX if you need any assets :-)

grafixeyehero:
@thornbill i think this relating with a missing icon512

Issue #27, opened 2019-01-20
Amazon logo displaying square icon when it should be rectangular


dinki:
![gnnuipjtilqjbclheqkzfghm](https://user-images.githubusercontent.com/2521542/51443698-b81a5280-1cb2-11e9-84db-23bacbc8f338.png)

Amazon Fire TV Jellyfin app icon is a bit ugly in the rectangle box... Should extend the colors to fill the box..

PR #70, opened 2019-03-10
Fix incorrect icon usage

grafixeyehero:
fix Google Android TV using incorrect icon  #34


grafixeyehero:
@thornbill Amazon devices fetch icons and banners in a different way compared to the Android standard methods. i found this article https://forums.developer.amazon.com/questions/483/how-to-provide-the-firetv-and-firetv-stick-app-ban.html

thornbill:
I found another link in the [Amazon developer forums](https://forums.developer.amazon.com/answers/50800/view.html) that also indicates they use the icon provided when uploading the app for the home screen and not the banner image in the manifest. So this will reintroduce #27 but that seem inevitable until we get the app in Amazon's store.

PR #75, opened 2019-03-14
Update default Jellyfin theme

thornbill:
* Updates layout of the startup fragment / loading screen
* Updates colors to match Jellyfin's color scheme
* Changes direct references to resources and colors to using attribute values from theme. This will allow for dynamic theme switching in the future. :sparkles:
* Disables sideline/fastlane color preference (to be replaced by theme switching)
* Updates layouts to use start/end instead of left/right for improved RTL support
* Removes unused resources and `GridItemPresenter` classes


app/src/main/java/org/jellyfin/androidtv/browsing/StdGridFragment.java
```diff
@@ -702,11 +691,7 @@ public void run() {
             mHandler.post(new Runnable() {
                 @Override
                 public void run() {
-                    if (mBackgroundUrl != null) {
                         updateBackground(mBackgroundUrl);
```
dkanada: indentation is off

app/src/main/java/org/jellyfin/androidtv/details/ItemListActivity.java
```diff
@@ -759,14 +753,13 @@ private void stopRotate() {
 
     protected void updateBackground(String url) {
         if (url == null) {
-            BackgroundManager.getInstance(this).setDrawable(mDefaultBackground);
+            BackgroundManager.getInstance(this).setDrawable(null);
```
dkanada: seems like a lot of duplicate code but I wouldn't want to consolidate it either haha

app/src/main/res/layout/guide_filter_popup.xml
```diff
@@ -10,7 +9,7 @@
         android:layout_alignParentTop="true"
         android:layout_centerHorizontal="true"
         android:textSize="30sp"
-        android:textColor="@android:color/holo_blue_light"
+        android:textColor="?attr/defaultBrandColor"
```
dkanada: Maybe a different label to avoid the word brand?

app/src/main/res/xml/preferences.xml
```diff
@@ -42,7 +42,7 @@
             android:positiveButtonText="@null"
             app:numColumns="5"
             app:itemLayout="@layout/dash_grid_item_color"
-            />
+            /> -->
```
dkanada: I would just remove but it doesn't really matter.

app/src/main/res/layout/guide_filter_popup.xml
thornbill: I blame Google for this one. It comes from the base Android theme we're extending. :joy:

app/src/main/java/org/jellyfin/androidtv/details/ItemListActivity.java
thornbill: Yeah. I briefly looked at moving this to the parent activity classes, but they are "borrowed" from AOSP so a larger refactor should probably happen at some point to remove the Apache licensed code.

app/src/main/java/org/jellyfin/androidtv/browsing/StdGridFragment.java
thornbill: Nice catch. It's fixed now.

app/src/main/res/xml/preferences.xml
thornbill: I removed this and the related classes. Turns out `ColorPreference` was also "borrowed" from AOSP.

PR #76, opened 2019-03-19
Update apiclient

thornbill:
* Replaces apiclient jar with jitpack version
* Restores classes removed from apiclient in `compat` package and flags the classes as deprecated
* Removes conflicting and unused dependencies

Depends on https://github.com/jellyfin/jellyfin-apiclient-java/pull/8


thornbill:
I got the impression some of this functionality moved to new classes in the apiclient, but the git history was not helpful in determining what it was replaced with. I did restore a few classes that seemed like they would still be needed in the apiclient in https://github.com/jellyfin/jellyfin-apiclient-java/pull/8. :man_shrugging:

PR #77, opened 2019-03-22
Remove send log buttons

### LINKED ISSUE #45: Send Logs does nothing
The "Send Logs" option in the app does nothing since we removed the `LogReporter` in https://github.com/jellyfin/jellyfin-androidtv/pull/9. We should evaluate if there is a way to do this without depending on a centralized service. If not the UI for this should be removed.

BnMcG:
Resolves #45 

Remove "send log" buttons that no longer work. Tested on Fire TV Gen 2. Removes buttons from both the login and the home screens.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
@@ -137,13 +137,11 @@ public class HomeFragment extends StdBrowseFragment {
     public void onResume() {
         super.onResume();
 
-        addLogsButton();
         //make sure rows have had a chance to be created
         new Handler().postDelayed(new Runnable() {
             @Override
             public void run() {
                 addNowPlaying();
-                addLogsButton();
                 //check for resume row and add if not there
                 if (!hasResumeRow()) addContinueWatching();
             }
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
```diff
@@ -410,9 +397,6 @@ public void onItemClicked(Presenter.ViewHolder itemViewHolder, Object item,
                         Intent settings = new Intent(getActivity(), SettingsActivity.class);
                         getActivity().startActivity(settings);
                         break;
-                    case REPORT:
-                        Utils.reportError(getActivity(), "Send Log to Dev");
```
cvium: Should this function also be removed?

thornbill: I'm fine with it either way. That method just handles displaying a modal confirmation message, so we may reuse it.

app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
```diff
@@ -365,20 +363,9 @@ protected void addAdditionalRows(ArrayObjectAdapter rowAdapter) {
         toolsRow = new ArrayObjectAdapter(mGridPresenter);
         toolsRow.add(new GridButton(SETTINGS, mApplication.getString(R.string.lbl_app_settings), R.drawable.gears));
         toolsRow.add(new GridButton(LOGOUT, mApplication.getString(R.string.lbl_logout) + TvApp.getApplication().getCurrentUser().getName(), R.drawable.logout));
-
-        sendLogsButton = new GridButton(REPORT, mApplication.getString(R.string.lbl_send_logs), R.drawable.upload);
```
thornbill: Can we remove the `sendLogsButton` variable entirely?

app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
BnMcG: I left it in as it's referenced elsewhere - I'm happy to remove it though.

app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
BnMcG: Derp. I've now removed the instance variable for sendLogsButton, and also the "REPORT" constant. Android Studio also suggested converting toolsRow to a local variable, so I made that change, too.

PR #111, opened 2019-06-02
Small changes to the build

jbcorreia:
Changed build so that it works with x86 too - useful for testing with an Emulator
Allow unencrypted connections on latest Android API


jbcorreia:
It had to do with the latest gradle version, it's now fixed.

thornbill:
The build works for me now.

I think we need to ensure we can support x86 devices before we add it as a build target and start distributing apks for x86. My main concern is the build of libvlc currently in the repo only supports arm devices. We need to address the fact that we are including libvlc as a jar anyway. Hopefully we can add x86 support for libvlc as part of those changes.

thornbill:
These changes to enable X86 support should not be needed once #121 is merged.

I pulled out the change to enable cleartext traffic into #120 as it was needed when running an emulator with a newer API version.

PR #141, opened 2019-08-22
Layout changes to improve music libraries

dkanada:


dkanada:
I was going to change the icons here as well but it's getting a bit large. I'll probably only change the most common ones and maybe do a few more improvements to the layouts.

PR #168, opened 2019-10-16
Server sign in updates

thornbill:
* Updates location of `GsonJsonSerializer`
* Refactors server sign in code to remove duplication and consolidate code in `AuthenticationHelper`
* Adds minimum server version check since the apiclient no longer supports servers earlier than 10.3.0.

**Depends on: https://github.com/jellyfin/jellyfin-apiclient-java/pull/27**

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -114,6 +114,7 @@
     <string name="lbl_yes">Yes</string>
     <string name="msg_error_connecting_server">Error Connecting to Server</string>
     <string name="msg_error_server_unavailable">Server unavailable</string>
+    <string name="msg_error_server_version">Unsupported server version. Server version must be %1$s or later.</string>
     <string name="msg_error_signin">Error Signing In</string>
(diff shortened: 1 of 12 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -113,6 +113,9 @@
     <string name="msg_report_not_sent">Report NOT sent</string>
     <string name="lbl_yes">Yes</string>
     <string name="msg_error_connecting_server">Error Connecting to Server</string>
+    <string name="msg_error_server_unavailable">Server unavailable</string>
+    <string name="msg_error_server_version">Unsupported server version. Server version must be %1$s or later.</string>
+    <string name="msg_error_signin">Error Signing In</string>
```
dkanada: If this is a toast I would make it a full sentence.

thornbill: @dkanada any suggestions on what to make this? This error message is displayed as a toast notification if automatic login with saved credentials fails.

PR #170, opened 2019-10-17
Add Azure Pipelines config

thornbill:
* Adds an Azure Pipelines config for building the app and running unit tests

See: https://github.com/thornbill/jellyfin-androidtv/pull/1/checks

@jellyfin/core someone with access will need to set this up on the Azure side.

### DIFF
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -1,0 +1,15 @@
+trigger:
+- master
+
+pool:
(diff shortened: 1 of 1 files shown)


anthonylavado:
I believe at least @EraYaN will have access (I should probably get it too)

EraYaN:
@anthonylavado PM me your MS account email on matrix and I’ll add you to the admin team. I’ll add this one in a bit.

PR #174, opened 2019-10-28
Release new version

dkanada:


app/build.gradle
```diff
@@ -8,7 +8,7 @@ android {
     defaultConfig {
         minSdkVersion 21
         targetSdkVersion 28
-        versionCode 901
+        versionCode 902
```
thornbill: This should be `1000` I think?

Issue #177, opened 2019-10-31
Crash when adding server


moreje:
I installed the 0.10 apk for Android TV on my Philips Android TV.
The app crash and exit when adding a server...

tomkriz:
Same here. Recently got the newest Nvidia Shield Pro, Android SHIELD TV 9.1.0 which should be latest updates.

JellyFin 0.9.1 works. Since this was a new install, I thought maybe installing v0.10 after configuring on 0.9.1 would work, but it didn't. I wasn't even able to add a server, it just crashed on open after trying that.

anthonylavado:
Just to confirm, what version of Jellyfin server are you running? Is it the most recent, 10.4.0?

rdissertori:
same here using docker
```
Version: 10.4.0
Operating System: Linux
Architecture: X64
```

moreje:
same here: 10.4.0 inside a container

rdissertori:
just reporting back: that hotfix did it for me

anthonylavado:
@cludch Following up here from https://github.com/jellyfin/jellyfin-androidtv/issues/186#issuecomment-549970414.

You mentioned you're using a reverse proxy. Are you on version 10.4.1? There's a note about the baseurl path having to be changed: https://github.com/jellyfin/jellyfin/releases/tag/v10.4.1

bwaddington:
Version: 10.4.1
Operating System: Linux
Architecture: X64

jellyfin-androidtv:  v0.10.1

i had a problem adding the server keep cashing out with address 
****.duckdns.org    
but when i put the port on it connected

****.duckdns.org:8096

spoolr:
> i had a problem adding the server keep cashing out with address
> ****.duckdns.org
> but when i put the port on it connected
> 
> ****.duckdns.org:8096

I don't know why, but they removed the default port with "Merge pull request #172 from thornbill/default-port"

thornbill:
Well it was poorly implemented and we are going the route of “what you enter in a client should match what you enter in a browser” to make things more consistent. It definitely should not crash when it tries to connect using an invalid port though.

Cludch:
@anthonylavado 
Yes I am on 10.4.1 on the server-side. I use a subdomain for jellyfin, so I simply enter `https://jellyfin.domain.com` within the client. I also tried leaving the `https://` part out. Still crashing.
I could try to use the internal hostname and port `8096` to see if it's an issue with the reverse proxy.

[...]

PR #182, opened 2019-11-03
Disable DelayedMessage to prevent crash on login

thornbill:
Temporary workaround for #177


thornbill:
Just to provide some additional context, this seems to be failing on server discovery because that does not happen on the main ui thread.

thornbill:
I was thinking we could comment it out now and then fix it properly when time allows.

garikaib:
Having the same issue, is there an APK build with this fix applied?

thornbill:
0.10.1 includes this removal. 0.10.2 includes the proper fix from https://github.com/jellyfin/jellyfin-androidtv/pull/189.

garikaib:
Just tested that and for some reason, I cannot add the server. I keep getting a server unavailable error. Looking at the server logs it seems the client is not even attempting to connect.

PR #188, opened 2019-11-07
Fix version check when no servers in response

thornbill:
Fixes a crash when the response does not include a server for some reason (invalid url, server offline, etc.).

Refs: #177

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/startup/StartupActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/startup/StartupActivity.java
@@ -181,7 +181,8 @@ public class StartupActivity extends Activity {
                     }
 
                     // Check the server version
-                    if (!AuthenticationHelper.isSupportedServerVersion(response.getServers().get(0))) {
+                    if (!response.getServers().isEmpty() &&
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/startup/StartupActivity.java
```diff
@@ -181,7 +181,8 @@ public void onResponse(ConnectionResult response) {
                     }
 
                     // Check the server version
-                    if (!AuthenticationHelper.isSupportedServerVersion(response.getServers().get(0))) {
+                    if (!response.getServers().isEmpty() &&
+                            !AuthenticationHelper.isSupportedServerVersion(response.getServers().get(0))) {
```
dkanada: Wouldn't this work better in the section above so users get an error message?

thornbill: The `Unavailable` check above should be triggered before it reaches here. The other case was the one causing the `Exception` because that check was happening after the version check. It's still a good idea to make sure the list isn't empty before trying to grab the first value imo.

PR #196, opened 2019-11-13
Fix search with denied permissions

thornbill:
Fixes an issue where the `onRequestPermissionsResult` callback is never called and the search activity was never displayed if the `RECORD_AUDIO` permission was denied.

cc @joshuaboniface

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/base/BaseActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/base/BaseActivity.java
@@ -1,6 +1,7 @@
 package org.jellyfin.androidtv.base;
 
 import android.app.Activity;
+import android.content.pm.PackageManager;
 import android.graphics.Rect;
 import android.os.Bundle;
 import android.os.Handler;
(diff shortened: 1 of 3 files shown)


thornbill:
The build failure is due to #195.

app/src/main/java/org/jellyfin/androidtv/base/BaseActivity.java
```diff
@@ -178,4 +191,16 @@ public boolean onSearchRequested() {
         TvApp.getApplication().showSearch(this, false);
         return true;
     }
+
+    @Override
+    public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
+        // Search functionality requests access to RECORD_AUDIO
+        if (requestCode == TvApp.SEARCH_PERMISSION) {
```
dkanada: Why move this method to the base class? It should work, just curious since it might also work in the main class without any public method calls.

thornbill: In order for this method to be called, the class must be an instance of `Activity` which `TvApp` is not. :smile:

Issue #197, opened 2019-11-14
Settings visibility and aspect ratio for Live TV tiles


Nickbert7:
The settings menu within the libraries and the guide are not well readable:
![image](https://user-images.githubusercontent.com/14938497/68860966-de4e2a80-06ea-11ea-8281-b2cff4b992a7.png)
![image](https://user-images.githubusercontent.com/14938497/68860973-e1e1b180-06ea-11ea-8935-35a8e3c62a13.png)

Also the aspect ratio for the Live TV and Recordings tiles are incorrect
![image](https://user-images.githubusercontent.com/14938497/68861213-616f8080-06eb-11ea-87d3-3ff7d9151353.png)

Nickbert7:
The menue is fixed, but the tiles are still not correct.
Should I create a separate issue?

nielsvanvelzen:
A new issue for the tile issues is preferred (if there's none yet), thanks.

PR #198, opened 2019-11-14
Fix style issues and untranslatable text

thornbill:
* Fixes transparent popup backgrounds and grey text color as reported in #197
* Adds support for translating additional strings

### DIFF
--- a/app/src/main/res/layout/live_tv_guide.xml
+++ b/app/src/main/res/layout/live_tv_guide.xml
@@ -21,7 +21,7 @@
             android:layout_height="wrap_content"
             android:focusable="false"
             android:focusableInTouchMode="false"
-            android:id="@+id/channels"/>
+            android:id="@+id/channels" />
     </ScrollView>
 
     <org.jellyfin.androidtv.ui.ObservableScrollView
@@ -76,7 +76,7 @@
         android:layout_alignParentStart="true"
         android:background="@color/timeline_bg"
         android:gravity="center"
-        android:textSize="14sp"/>
+        android:textSize="14sp" />
 
     <ImageView
         android:layout_width="match_parent"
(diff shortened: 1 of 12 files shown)


app/src/main/res/layout/live_tv_guide.xml
```diff
@@ -67,35 +69,33 @@
...
-
-    <org.jellyfin.androidtv.ui.ClockUserView
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:id="@+id/clock"
-        android:layout_alignTop="@+id/title"
-        android:layout_marginRight="40sp"
-        android:layout_alignParentEnd="true"
-        android:alpha=".9" />
+        android:textSize="14sp"/>
```
dkanada:
```suggestion
        android:textSize="14sp" />
```

app/src/main/res/values/strings.xml
```diff
@@ -330,4 +330,17 @@
...
+    <string name="lbl_news">News</string>
+    <string name="lbl_new_only">New Only</string>
+    <string name="lbl_audio_delay">Audio Delay</string>
+    <string name="lbl_new_episodes">New Episodes</string>
+    <string name="lbl_premieres">Premieres</string>
+    <string name="lbl_repeat_episodes">Repeat Episodes</string>
+    <string name="lbl_live_broadcasts">Live Broadcasts</string>
+    <string name="lbl_hd_programs">HD Programs</string>
+    <string name="lbl_colored_backgrounds">Enable Color Coded Backgrounds</string>
+    <string name="lbl_start_favorites">Place favorites at beginning</string>
```
dkanada: I would use title casing for the favorite order string if it's a preference. Also, I would change the colored backgrounds preference to something like `Generate Background Color from Images` because the current string is pretty vague.

app/src/main/res/layout/live_tv_guide.xml
thornbill: Fixed.

app/src/main/res/values/strings.xml
thornbill: I updated the casing, but the "color coded backgrounds" refers to the live tv guide feature where the background of the program differs depending on the genre or category. A very similar wording is used in web for this feature.

dkanada: Ah okay, I misunderstood the meaning. It still isn't very clear but I guess it's hard to explain what this option does. I have been removing the `Enable` prefix in front of boolean settings though since the checkbox already indicates the behavior.

Issue #81, opened 2019-04-02
Latest TV Shows doesn't show on home screen, but Latest Movies does


ragoneone:
Latest TV Shows doesn't show on home screen, but Latest Movies does.  This is only on the Android TV app.  The standard Android app shows Latest TV on my phone and the wen interface shows it too.  When I use the Android TV app on my Nvidia Shield, it doesn't appear.

choakem:
This is probably the biggest issue I have right now with androidtv. Not bad considering how hound the apk is...

thornbill:
Something similar to this method for getting the latest movies will need to be added.

See https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java#L234-L242

thornbill:
Fixed by #201

PR #201, opened 2019-11-15
Added Latest TV Shows to Home Fragment

AndreasGB:
Hi there,

I took a quick stab at #81. I was a bit unsure at which position to insert the row in the different modes, but hope this is fine.

Cheers

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
@@ -179,8 +179,8 @@ public class HomeFragment extends StdBrowseFragment {
                         case TvShows:
                             addNextUp();
                             addPremieres();
-                            addLatestTVShows();
                             addLatestMovies();
+                            addLatestTVShows();
                             addOnNow();
                             break;
 
@@ -216,7 +216,7 @@ public class HomeFragment extends StdBrowseFragment {
 
     }
 
-    protected StdItemQuery getResumeQuery() {
+    private StdItemQuery getResumeQuery() {
         StdItemQuery resumeItems = new StdItemQuery();
         resumeItems.setMediaTypes(new String[] {"Video"});
         resumeItems.setRecursive(true);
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
```diff
@@ -179,6 +179,7 @@ public void onResponse(ItemsResult response) {
                         case TvShows:
                             addNextUp();
                             addPremieres();
+                            addLatestTVShows();
```
thornbill: I would move this after Latest Movies to be consistent with the other cases.

app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
```diff
@@ -238,6 +241,15 @@ protected void addLatestMovies() {
 
     }
 
+    protected void addLatestTVShows() {
```
thornbill: This method can be private. (Not sure why the others aren't, but no need to continue that here :smile:)

app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
```diff
@@ -238,6 +241,15 @@ protected void addLatestMovies() {
...
+        LatestItemsQuery latestTVShows = new LatestItemsQuery();
+        latestTVShows.setFields(new ItemFields[] {ItemFields.PrimaryImageAspectRatio, ItemFields.Overview});
+        latestTVShows.setIncludeItemTypes(new String[]{"Series"});
```
thornbill: To get results that are more consistent with the web interface, it looks like we should include `Episode` instead of `Series` and set `latestTVShows.setGroupItems(true);`.

AndreasGB: Is there a way to force widescreen tiles? 

Doing it this way leads to a bug where if there is only one episode left in a season, the widescreen card is shown, for the grouped ones the portrait.

app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
AndreasGB: Okay, will make the others private as well then.

app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
AndreasGB: Okay, will do.

PR #207, opened 2019-11-20
Upgrading Gradle Version to 5.4.1

AndreasGB:
Also adding myself to CONTRIBUTORS.md now, as this ate way too much time :D

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -42,7 +42,14 @@ dependencies {
     amazonImplementation 'com.github.amzn:exoplayer-amazon-port:amazon-r1.5.6-SNAPSHOT'
     googleImplementation 'com.devbrackets.android:exomedia:2.5.6'
     googleImplementation 'com.google.android.exoplayer:exoplayer:r1.5.6'
-    implementation 'com.github.jellyfin:jellyfin-apiclient-java:master-SNAPSHOT'
+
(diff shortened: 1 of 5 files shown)


app/build.gradle
```diff
@@ -39,7 +39,9 @@ android {
 dependencies {
     implementation fileTree(include: ['*.jar'], dir: 'libs')
     amazonImplementation 'com.github.jellyfin:exomedia:amazon~2.5.6-SNAPSHOT'
+    amazonImplementation 'com.github.amzn:exoplayer-amazon-port:amazon-r1.5.6-SNAPSHOT'
     googleImplementation 'com.devbrackets.android:exomedia:2.5.6'
+    googleImplementation 'com.google.android.exoplayer:exoplayer:r1.5.6'
```
thornbill: Maybe we should add a comment to note why these are needed.

AndreasGB: Otoh, if anyone removes them, they will find out why pretty quickly :D

AndreasGB:
So, we now have the ability to modify the :library module of the apiclient in the same project with this new change by making a composite build. Its PR needs to get merged before this one though as they both need to be on the same gradle and plugin versions.

AndreasGB:
@anthonylavado The blocking PR has merged, can you please remove the label? Thanks :)

PR #210, opened 2019-11-25
Add home sections

nielsvanvelzen:
I'm currently working on moving my personal library from Plex to Jellyfin and one of the things that annoyed me is that the sections I made in the web version weren't available in the TV app. So I implemented it!

I tried to make it look exactly the same as the web version but some things are still different:
 - The library buttons are not implemented as I tried but couldn't find a nice way to implement it
 - The "Latest Additions" section is split in the web version with one row for each media type, my implementation shows everything in a single row
 - Editing the sections is not supported (I might add it another time in a separate pull request)

It took quite some time to get familiar with this part of the code because I haven't used Java in a few years (I use Kotlin primarily) and the structure of the app doesn't make a lot of sense to me because there is strong cohesion between most classes.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
@@ -5,79 +5,100 @@ import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Bundle;
 import android.os.Handler;
-import androidx.leanback.widget.ArrayObjectAdapter;
-import androidx.leanback.widget.HeaderItem;
-import androidx.leanback.widget.ListRow;
-import androidx.leanback.widget.OnItemViewClickedListener;
-import androidx.leanback.widget.Presenter;
-import androidx.leanback.widget.Row;
-import androidx.leanback.widget.RowPresenter;
-import android.widget.Toast;
 
 import org.jellyfin.androidtv.R;
 import org.jellyfin.androidtv.TvApp;
 import org.jellyfin.androidtv.base.CustomMessage;
 import org.jellyfin.androidtv.base.IMessageListener;
+import org.jellyfin.androidtv.constants.HomeSectionType;
 import org.jellyfin.androidtv.integration.RecommendationManager;
 import org.jellyfin.androidtv.itemhandling.ItemRowAdapter;
 import org.jellyfin.androidtv.livetv.LiveTvGuideActivity;
 import org.jellyfin.androidtv.model.ChangeTriggerType;
-import org.jellyfin.androidtv.model.DisplayPriorityType;
+import org.jellyfin.androidtv.model.LogonCredentials;
 import org.jellyfin.androidtv.playback.AudioEventListener;
 import org.jellyfin.androidtv.playback.MediaManager;
-import org.jellyfin.androidtv.presentation.GridButtonPresenter;
+import org.jellyfin.androidtv.presentation.CardPresenter;
+import org.jellyfin.androidtv.presentation.PositionableListRowPresenter;
 import org.jellyfin.androidtv.presentation.ThemeManager;
 import org.jellyfin.androidtv.querying.QueryType;
 import org.jellyfin.androidtv.querying.StdItemQuery;
 import org.jellyfin.androidtv.querying.ViewQuery;
-import org.jellyfin.androidtv.settings.SettingsActivity;
-import org.jellyfin.androidtv.model.LogonCredentials;
-import org.jellyfin.androidtv.startup.SelectUserActivity;
-import org.jellyfin.androidtv.ui.GridButton;
 import org.jellyfin.androidtv.util.apiclient.AuthenticationHelper;
-
-import java.io.IOException;
-
-import org.jellyfin.apiclient.interaction.EmptyResponse;
 import org.jellyfin.apiclient.interaction.Response;
+import org.jellyfin.apiclient.model.entities.DisplayPreferences;
 import org.jellyfin.apiclient.model.entities.LocationType;
+import org.jellyfin.apiclient.model.entities.MediaType;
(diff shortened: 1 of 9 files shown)


thornbill:
Thanks for the PR! The code is certainly a bit of a mess structurally. We’ve been trying to clean it up some as we go, but it’s a slow process. Just a heads up it may be a few days before this gets reviewed.

nielsvanvelzen:
I rebased the branch as there were some conflicts in strings.xml.

nielsvanvelzen:
I realized that the "latest additions" row is really annoying when all media types are showing. I also noticed that the feature of the web-home screen where you can select which (video-) libraries show in a single row doesn't work in my version.

This change should probably not be merged unless some more features are added, I'm not sure if I have time in the next few weeks to implement this as I expect it requires quite a lot of changes in the code to add this without using hacks to get it working.

thornbill:
@nielsvanvelzen I noticed a few minor issues in my testing also.

* User settings for sections need merged with the defaults (the preference only seems to contain a value if it does not match the default)
* Card heights are inconsistent in some rows
* Selecting logout does not seem to work

If you don't mind, I may continue your work here as this is a really nice feature! I should have some availability this week to work on it.

nielsvanvelzen:
I don't mind! I assume you have more knowledge about the code base than me, I'm curious how you would add the missing features.

nielsvanvelzen:
I took a look at where the data to separate the "latest media" section comes from by looking trough network requests in the web client:

1. Get /Users/[id]/Views (now: views) endpoint
2. Get /Users/[id] (now: configuration) endpoint
3. Remove views inside the LatestItemsExcludes property of the configuration
4. For all views create a latestitemsquery query with parent id set to view id
5. Create rows for all queries

I don't know quite yet how to implement this. As requesting the needed data makes it asynchronous and this would cause the row ordering to get inconsistent.

thornbill:
imho even if we can't apply the more complex rules around the latest items, if we can just include Latest Movies and Latest TV Shows as separate rows in the correct position, this would still be a major improvement.

nielsvanvelzen:
I don't think that would really work, I currently have 2 separate tv show libraries, a movie library and a music library. Everyone has different types of libraries.
We can't show a row with latest music when you have no music library, if we did it would be empty. To prevent that from happening we still need to check first by requesting the views.

nielsvanvelzen:
Should be ready for review now!

app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
```diff
@@ -132,283 +150,223 @@ public void onResume() {
...
-    private void addLatestMovies() {
-        LatestItemsQuery latestMovies = new LatestItemsQuery();
-        latestMovies.setFields(new ItemFields[] {ItemFields.PrimaryImageAspectRatio, ItemFields.Overview});
-        latestMovies.setIncludeItemTypes(new String[]{"Movie"});
-        latestMovies.setImageTypeLimit(1);
-        latestMovies.setLimit(50);
-        mRows.add(new BrowseRowDef(mApplication.getString(R.string.lbl_latest_movies), latestMovies, new ChangeTriggerType[]{ChangeTriggerType.LibraryUpdated, ChangeTriggerType.MoviePlayback}));
+                    @Override
+                    public void onError(Exception exception) {
+                        exception.printStackTrace();
```
thornbill: The exception here should be logged using the application logger. `mApplication.getLogger()`

app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragmentFooterRow.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public void onItemClicked(Presenter.ViewHolder itemViewHolder, Object item, RowPresenter.ViewHolder rowViewHolder, Row row) {
+        if (!(item instanceof GridButton)) return;
+
+        switch (((GridButton) item).getId()) {
+            case LOGOUT:
+                TvApp app = TvApp.getApplication();
+
+                if (app.getIsAutoLoginConfigured()) {
```
thornbill: This may be beyond the scope of this PR, but I have been trying to understand why this behaves differently when autologin is configured. Why would we not just always show the user selection screen when logout is selected? Any ideas?

nielsvanvelzen: I think this is because auto-login skips the user screen so the only way to change users is using the logout button. I'd rather have the functionality work the same no matter what your settings are. So I'll change it to always open the user screen.

In the future we might rename it to "Change User" too because it doesn't actually clear credentials.

app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
nielsvanvelzen: Fixed

nielsvanvelzen:
I just noticed that we should probably remove the ThemeManager as the "recommendations" are removed from the HomeFragment now. I made a separate issue for that: #235

thornbill:
This is looking great! I did notice 2 minor issues in testing.

1. It seems like web always excludes "Live TV" from the latest media section and it is being included here.
2. It seems that if there is a mixture of "primary" and "banner" images in a latest category, then the cards are different heights.

The second issue could have been preexisting as the logic around the card image size is a bit of a mess, so I don't consider that a blocker.

nielsvanvelzen:
I don't have live tv set-up so it's hard for me to do anything about the first issue as I can't test it. Could you look into that?

The second issue is quite hard to workaround as the way the presenters & adapters are programmed are very hard to modify without breaking anything. I'd say we should change that in a separate issue. I'd like to only display the primary image and never the thumb (or backdrop) but I couldn't find a good way to change it for the home sections only.

thornbill:
Sure I will take a look at removing Live TV.

PR #213, opened 2019-11-27
Fix switching to encoded subtitles causing the video player to hang.

nielsvanvelzen:
The .switchSubtitleStream() function calls the stop & play functions which will then call the switch function and this ends up looping.

Adding an additional check if the stream we want to switch to isn't the currently used stream prevents this.


nielsvanvelzen:
I tried that at first but I still got some issues so I used the current solution. I'll have another look at it later this week to see if I can fix it outside of the `switchSubtitleStream` method.

PR #216, opened 2019-11-29
Change source language from en to en-us

theg00s3:
This *should* work


dkanada:
This should probably just be set in Weblate rather than changed in the source.

theg00s3:
good idea
________________________________

dkanada:
@anthonylavado could you make the change?

anthonylavado:
@dkanada @theg00s3 I actually don't think I can change the name in Weblate. Because no language code is picked up, it is just assuming "English". I could change the definition for that to be English US, but I think it might not break other things.

dkanada:
@anthonylavado so this pull request should get merged before you can set the default in Weblate then?

thornbill:
Based on the Android documentation (https://developer.android.com/guide/topics/resources/localization) we must have the default string values in the default (locale agnostic) location. This is reinforced by the CI build being broken by this change.

PR #217, opened 2019-11-29
Replace string usages for BaseItemDto.Type with the EBaseItemType

AndreasGB:
Also fixing dependency substitution now that we use explicit versions

This depends on the related PR jellyfin/jellyfin-apiclient-java#30 , which should be made available as release v0.5.0 before this is merged.

I'm sorry for the person that reviews this and makes sure that all conditional replacements are correct, it was a pain for me to write and review before sending the PR as well :(

Luckily this has to be done only once :)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/itemhandling/ItemRowAdapter.java
+++ b/app/src/main/java/org/jellyfin/androidtv/itemhandling/ItemRowAdapter.java
@@ -38,6 +38,7 @@ import org.jellyfin.apiclient.interaction.Response;
 import org.jellyfin.apiclient.model.apiclient.ServerInfo;
 import org.jellyfin.apiclient.model.dto.BaseItemDto;
 import org.jellyfin.apiclient.model.dto.BaseItemPerson;
+import org.jellyfin.apiclient.model.dto.EBaseItemType;
 import org.jellyfin.apiclient.model.dto.UserDto;
 import org.jellyfin.apiclient.model.livetv.ChannelInfoDto;
 import org.jellyfin.apiclient.model.livetv.LiveTvChannelQuery;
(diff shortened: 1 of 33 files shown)


thornbill:
I'm going to rebase this on master and force push to resolve the conflicts.

app/src/main/java/org/jellyfin/androidtv/itemhandling/ItemRowAdapter.java
```diff
@@ -812,7 +813,7 @@ public void onResponse(ItemsResult response) {
                     for (BaseItemDto item : response.getItems()) {
                         //re-map the display prefs id to our actual id
                         item.setDisplayPreferencesId(item.getId());
-                        if (!ignoreTypeList.contains(item.getCollectionType()) && !ignoreTypeList.contains(item.getType())) {
```
thornbill: Why was the second half of this check removed?

AndreasGB: IIRC it could never be true, so I removed it. I'll check again.

AndreasGB: So, `ignoreTypeList` is initialized as following:

```
    private static String[] ignoreTypes = new String[]{"books", "games"};
    private static List<String> ignoreTypeList = Arrays.asList(ignoreTypes);
```

and I think those are collection types (which I should probably turn into enums as well in a future PR), and doesn't contain any item types. Another case where the usage of strings as types confuses things.

thornbill: Good catch! Support for games has also been removed from the server so we should be able to remove it at some point also.

PR #218, opened 2019-11-29
Fixed playlist query

joern-h:
Issue: 
Playlists in music library are missing, only the playlist with the favorites was present.

Before

Fix: 
The playlist query with MediaTypes=Audio dosent return any elements. Removing this argument solves it. This argument also does not exist in the jellyfin-web playlist query.

After


joern-h:
This commit fixes the Shuffle-All button on playlists.

PR #228, opened 2019-12-14
Rename "build" file to "build.sh"

nielsvanvelzen:
Currently this isn't causing any issues but in the future when we might add additional plugins to the build.gradle file the "build" file might cause issues because Gradle needs a build directory in the project root.

Came across this issue when I tried enabling Kotlin support (for testing purposes) and got some weird IO errors.


thornbill:
Just for reference here is the relevant documentation for Gradle. https://docs.gradle.org/current/userguide/directory_layout.html#dir:project_root

joshuaboniface:
This is fine, just a small adjustment to the build infra to support it.

PR #229, opened 2019-12-14
Delete empty strings introduced by weblate from translations

### LINKED ISSUE #227: Texts are not showing when using Japanese system language.
I installed jellyfin on my fireTV stick.
and there was no text on the app.
I changed the device's system language to English. all the texts are shown properly.

After I set server login information, I set the system language to  Japanese again.
The only text I can see was "設定" which means settings.

I guess this is caused by missing text data. So, It would be better to show English text when there is no local language data.

Thank you.

AndreasGB:
This is supposed to fix #227 by deleting the empty strings that weblate had introduced.

I'm unsure wheter weblate would re-introduce these empty strings once a new one is translated, but I guess we can wait and see.


AndreasGB:
Fixed the remaining broken languages

AndreasGB:
That should have been all of them

PR #200, opened 2019-11-15
Add ACRA for crash reporting

thornbill:
Closes #193

PR #231, opened 2019-12-15
Enable Kotlin support

nielsvanvelzen:
As there was no objection to adding new functionality to the app in Kotlin I hereby introduce kotlin support for Jellyfin Android TV!

Changes:
- Adds the Kotlin extensions to the gradle build file
- Adds kotlin stdlib (for java 7) to the gradle dependencies
- Adds android ktx core, this adds a few additional functions to various parts of android for kotlin (https://developer.android.com/kotlin/ktx)
- Converted MainActivity to kotlin to validate everything is working
- Added .editorconfig file with code style for kotlin files only (generated by android studio from my own presets, currently set to tabs whereas the java code base currently uses spaces)

Notes:
- When #200 is merged we should use the stdlib for java 8
- As this branch started out for my own testing purposes I added the .editorconfig file but I can remove or change it if unwanted
- If it is decided to keep the editorconfig we should expand it for java, xml and groovy (gradle files)
- Depends on #228


thornbill:
Tagging as blocked since it is dependent on #228

PR #232, opened 2019-12-18
Add new and unfinished episodes to "watch next" section on Android TV

### LINKED ISSUE #74: Play Next and Channel section
At the moment the Android TV app doesn't show up in the Play Next section or available in the list of apps to include in the Channel section. This would be nice to have in future versions.

nielsvanvelzen:
This PR will add functionality to Jellyfin that will update the "watch next" section on leanback supported devices (Android TV).

Partially fixes #74 (watch next part), depends on #231 (kotlin support)

**What is does**
- Adds episodes to the "watch next" section when visiting the home screen

**What it doesn't**
- Does not add movies or other media
- Does not add custom channels (but could be expanded to support it)
- Does not update in the background (we need to refactor authentication code for this to work, as the sign in happens in the StartupActivity instead of the TvApp class)
- Does not update automatically when a user changes media (like marking something as played etc.) as it was not easy to hook into these events

I think this is a good MVP for the feature that could be expanded upon in the future


nielsvanvelzen:
Screenshot of new/unwatched episode:
image

Preview unfinished episode:
image

nielsvanvelzen:
CI failed due too out of memory, are you able to run the checks again? @thornbill

thornbill:
@nielsvanvelzen I had to add a jvm option to increase the max heap size. Without it the build was failing on CI and my local machine. 768m seems to work, 512m was still failing. I'm really not sure how much ram is available to use in the Azure Pipelines containers.

nielsvanvelzen:
I'm glad it is working now!

thornbill:
Did you happen to see the [RecommendationManager](https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/integration/RecommendationManager.java) when you were working on this? I don’t think it works on Android TV. It’s likely using a method that’s no longer supported by the platform. I think it should be removed in favor of your solution here.

nielsvanvelzen:
I saw the RecommendationManager after I did most of the work on this PR but it seems like it doesn't do anything? It creates notifications but as far as I know there is no support for notifications in Android TV, could be a Fire TV thing maybe?

I can remove it but I'd like to have a better idea of what it is supposed to do first.

nielsvanvelzen:
So looking at the code it might have been used to provide jellyfin content to other apps because there was a content provider registered. But if that were the case I think it's broken because I can't get that to work.

It also created notifications (although as I said, android tv doesn't support notification afaik) but those never showed up anywhere.

So I removed it, if someone complains later that suddenly a feature with notifications/content providing is missing we can always get it back from git

nielsvanvelzen:
I suspect [this](https://github.com/jellyfin/jellyfin-androidtv/pull/232/files#diff-8964488871457516da2cceb030dcfd28R38) line will make sure FireTV devices will be ignored. As I'm quite sure FireTV is a fork of Android, not Android TV.
I'll fix the when I get home in about 8 hours, seems like it's just merging the dependencies.

PR #233, opened 2019-12-20
Deprecation fixes

nielsvanvelzen:
Fixes most deprecations warnings from android/androidx. Each change was tested and the behavior is still the same in my atv emulator.


nielsvanvelzen:
Good catch! As I have the auto sign-in option turned on this activity never showed up for me. When using androidx fragment you have to use the `FragmentActivity` class instead of `Activity`. I somehow missed some classes.

I converted the remaining activities.

PR #234, opened 2019-12-21
Remove unused resources

nielsvanvelzen:
This PR removes:
 - Unused drawables
 - Unused layout files
 - Unused strings (in all languages)
 - Unused array definitions
 - Unused ids (never referenced)
 - Unused legacy theme

This will clean up the app a tiny bit. Most work is done using the "Remove Unused Resources" refactor function in Android Studio with additional removal and changes done manually.


thornbill:
It looks like a translation for one of the removed strings has been added to master which is causing the CI build failure. This will need rebased on the latest master and that string removed.

```
  /home/vsts/work/1/s/app/src/main/res/values-da/strings.xml:156: Error: "lbl_select_color" is translated here but not found in default locale [ExtraTranslation]
      <string name="lbl_select_color">Vælg Farve</string>
```

PR #239, opened 2019-12-29
Refactor search

nielsvanvelzen:
I noticed my search activity did not look like the leanback search interface and after digging into the code I found out this was caused by a lack of recording permission (I denied it myself).

There were 2 search interfaces: one if speech was allowed and one if it was not. This is dumb because they both supported manual input.

Changes in this PR:
- Remove the no-speech search
- Always use the leanback search
- Let the leanback search handle permissions (tested; that works)
- Rename the fragment from `MySearchFragment` to just `SearchFragment`
- Removed some additional checks and unnecessary code in TvApp related to search
- Refactor code for search activity & fragment
- When search query is submitted (enter key pressed) search instantly instead of delayed


nielsvanvelzen:
Tested this in an emulator with Android 5.1 too as there was a check for android version that I removed and it works perfectly fine

nielsvanvelzen:
I was not aware that special changes need to be made to support Fire TV. I have no such devices and it seems like there is no emulator for Fire TV so the changed I made have not been tested.

I re-added a check to verify speech recognition support assuming Amazon returns false. This will disable voice and even hide the icon in the UI.

nielsvanvelzen:
Unfortunately disabling speech is not that easy. I tried hiding the icon but that broke the recyclerview causing no results to appear. Setting the alpha to 0 worked but it wouldn't focus the textbox by default. It was also possible to still focus on the "invisible" icon.

The simplest way to disable speech is by disabling the speech recognizer but that causes an icon that doesn't work to still show, I don't think that is a good user experience. For now I've made an [issue](https://issuetracker.google.com/issues/146976558) in the google issue tracker.

So it might be better to still have a separate fragment for Fire TV users that is custom made for text-only search. I can add the old fragment back with some refactoring but I'd like to hear your opinion first @thornbill.

thornbill:
I would agree that doesn’t sound like the best experience. Perhaps it is best to keep a dedicated text search fragment for Amazon for now.

app/src/main/java/org/jellyfin/androidtv/TvApp.java
```diff
@@ -237,27 +227,7 @@ public void setPlaybackController(PlaybackController playbackController) {
...
-                        .setPositiveButton("Ok", new DialogInterface.OnClickListener() {
-                            public void onClick(DialogInterface dialog, int whichButton) {
-                                ActivityCompat.requestPermissions(activity, new String[] {Manifest.permission.RECORD_AUDIO}, SEARCH_PERMISSION);
-                            }
-                        }).show();
-            } else {
-                ActivityCompat.requestPermissions(activity, new String[] {Manifest.permission.RECORD_AUDIO}, SEARCH_PERMISSION);
-            }
-        } else {
             showSearchWithPermission(activity, musicOnly);
```
thornbill: Perhaps we should remove the `showSearchWithPermission` method and just move the implementation here. It no longer seems necessary since the permissions aren’t handled here now.

nielsvanvelzen: Good one!

app/src/main/java/org/jellyfin/androidtv/search/TextSearchFragment.java
```diff
@@ -0,0 +1,73 @@
...
+        super.onActivityCreated(savedInstanceState);
+
+        // Add event listeners
+        EditText searchBar = getActivity().findViewById(R.id.search_bar);
+        searchBar.addTextChangedListener(this);
+        searchBar.setOnEditorActionListener(this);
+
+        // Set up result fragment
+        RowsSupportFragment rowsSupportFragment = (RowsSupportFragment) getChildFragmentManager().findFragmentById(R.id.results_frame);
+        rowsSupportFragment.setAdapter(searchProvider.getResultsAdapter());
```
thornbill: The click listener needs added here to be able to select search results.

nielsvanvelzen: No idea how I forgot this, probably made a mistake when copy pasting it. It is fixed now :)

PR #242, opened 2019-12-30
Add enter_manually button to users list

### LINKED ISSUE #236: Move Enter Manually for User Name
Similar to #167, a user on Reddit reported they weren't able to log in. It turns out that having "Enter manually" down by the "Select server" button caused some confusion. They weren't aware that was the button needed to enter a username for a hidden user.

I see where the confusion could happen: the same button, with the same appearance and location, is used on the previous screen to enter a manual server address.

Perhaps we can move it up on the user screen to the list of users? If users are visible, it should be at the end of the list. If there are not, it would be the only option.

Nautigsam:
This PR is intended to fix #236 

This is a WIP since I do not know hot to mix properly user cards and "enter manually" button on the same row. Could I have some input on it?

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/startup/SelectUserFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/startup/SelectUserFragment.java
@@ -2,6 +2,8 @@ package org.jellyfin.androidtv.startup;
 
 import android.content.Intent;
 import android.os.Bundle;
+import android.widget.Toast;
+
 import androidx.leanback.widget.ArrayObjectAdapter;
 import androidx.leanback.widget.HeaderItem;
 import androidx.leanback.widget.ListRow;
@@ -9,7 +11,6 @@ import androidx.leanback.widget.OnItemViewClickedListener;
 import androidx.leanback.widget.Presenter;
 import androidx.leanback.widget.Row;
 import androidx.leanback.widget.RowPresenter;
-import android.widget.Toast;
 
 import org.jellyfin.androidtv.R;
 import org.jellyfin.androidtv.TvApp;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/startup/SelectUserFragment.java
```diff
@@ -49,13 +49,19 @@ protected void addAdditionalRows(ArrayObjectAdapter rowAdapter) {
         HeaderItem usersHeader = new HeaderItem(rowAdapter.size(), mApplication.getString(R.string.lbl_select_user));
         ItemRowAdapter usersAdapter = new ItemRowAdapter(mServer, new CardPresenter(), rowAdapter);
         usersAdapter.Retrieve();
```
Nautigsam: I do not know how to add a simple button to this `ItemRowAdapter`. Is it even possible?

dkanada: It looks like you found a method, right?

Nautigsam: No, when there are no users I create another adapter with `GridButtonPresenter`. When there are existing users, I don't know how to add this button since the adapter uses `CardPresenter`.

nielsvanvelzen: I think you need to create your own adapter & presenter that doesn't use the current one. It gives more flexibility and allows for better customization.

Might be a bit of work though.

Nautigsam:
I close this since I need the ongoing refactoring of `ItemRowAdapter`.

PR #243, opened 2019-12-31
Fix top margin on card title

thornbill:
This fixes a small display issue where a 1px bar would be shown above the title when it has the marque effect active. See the bar above the "ne" in the following image.

screen


nielsvanvelzen:
I didn't even notice this!

PR #244, opened 2020-01-01
Upgrade to ExoPlayer 2

nielsvanvelzen:
The goal of this PR is to upgrade ExoPlayer to the latest revision supported by the Amazon port, which is currently [2.10.6](https://github.com/amzn/exoplayer-amazon-port/tree/amazon/r2.10.6). In addition ExoMedia will be removed.

The goal is **not** to refactor the player code. Everything should work as it did before with minimal changes. In the future we should consider rebuilding all player code however.

Current progress:
 - Audio playback kinda works, progression is not tracked properly
 - Video playback works quite well with some minor issues

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -27,26 +27,12 @@ android {
             debuggable true
         }
     }
-
-    flavorDimensions "version"
-    productFlavors {
-        google {
-            dimension = "version"
-            versionName = version + "g"
-        }
-        amazon {
-            dimension = "version"
-            versionName = version + "a"
-        }
-    }
 }
 
 dependencies {
     def acraVersion = '5.4.0'
 
     implementation fileTree(include: ['*.jar'], dir: 'libs')
-    amazonImplementation 'com.amazon.android:exoplayer:2.10.6'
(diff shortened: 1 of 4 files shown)


nielsvanvelzen:
Everything seems to work fine for me, tested it on my television while actually watching series (instead of checking if things work in the emulator) and everything worked flawlessly. 

Video quality is much better and smoother, upscaling is a lot nicer (which was the main reason for me to try upgrading).

For the reviewer(s): Playback for both video and audio changed. So both should be tested thoroughly, I'm mostly concerned about events (playback ending and play/pause) working differently than before.

Florianisme:
Will test this evening on my TV!

Florianisme:
Worked flawlessly! Butter smooth (at 4k) and no issues with the audio (5.1). Resuming after stopping worked aswell

thornbill:
This may be a good time to try killing the Google/Amazon build variants. The Amazon build of ExoPlayer should work for both and would drastically reduce build times.

nielsvanvelzen:
Increasing build times is always a good thing! Hopefully everything works as expected

thornbill:
I'm doing some testing on my assortment of Fire TV devices, so far so good! :tada:

PR #246, opened 2020-01-02
Focus resume or otherwise play button by default

### LINKED ISSUE #126: Minor Issue, 'Play' not highlighted by default when opening movie info
This only happens when the storyline is too long eg 
MVIMG_20190720_225100

Device: Nvidia Shield, jellyfin version 0.9

Florianisme:
Request the focus on the resume button if it is visible, otherwise the play button is focused by default.
Fixes #126


nielsvanvelzen:
Awesome, I ran into this problem yesterday so really awesome it's already fixed :)

Issue #146, opened 2019-09-01
[Feature Request] Better Seeking & GUI


grravity:
The seeking controls are quite bad imo. Using left or right on the controller is not very smooth and holding it down I'd expect it to seek faster the longer you hold it. 

The gui for seeking is also quite poor imo, I feel something similar to Youtube's seek bar, where you press up on the controller, and a dot appears on the seekbar and you can scroll through and pick your exact time. This would be ideal imo.

Would also love to see seek previews (Similar to youtube, a box above where you are seeking that displays a frame of the film)

Florianisme:
I'm currently working out a redesign and implementation of a more modern and intuitive player UI

nielsvanvelzen:
The new overlay was added in #251 and seeking was slightly improved in #581.

PR #247, opened 2020-01-02
Player overlay redesign

Florianisme:
Addresses #146
I think the overall player design might also be reorganized and cleaned up a bit.
 
Redesign will include:

- Reworking of playback progress and seeking
- Redesign of button layout

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -8,6 +8,7 @@
  - [Letroll](https://github.com/letroll)
  - [AndreasGB](https://github.com/andreasgb)
  - [joern-h](https://github.com/joern-h)
+ - [Niels van Velzen](https://github.com/nielsvanvelzen)
 
 # Emby Contributors
 
--- a/app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/browsing/HomeFragment.java
@@ -5,42 +5,30 @@ import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Bundle;
 import android.os.Handler;
-import androidx.leanback.widget.ArrayObjectAdapter;
-import androidx.leanback.widget.HeaderItem;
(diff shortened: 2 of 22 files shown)


Florianisme:
Screenshot_1577981902

Current WIP, I'd like to clean up the button row a bit to make it feel more "Android vanilla" like, don't know how I'm going to archieve that yet

nielsvanvelzen:
If I were to redesign the player I would do the following:
- Remove the poster, it uses a lot of space and I already know what I'm watching
- Remove the release year, I don't really care about that while watching
- Make the controls compacter, it's blocking the video
- "floating" controls (background is now full width, make it smaller) look better
- Add "backward" control. Why can I go forward but not backward?

I think you can get some inspiration from the [leanback documentation](https://developer.android.com/training/tv/playback/transport-controls)

Some other features that would be awesome:
- When controls not visible, using the forward and backward controls on the remote should only show the seekbar, not the entire control view
- Separate the "up next" to it's own activity (I hate how it shows on top of the video so I actually disabled it) - _this one I might do myself in the future_

Florianisme:
@nielsvanvelzen very good points! I didn't want to be too radical with the changes but I think it would be great if @thornbill could give some feedback on our ideas before I start working on them! 

I think that the media player overlay could somewhat behave like the YouTube player for Android TV. Only the main media controls (eg. play, pause, forward, backward) are shown until you press the three dots menu.
We could hide the subtiltle and audio submenus behind that. 
What do you think of that approach?

nielsvanvelzen:
I actually switch subtitles a lot in the player so I'd rather keep them unless we add them to the details activity (so I can set it beforehand).

Would also be nice if we could show the queue and chapters when pressing the "down" button (like Plex does)

Depending on your goals we could keep this PR just a simple "make is look good instead of terrible" and create an issue where we can discuss a new design before implementing it.

Florianisme:
I F*d up my version control, will close this PR and reopen

PR #249, opened 2020-01-02
Add option to choose the preferred media player in settings

nielsvanvelzen:
This PR removes the option "Use External Player for Videos" and replaces it with a new option where the user can choose between the following items:
- Auto
  The old behavior when not using external player

- ExoPlayer
  Force ExoPlayer for playback

- libVLC
  Force libVLC for playback

- External
  The old behavior when using external player

A new preference config version is introduced to migrate the existing setting to the new one.

This change could help when debugging codecs in various players and gives the user more control over playback.


nielsvanvelzen:
I'm not really sure how playback for live tv works and am unable to test it so I did not change those options.

app/src/main/res/values/strings.xml
```diff
@@ -305,11 +305,9 @@
     <string name="lbl_next_episode">Next Episode</string>
     <string name="lbl_album_name">Album name</string>
     <string name="lbl_song_title">Song Title</string>
-    <string name="pref_use_external_player_title">Use External Player for Videos</string>
-    <string name="pref_use_external_player_summary">Requires a capable app to play video types</string>
```
thornbill: These will need to be removed for all locales.

app/src/main/java/org/jellyfin/androidtv/startup/StartupActivity.java
```diff
@@ -70,6 +70,12 @@ protected void onCreate(Bundle savedInstanceState) {
             application.getPrefs().edit().putBoolean("pref_enable_info_panel", false).apply();
             application.getSystemPrefs().edit().putString("sys_pref_config_version", "4").apply();
         }
+        if (Integer.parseInt(application.getConfigVersion()) < 5) {
```
thornbill: Does it make sense to maintain a config version? I have been thinking it may be less overhead to just check if a specific preference exists and set the default value if it does not.

nielsvanvelzen: I think it does make sense! Imagine a setting being renamed or (like in this case) a setting replacing another one. It would be annoying if the app auto-updates via the play store and suddenly your behavior changed because there is no migration. In this case external player users wouldn't use the external player anymore.

I do however think that the way this is done is ugly, in the future we might want to separate the preferences to a custom class which brings type safety (using enums instead of strings etc.) and handles migration a bit better.

This class could have getters & setters that need to be called and could use shared preferences as storage or maybe even sqlite.

Florianisme: I've created a simple class for usage in my apps with these simple getters and setters and I've used it a lot. Might save a little bit of work https://github.com/Florianisme/SharedPreferences

nielsvanvelzen: I would prefer the different settings being defined too, so no "getBoolean" or "getString" but "getPreferredPlayer" and "getConfigVersion".

Something like:
```kotlin
enum class PlayerType {
	AUTO,
	VLC,
	EXOPLAYER,
	EXTERNAL
}

class Preferences(
	private val context: Context
) {
	private val sharedPreferences by lazy {
		PreferenceManager.getDefaultSharedPreferences(context)
	}

	/* Preferences */
	var preferedPlayer by enumPreference("pref_video_player", PlayerType.AUTO)
	var configVersion by intPreference("sys_pref_config_version", 1)

	/** Storage functions */
	private fun intPreference(key: String, default: Int) = object : ReadWriteProperty<Preferences, Int> {
		override fun getValue(thisRef: Preferences, property: KProperty<*>): Int {
			return sharedPreferences.getInt(key, default)
		}

		override fun setValue(thisRef: Preferences, property: KProperty<*>, value: Int) {
			sharedPreferences.edit().putInt(key, value).apply()
		}
	}

	private inline fun <reified T: Enum<T>> enumPreference(key: String, default: T?) = object : ReadWriteProperty<Preferences, T?> {
		override fun getValue(thisRef: Preferences, property: KProperty<*>): T? {
			val stringValue = sharedPreferences.getString(key, null)

			return if (stringValue == null) default
			else T::class.java.enumConstants.find { it.name == stringValue }
		}

		override fun setValue(thisRef: Preferences, property: KProperty<*>, value: T?) {
			if (value == null) sharedPreferences.edit().remove(key).apply()
			else sharedPreferences.edit().putString(key, value.toString()).apply()
		}
	}
}
```

_I actually wanted to make a quick example but got carried away so this should probably work already lol_

nielsvanvelzen: Now that I spend some effort I made an issue for it #250

app/src/main/res/values/strings.xml
nielsvanvelzen: Fixed, first time working on an android app with multiple languages so I forgot to remove the strings from other files. I noticed the string editor in android studio helps with that fortunately

thornbill:
I noticed some issues when forcing libvlc and the video has to be transcoded:
* Position reporting not updated
* Seek bar does not update on skipping

Maybe we should add some form of disclaimer that this option could cause playback issues if set to exoplayer or libvlc.

nielsvanvelzen:
So the playback of transcoded video is working but the progress is not updated properly? If that's the only option we should be able to fix that. I'll look into adding a warning when not choosing the `auto` or `external` option.

thornbill:
Right playback is working but when forcing libvlc and it’s transcoding the progress bar does not show the progress and you are unable to resume playback from where you left off. I’m guessing something is up with the progress reporting that is responsible for both of those issues.

nielsvanvelzen:
I think we should report the progress reporting from libVLC while transcoding in a separate issue from this one so we can fix it after this PR and the ExoPlayer 2 PR is merged. (to prevent conflicts).

Regarding the warning/disclaimer, it seems like it's not possible to add an additional description to list preferences so the only option would be to show a warning after selecting a preferred player. I don't think we should go that far about warning users. We should instead make sure that no matter what playback backend is used their will be no problems.

The only issues I can think of that could happen are:
 - Wrong implementation in the library (in this case VLC with reporting the current progress) - This should absolutely not happen and needs to be fixed anyway
 - Wrong codec used - Profiles need to be updated so it transcodes

These issues can still happen when the "auto" option is used.

PR #251, opened 2020-01-03
Player overlay redesign

Florianisme:
Addresses #146 
VIdeo Player overlay will be redesigned making it match the vanilla Android TV style.
Existing features such as subtitles and audio track control should not be affected by this PR.


thornbill:
I know you had asked in the other PR... I am definitely good with improving the design of the playback controls. I do think subtitles and language should be top level controls. Less common options should be fine in a sub menu.

Florianisme:
I'm currently using the leanback overlay and this is what it looks like:
Screenshot_1578150909

What do you think about it?

nielsvanvelzen:
I like it, how does it look for chapter selection?

Florianisme:
@nielsvanvelzen not yet implemented but I should be able to just implement it like the audio track switching from my last commit

Florianisme:
Next task is migrating the Live TV controls to the new overlay and start removing the old obsolete code and layout

Florianisme:
All Live TV buttons (except for the record button) were merged to the new overlay.
This is what it looks like in TV and normal movie mode:
Screenshot_1578250529
Screenshot_1578250577
Oh and the popup menus are working aswell.
Moving the buttons to different positions is obviously no issue.

ThatNerdyPikachu:
Question: why is there a "full screen" button there, and what does it do? (This is also in the old overlay.)

AndreasGB:
I think it adjusts whether the video is zoomed in to get rid of black bars, but I have never used it either.

ThatNerdyPikachu:
Oh, right. That is what it does, just tried it. Thanks!

Florianisme:
> Question: why is there a "full screen" button there, and what does it do? (This is also in the old overlay.)

Yeah, maybe we could swap that out with a less misleading icon such as this (it's literally called "aspect-ratio"):
aspect-ratio

AndreasGB:
I think that would make sense if this is what other TV apps are using. I can't check right now since I'm not at home.

dkanada:
What are the back arrow and three dot menu for on the live TV interface?

Florianisme:
Back arrow is for switching to the previous channel and the three dot button is for showing the quick channel changer

dkanada:
Is there no button to switch to the next channel? Also, that three dot button doesn't seem to look that nice, isn't the material icon different than that?

Florianisme:
Let's move this general discussion to the issue #257

Mantu120:
@Florianisme 
Can you add a "Chapter Selection" button after "Audio Track Selection" button?
It will be good for selecting Chapters for Movie file.

Florianisme:
@Mantu120 yes, it'll have the same behavior as before

Florianisme:
I think everything works now so far. So I guess extensive testing can begin :)

app/src/main/java/org/jellyfin/androidtv/util/Utils.java
```diff
@@ -206,6 +206,10 @@ public static PopupMenu createPopupMenu(Activity activity, View view, int gravit
         return new PopupMenu(activity, view, gravity);
     }
 
+    public static PopupMenu createPopupMenu(Context context, View view, int gravity) {
```
nielsvanvelzen: I think the `createPopMenu` function which accepts an `Activity` instance can be removed now

app/src/main/java/org/jellyfin/androidtv/playback/overlay/actions/AdjustAudioDelayAction.java
```diff
@@ -0,0 +1,14 @@
+package org.jellyfin.androidtv.playback.overlay.actions;
+
+import android.content.Context;
+
+import org.jellyfin.androidtv.R;
+import org.jellyfin.androidtv.playback.overlay.CustomPlaybackTransportControlGlue;
+
+public class AdjustAudioDelayAction extends CustomAction {
```
nielsvanvelzen: A lot of actions have the exact same code with only the name and icon changed. Wouldn't it be better to create an `IconAction` which accepts the icon as parameter in the constructor? This way you can re-use that class for all simple actions.

Maybe even add it to the `CustomAction` class so you don't even need to extend it.

app/src/main/java/org/jellyfin/androidtv/playback/overlay/actions/RecordAction.java
```diff
@@ -0,0 +1,23 @@
+package org.jellyfin.androidtv.playback.overlay.actions;
+
+import android.content.Context;
+import android.graphics.drawable.Drawable;
+
+import org.jellyfin.androidtv.R;
+import org.jellyfin.androidtv.playback.overlay.CustomPlaybackTransportControlGlue;
+
+public class RecordAction extends CustomAction {
```
nielsvanvelzen: (addition to the previous suggestion for the action classes) adding a second icon parameter to the (to-be-made) constructor could trigger the active/inactive behavior. When the second icon is null it will always use the first one.

app/src/main/java/org/jellyfin/androidtv/playback/overlay/actions/CustomAction.java
```diff
@@ -0,0 +1,37 @@
...
+    private Context context;
+    private CustomPlaybackTransportControlGlue customPlaybackTransportControlGlue;
+
+    public CustomAction(Context context, CustomPlaybackTransportControlGlue customPlaybackTransportControlGlue) {
+        this(0);
+        this.context = context;
+        this.customPlaybackTransportControlGlue = customPlaybackTransportControlGlue;
+    }
+
+    private CustomAction(int id) {
```
nielsvanvelzen: I don't think I saw this constructor use in any other place. I'd suggest moving the `super` call to the other constructor.

app/src/main/java/org/jellyfin/androidtv/playback/overlay/CustomActionClickedHandler.java
```diff
@@ -0,0 +1,147 @@
...
+import org.jellyfin.androidtv.playback.PlaybackController;
+import org.jellyfin.androidtv.playback.VideoManager;
+import org.jellyfin.androidtv.ui.AudioDelayPopup;
+import org.jellyfin.androidtv.ui.ValueChangedListener;
+import org.jellyfin.androidtv.util.Utils;
+import org.jellyfin.apiclient.model.entities.MediaStream;
+
+import java.util.List;
+
+public class CustomActionClickedHandler {
```
nielsvanvelzen: I don't really like how this one class is responsible for all click handlers, I'd rather see them added to their respective action class.

app/src/main/java/org/jellyfin/androidtv/playback/overlay/CustomPlaybackTransportControlGlue.java
```diff
@@ -0,0 +1,239 @@
...
+            getPlayerAdapter().rewind();
+        } else if (action == fastForwardAction) {
+            getPlayerAdapter().fastForward();
+        } else if (action == skipNextAction) {
+            getPlayerAdapter().next();
+        }
+        notifyActionChanged(action);
+    }
+
+    public void onCustomActionClicked(Action action, View view) {
```
nielsvanvelzen: See my previous comment on the way these handlers are implemented. Would be better if you could just do `action.onClicked();` or something like that

app/src/main/java/org/jellyfin/androidtv/playback/overlay/CustomSeekProvider.java
```diff
@@ -0,0 +1,27 @@
+package org.jellyfin.androidtv.playback.overlay;
+
+import androidx.leanback.widget.PlaybackSeekDataProvider;
+
+public class CustomSeekProvider extends PlaybackSeekDataProvider {
```
nielsvanvelzen: I'm not really sure what this class does. Is it providing the possible "seek positions" to the player? And if so, does this mean that whatever I watch it can only seek to 30 positions? Which would mean that if I watched a movie of 30 hours I would skip one hour every time?

app/src/main/java/org/jellyfin/androidtv/playback/overlay/VideoPlayerAdapter.java
```diff
@@ -0,0 +1,148 @@
...
+
+    void setMasterOverlayFragment(CustomPlaybackOverlayFragment customPlaybackOverlayFragment) {
+        this.customPlaybackOverlayFragment = customPlaybackOverlayFragment;
+    }
+
+    CustomPlaybackOverlayFragment getMasterOverlayFragment() {
+        return customPlaybackOverlayFragment;
+    }
+
+    boolean canRecordLieTv() {
```
nielsvanvelzen: Typo in `live`

app/src/main/java/org/jellyfin/androidtv/playback/overlay/CustomSeekProvider.java
nielsvanvelzen: I just confirmed this myself while testing. I think adding stepts for every `x` seconds would be better.

nielsvanvelzen:
So far it looks very good. Some things I noticed while testing in my development emulator:
- When seeking to the end the player crashes (can't really find a good stacktrace for this right now)
- When choosing a chapter pressing the LEFT or RIGHT button at the edge (so left when first item is selected) causes the player controls to show in the background
- One particular show in my library was unable to play, the player instantly crashed when opening. I tried it multiple times (restarting the app too). At some point it suddenly worked somehow. [stacktrace.txt](https://github.com/jellyfin/jellyfin-androidtv/files/4048022/stacktrace.txt)

app/src/main/java/org/jellyfin/androidtv/playback/overlay/actions/RecordAction.java
Florianisme: Good point, but as you can see, I check which type of action is clicked in the onCustomActionClicked method in the CustomPlaybackTransportControlGlue class. But maybe I can instead use some sort of enum (eg. ACTION_SUBTITLES) which I pass in the IconAction object on initialization which I can later check against.

app/src/main/java/org/jellyfin/androidtv/playback/overlay/CustomPlaybackTransportControlGlue.java
Florianisme: Good point!
But then I will have to keep an action class for every action I add.

app/src/main/java/org/jellyfin/androidtv/playback/overlay/CustomSeekProvider.java
Florianisme: Yes, I also think I saw a code example of a quick seeker, which means the faster you click (or the longer you hold to the right), the faster you seek

Florianisme:
First of all, thanks a lot for the comprehensive review!
 
> When seeking to the end the player crashes (can't really find a good stacktrace for this right now)

I cant reproduce it responsibly but when I do it says that 
[stacktrace.txt](https://github.com/jellyfin/jellyfin-androidtv/files/4048400/stacktrace-seeking.txt) ExoPlayer was null so maybe there is something conflicting with your recent PR and these changes.

> When choosing a chapter pressing the LEFT or RIGHT button at the edge (so left when first item is selected) causes the player controls to show in the background

Yes I also noticed that before but I havent found out why it worked before when the old overlay was basically behind the chapter selection aswell.

> One particular show in my library was unable to play, the player instantly crashed when opening. I tried it multiple times (restarting the app too). At some point it suddenly worked somehow. [stacktrace.txt](https://github.com/jellyfin/jellyfin-androidtv/files/4048022/stacktrace.txt)

Maybe the check is too early before everything is initialized properly. I thought about only adding the title, current pos and all the actions after getting a callback from the CustomPlaybackOverlayFragment when everything is done (kinda like the old `updateDisplay()` method in there)

> edit: I also see a lot of unused imports, try to use the "Optimize Imports" quickfix for this in Android Studio.

I never notice that, will do :)

nextlooper42:
I hope that audio and subtitle button will remain onscreen all time, every non english person use them tbh :D

app/src/main/java/org/jellyfin/androidtv/playback/overlay/CustomSeekProvider.java
nielsvanvelzen: Can you update this one so it works better for movies / long episodes? 30 steps is enough for a 30 minute show but it isn't for a 3 hour movie. The behavior you mentioned you saw in an example would be awesome to have

Florianisme:
No, I was waiting for your response regarding the single action items.
> See my previous comment on the way these handlers are implemented. Would be better if you could just do `action.onClicked();` or something like that

If I do that I'll have to keep a class per action. If not all actions can be merged into a single action class..

nielsvanvelzen:
You can add the click function as parameters. In Java this can be done like this:

```java
public interface ActionClickHandler {
    void onClicked(Context context);
}

public class TestAction extends CustomAction {
    public TestAction(Context context, ActionClickHandler onClick) {}
}
```

In Kotlin it could look like this:
```kotlin
class TestAction(
        private val context: Context,
        private val onClick: (context: Context) -> Unit
) extends CustomAction()
```

If you do it like this all actions can be implemented "inline" and don't need their own classes.

Florianisme:
@nielsvanvelzen I know what an interface is but I won't implement the logic for clicking on an action in its initialization. That would be a big mess.

nielsvanvelzen:
In that case I'd suggest creating separate action classes with their own click implementation.

Florianisme:
I'd like to clean up some classes and code from the old overview as a followup PR. I'll do some testing aswell!

nielsvanvelzen:
What do you mean with "overview"? If you mean the details screen: we're already working on that!
If you meant to say "overlay", awesome 🎉

Florianisme:
Ah sorry, yes I meant the old overlay code. Saw your PR, looks really promising!

giejay:
I'm assuming this PR does not include being able to set the subtitle delay like on the native Android app?

nielsvanvelzen:
@giejay no it doesn't. I'm not aware of a feature request for it either.

giejay:
I dont think there is a feature request currently so I will see if I can make some time to port it from the Android app. Just making sure it's not already done;)

thornbill:
I have (finally) started doing some testing on this now that 0.11.0 has been released. Overall it is looking great!

Here are a couple of things I noticed initially:
* The branch needs updated with the latest changes from master. (Building this branch is failing now due to #288)
* The state of the play/pause button on the overlay does not update if using the play/pause button on a remote.
* Pushing the "down" button on my remote while watching live tv seems to cause a conflict with the on screen controls and the channel selector. Both attempt to display and then neither stay on screen.

I will do some additional testing just to make sure everything else seems to be working well.

Florianisme:
Thanks for your initial feedback! 😄 
> * The branch needs updated with the latest changes from master. (Building this branch is failing now due to #288)

Do I just pull from upstream into my working copy and commit everything? There will be hundreds of other commits then right?
> * The state of the play/pause button on the overlay does not update if using the play/pause button on a remote.

I will address that next, thanks!
> * Pushing the "down" button on my remote while watching live tv seems to cause a conflict with the on screen controls and the channel selector. Both attempt to display and then neither stay on screen.

Yeah, I've had some similar issues before but not this specific problem, will fix!

thornbill:
You should be able to `git fetch --all` and `git merge upstream/master` and then push the update. GitHub should not show the commits from master, but sometimes they do show up for some reason. The other option would be to rebase and force push, but that messes with previous reviews since it overwrites the branch history.

nielsvanvelzen:
Hey @Florianisme can you resolve the conflicts again? We will merge it when that's done.

PR #260, opened 2020-01-09
Refactor preferences

### LINKED ISSUE #250: Create facade before preferences
The application talks directly with the preference store and this works fine at the moment. But the keys (and "enum" values) are always strings. This can cause problems when a string is not speller correctly or if a developer assumes a certain value exists but it actually doesn't.

For example: I recently added a "enum" that saves the preferred video player. This could save the string "vlc" or "exoplayer". But what if _somehow_ the value isn't set properly? It could break something!

The idea here is to create a facade that brings all preferences to a single file with typesafety, neatly ordered and documented.

An example implementation (in kotlin) could look like this:

```kotlin
enum class PlayerType {
	AUTO,
	VLC,
	EXOPLAYER,
	EXTERNAL
}

class Preferences(
	private val context: Context
) {
	private val sharedPreferences by lazy {
		PreferenceManager.getDefaultSharedPreferences(context)
	}

	/* Preferences */
	var preferedPlayer by enumPreference("pref_video_player", PlayerType.AUTO)
	var configVersion by intPreference("sys_pref_config_version", 1)

	/** Storage functions */
	private fun intPreference(key: String, default: Int) = object : ReadWriteProperty<Preferences, Int> {
		override fun getValue(thisRef: Preferences, property: KProperty<*>): Int {
			return sharedPreferences.getInt(key, default)
		}

		override fun setValue(thisRef: Preferences, property: KProperty<*>, value: Int) {
			sharedPreferences.edit().putInt(key, value).apply()
		}
	}

	private inline fun <reified T: Enum<T>> enumPreference(key: String, default: T?) = object : ReadWriteProperty<Preferences, T?> {
		override fun getValue(thisRef: Preferences, property: KProperty<*>): T? {
			val stringValue = sharedPreferences.getString(key, null)

			return if (stringValue == null) default
			else T::class.java.enumConstants.find { it.name == stringValue }
		}

		override fun setValue(thisRef: Preferences, property: KProperty<*>, value: T?) {
			if (value == null) sharedPreferences.edit().remove(key).apply()
			else sharedPreferences.edit().putString(key, value.toString()).apply()
		}
	}
}
```

Working with this can be done in Kotlin:
```kotlin
val preferences = Preferences(context) // Ideally a stored instance in TvApp or something
preferences.preferedPlayer = PlayerType.VLC
```

Or Java:
```kotlin
Preferences preferences = new Preferences(context); // Ideally a stored instance in TvApp or something
preferences.setPreferedPlayer(PlayerType.VLC);
```

I think this could improve the way preferences are used and helps new developers to learn what preferences are available and how to use them (if kdoc/javadoc is properly added).

_This issue started when @thornbill asked if preference migration was actually useful. The conversation started [right here](https://github.com/jellyfin/jellyfin-androidtv/pull/249)_

nielsvanvelzen:
Fixes #250


nielsvanvelzen:
I'm considering this PR done. Ideally I wanted to store integers/longs not as string but that caused issues in the preference fragment (it only supports lists for strings...).

So for now this is it. The migration code will migrate the settings from the currently released app version, not the master branch.

thornbill:
CI build is failing with the following error:

```
  Errors found:
  
  /home/vsts/work/1/s/app/src/main/java/org/jellyfin/androidtv/playback/ExternalPlayerActivity.java:68: Error: Overriding method should call super.onActivityResult [MissingSuperCall]
      protected void onActivityResult(int requestCode, int resultCode, Intent data) {
```

app/src/main/java/org/jellyfin/androidtv/preferences/ui/UserPreferencesFragment.kt
```diff
@@ -0,0 +1,145 @@
...
+import android.os.Bundle
+import androidx.preference.*
+import org.jellyfin.androidtv.R
+import org.jellyfin.androidtv.TvApp
+import org.jellyfin.androidtv.model.LogonCredentials
+import org.jellyfin.androidtv.util.DeviceUtils
+import org.jellyfin.androidtv.util.apiclient.AuthenticationHelper
+import java.io.IOException
+
+class UserPreferencesFragment : PreferenceFragmentCompat(), OnSharedPreferenceChangeListener {
```
thornbill: Should we be using `LeanbackPreferenceFragmentCompat`?

nielsvanvelzen: LeanbackPreferenceFragmentCompat unfortunately doesn't work, when I tried it previously it crashed the application. So for now I didn't change the ui much. I might do that in the future though.

thornbill: I don’t like that this doesn’t match the app theme. Maybe we should extend the default theme you used to apply the Jellyfin colors as a temporary solution?

nielsvanvelzen: I managed to get the leanback style preferences working. It now looks like this:

image

thornbill:
@nielsvanvelzen could you take a look at these conflicts? The code changes look fine! I will do some testing once the conflicts are fixed. :smile:

nielsvanvelzen:
I wanted to rebase it at first but didn't trust the result so I did a merge.

app/src/main/java/org/jellyfin/androidtv/startup/StartupActivity.java
```diff
@@ -57,26 +57,6 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        }
-        if (Integer.parseInt(application.getConfigVersion()) < 5) {
-            boolean useExternal = application.getPrefs().getBoolean("pref_video_use_external", false);
-
-            application.getPrefs().edit().putString("pref_video_player", useExternal ? "external" : "auto").apply();
-            application.getSystemPrefs().edit().putString("sys_pref_config_version", "5").apply();
-        }
-
         //Ensure we have prefs
         PreferenceManager.setDefaultValues(this, R.xml.preferences, false);
```
nielsvanvelzen: Might want to move this to the newly created stores

app/src/main/java/org/jellyfin/androidtv/preferences/UserPreferences.kt
```diff
@@ -0,0 +1,169 @@
...
+	/**
+	 * Sign out automatically after x milliseconds
+	 */
+	var autoSignOutTimeout by stringPreference("pref_auto_logoff_timeout", "3600000")
+
+	/* Display */
+	/**
+	 * Enable background images while browsing
+	 */
+	var backdropEnabled by booleanPreference("pref_show_backdrop", false)
```
thornbill: It seems all boolean preferences are being defaulted to `false` now. Is this correct?

nielsvanvelzen: I might have forgotten to set the default values for the booleans. I'll check them tomorrow.

PR #265, opened 2020-01-17
Playing Next screen

nielsvanvelzen:
This PR removes the "playing next" dialog that is shown when the media you're watching is ending and adds a new screen that shows up after the media ended.

It looks like this:
image
image
_The second one is slightly older, I think some small things like margins have changed_

As you can see the thumbnail for the next playing item is shown with some information (title/description) on the right. On top of the "box" is a progress bar that will decrease. When it hits 0 it will play the next item.

## Why
I made it because I don't like things showing up on top of my video and this also reduces the mess of the video player code (less refactoring in the future!). I also think a screen like this helps the user to get a moment to get a quick snack or something.

## Future changes
In the future I'd like to add a setting to control how long the "next up" screen will be shown (slightly depends on #260). If users want the old behavior we could show a smaller version on top of the video when the setting is set to a negative value. (-5 means "5 seconds before video ends" but +5 means "show for 5 seconds after video ends")

The feature is implemented in a modular way so it can by easily re-used if we change things in the video-player. This is also the reason it calls the API by itself to request item information.


thornbill:
This looks pretty awesome! It will be a few days before I am able to review the code changes.

At some point I would like an option for a more minimal display version too.

AndreasGB:
While it looks nice so far, I have two UX concerns

* There is no way to cancel the timeout until the next episode is played and no buttons informing the user of an ability to intervene. Ideal to me would be an auto-filling continue button like I think Netflix has, which doesn't fill automatically anymore once it has lost focus once, and a second cancel button.

* The next episode is played too quickly by default, I would give the users more time to find their remote in case they want to cancel

Didn't look at the code yet

nielsvanvelzen:
For your first point: You can actually cancel it with the "back" button on the remote to stop or the "ok" button to instantly play the next item.
When the timer is cancelled pressing "back" again will close the nextup screen.
Second point: After #260 is merged I can add an option to change the duration. IMO when watching a whole season you want it to be fast. Might add a few seconds for the default though.

app/src/main/res/layout/fragment_next_up_buttons.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="wrap_content"
+    android:layout_height="wrap_content"
+    android:orientation="horizontal"
+    android:padding="16dp">
```
AndreasGB: I would remove the padding here to make it better to fit into other views, who then themselves can decide how to margin it

app/src/main/res/layout/fragment_next_up.xml
```diff
@@ -0,0 +1,76 @@
...
+            android:ellipsize="end"
+            android:maxLines="6"
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toEndOf="@id/image"
+            app:layout_constraintTop_toBottomOf="@id/title"
+            tools:text="Episode description" />
+    </androidx.constraintlayout.widget.ConstraintLayout>
+
+    <org.jellyfin.androidtv.playback.nextup.NextUpButtons
```
AndreasGB: Would move this into the Overlay layout

app/src/main/res/layout/fragment_next_up.xml
```diff
@@ -0,0 +1,76 @@
...
+            tools:text="Episode title" />
+
+        <TextView
+            android:id="@+id/description"
+            style="@style/TextAppearance.Leanback.DetailsDescriptionBody"
+            android:layout_width="0dp"
+            android:layout_height="0dp"
+            android:layout_marginStart="16dp"
+            android:ellipsize="end"
+            android:maxLines="6"
```
AndreasGB: Would set this to 4 to end above the `NextUpButtons`

app/src/main/res/layout/fragment_next_up.xml
```diff
@@ -0,0 +1,76 @@
...
+            app:layout_constraintBottom_toBottomOf="parent"
```
AndreasGB: `app:layout_constraintBottom_toTopOf="@id/fragment_next_up_buttons"` to avoid underflowing the buttons with text

app/src/main/java/org/jellyfin/androidtv/playback/nextup/NextUpActivity.kt
```diff
@@ -0,0 +1,65 @@
...
+			Log.e(LOG_TAG, "No id found in bundle for NextUpActivity.")
+			finish()
+		}
+
+		// Add background manager
+		BackgroundManager.getInstance(this).attach(window)
+
+		// Load item info
+		GlobalScope.launch(Dispatchers.Main) {
+			val data = loadItemData(id) ?: return@launch
```
AndreasGB: Maybe log an error in the `null` case and `finish()`

app/src/main/res/layout/vlc_player_interface.xml
```diff
@@ -58,114 +58,170 @@
...
     </FrameLayout>
 
-    <include
-        android:id="@+id/smNextUpPanel"
-        layout="@layout/small_upnext_panel"
-        android:layout_width="match_parent"
-        android:layout_height="75dp"
-        android:layout_gravity="bottom" />
+
+    <ProgressBar
```
AndreasGB: This is not present in the current version of the overlay, I think this file is broken. This ProgressBar was removed in https://github.com/jellyfin/jellyfin-androidtv/commit/ed01855a66309d1092b86a45e007aea0b2ff1422#diff-92d250046e8bf52735338f876b5b0009

app/src/main/java/org/jellyfin/androidtv/playback/nextup/NextUpButtons.kt
```diff
@@ -0,0 +1,66 @@
...
+	fun startTimer() {
+		val duration = TvApp.getApplication().userPreferences.nextUpTimeout.toLong()
+
+		// Cancel current timer if one is already set
+		countdownTimer?.cancel()
+
+		// Create timer
+		countdownTimer = object : CountDownTimer(duration, 1) {
+			override fun onTick(millisUntilFinished: Long) {
+				fragment_next_up_buttons_play_next_progress.max = duration.toInt()
```
AndreasGB: This is probably a little bit nit-picky, but do we actually need to set max on every tick?

app/src/main/java/org/jellyfin/androidtv/playback/nextup/NextUpButtons.kt
```diff
@@ -0,0 +1,66 @@
...
+
+			override fun onFinish() {
+				// Perform a click so the event handler will activate
+				view.fragment_next_up_buttons_play_next.performClick()
+			}
+		}.start()
+	}
+
+	fun stopTimer() {
+		countdownTimer?.cancel()
```
AndreasGB: We could also reset the progressbar here, looks a bit strange when it's hanging halfway there

app/src/main/java/org/jellyfin/androidtv/util/apiclient/ApiClientExtensions.kt
```diff
@@ -27,3 +29,12 @@ suspend fun ApiClient.getUserViews(): ItemsResult? = suspendCoroutine { continua
 		override fun onError(exception: Exception?) = continuation.resume(null)
 	})
 }
+/**
+ * Adds a coroutine capable version of the "getNextUpEpisodes" function
```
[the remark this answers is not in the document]
nielsvanvelzen: no it's not

app/src/main/res/layout/fragment_next_up.xml
nielsvanvelzen: This should be done at a later moment when we're supporting a negative timeout for the buttons. For now it's embedded in the fragment so the code is entirely separated.

app/src/main/java/org/jellyfin/androidtv/playback/nextup/NextUpButtons.kt
nielsvanvelzen: So it says "every tick" but it won't actually call it every tick. It only calls the onTick function when it is able to. It's entirely based on the system performance.

app/src/main/res/layout/fragment_next_up.xml
nielsvanvelzen: Because the buttons are not defined in the same constraintlayout it's not possible. With the max lines set to 4 it should be good now though.

app/src/main/res/drawable/button_default_highlight.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+    <corners android:radius="3dp" />
+    <solid android:color="#E6CCCCCC" />
```
thornbill: Should we assign these color values to resources to allow for theming in the future?

nielsvanvelzen: Good idea, I will do that so we're 0.1% closer to a hotdog theme.

PR #268, opened 2020-01-21
Add HomeFragmentRow for live tv buttons

thornbill:
This restores some missing buttons for Live TV that were removed by the home section changes. I moved those buttons and the guide button to a separate row to reflect the web ui.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentLiveTVRow.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentLiveTVRow.kt
@@ -1,0 +1,75 @@
+package org.jellyfin.androidtv.ui.home
+
+import android.app.Activity
+import android.content.Intent
+
+import androidx.leanback.widget.ArrayObjectAdapter
+import androidx.leanback.widget.HeaderItem
+import androidx.leanback.widget.ListRow
+import androidx.leanback.widget.OnItemViewClickedListener
+import androidx.leanback.widget.Presenter
+import androidx.leanback.widget.Row
+import androidx.leanback.widget.RowPresenter
+
+import org.jellyfin.androidtv.R
+import org.jellyfin.androidtv.TvApp
+import org.jellyfin.androidtv.browsing.BrowseRecordingsActivity
+import org.jellyfin.androidtv.browsing.BrowseScheduleActivity
+import org.jellyfin.androidtv.browsing.UserViewActivity
+import org.jellyfin.androidtv.livetv.LiveTvGuideActivity
+import org.jellyfin.androidtv.presentation.CardPresenter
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentLiveTVRow.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.jellyfin.androidtv.browsing.BrowseScheduleActivity
+import org.jellyfin.androidtv.browsing.UserViewActivity
+import org.jellyfin.androidtv.livetv.LiveTvGuideActivity
+import org.jellyfin.androidtv.presentation.CardPresenter
+import org.jellyfin.androidtv.presentation.GridButtonPresenter
+import org.jellyfin.androidtv.ui.GridButton
+import org.jellyfin.apiclient.model.dto.BaseItemDto
+
+class HomeFragmentLiveTVRow(val activity: Activity?) : HomeFragmentRow(), OnItemViewClickedListener {
+	override fun addToRowsAdapter(cardPresenter: CardPresenter?, rowsAdapter: ArrayObjectAdapter?) {
```
nielsvanvelzen: When overriding Java methods in Kotlin you can actually choose if an parameter is non-null. Removing the question marks from the functions parameters so you don't need to use the non-null operator (`!!`) in the implementation.

_Same for some other places in this file_

app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentLiveTVRow.kt
```diff
@@ -0,0 +1,76 @@
...
+		if (item !is GridButton) return
+
+		when (item.id) {
+			TvApp.LIVE_TV_GUIDE_OPTION_ID -> {
+				val guide = Intent(activity, LiveTvGuideActivity::class.java)
+				activity!!.startActivity(guide)
+			}
+			TvApp.LIVE_TV_RECORDINGS_OPTION_ID -> {
+				val recordings = Intent(activity, BrowseRecordingsActivity::class.java)
+				val folder = BaseItemDto()
```
nielsvanvelzen: Can be written like so (not necessary though, just letting you know):

```kotlin
val folder = BaseItemDto().apply {
	id = ""
	name = activity?.getString(R.string.lbl_recorded_tv)
}
```

When you want to go all-out this one won't use any variables (Should probably add some new lines to make it readable though):
```kotlin
activity!!.startActivity(Intent(activity, BrowseRecordingsActivity::class.java).apply {
	putExtra("Folder", TvApp.getApplication().serializer.SerializeToString(BaseItemDto().apply {
		id = ""
		name = activity.getString(R.string.lbl_recorded_tv)
	}))
})
```
Kotlin can do a lot of fun things like this :)

PR #270, opened 2020-01-23
Initial structure idea for the redesign of the FullDetailsActivity

AndreasGB:
Basically a very rough draft of an idea on how to structure this. Sorry for my bad Kotlin :D

Basically only works when selecting an Episode of a Show in Up Next

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/details/BaseDetailsFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/BaseDetailsFragment.kt
@@ -1,0 +1,14 @@
+package org.jellyfin.androidtv.details
+
+import android.os.Bundle
+import androidx.leanback.app.DetailsSupportFragment
+import androidx.leanback.widget.*
+
+open class BaseDetailsFragment : DetailsSupportFragment() {
+	protected lateinit var rowsAdapter: ArrayObjectAdapter
+
+	override fun onCreate(savedInstanceState: Bundle?) {
+		super.onCreate(savedInstanceState)
+	}
+
+}
--- a/app/src/main/java/org/jellyfin/androidtv/details/BaseDetailsFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/BaseDetailsFragment.kt
@@ -1,14 +1,16 @@
 package org.jellyfin.androidtv.details
 
-import android.os.Bundle
 import androidx.leanback.app.DetailsSupportFragment
 import androidx.leanback.widget.*
+import com.squareup.picasso.Picasso
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 
 open class BaseDetailsFragment : DetailsSupportFragment() {
 	protected lateinit var rowsAdapter: ArrayObjectAdapter
 
-	override fun onCreate(savedInstanceState: Bundle?) {
-		super.onCreate(savedInstanceState)
-	}
 
+	protected suspend fun getImageFromURL(url: String) = withContext(Dispatchers.IO) {
+		Picasso.with(activity).load(url).get()
+	}
 }
--- a/app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
@@ -1,0 +1,56 @@
+package org.jellyfin.androidtv.details
+
+import android.os.Bundle
+import android.util.Log
+import androidx.fragment.app.FragmentActivity
+import androidx.leanback.app.BackgroundManager
+import com.squareup.picasso.Picasso
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import org.jellyfin.androidtv.R
+import org.jellyfin.androidtv.TvApp
+import org.jellyfin.androidtv.util.apiclient.asEpisode
+import org.jellyfin.androidtv.util.apiclient.getItem
+import org.jellyfin.apiclient.model.dto.BaseItemDto
+import org.jellyfin.apiclient.model.dto.BaseItemType
+import org.jellyfin.apiclient.model.dto.ImageOptions
+
+class DetailsActivity : FragmentActivity() {
+	private lateinit var fragment: BaseDetailsFragment
+
+	override fun onCreate(savedInstanceState: Bundle?) {
+		super.onCreate(savedInstanceState)
+
+		val id = intent.getStringExtra("id")
+		if (id == null) finish()
+
+		BackgroundManager.getInstance(this).attach(window)
+
+		GlobalScope.launch(Dispatchers.Main) {
+			val baseItem = getBaseItemDtoForID(id) ?: return@launch
+
+			when (baseItem.baseItemType!!) {
+				BaseItemType.Episode -> {
+					val primaryImageUrl = TvApp.getApplication().apiClient.GetImageUrl(baseItem, ImageOptions())
+					val primaryImageBitmap = getImageFromURL(primaryImageUrl)
+					val episode = baseItem.asEpisode(primaryImageBitmap)
+
+					fragment = EpisodeDetailsFragment(episode)
+				}
+				else -> TODO()
+			}
+
+			supportFragmentManager.beginTransaction().add(android.R.id.content, fragment).commit()
+		}
+	}
+
+	private suspend fun getBaseItemDtoForID(id: String) = withContext(Dispatchers.IO) {
+		return@withContext TvApp.getApplication().apiClient.getItem(id)
+	}
+
+	private suspend fun getImageFromURL(url: String) = withContext(Dispatchers.IO) {
+		return@withContext Picasso.with(this@DetailsActivity).load(url).get()
+	}
+}
--- a/app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
@@ -17,6 +17,8 @@ import org.jellyfin.apiclient.model.dto.BaseItemDto
 import org.jellyfin.apiclient.model.dto.BaseItemType
 import org.jellyfin.apiclient.model.dto.ImageOptions
 
+private const val LOG_TAG = "DetailsActivity"
+
 class DetailsActivity : FragmentActivity() {
 	private lateinit var fragment: BaseDetailsFragment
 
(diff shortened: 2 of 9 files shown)


app/src/main/java/org/jellyfin/androidtv/model/itemtypes/Episode.kt
```diff
@@ -0,0 +1,14 @@
...
+import android.graphics.Bitmap
+
+data class Episode (
+	val id: String,
+	val canDelete: Boolean,
+	val canDownload: Boolean,
+	//TODO: Chapters: ArrayList<ChapterInfoDto>
+	val communityRating: Float,
+	val name: String,
+	val overview: String,
```
nielsvanvelzen: I think we should use our own naming and I prefer "description" instead of "overview"

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/Episode.kt
```diff
@@ -0,0 +1,14 @@
+package org.jellyfin.androidtv.model.itemtypes
+
+import android.graphics.Bitmap
+
+data class Episode (
+	val id: String,
+	val canDelete: Boolean,
```
nielsvanvelzen: The properties for "canDelete" and "canDownload" are user-permissions and can't be set per-item or per-library afaik. So I'd remove them from the episode pojo and move them somewhere else

Not sure if you can find this information in the UserDto though

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/Episode.kt
```diff
@@ -0,0 +1,14 @@
...
+	val canDownload: Boolean,
+	//TODO: Chapters: ArrayList<ChapterInfoDto>
+	val communityRating: Float,
```
nielsvanvelzen: Should probably use a double

app/src/main/java/org/jellyfin/androidtv/details/DetailsDescriptionPresenter.kt
```diff
@@ -0,0 +1,15 @@
+package org.jellyfin.androidtv.details
+
+import androidx.leanback.widget.AbstractDetailsDescriptionPresenter
+
+class DetailsDescriptionPresenter(val title: String, val subtitle: String, val body: String) : AbstractDetailsDescriptionPresenter() {
+	override fun onBindDescription(viewHolder: ViewHolder, itemData: Any) {
+		val us = this
+
+		viewHolder.apply {
```
nielsvanvelzen: When using `.also` the `this` scope won't change so you don't need the `val us = this` hack.

app/src/main/java/org/jellyfin/androidtv/util/apiclient/BaseItemDtoExtensions.kt
```diff
@@ -0,0 +1,15 @@
+package org.jellyfin.androidtv.util.apiclient
+
+import android.graphics.Bitmap
+import org.jellyfin.androidtv.model.itemtypes.Episode
+import org.jellyfin.apiclient.model.dto.BaseItemDto
+
+fun BaseItemDto.asEpisode(primaryImage: Bitmap?) : Episode {
```
nielsvanvelzen: I'm now thinking we should probably just use the image urls in those objects for now. The episode class could contain a function to get the bitmap.

Also, episodes should probably use the "thumb" image instead of primary (I might be wrong about this though)

app/src/main/java/org/jellyfin/androidtv/details/BaseDetailsFragment.kt
```diff
@@ -0,0 +1,14 @@
+package org.jellyfin.androidtv.details
+
+import android.os.Bundle
+import androidx.leanback.app.DetailsSupportFragment
+import androidx.leanback.widget.*
+
+open class BaseDetailsFragment : DetailsSupportFragment() {
+	protected lateinit var rowsAdapter: ArrayObjectAdapter
+
+	override fun onCreate(savedInstanceState: Bundle?) {
```
nielsvanvelzen: Empty function, only calls that only calls it's parent could probably be removed

app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.jellyfin.apiclient.model.dto.ImageOptions
+
+class DetailsActivity : FragmentActivity() {
+	private lateinit var fragment: BaseDetailsFragment
+
+	override fun onCreate(savedInstanceState: Bundle?) {
+		super.onCreate(savedInstanceState)
+
+		val id = intent.getStringExtra("id")
+		if (id == null) finish()
```
nielsvanvelzen: Might want to add a log statement here for debugging so you won't find yourself in a 10 hour debug session asking yourself why the activity never shows up

app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
```diff
@@ -0,0 +1,56 @@
...
+
+		BackgroundManager.getInstance(this).attach(window)
```
nielsvanvelzen: I don't think you need the backgroundmanager for the detailsactivity

app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
```diff
@@ -0,0 +1,56 @@
...
+				}
+				else -> TODO()
+			}
+
+			supportFragmentManager.beginTransaction().add(android.R.id.content, fragment).commit()
+		}
+	}
+
+	private suspend fun getBaseItemDtoForID(id: String) = withContext(Dispatchers.IO) {
+		return@withContext TvApp.getApplication().apiClient.getItem(id)
```
nielsvanvelzen: You can remove the `return@withContext` part. Kotlin automatically uses the returned value of the last expression.

app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
```diff
@@ -0,0 +1,56 @@
...
+
+		val id = intent.getStringExtra("id")
+		if (id == null) finish()
+
+		BackgroundManager.getInstance(this).attach(window)
+
+		GlobalScope.launch(Dispatchers.Main) {
+			val baseItem = getBaseItemDtoForID(id) ?: return@launch
+
+			when (baseItem.baseItemType!!) {
```
nielsvanvelzen: You can actually do something like:
```kotlin
fragment = when(){....}
```

As long as each case returns something it should work

app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
```diff
@@ -0,0 +1,59 @@
...
+	private fun buildDetails() {
+
+		Log.i("EpisodeDetailsFragment", data.name)
+
+		val selector = ClassPresenterSelector().apply {
+			// Attach your media item details presenter to the row presenter:
+
+			val detailsDescriptionPresenter = DetailsDescriptionPresenter(title = data.name, subtitle = "TODO", body = data.overview)
+
+			if (data.primaryImage != null) {
```
nielsvanvelzen: I'd put the if statement inline so you don't need to repeat the same code twice:
```kotlin
FullWidthDetailsOverviewRowPresenter(detailsDescriptionPresenter, data.primaryImage?.let { DetailsOverviewLogoPresenter() })
```

app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
```diff
@@ -0,0 +1,59 @@
...
+				FullWidthDetailsOverviewRowPresenter(detailsDescriptionPresenter, DetailsOverviewLogoPresenter()).also {
+					addClassPresenter(DetailsOverviewRow::class.java, it)
+				}
+			} else {
+				FullWidthDetailsOverviewRowPresenter(detailsDescriptionPresenter).also {
```
nielsvanvelzen:
```kotlin
addClassPresenter(DetailsOverviewRow::class.java, FullWidthDetailsOverviewRowPresenter(detailsDescriptionPresenter))
```
Looks better to me

app/src/main/java/org/jellyfin/androidtv/details/BaseDetailsFragment.kt
AndreasGB: Sure, this is a leftover, but might as well stay there because once we have more sub-classes modeled out, we'll probably move stuff up there as well

app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
AndreasGB: Ah, I think I stole that from your next episode PR, might want to add that there as well then :)

app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
AndreasGB: Thanks, will be included in the next commit

app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
AndreasGB: Hmm, but if we want to have a backdrop image in some, wouldn't we need to the BackgroundManager as well?

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/Episode.kt
AndreasGB: Ah, then I'll remove them. Supporting this is probably also pointless for a TV experience (at least at first)

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/Episode.kt
AndreasGB: BaseItemDto has a float though, should I still change it?

app/src/main/java/org/jellyfin/androidtv/details/BaseDetailsFragment.kt
AndreasGB: I removed it now

app/src/main/java/org/jellyfin/androidtv/util/apiclient/BaseItemDtoExtensions.kt
AndreasGB: My new idea for this is now to move the download function for images into BaseDetailsFragment and then let each item-specific class do its own handling.

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/Episode.kt
nielsvanvelzen: Hmm... It doesn't really make sense to store a score as float IMO and I expect the new API to use doubles too so we can convert it for now

app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
nielsvanvelzen: haha good catch, I will do that

app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
nielsvanvelzen: The detailsactivity uses a parallel background thingy iirc

app/src/main/java/org/jellyfin/androidtv/util/apiclient/BaseItemDtoExtensions.kt
nielsvanvelzen: Yeah that would work for now

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/Episode.kt
AndreasGB: Okay, will change it then

app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
AndreasGB: Hmm, then I'll leave it out

PR #275, opened 2020-01-24
Added play and resume actions

AndreasGB:
Basically we now have Play and Resume buttons that work, but the way they work is ugly.

I think we may need to re-think the item types and model them into a class structure as on the server where they have a common inheritance root, before this can become prettier

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
@@ -8,8 +8,13 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.launch
 import org.jellyfin.androidtv.TvApp
+import org.jellyfin.androidtv.details.actions.BaseAction
+import org.jellyfin.androidtv.details.actions.PlayFromBeginningAction
+import org.jellyfin.androidtv.details.actions.ResumeAction
 import org.jellyfin.androidtv.model.itemtypes.Episode
 import org.jellyfin.apiclient.model.dto.ImageOptions
+
+private const val LOG_TAG = "EpisodeDetailsFragment"
 
 class EpisodeDetailsFragment(private val data: Episode) : BaseDetailsFragment() {
 
@@ -33,23 +38,37 @@ class EpisodeDetailsFragment(private val data: Episode) : BaseDetailsFragment() 
 			// Attach your media item details presenter to the row presenter:
 			val detailsDescriptionPresenter = DetailsDescriptionPresenter(title = data.name, subtitle = "TODO", body = data.description)
 
+			val overviewRowPresenter = FullWidthDetailsOverviewRowPresenter(
+				detailsDescriptionPresenter,
+				primaryImageBitmap?.let { DetailsOverviewLogoPresenter() })
 
-			addClassPresenter(DetailsOverviewRow::class.java,
-				FullWidthDetailsOverviewRowPresenter(
-					detailsDescriptionPresenter,
-					primaryImageBitmap?.let { DetailsOverviewLogoPresenter() }))
+			overviewRowPresenter.onActionClickedListener = OnActionClickedListener {
+				val action = it as BaseAction
+				action.onClick()
+			}
+
+			addClassPresenter(DetailsOverviewRow::class.java, overviewRowPresenter)
 
 
 			addClassPresenter(ListRow::class.java, ListRowPresenter())
 		}
 		rowsAdapter = ArrayObjectAdapter(selector)
 
-		val detailsOverview = DetailsOverviewRow("Media Item Details").apply {
-			imageDrawable = BitmapDrawable(resources, primaryImageBitmap)
-			// Add images and action buttons to the details view
-			addAction(Action(1, "Continue from 12:10"))
-			addAction(Action(2, "Play from Beginning"))
+		val actionsAdapter = SparseArrayObjectAdapter()
+		if (data.canResume) actionsAdapter.set(0, ResumeAction(context!!, data.playbackPositionTicks, data.id))
+		actionsAdapter.set(1, PlayFromBeginningAction(context!!, data.id))
+		actionsAdapter.set(2, Action(1, "Set Watched"))
+		actionsAdapter.set(3, Action(1, "Add Favorite"))
+		actionsAdapter.set(4, Action(1, "Add to Queue"))
+		actionsAdapter.set(5, Action(1, "Go to Series"))
+		actionsAdapter.set(6, Action(1, "Delete"))
+
+		val detailsOverview = DetailsOverviewRow("Media Item Details").also {
+			it.imageDrawable = BitmapDrawable(resources, primaryImageBitmap)
+			it.actionsAdapter = actionsAdapter
 		}
+
+
 		rowsAdapter.add(detailsOverview)
 
 		// Add a Related items row
@@ -63,4 +82,6 @@ class EpisodeDetailsFragment(private val data: Episode) : BaseDetailsFragment() 
 
 		adapter = rowsAdapter
 	}
+
+
 }
--- a/app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
@@ -16,7 +16,7 @@ import org.jellyfin.apiclient.model.dto.ImageOptions
 
 private const val LOG_TAG = "EpisodeDetailsFragment"
 
-class EpisodeDetailsFragment(private val data: Episode) : BaseDetailsFragment() {
+class EpisodeDetailsFragment(private val episode: Episode) : BaseDetailsFragment() {
 
 	override fun onCreate(savedInstanceState: Bundle?) {
 		super.onCreate(savedInstanceState)
@@ -29,22 +29,26 @@ class EpisodeDetailsFragment(private val data: Episode) : BaseDetailsFragment() 
 
 	private suspend fun buildDetails() {
 
-		Log.i("EpisodeDetailsFragment", data.name)
+		Log.i("EpisodeDetailsFragment", episode.name)
 
-		val primaryImageUrl = TvApp.getApplication().apiClient.GetImageUrl(data.id, ImageOptions())
+		val primaryImageUrl = TvApp.getApplication().apiClient.GetImageUrl(episode.id, ImageOptions())
 		val primaryImageBitmap = getImageFromURL(primaryImageUrl)
 
 		val selector = ClassPresenterSelector().apply {
 			// Attach your media item details presenter to the row presenter:
-			val detailsDescriptionPresenter = DetailsDescriptionPresenter(title = data.name, subtitle = "TODO", body = data.description)
+			val detailsDescriptionPresenter = DetailsDescriptionPresenter(title = episode.name, subtitle = "TODO", body = episode.description)
 
 			val overviewRowPresenter = FullWidthDetailsOverviewRowPresenter(
 				detailsDescriptionPresenter,
 				primaryImageBitmap?.let { DetailsOverviewLogoPresenter() })
 
 			overviewRowPresenter.onActionClickedListener = OnActionClickedListener {
-				val action = it as BaseAction
-				action.onClick()
+				if (it is BaseAction) {
+					val action = it as BaseAction
+					action.onClick()
+				} else {
+					Log.e(LOG_TAG, "The clicked Action did not derive from BaseAction, this is unsupported!")
+				}
 			}
 
 			addClassPresenter(DetailsOverviewRow::class.java, overviewRowPresenter)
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/jellyfin/androidtv/model/itemtypes/Episode.kt
```diff
@@ -7,5 +7,7 @@ data class Episode (
 	//TODO: Chapters: ArrayList<ChapterInfoDto>
 	val communityRating: Double,
 	val name: String,
-	val description: String
+	val description: String,
+	val canResume: Boolean,
```
nielsvanvelzen: Might be worth splitting the user specific data to a separate class. Something like "Episode" and "UserEpisode". This way we can (in the future) cache episode data in the app when switching users.

Not necessary though

app/src/main/java/org/jellyfin/androidtv/details/actions/ResumeAction.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.jellyfin.androidtv.util.TimeUtils
+import org.jellyfin.androidtv.util.apiclient.PlaybackHelper
+import org.jellyfin.apiclient.interaction.Response
+import org.jellyfin.apiclient.model.dto.BaseItemDto
+import org.jellyfin.apiclient.model.dto.BaseItemType
+
+private const val LOG_TAG = "ResumeAction"
+
+class ResumeAction(context: Context, val playbackPositionTicks: Long, val itemID: String) : PlaybackAction(ActionID.RESUME.id, context) {
+	val actualPlaybackPositionInMillis: Long
```
nielsvanvelzen:
```kotlin
val actualPlaybackPositionInMillis by lazy {
    playbackPositionTicks / 10000 - TvApp.getApplication().resumePreroll
}
```

The "by" syntax will move the setter & getter to a separate implementation. The "lazy" function will run the expression when the value of the property is first requested.

This would replace the set in the "init" function

Reference: https://kotlinlang.org/docs/reference/delegated-properties.html

app/src/main/java/org/jellyfin/androidtv/details/actions/ResumeAction.kt
```diff
@@ -0,0 +1,38 @@
...
+
+	init {
+		this.actualPlaybackPositionInMillis = playbackPositionTicks / 10000 - TvApp.getApplication().resumePreroll
+		this.label1 = context.getString(R.string.lbl_resume_from, TimeUtils.formatMillis(actualPlaybackPositionInMillis))
+	}
+
+	override fun onClick() {
```
nielsvanvelzen:
```kotlin
override fun onClick() = GlobalScope.launch(Dispatchers.Main) {
	Log.i(LOG_TAG, "Resume Clicked!")
	playItemWithID(itemID, actualPlaybackPositionInMillis, false)
}
```

app/src/main/java/org/jellyfin/androidtv/details/actions/PlaybackAction.kt
```diff
@@ -0,0 +1,45 @@
...
+		val baseItem = TvApp.getApplication().apiClient.getItem(id)
+		if (baseItem == null) {
+			Log.e(LOG_TAG, "Failed to get a base item for the given ID")
+			return@withContext
+		}
+		play(baseItem, pos, shuffle)
+	}
+
+	private fun play(item: BaseItemDto, pos: Long, shuffle: Boolean) {
+		PlaybackHelper.getItemsToPlay(item, pos == 0L && item.baseItemType == BaseItemType.Movie, shuffle, object : Response<List<BaseItemDto>>() {
```
nielsvanvelzen: Should probably add an extension function to PlaybackHelper for coroutines instead of callback here (just like for the API).

app/src/main/java/org/jellyfin/androidtv/details/actions/PlayFromBeginningAction.kt
```diff
@@ -0,0 +1,23 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import org.jellyfin.androidtv.R
+
+private const val LOG_TAG = "PlayFromBeginningAction"
+
+class PlayFromBeginningAction(context: Context, val itemID: String) : PlaybackAction(ActionID.PLAY_FROM_BEGINNING.id, context) {
+	init {
+	    label1 = context.getString(R.string.lbl_play)
```
nielsvanvelzen: Why is this called `label` is that a google thingy or our thingy?

app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
```diff
@@ -33,23 +38,37 @@ class EpisodeDetailsFragment(private val data: Episode) : BaseDetailsFragment()
 			// Attach your media item details presenter to the row presenter:
 			val detailsDescriptionPresenter = DetailsDescriptionPresenter(title = data.name, subtitle = "TODO", body = data.description)
 
+			val overviewRowPresenter = FullWidthDetailsOverviewRowPresenter(
+				detailsDescriptionPresenter,
+				primaryImageBitmap?.let { DetailsOverviewLogoPresenter() })
+
+			overviewRowPresenter.onActionClickedListener = OnActionClickedListener {
+				val action = it as BaseAction
```
nielsvanvelzen: Maybe add a typecheck if it actually is a BaseAction (just to be sure)

app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
```diff
@@ -8,9 +8,14 @@ import kotlinx.coroutines.Dispatchers
...
 import org.jellyfin.androidtv.TvApp
+import org.jellyfin.androidtv.details.actions.BaseAction
+import org.jellyfin.androidtv.details.actions.PlayFromBeginningAction
+import org.jellyfin.androidtv.details.actions.ResumeAction
 import org.jellyfin.androidtv.model.itemtypes.Episode
 import org.jellyfin.apiclient.model.dto.ImageOptions
 
+private const val LOG_TAG = "EpisodeDetailsFragment"
+
 class EpisodeDetailsFragment(private val data: Episode) : BaseDetailsFragment() {
```
nielsvanvelzen: I'm now thinking it might be nicer to use "episode" instead of "data". Looks more clear when using the actual data.

app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
```diff
@@ -33,23 +38,37 @@ class EpisodeDetailsFragment(private val data: Episode) : BaseDetailsFragment()
...
 			addClassPresenter(ListRow::class.java, ListRowPresenter())
 		}
 		rowsAdapter = ArrayObjectAdapter(selector)
 
-		val detailsOverview = DetailsOverviewRow("Media Item Details").apply {
-			imageDrawable = BitmapDrawable(resources, primaryImageBitmap)
-			// Add images and action buttons to the details view
-			addAction(Action(1, "Continue from 12:10"))
-			addAction(Action(2, "Play from Beginning"))
+		val actionsAdapter = SparseArrayObjectAdapter()
```
nielsvanvelzen: I think an ArrayObjectAdapter works better so we don't need to add the integers for ordering (just make sure everything is added in the same place)

nielsvanvelzen: Also, using the .apply or .also function is possible here too

app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
```diff
@@ -33,23 +38,37 @@ class EpisodeDetailsFragment(private val data: Episode) : BaseDetailsFragment()
...
+		if (data.canResume) actionsAdapter.set(0, ResumeAction(context!!, data.playbackPositionTicks, data.id))
+		actionsAdapter.set(1, PlayFromBeginningAction(context!!, data.id))
+		actionsAdapter.set(2, Action(1, "Set Watched"))
+		actionsAdapter.set(3, Action(1, "Add Favorite"))
+		actionsAdapter.set(4, Action(1, "Add to Queue"))
+		actionsAdapter.set(5, Action(1, "Go to Series"))
+		actionsAdapter.set(6, Action(1, "Delete"))
+
+		val detailsOverview = DetailsOverviewRow("Media Item Details").also {
```
nielsvanvelzen: Btw, I'm ignoring the hardcoded strings for now but they need to be addressed at some point

AndreasGB: Sure, in the real Actions I'm resolving the strings properly

app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
AndreasGB: Will do

app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
AndreasGB: Will do

app/src/main/java/org/jellyfin/androidtv/details/actions/PlayFromBeginningAction.kt
AndreasGB: That's a [Google thing](https://developer.android.com/reference/androidx/leanback/widget/Action) sadly, I dislike the name as well.

app/src/main/java/org/jellyfin/androidtv/details/actions/ResumeAction.kt
AndreasGB: Hmm, like this it complains that the Return Type is a Job, so it seems like we would also need to adjust the base class. Shall we do this?

app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
AndreasGB: The documentation (recommended it)[https://developer.android.com/reference/android/support/v17/leanback/widget/DetailsOverviewRow], but for now it doesn't make much of a difference. Should I still change it to ArrayObjectAdapter?

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/Episode.kt
AndreasGB: Right, we need to very much rethink the whole model though, should we do this here, via Issue or Matrix?

app/src/main/java/org/jellyfin/androidtv/details/actions/ResumeAction.kt
AndreasGB: Ah, nice. Done.

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/Episode.kt
nielsvanvelzen: Probably in the issue or a new PR

app/src/main/java/org/jellyfin/androidtv/details/actions/ResumeAction.kt
nielsvanvelzen: Oh in that case we can do 2 things:

End the expression with `Unit` (the void type of Kotlin) or don't change it.

app/src/main/java/org/jellyfin/androidtv/details/actions/PlayFromBeginningAction.kt
nielsvanvelzen: The should've called it "title" and "subtitle" or something. We can't do anything about it though

app/src/main/java/org/jellyfin/androidtv/details/actions/ResumeAction.kt
AndreasGB: Okay, thanks. Will leave it.

app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
nielsvanvelzen: I think the docs recommend it because they are expecting the actions to be added/removed from various places. But we're doing it in a single place so a normal adapter should work fine.

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/Episode.kt
AndreasGB: Okay, let's move this to a new PR

app/src/main/java/org/jellyfin/androidtv/util/apiclient/PlaybackHelperExtensions.kt
```diff
@@ -0,0 +1,21 @@
+package org.jellyfin.androidtv.util.apiclient
```
nielsvanvelzen: Should probably move this out of the apiclient package

app/src/main/java/org/jellyfin/androidtv/details/EpisodeDetailsFragment.kt
```diff
@@ -24,32 +29,53 @@ class EpisodeDetailsFragment(private val data: Episode) : BaseDetailsFragment()
...
 		}
 		rowsAdapter = ArrayObjectAdapter(selector)
 
-		val detailsOverview = DetailsOverviewRow("Media Item Details").apply {
-			imageDrawable = BitmapDrawable(resources, primaryImageBitmap)
-			// Add images and action buttons to the details view
-			addAction(Action(1, "Continue from 12:10"))
-			addAction(Action(2, "Play from Beginning"))
+		val actionsAdapter = SparseArrayObjectAdapter()
+		actionsAdapter.apply {
```
nielsvanvelzen: .apply could be added directly after creating the instance.

`val actionsAdapter = ArrayObjectAdapter().apply{ /*code */ }`

AndreasGB: Okay, will change it and remove the SparseArrayObjectAdapter

PR #276, opened 2020-01-25
Introduce new Items Model

AndreasGB:
This is my idea of how we could better handle different item types, currently there's only Episode implemented, but the others would be done similarly.


app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
```diff
@@ -33,14 +29,13 @@ class DetailsActivity : FragmentActivity() {
...
-				BaseItemType.Episode -> {
-					val episode = baseItem.asEpisode()
-					EpisodeDetailsFragment(episode)
+			lifted.acceptVisitor(object : IBaseItemVisitor {
+				override fun visit(item: Episode) {
+					fragment = EpisodeDetailsFragment(item)
 				}
-				else -> TODO()
-			}
+			})
```
nielsvanvelzen: I would prefer it like this, or is there a reason the visitor pattern would be better in this case?
```kotlin
fragment = when(lifted) {
	is Episode -> EpisodeDetailsFragment(lifted)
	else -> TODO()
}
```

AndreasGB: It forces you to notice where you missed some changes when a new Item type is introduced, as the interface changes and you have to adjust it everywhere.

AndreasGB:
Okay, PlayableItem refactored out

app/src/main/java/org/jellyfin/androidtv/details/DetailsActivity.kt
nielsvanvelzen: Did some quick searching and it seems like using sealed classes is a way to do it "the kotlin way". It would look basically the same as my example except for the fact that there is no "else" block.

- https://kotlinlang.org/docs/reference/sealed-classes.html
- https://stackoverflow.com/questions/33602705/best-way-to-implement-visitor-pattern-in-kotlin

PR #278, opened 2020-01-25
Movies details

nielsvanvelzen:
# Still a draft but I made a mistake, don't review yet :)

Goal is to create a reference implementation for details by creating the movie details page.

Also changes some other things


nielsvanvelzen:
Merging this early so @AndreasGB won't have to deal with conflicts

PR #279, opened 2020-01-27
Remove removed strings

thornbill:
Since we can't seem to do this with Weblate. I think the only way forward is to remove these strings and hope the changes get pulled into Weblate before any further edits are made.

where


nielsvanvelzen:
Weblate didn't update so I expect this can still be merged without problems.

PR #280, opened 2020-01-29
Implement a toggle watched action

AndreasGB:
There *could* be a crash right now if the user marks an episode played and then immediately quits the activity. Question is, do we consider this likely enough to go for `LiveData` and the likes or consider this fine for the MVP?


nielsvanvelzen:
I think we should split the "buildDetails" function in two parts:
 - Constructing the layout
 - Setting values

Next we should add those functions in the "BaseDetailsFragment" and send the fragment instance to the action. The action could then call the "set values" method to force an update for the buttons.

AndreasGB:
Since this would also mean bigger changes to the base fragment, should I wait until your #278 is merged or could you merge your partial state?

AndreasGB:
Closing in favor of #285

PR #285, opened 2020-02-03
Details refactor

nielsvanvelzen:
I did quite a lot on this branch.. whoops

Changed include:
- Observer implementation for item classes
- Better looking code for the detail fragments
- Custom action adapter
  - Supports resetting all items
  - Shows icons above text
  - Adds a new "ToggleAction" class that will be colored on screen when active
- Added new icons 
  - Replace bitmap resume with vector icon (custom made)
  - Add menu vector icon

I probably broke things too!


AndreasGB:
I'll have a look at it and will very probably then #280 then :D

nielsvanvelzen:
Yeah sorry about that! I included your changes in this PR (with some changes).

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/LiftedItems.kt
```diff
@@ -4,20 +4,23 @@ import org.jellyfin.apiclient.model.dto.BaseItemDto
...
 	val description: String? = original.overview
 	val images: ImageCollection = ImageCollection(original)
 	val added: Date = original.dateCreated
+	var favorite: Boolean by Delegates.observable(original.userData.isFavorite, ::observer)
 }
 
 sealed class PlayableItem(original: BaseItemDto) : BaseItem(original) {
-	val canResume: Boolean = original.canResume
 	val playbackPositionTicks: Long = original.userData.playbackPositionTicks
+	val canResume: Boolean = original.canResume && playbackPositionTicks > 0
```
AndreasGB: I think the `playbackPositionTicks > 0` check should be sufficient, since the `original.canResume` just does a basically equivalent check.

```java
	public final boolean getCanResume()
	{
		return getUserData() != null && getUserData().getPlaybackPositionTicks() > 0;
	}
```


https://github.com/jellyfin/jellyfin-apiclient-java/blob/master/library/src/main/java/org/jellyfin/apiclient/model/dto/BaseItemDto.java#L2311

app/src/main/java/org/jellyfin/androidtv/util/apiclient/ApiClientExtensions.kt
```diff
@@ -29,3 +30,25 @@ suspend fun ApiClient.getItem(id: String): BaseItemDto? = suspendCoroutine { con
...
+		override fun onResponse(response: UserItemDataDto?) { continuation.resume(response!!) }
+		override fun onError(exception: Exception?) { continuation.resume(null) }
+	})
+}
+
+
+suspend fun ApiClient.updateFavoriteStatus(itemId: String, userId: String, isFavorite: Boolean) : UserItemDataDto? = suspendCoroutine { continuation ->
+	UpdateFavoriteStatusAsync(itemId, userId, isFavorite, object : Response<UserItemDataDto>() {
+		override fun onResponse(response: UserItemDataDto?) { continuation.resume(response!!) }
+		override fun onError(exception: Exception?) { continuation.resume(null) }
```
AndreasGB: Maybe we should add proper error handling for all these coroutines before we release

app/src/main/java/org/jellyfin/androidtv/details/actions/ToggleWatchedAction.kt
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	override fun onClick() {
+		GlobalScope.launch(Dispatchers.Main) {
+			val apiClient = TvApp.getApplication().apiClient
+
+			if (item.played) apiClient.markUnplayed(item.id, TvApp.getApplication().currentUser.id)
+			else apiClient.markPlayed(item.id, TvApp.getApplication().currentUser.id, null)
+
+			item.played = !item.played
```
AndreasGB: I missed this in my review, we probably should overwrite everything in the item with information that the server sends us in response to marking played / unplayed.

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/LiftedItems.kt
nielsvanvelzen: I didn't even look at the baseItem. I assumed it was just passed from the server!

app/src/main/java/org/jellyfin/androidtv/util/apiclient/ApiClientExtensions.kt
nielsvanvelzen: I'll make a note in the issue for that.

PR #295, opened 2020-02-05
Implement Cast display for Movies

AndreasGB:
Basically this is very much not yet done, but more to verify whether this is an acceptable path to handling item selection, or whether we want to go with a more complex scheme like a custom row that handles its own clicks or each item on its own.

Also, I experimented with square cards and it already did not look good as some actor pictures are off-center, so circle crop would definitely ruin them. Will need to stay in portrait mode.

Another question is whether we would want to use our own layout or not, because to me the default `ImageCardView` seems fine for now.

This is what it looks like currently
device-2020-02-05-205625

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/base/ClickablePresenter.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/base/ClickablePresenter.kt
@@ -1,7 +1,0 @@
-package org.jellyfin.androidtv.base
-
-import androidx.leanback.widget.Presenter
-
-abstract class ClickablePresenter : Presenter() {
-	abstract fun onItemClicked(item: Any?)
-}
--- a/app/src/main/java/org/jellyfin/androidtv/base/ClickablePresenter.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/base/ClickablePresenter.kt
@@ -1,0 +1,7 @@
+package org.jellyfin.androidtv.base
+
+import androidx.leanback.widget.Presenter
+
+abstract class ClickablePresenter : Presenter() {
+	abstract fun onItemClicked(item: Any?)
+}
--- a/app/src/main/java/org/jellyfin/androidtv/base/IItemClickListener.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/base/IItemClickListener.kt
@@ -1,7 +1,5 @@
 package org.jellyfin.androidtv.base
 
-import androidx.leanback.widget.Presenter
-
 interface IItemClickListener {
 	fun onItemClicked(item: Any?)
 }
--- a/app/src/main/java/org/jellyfin/androidtv/base/IItemClickListener.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/base/IItemClickListener.kt
@@ -1,0 +1,7 @@
+package org.jellyfin.androidtv.base
+
+import androidx.leanback.widget.Presenter
+
+interface IItemClickListener {
+	fun onItemClicked(item: Any?)
+}
--- a/app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
@@ -1,16 +1,22 @@
 package org.jellyfin.androidtv.details
 
+import android.content.Intent
+import android.util.Log
 import androidx.leanback.widget.*
 import org.jellyfin.androidtv.R
 import org.jellyfin.androidtv.details.actions.*
 import org.jellyfin.androidtv.model.itemtypes.Movie
+import org.jellyfin.androidtv.model.itemtypes.Person
 import org.jellyfin.androidtv.presentation.InfoCardPresenter
 
-class MovieDetailsFragment(item: Movie) : BaseDetailsFragment<Movie>(item) {
+private const val LOG_TAG = "MovieDetailsFragment"
+
+class MovieDetailsFragment(item: Movie) : BaseDetailsFragment<Movie>(item), OnItemViewClickedListener {
+
 	private val detailsRow by lazy { DetailsOverviewRow(Unit).apply { actionsAdapter = ActionAdapter() } }
 //	private val chaptersRow by lazy { Row() }
 //	private val staffRow by lazy { Row() }
-//	private val charactersRow by lazy { Row() }
+	private val charactersRow by lazy { ListRow(HeaderItem("Cast/Crew"), ArrayObjectAdapter(PersonPresenter())) }
 //	private val relatedRow by lazy { Row() }
 	private val mediaInfoRow by lazy { ListRow(HeaderItem("Media info"), ArrayObjectAdapter(InfoCardPresenter())) }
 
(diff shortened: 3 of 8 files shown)


app/src/main/java/org/jellyfin/androidtv/model/itemtypes/LiftedItems.kt
```diff
@@ -36,4 +39,14 @@ class Episode(original: BaseItemDto) : PlayableItem(original) {
 	}
 }
 
-class Movie(original: BaseItemDto) : PlayableItem(original)
+class Person(original: BaseItemPerson) {
```
nielsvanvelzen: Can be moved to a seperate file as it doesn't use our base class (yet?)

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/LiftedItems.kt
```diff
@@ -36,4 +39,14 @@ class Episode(original: BaseItemDto) : PlayableItem(original) {
...
+	val id: String = original.id
+	val name: String = original.name
+	val role: String = original.role
+	val type: PersonType = original.personType
+	val primaryImage: ImageCollection.Image? = original.primaryImageTag?.let { ImageCollection.Image(original.id, ImageType.Primary, it) }
```
nielsvanvelzen: Don't persons support multiple images? In that case we could use the imagecollection

AndreasGB: I doesn't seem like they do, Persons are weirdly different from all other items. (https://github.com/jellyfin/jellyfin-apiclient-java/blob/36cd46c6682fd9506385f87510e182fa332e5210/library/src/main/java/org/jellyfin/apiclient/model/dto/BaseItemPerson.java)

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/LiftedItems.kt
AndreasGB: Oh, forgot to ask that. I wonder if persons should remain separate or become an item as well. Currently they are their entirely own thing in the apiclient

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/LiftedItems.kt
nielsvanvelzen: I looked at the network responses in the web version and it looks like the "BaseItemPerson" is not actually the base type "person". It's an object used with minimal data showing the relation between a series/movie and a person.

We should probably rename it

AndreasGB: Oh whoops. `FullDetailsActivity` uses it to display this row, so I assumed it is the only representation of persons :D Will do the rename later.

app/src/main/java/org/jellyfin/androidtv/details/PersonPresenter.kt
```diff
@@ -1,28 +1,42 @@
...
+
+		context.startActivity(intent)
+	}
+
 	override fun onCreateViewHolder(parent: ViewGroup?): ViewHolder {
-		return ViewHolder(ImageCardView(parent!!.context))
+		return ViewHolder(ImageCardView(ContextThemeWrapper(parent!!.context, R.style.MarqueeImageCardViewTheme)))
 	}
 
 	override fun onBindViewHolder(viewHolder: ViewHolder?, item: Any?) {
```
nielsvanvelzen: `override fun onBindViewHolder(viewHolder: ViewHolder, item: Any) {`
should work, so you won't need the `requireNotNull` calls

app/src/main/java/org/jellyfin/androidtv/base/ClickablePresenter.kt
```diff
@@ -0,0 +1,7 @@
+package org.jellyfin.androidtv.base
+
+import androidx.leanback.widget.Presenter
+
+abstract class ClickablePresenter : Presenter() {
```
nielsvanvelzen: Interface might be better if it only defines the "clickability"

app/src/main/java/org/jellyfin/androidtv/base/ClickablePresenter.kt
```diff
@@ -0,0 +1,7 @@
...
+	abstract fun onItemClicked(item: Any?)
```
nielsvanvelzen: Might be worth using a generic on the class/interface so we don't need to use the `any` type

app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
```diff
@@ -1,22 +1,19 @@
...
 private const val LOG_TAG = "MovieDetailsFragment"
 
-class MovieDetailsFragment(item: Movie) : BaseDetailsFragment<Movie>(item), OnItemViewClickedListener {
+class MovieDetailsFragment(item: Movie) : BaseDetailsFragment<Movie>(item) {
 
 	private val detailsRow by lazy { DetailsOverviewRow(Unit).apply { actionsAdapter = ActionAdapter() } }
 //	private val chaptersRow by lazy { Row() }
 //	private val staffRow by lazy { Row() }
-	private val charactersRow by lazy { ListRow(HeaderItem("Cast/Crew"), ArrayObjectAdapter(PersonPresenter())) }
+	private val charactersRow by lazy { ListRow(HeaderItem("Cast/Crew"), ArrayObjectAdapter(PersonPresenter(this.activity!!))) }
```
nielsvanvelzen: Use `this.context` instead of `this.activity`

app/src/main/java/org/jellyfin/androidtv/base/ClickablePresenter.kt
AndreasGB: Doesn't usage of generics yield completely different classes? In C# at least it does iirc. If so that mean as a consequence, that we can no longer call the `onItemClicked` method on a common parent type, no?

nielsvanvelzen: I don't know how C# handles and have no idea if it actually is possible. If it does it will make the code more typesafe as I don't like having to cast "Any" to the objects we expect. Changes are it will introduce issues later.

AndreasGB: While I agree that it would be nice, I currently don't see a way to make it happen. I think somewhere a down cast has to happen, as we are given an Any for item. Maybe I'm too sleepy to see it right now though. Can you please have a look tomorrow as well?

AndreasGB:
Hm, seems like getting dependencies failed for the CI build?

nielsvanvelzen:
CI seems fine (but I suspect you triggered a new build)

app/src/main/java/org/jellyfin/androidtv/base/IItemClickListener.kt
```diff
@@ -0,0 +1,7 @@
+package org.jellyfin.androidtv.base
+
+import androidx.leanback.widget.Presenter
```
nielsvanvelzen: We don't need this import. When using Android Studio you can have it fix your imports when creating commits. (option for this is in the commit dialog)

AndreasGB: Okay, setting that now for the future, and removing this import. Thanks :)

PR #301, opened 2020-02-05
Introduce new config version which when applied enables the acra prompt

AndreasGB:
This introduces a new config version which explicitly sets acra into the prompt mode once it is applied. 

**Risks**
* Users that have already manually disabled ACRA reporting have it re-enabled


AndreasGB:
Closing as this might actually only be a UI problem

PR #306, opened 2020-02-06
Partial fix for subtitles

### LINKED ISSUE #290: Double subtitle in ExoPlayer at v0.11.0
I just tested v0.11.0 on my Android TV, and found double subtitle issue if ExoPlayer is seleted.
 
subtitle_issue

nielsvanvelzen:
Fixes #290 

This PR will stop exoplayer from choosing subtitles on it's own. This fixes the double subtitles or subtitles when none are requested when starting a video.

Changing the subtitle track will still cause issues.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/playback/VideoManager.java
@@ -1,21 +1,27 @@
 package org.jellyfin.androidtv.playback;
 
 import android.app.Activity;
+import android.content.Context;
 import android.content.res.Configuration;
 import android.graphics.PixelFormat;
 import android.net.Uri;
 import android.os.Handler;
+import android.os.Looper;
 import android.view.SurfaceHolder;
 import android.view.SurfaceView;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.FrameLayout;
 
+import com.google.android.exoplayer2.DefaultRenderersFactory;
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
According to @AndreasGB the bug was not fixed for him yet. Need to do some more debugging I guess...

AndreasGB:
I may be misunderstanding documentation here, but it seems like this is an incorrect way to disable the subtitles. First, the constant [`C.TRACK_TYPE_VIDEO`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/C.html#TRACK_TYPE_VIDEO) which happens to be 2 seems to be misused here, as it is used to indicate Type, but used as an index into the array of renderers by [`setRendererDisabled(...)`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.ParametersBuilder.html#setRendererDisabled-int-boolean-)

For the one testfile, this seems to lign up correctly, but I think this is relying on coincidence that there are not more renderers by chance.

The docs on the [`DefaultTrackSelector`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/trackselection/DefaultTrackSelector.html) recommend to be able to select specific tasks to listen to player events and then adjust selection by defining an override with `null` it seems. See `Selecting sepecific tracks` and `Disabling renderers`.

thornbill:
I squashed all the commits to make this easier to cherry pick for the release. Going to push this through now. :rocket:

PR #314, opened 2020-02-07
Only enable ACRA in release builds

AndreasGB:
I just had ACRA run in a debug build, this check is supposed to only enable it in non-debug builds.


thornbill:
Why would we not also want crashes from debug builds? I find this can be helpful. My main tv is always running a debug build and the crashes are handled separately in tracepot.

nielsvanvelzen:
When running in debug the user can disable it anyway so I'd say leave it as it is. Just like thornbill I run my self built apk in debug mode.

AndreasGB:
Okay, closing then :)

PR #315, opened 2020-02-07
Fix broken logger calls

AndreasGB:
So, turns out that there were a *lot* of incorrect usages of the `ILogger` interface that may lead to crashes due to format string injections via them.


AndreasGB:
Hmm, apparently there is a conflict, don't review yet...

PR #321, opened 2020-02-08
Chapter cards

AndreasGB:
Basically very similar to the Persons

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ChapterInfo.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ChapterInfo.kt
@@ -1,0 +1,10 @@
+package org.jellyfin.androidtv.model.itemtypes
+
+import org.jellyfin.apiclient.model.dto.ChapterInfoDto
+import org.jellyfin.apiclient.model.entities.ImageType
+
+class ChapterInfo(originalChapterInfoDto: ChapterInfoDto, val baseItem: BaseItem, index: Int) {
+	val startPositionTicks: Long = originalChapterInfoDto.startPositionTicks / 10000
+	val name: String = originalChapterInfoDto.name
+	val image: ImageCollection.Image? = originalChapterInfoDto.imageTag?.let { ImageCollection.Image(baseItem.id, ImageType.Chapter, it, index) }
+}
--- a/app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ImageCollection.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ImageCollection.kt
@@ -21,7 +21,7 @@ class ImageCollection(original: BaseItemDto) {
 			TvApp.getApplication().apiClient.GetImageUrl(itemId, ImageOptions().also {
 				it.imageType = type
 				it.tag = tag
-				if (index != null) it.imageIndex = index
+				it.imageIndex = index
 			})
 		}
 
--- a/app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ImageCollection.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ImageCollection.kt
@@ -15,25 +15,14 @@ class ImageCollection(original: BaseItemDto) {
 	val logo = original.imageTags[ImageType.Logo]?.let { Image(original.id, ImageType.Logo, it) }
 	val backdrops = original.backdropImageTags.map { Image(original.id, ImageType.Backdrop, it) }.toList()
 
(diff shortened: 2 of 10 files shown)


app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ChapterInfo.kt
```diff
@@ -0,0 +1,10 @@
+package org.jellyfin.androidtv.model.itemtypes
+
+import org.jellyfin.apiclient.model.dto.ChapterInfoDto
+import org.jellyfin.apiclient.model.entities.ImageType
+
+class ChapterInfo(originalChapterInfoDto: ChapterInfoDto, val baseItem: BaseItem, index: Int) {
```
nielsvanvelzen: Can't you get the image from the data inside ChapterInfoDto?

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ImageCollection.kt
```diff
@@ -15,14 +16,28 @@ class ImageCollection(original: BaseItemDto) {
 	val backdrops = original.backdropImageTags.map { Image(original.id, ImageType.Backdrop, it) }.toList()
 
 	class Image(private val itemId: String, private val type: ImageType, private val tag: String) {
```
nielsvanvelzen: Add the `index` option as nullable to the constructor

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ImageCollection.kt
```diff
@@ -15,14 +16,28 @@ class ImageCollection(original: BaseItemDto) {
...
+			if (type == ImageType.Chapter)
+				TvApp.getApplication().apiClient.GetImageUrl(itemId, ImageOptions().also {
+					it.imageType = type
+					it.tag = tag
+					it.imageIndex = index
+				})
+			else
 			TvApp.getApplication().apiClient.GetImageUrl(itemId, ImageOptions().also {
 				it.imageType = type
 				it.tag = tag
```
nielsvanvelzen: Just add the index here:
`if(imageIndex != null) it.index = imageIndex`

Might even be possible to do it without  the if statement (I think null is the default value)

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ImageCollection.kt
```diff
@@ -15,14 +16,28 @@ class ImageCollection(original: BaseItemDto) {
...
 			})
 		}
 
+		constructor(itemId: String, type: ImageType, tag: String, index: Int) : this(itemId, type, tag) {
```
nielsvanvelzen: Remove this constructor in favor of the other one I mentioned

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ChapterInfo.kt
AndreasGB: Sadly not. A ChapterInfoDto doesn't contain the ID of the item it belongs to, and the index is also required by the server, inquiring with tag is not enough :/

nielsvanvelzen: Oh that's unfortunate. Can't do anything about it though :(

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ImageCollection.kt
```diff
@@ -14,15 +15,18 @@ class ImageCollection(original: BaseItemDto) {
...
 	val backdrops = original.backdropImageTags.map { Image(original.id, ImageType.Backdrop, it) }.toList()
 
-	class Image(private val itemId: String, private val type: ImageType, private val tag: String) {
+	class Image(private val itemId: String, private val type: ImageType, private val tag: String, private val index: Int? = null) {
+
 		val url: String by lazy {
 			TvApp.getApplication().apiClient.GetImageUrl(itemId, ImageOptions().also {
 				it.imageType = type
 				it.tag = tag
+				if (index != null) it.imageIndex = index
```
nielsvanvelzen: You can remove the if statement as the default value if `imageIndex` is null

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/ChapterInfo.kt
AndreasGB: Yeh, bummed me out as well that this old API is sabotaging us here. At least it's clean on the lifted ones

PR #326, opened 2020-02-09
Implementation of Similar Items row

AndreasGB:
We probably need new Images as fallback though, the current ones I'm referencing here are super ugly :D

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/details/ItemPosterPresenter.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/ItemPosterPresenter.kt
@@ -1,0 +1,59 @@
+package org.jellyfin.androidtv.details
+
+import android.content.Context
+import android.content.Intent
+import android.graphics.drawable.BitmapDrawable
+import android.view.ViewGroup
+import androidx.appcompat.view.ContextThemeWrapper
+import androidx.leanback.widget.ImageCardView
+import androidx.leanback.widget.Presenter
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import org.jellyfin.androidtv.R
+import org.jellyfin.androidtv.TvApp
+import org.jellyfin.androidtv.base.IItemClickListener
+import org.jellyfin.androidtv.model.itemtypes.BaseItem
+
+private const val LOG_TAG = "ItemPosterPresenter"
+
+class ItemPosterPresenter(private val context: Context) : Presenter(), IItemClickListener {
+
+	override suspend fun onItemClicked(item: Any?) {
+		requireNotNull(item)
+		val baseItem = item as BaseItem
+		val intent = Intent(context, DetailsActivity::class.java)
+		intent.putExtra("id", baseItem.id)
+
+		context.startActivity(intent)
+	}
+
+	override fun onCreateViewHolder(parent: ViewGroup?): ViewHolder {
+		return ViewHolder(ImageCardView(ContextThemeWrapper(parent!!.context, R.style.MarqueeImageCardViewTheme)))
+	}
+
+	override fun onBindViewHolder(viewHolder: ViewHolder, item: Any) {
+		val baseItem = item as BaseItem
+		val cardView = viewHolder.view as ImageCardView
+
+		cardView.titleText = baseItem.name
+		cardView.contentText = baseItem.description
+		cardView.isFocusable = true
+		cardView.isFocusableInTouchMode = true
+		cardView.setMainImageDimensions(200, 300)
+		cardView.mainImage = TvApp.getApplication().getDrawableCompat(R.drawable.tile_port_video)
+
+		if (baseItem.images.primary != null) {
+			GlobalScope.launch(Dispatchers.Main) {
+				cardView.mainImage = BitmapDrawable(baseItem.images.primary.getBitmap(TvApp.getApplication()))
+			}
+		}
+
+	}
+
+	override fun onUnbindViewHolder(viewHolder: ViewHolder?) {
+		val cardView = viewHolder!!.view as ImageCardView
+
+		// TODO: Somehow release BitmapDrawable?
+	}
+}
--- a/app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
@@ -2,9 +2,11 @@ package org.jellyfin.androidtv.details
 
 import androidx.leanback.widget.*
 import org.jellyfin.androidtv.R
+import org.jellyfin.androidtv.TvApp
 import org.jellyfin.androidtv.details.actions.*
 import org.jellyfin.androidtv.model.itemtypes.Movie
 import org.jellyfin.androidtv.presentation.InfoCardPresenter
+import org.jellyfin.androidtv.util.apiclient.getSimilarItems
 
 private const val LOG_TAG = "MovieDetailsFragment"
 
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/jellyfin/androidtv/model/itemtypes/LiftedItems.kt
```diff
@@ -2,9 +2,12 @@ package org.jellyfin.androidtv.model.itemtypes
 
 import org.jellyfin.apiclient.model.dto.BaseItemDto
 import org.jellyfin.apiclient.model.dto.BaseItemType
+import org.jellyfin.apiclient.model.querying.ItemFields
 import java.util.*
 import kotlin.properties.Delegates
 
+val FieldsRequiredForLift = arrayOf(ItemFields.DateCreated, ItemFields.MediaSources, ItemFields.MediaStreams, ItemFields.People)
```
nielsvanvelzen: I'd hardcode this in the function that uses it

app/src/main/java/org/jellyfin/androidtv/util/apiclient/ApiClientExtensions.kt
```diff
@@ -45,10 +48,26 @@ suspend fun ApiClient.markUnplayed(itemId: String, userId: String) : UserItemDat
...
 		override fun onResponse(response: UserItemDataDto?) { continuation.resume(response!!) }
 		override fun onError(exception: Exception?) { continuation.resume(null) }
 	})
 }
+
+suspend fun ApiClient.getSimilarItems(item: BaseItem): List<BaseItem>? = suspendCoroutine {continuation ->
+	val query = SimilarItemsQuery()
+	query.setUserId(TvApp.getApplication().currentUser.id)
+	query.setId(item.id)
+	query.setLimit(10)
```
nielsvanvelzen: I'd add the limit as parameter with 10 as default value

Maybe increase it to 25 or something too

app/src/main/java/org/jellyfin/androidtv/util/apiclient/ApiClientExtensions.kt
```diff
@@ -45,10 +48,26 @@ suspend fun ApiClient.markUnplayed(itemId: String, userId: String) : UserItemDat
...
-
 suspend fun ApiClient.updateFavoriteStatus(itemId: String, userId: String, isFavorite: Boolean) : UserItemDataDto? = suspendCoroutine { continuation ->
 	UpdateFavoriteStatusAsync(itemId, userId, isFavorite, object : Response<UserItemDataDto>() {
 		override fun onResponse(response: UserItemDataDto?) { continuation.resume(response!!) }
 		override fun onError(exception: Exception?) { continuation.resume(null) }
 	})
 }
+
+suspend fun ApiClient.getSimilarItems(item: BaseItem): List<BaseItem>? = suspendCoroutine {continuation ->
+	val query = SimilarItemsQuery()
```
nielsvanvelzen: You can use .apply or .also here

app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
```diff
@@ -66,6 +68,14 @@ class MovieDetailsFragment(item: Movie) : BaseDetailsFragment<Movie>(item) {
 			item.cast.forEach(it::add)
 		}
 
+		val similarItems = TvApp.getApplication().apiClient.getSimilarItems(item)
+		val similarMovies = similarItems?.filter { it is Movie }
```
nielsvanvelzen: Maybe add an expected type to the `getSimilarItems` so it won't respond with data you don't need

app/src/main/java/org/jellyfin/androidtv/details/ItemPosterPresenter.kt
```diff
@@ -0,0 +1,59 @@
...
+
+		if (baseItem.images.primary != null) {
+			GlobalScope.launch(Dispatchers.Main) {
+				cardView.mainImage = BitmapDrawable(baseItem.images.primary.getBitmap(TvApp.getApplication()))
+			}
+		}
+
+	}
+
+	override fun onUnbindViewHolder(viewHolder: ViewHolder?) {
```
nielsvanvelzen: This doesn't do anything...

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/LiftedItems.kt
AndreasGB: We might need this in multiple places though and it may need to be updated once we implement more item types, which is why I put it locally close to where the classes are defined.

nielsvanvelzen: Fair enough, can you change it to use the [proper naming convention](https://kotlinlang.org/docs/reference/coding-conventions.html#property-names)?

app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
AndreasGB: Hmm, we could use generics and then integrate the filter into the method, but I'm not sure there how we would allow multiple types to pass.

app/src/main/java/org/jellyfin/androidtv/details/ItemPosterPresenter.kt
AndreasGB: True

app/src/main/java/org/jellyfin/androidtv/util/apiclient/ApiClientExtensions.kt
AndreasGB: Sure, can do that

app/src/main/java/org/jellyfin/androidtv/util/apiclient/ApiClientExtensions.kt
AndreasGB: Will do

app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
nielsvanvelzen: Doesn't the query class support setting a type?

app/src/main/java/org/jellyfin/androidtv/model/itemtypes/LiftedItems.kt
AndreasGB: Sure, will do

app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
AndreasGB: I don't think so

nielsvanvelzen: Nope it does not support that.

nielsvanvelzen: In that case just one-line it like this:
```kotlin
val similarItems = TvApp.getApplication().apiClient.getSimilarItems(item)?.filter { it is Movie }
```

app/src/main/java/org/jellyfin/androidtv/details/ItemPosterPresenter.kt
AndreasGB: I would btw like to keep it for the TODO it contains

app/src/main/java/org/jellyfin/androidtv/util/apiclient/ApiClientExtensions.kt
```diff
@@ -55,12 +55,14 @@ suspend fun ApiClient.updateFavoriteStatus(itemId: String, userId: String, isFav
...
-	val query = SimilarItemsQuery()
-	query.setUserId(TvApp.getApplication().currentUser.id)
-	query.setId(item.id)
-	query.setLimit(10)
-	query.fields = FieldsRequiredForLift
+suspend fun ApiClient.getSimilarItems(item: BaseItem, limit: Int = 25): List<BaseItem>? = suspendCoroutine { continuation ->
+	val query = SimilarItemsQuery().apply {
+		id = item.id
+		userId = TvApp.getApplication().currentUser.id
+		this.limit = limit
```
nielsvanvelzen: I _think_ this sets the value of query.limit to query.limit

AndreasGB: IntelliJ syntax highlighting highlighted the other limit, I'll check with the debugger, just to be sure

AndreasGB: Works

PR #327, opened 2020-02-09
Update Gradle to Kotlin DSL

nielsvanvelzen:
- ❤️ Kotlin
- 💔 Groovy

_Might be easier to review the commits instead of all changes at once_


app/build.gradle
```diff
@@ -1,72 +0,0 @@
...
-    }
-
-    buildTypes {
-        release {
-            minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
-        }
-        debug {
-            applicationIdSuffix ".debug"
-            debuggable true
```
thornbill: Was this redundant?

app/src/main/java/org/jellyfin/androidtv/ui/ExpandableTextView.java
```diff
@@ -4,22 +4,23 @@
...
 import org.jellyfin.androidtv.details.ExpandedTextActivity;
 
+import androidx.core.app.ActivityOptionsCompat;
+
 /**
  * Created by spam on 9/30/2016.
  */
 
-public class ExpandableTextView extends AppCompatTextView {
+public class ExpandableTextView extends TextView {
```
thornbill: Why are we moving away from using the Androidx appcompat classes?

settings.gradle.kts
```diff
@@ -0,0 +1,12 @@
+include(":app")
+
+val enableDependencySubstitution = true
```
thornbill: Could we make this check an environment variable?

app/build.gradle
nielsvanvelzen: The "debug" version has debug on by default. I think there's some more redundant settings but I didn't look up everything.

> When you create a new module, Android Studio automatically creates the debug and release build types for you. Although the debug build type doesn't appear in the build configuration file, Android Studio configures it with debuggable true. 

[source](https://developer.android.com/studio/build/build-variants)

app/src/main/java/org/jellyfin/androidtv/ui/ExpandableTextView.java
nielsvanvelzen: There is only a few custom views that use the AppCompat classes. However if we truly used AppCompat we would need to change almost all UI related code (XML and Java) to use AppCompat.

The views we did use AppCompat for did not really use any features that were not available in the first versions of Android so it was not necessary to use it.

We can always add back AppCompat in the future but it would need to be done properly for all java/xml. You can see what I mean by using the "Migrate to Appcompat..." function in Android Studio. It gives you an impression of how much code was not AppCompat yet.

settings.gradle.kts
nielsvanvelzen: My initial idea was to add a property that could be added in the local.properties but I didn't get it working properly. Environment variable might be an idea too.

nielsvanvelzen: Added this one now. Adding `enable.dependency.substitution=false` to the `local.properties` file will disable dependency substitution. It's enabled by default.

PR #330, opened 2020-02-10
Make Debug Builds distinguishable from Release

AndreasGB:


Florianisme:
Can we also add a new "debug" build flavor? That way I can install the official jellyfin app from the play store besides the debug build on my TV. Now that the debug build even has a new icon, distinguishing them from each other should be no problem?

nielsvanvelzen:
The debug version already uses a separate package-id so installing both at the same time is possible.

AndreasGB:
I'll include your new image soon, @nielsvanvelzen

nielsvanvelzen:
I've checked the Weblate settings to make sure this doesn't screw up something over there and that should be good. It's also possible to [set the name using Gradle](https://stackoverflow.com/a/32562324), might be a preferable option instead of creating an XML file for a single string.

AndreasGB:
So, which of the three options should we pick? I don't have a preference :D

nielsvanvelzen:
I think the option where we add a new string to the strings.xml called "app_name_debug" and use Gradle to set it for debug builds would be great.

PR #307, opened 2020-02-06
Custom detail layout

nielsvanvelzen:
Unfinished but I can already share the first screenshots!

image
image

PR #331, opened 2020-02-11
My media channel

### LINKED ISSUE #74: Play Next and Channel section

### LINKED ISSUE #300: Next Up Not Supported?
From a [user comment](https://www.reddit.com/r/jellyfin/comments/ezgy3q/new_android_tv_release_v0110_improved_video/fgnage1) on Reddit:
>I've been looking forward to this release for the channels and up next support but neither of them are working. Any ideas?

>...The application Jellyfin is not showing as an application with channels available and in the Next Up menu it is not shown as an application that supports the Next Up channel.

nielsvanvelzen:
This adds a new "channel" (row) to the leanback home screen that displays all libraries. It's possible to open a library by clicking on it but the codebase is one big mess so it doesn't work that well right now. 

This also fixes Jellyfin not displaying when choosing which app can use the "watch next" channel. (Apparently you need to have at least one custom channel for your app to show in watch next.. good job Google!)

Because I wanted to use a crispy logo I added new vector images (both release&debug) and updated the app itself to use that one too.

Closes #300
Closes #74


app/src/main/java/org/jellyfin/androidtv/util/UriExtensions.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import android.content.ContentResolver
+import android.net.Uri
+import androidx.annotation.AnyRes
+import org.jellyfin.androidtv.TvApp
+
+/**
+ * Get the uri for a given resource
+ */
+fun uriFromResourceId(@AnyRes resid: Int): Uri {
```
thornbill: Seems like this could be used to replace this [method](https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/util/ImageUtils.java#L280-L289) or perhaps be made to share some common code given that it returns a `String` and this returns a `Uri`.

app/src/main/java/org/jellyfin/androidtv/util/NumberExtensions.kt
```diff
@@ -0,0 +1,9 @@
+package org.jellyfin.androidtv.util
+
+import org.jellyfin.androidtv.TvApp
+
+/**
+ * Current (pixel) value as display pixels
+ */
+val Int.dp: Int
```
thornbill: This seems to be doing the same thing as this [method](https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/util/Utils.java#L69-L72).

nielsvanvelzen: I actually introduced this change in another PR (https://github.com/jellyfin/jellyfin-androidtv/pull/307/commits/1e1d83b487ad748bcbfb53bfdb36113aaafe747d#diff-3d7ec439941e8f132c864ebd3a25191e). I didn't know we already had a method to do this calculation.

This variant uses a Kotlin extension function so we can write `7.dp` and it will convert the integer into pixels (for my emulator it would return `14`). I think it looks better than `Utils.convertDpToPixel(context, 7)`.

It's not really possible to use the extension function in the Java classes so I changed the extension function to use the Utils class instead.

app/src/main/java/org/jellyfin/androidtv/util/UriExtensions.kt
nielsvanvelzen: Good one, I changed the method in the ImageUtils class to properly specify the path (it was missing the "resource type name") and used that one. I tested it and it still works for the other usages as well.

nielsvanvelzen:
Btw, don't merge this yet as I want to look into fixing opening the browse screen properly for all types of libraries.

nielsvanvelzen:
It's ready now @thornbill 🎉

fsaris:
Is this feature already available in current release or do we have to wait for v0.12.0?

How is user management handled here? As jellyfin automatically logsoff how can we keep the next up etc up-to-date?

nielsvanvelzen:
It will be available in 0.12 not before that. It will update the information the same way as the "play next" section is updated; When a logged in user opens the home page it will refresh the data.

Unfortunately at this point in time we can not run it automatically in the background, this is however something I'd like to add in the future.

tomaswl1:
Hi Nielsvanvelzen in the ChannelManager.kt code the nextup code of it if you dont have any next up items in the app it will force close the app after around 20 seconds but fine if you do have nextup items and works as should on shield homescreen thought i would give you a heads up

nielsvanvelzen:
Hi @tomaswl1. Can you provide any logs of this issue?

tomaswl1:
will have a look once home mate tested multiple times defo crashing app around 20 to 30 seconds when dont have any nextup items in app but if quick enough and click watched so it generates the next up item app is fine and next up content shows on shield homescreen

PR #332, opened 2020-02-11
Add MultiBadgeCardView

AndreasGB:
This introduces a new CardView that can have badges in six locations.


app/src/main/java/org/jellyfin/androidtv/presentation/MultiBadgeImageCardView.kt
```diff
@@ -0,0 +1,134 @@
...
+			if (fade) {
+				fadeIn()
+			} else {
+				mFadeInAnimator.cancel()
+				main_image.alpha = 1f
+			}
+		}
+	}
+
+	fun setMainImageDimensions(width: Int, height: Int) {
```
nielsvanvelzen: Does this function properly convert display pixels to pixels?

AndreasGB: No, it takes raw pixels atm, should I move to dp?

PR #333, opened 2020-02-11
Introduce new debug versioning

AndreasGB:
This commit introduces a new versioning schema for debug versions, which makes it obvious which commit hash and branch a debug build was created from. 

This will allow better matching of crash reports of debug builds as well as possibly help us in setting up potential future nightly builds.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,7 +1,42 @@
+import java.io.ByteArrayOutputStream
+
 plugins {
 	id("com.android.application")
 	id("kotlin-android")
 	id("kotlin-android-extensions")
+}
+
+fun getGitHash(): String {
+	val stdout = ByteArrayOutputStream()
+	val action = Action<ExecSpec> {
+		commandLine("git", "rev-parse", "--short", "HEAD")
+		standardOutput = stdout
+	}
+
+	exec(action)
+	return stdout.toString().trim()
+}
+
+fun getBranchName(): String {
+	val stdout = ByteArrayOutputStream()
+	val action = Action<ExecSpec> {
+		commandLine("git", "rev-parse", "--abbrev-ref", "HEAD")
+		standardOutput = stdout
+	}
+
+	exec(action)
+	return stdout.toString().trim()
+}
+
+fun isTreeClean(): Boolean {
+	val stdout = ByteArrayOutputStream()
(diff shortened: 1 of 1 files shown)


app/build.gradle.kts
```diff
@@ -34,6 +69,7 @@ android {
 			// Use different application id to run release and debug at the same time
 			applicationIdSuffix = ".debug"
 			resValue("string", "app_name", "@string/app_name_debug")
+			versionNameSuffix =  "-debug-" + getBranchName() + "-" + getGitHash() + (if (isTreeClean()) "-treeClean" else "-treeDirty")
```
dkanada: I usually prefer to hide this information somewhere else so the name looks less awkward. It looks a lot better when you run the debug build more often than the production one.

AndreasGB: Hmmm, I get that, but how often does one really look at the version string?

dkanada: Ah, this is the version string. I have seen it added to the app name in the past, which is what I thought was happening here.

thornbill: I think I would prefer it only append `-dirty` if the tree is not clean and nothing when it is.

AndreasGB: Okay, will change that

dkanada:
@nielsvanvelzen we're only appending the commit hash, which I don't think makes tracking any easier, but this is standard practice for development builds and extremely useful information.

nielsvanvelzen:
I admit that I might have exaggerated about the "tracking" when applying commit hashes now that I thought a bit about it. I think it might indeed be useful to see which crash report come from release and which don't. Although using just a commit hash and dirty state mark might be enough, as we can always find the commit in the git logs without a branch name. When I did some digging we might be able to use the `git describe` command to get this information. (even the branch name).

If a single function could execute just a single git command (and optionally parse the result if more data was returned than needed) the performance won't suffer much. Proper detection if the git command is available should be added though, with a proper fallback (use "-debug-nogit" as suffix or something similar).

Might be worth looking at a [`buildSrc`](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources) module too for these utility functions (same as the dependency substitution)

AndreasGB:
Closing for now, we can revisit this once we decide to ship nightlies, before that it doesn't have much of an impact.

PR #336, opened 2020-02-13
Implement basic badges

AndreasGB:
When displaying an unwatched count, it currently is off-center to the left :/

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/details/ItemPosterPresenter.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/ItemPosterPresenter.kt
@@ -46,7 +46,7 @@ class ItemPosterPresenter(private val context: Context) : Presenter(), IItemClic
 			mainImageDrawable = TvApp.getApplication().getDrawableCompat(R.drawable.tile_port_video)
 
 			if (baseItem is PlayableItem)
-				setBadge(WatchedBadge(context, baseItem), MultiBadgeImageCardView.BadgeLocation.TOP_RIGHT)
+				setBadge(MultiBadgeImageCardView.BadgeLocation.TOP_RIGHT, WatchedBadge(context, baseItem))
 		}
 
 
--- a/app/src/main/java/org/jellyfin/androidtv/details/ItemPosterPresenter.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/ItemPosterPresenter.kt
@@ -14,6 +14,7 @@ import org.jellyfin.androidtv.TvApp
 import org.jellyfin.androidtv.base.IItemClickListener
 import org.jellyfin.androidtv.model.itemtypes.BaseItem
 import org.jellyfin.androidtv.model.itemtypes.PlayableItem
+import org.jellyfin.androidtv.ui.FavoriteBadge
 import org.jellyfin.androidtv.ui.MultiBadgeImageCardView
 import org.jellyfin.androidtv.ui.WatchedBadge
 import org.jellyfin.androidtv.util.dp
@@ -47,6 +48,8 @@ class ItemPosterPresenter(private val context: Context) : Presenter(), IItemClic
 
 			if (baseItem is PlayableItem)
 				setBadge(MultiBadgeImageCardView.BadgeLocation.TOP_RIGHT, WatchedBadge(context, baseItem))
+
+			setBadge(MultiBadgeImageCardView.BadgeLocation.BOTTOM_RIGHT, FavoriteBadge(context, baseItem))
 		}
 
 
--- a/app/src/main/java/org/jellyfin/androidtv/details/ItemPosterPresenter.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/ItemPosterPresenter.kt
@@ -13,7 +13,9 @@ import org.jellyfin.androidtv.R
 import org.jellyfin.androidtv.TvApp
 import org.jellyfin.androidtv.base.IItemClickListener
 import org.jellyfin.androidtv.model.itemtypes.BaseItem
+import org.jellyfin.androidtv.model.itemtypes.PlayableItem
 import org.jellyfin.androidtv.ui.MultiBadgeImageCardView
+import org.jellyfin.androidtv.ui.WatchedBadge
 import org.jellyfin.androidtv.util.dp
 
 private const val LOG_TAG = "ItemPosterPresenter"
@@ -42,6 +44,9 @@ class ItemPosterPresenter(private val context: Context) : Presenter(), IItemClic
 			isFocusableInTouchMode = true
 			setMainImageDimensions(100.dp, 150.dp)
 			mainImageDrawable = TvApp.getApplication().getDrawableCompat(R.drawable.tile_port_video)
+
+			if (baseItem is PlayableItem)
+				setBadge(WatchedBadge(context, baseItem), MultiBadgeImageCardView.BadgeLocation.TOP_RIGHT)
 		}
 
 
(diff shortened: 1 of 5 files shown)


app/src/main/res/layout/badge_watched.xml
```diff
@@ -0,0 +1,52 @@
...
+        app:layout_constraintTop_toTopOf="parent"
+        tools:visibility="visible"/>
+
+    <TextView
+        android:id="@+id/badge_counter"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:foregroundGravity="center"
+        android:gravity="center"
+        android:textAlignment="center"
```
nielsvanvelzen: The text alignment doesn't do anything because the view is as wide as the text

app/src/main/res/layout/badge_watched.xml
```diff
@@ -0,0 +1,52 @@
...
+        app:layout_constraintLeft_toLeftOf="parent"
+        app:layout_constraintRight_toRightOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        tools:visibility="visible"/>
+
+    <TextView
+        android:id="@+id/badge_counter"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:foregroundGravity="center"
```
nielsvanvelzen: Doesn't do much

app/src/main/res/layout/badge_watched.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
nielsvanvelzen: Because everything is literally stacked on top of each other it might be better to use a FrameLayout. Makes rendering a bit faster too.

app/src/main/java/org/jellyfin/androidtv/ui/WatchedBadge.kt
```diff
@@ -0,0 +1,62 @@
...
+		badge_counter.text = count.toString()
+	}
+
+	fun clear() {
+		badge_background.visibility = View.GONE
+		badge_watched_tick.visibility = View.GONE
+		badge_counter.visibility = View.GONE
+	}
+
+	private fun setItemData(item: PlayableItem?) {
```
nielsvanvelzen: This function accepts a nullable but when called always supplies a non-null item.

app/src/main/java/org/jellyfin/androidtv/details/ItemPosterPresenter.kt
```diff
@@ -42,6 +44,9 @@ class ItemPosterPresenter(private val context: Context) : Presenter(), IItemClic
 			isFocusableInTouchMode = true
 			setMainImageDimensions(100.dp, 150.dp)
 			mainImageDrawable = TvApp.getApplication().getDrawableCompat(R.drawable.tile_port_video)
+
+			if (baseItem is PlayableItem)
+				setBadge(WatchedBadge(context, baseItem), MultiBadgeImageCardView.BadgeLocation.TOP_RIGHT)
```
nielsvanvelzen: Might be a bit late but i'd swap the parameter so you specify the location first and the content second

AndreasGB: I think the current order is more "readable", like set badge watched badge to the top right, but that isn't a strong preference.

nielsvanvelzen: It's funny because I think it's more readable if you say "At the top right, I want to show the watched badge". It's not a big issue so you can leave it.

app/src/main/java/org/jellyfin/androidtv/ui/WatchedBadge.kt
```diff
@@ -0,0 +1,62 @@
...
+
+import android.content.Context
+import android.view.LayoutInflater
+import android.view.View
+import android.widget.FrameLayout
+import kotlinx.android.synthetic.main.badge_watched.view.*
+import org.jellyfin.androidtv.R
+import org.jellyfin.androidtv.model.itemtypes.PlayableItem
+
+class WatchedBadge(context: Context) : FrameLayout(context) {
```
nielsvanvelzen: If you specify the primary constructor as `(context: Context, private var item: PlayableItem? = null)` it will be a class property with a value of null by default.

app/src/main/java/org/jellyfin/androidtv/ui/WatchedBadge.kt
```diff
@@ -0,0 +1,62 @@
...
+	private var item: PlayableItem? = null
+
+	private val changeListener: () -> Unit = {
+		if (item != null) {
```
nielsvanvelzen: `item?.let(::setItemData)`

app/src/main/java/org/jellyfin/androidtv/ui/WatchedBadge.kt
```diff
@@ -0,0 +1,62 @@
...
+			setItemData(item!!)
+		}
+	}
+
+	init {
+		val inflater = LayoutInflater.from(context)
+		inflater.inflate(R.layout.badge_watched, this)
```
nielsvanvelzen: You can add a line here like `item?.let(::setItemData)` to update if you go for the primary constructor change. Could remove the secondary constructor entirely in that case.

AndreasGB: We still need one which takes only context as tools allegedly require them, but flipping them is a good idea.

app/src/main/java/org/jellyfin/androidtv/ui/FavoriteBadge.kt
```diff
@@ -0,0 +1,34 @@
...
+	private val changeListener: () -> Unit = {
+		item?.let(::setItemData)
+	}
+
+	init {
+		setImageDrawable(context.getDrawable(R.drawable.ic_heart_red))
+		item?.let(::setItemData)
+	}
+
+	constructor(context: Context) : this(context, null)
```
nielsvanvelzen: This constructor shouldn't be necessary because  the `item` parameter has a default value

app/src/main/java/org/jellyfin/androidtv/ui/WatchedBadge.kt
```diff
@@ -0,0 +1,57 @@
...
+		item?.let(::setItemData)
+	}
+
+	init {
+		val inflater = LayoutInflater.from(context)
+		inflater.inflate(R.layout.badge_watched, this)
+		item?.let(::setItemData)
+	}
+
+	constructor(context: Context) : this(context, null)
```
nielsvanvelzen: This constructor shouldn't be necessary because the item parameter has a default value

app/src/main/java/org/jellyfin/androidtv/ui/WatchedBadge.kt
```diff
@@ -0,0 +1,57 @@
...
+
+	fun setWatched() {
+		badge_background.visibility = View.VISIBLE
+		badge_watched_tick.visibility = View.VISIBLE
+		badge_counter.visibility = View.GONE
+	}
+
+	fun setRemainingCount(count: Int) {
```
nielsvanvelzen: I don't think this would happen fast but if the count is higher than 99 we should probably display `99+` so the text won't overflow.

PR #341, opened 2020-02-14
Use PreferredVideoPlayer enum for live tv preference

thornbill:
This updates the video player preferences for live tv to use the same method as for regular videos.

spiderman double

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preferences/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preferences/UserPreferences.kt
@@ -127,14 +127,9 @@ class UserPreferences(context: Context) : SharedPreferenceStore(PreferenceManage
 	var liveTvDirectPlayEnabled by booleanPreference("pref_live_direct", true)
 
 	/**
-	 * Use VLC for live TV playback
+	 * Preferred video player for live TV
 	 */
-	var liveTvUseVlc by booleanPreference("pref_enable_vlc_livetv", false)
-
-	/**
-	 * Use external player for live TV playback
-	 */
-	var liveTvUseExternalPlayer by booleanPreference("pref_live_tv_use_external", false)
+	var liveTvVideoPlayer by enumPreference("live_tv_video_player", PreferredVideoPlayer.AUTO)
 
 	/* ACRA */
 	/**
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/jellyfin/androidtv/preferences/UserPreferences.kt
```diff
@@ -170,5 +165,14 @@ class UserPreferences(context: Context) : SharedPreferenceStore(PreferenceManage
 			// Migrate to login behavior enum
 			putEnum("login_behavior", if (it.getString("pref_login_behavior", "0") == "1") LoginBehavior.AUTO_LOGIN else LoginBehavior.SHOW_LOGIN)
 		}
+
+		migration(toVersion = 4) {
```
nielsvanvelzen: I think we should add this new enum to the previous migration (`toVersion = 3`). My intention was to have 1 migration per app release. _The initial update will have 2 though, 1 for the setting change in 0.11 for people that upgrade from 0.10->0.12_

thornbill: :+1: Cool sounds good to me. I had wondered about that.

app/src/main/res/xml/preferences.xml
```diff
@@ -121,21 +121,14 @@
...
-            android:key="pref_enable_vlc_livetv"
-            android:summary="@string/desc_use_vlc_livetv"
-            android:title="@string/lbl_use_vlc_livetv" />
+        <ListPreference
+            android:entries="@array/pref_video_player_entries"
+            android:entryValues="@array/pref_video_player_values"
+            android:key="live_tv_video_player"
+            android:title="@string/pref_media_player"
+            app:useSimpleSummaryProvider="true" />
         <CheckBoxPreference
```
nielsvanvelzen: I don't know a lot about LiveTV but as far as I know this checkbox should only be enabled when the selected player is external. Because the `dependency` attribute only works when referencing checkboxes this should be added in code.
You can add it in the `UserPreferencesFragment.kt` file.

thornbill: I’m not sure why this was marked as a dependency. This controls direct playing in the embedded players. I have no clue why this isn’t handled automatically like it is for normal videos. 🤷‍♂

nielsvanvelzen: We can always make changes to that too in the future, I'll approve it for now :)

PR #347, opened 2020-02-17
Remaining actions for movie details

AndreasGB:


app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
```diff
@@ -49,12 +48,17 @@ class MovieDetailsFragment(item: Movie) : BaseDetailsFragment<Movie>(item) {
...
-			if (item.canResume) add(ResumeAction(context!!, item).apply { icon = context!!.getDrawable(R.drawable.ic_resume) })
-
-			add(PlayFromBeginningAction(context!!, item).apply { icon = context!!.getDrawable(R.drawable.ic_play) })
-			add(ToggleWatchedAction(context!!, item).apply { icon = context!!.getDrawable(R.drawable.ic_watch) })
-			add(ToggleFavoriteAction(context!!, item).apply { icon = context!!.getDrawable(R.drawable.ic_heart) })
-			add(Action(0, "More").apply { icon = context!!.getDrawable(R.drawable.ic_more) }) // Show menu with more options
+			if (item.canResume) add(ResumeAction(context!!, item))
+			add(PlayFromBeginningAction(context!!, item))
+			// TODO: Bring this back once we have a more understandable queue implementation for users
+			//add(ToggleWatchedAction(context!!, item))
```
nielsvanvelzen: I think you meant comment `AddToQueueAction`

app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
```diff
@@ -49,12 +48,17 @@ class MovieDetailsFragment(item: Movie) : BaseDetailsFragment<Movie>(item) {
...
+			add(ToggleFavoriteAction(context!!, item))
+			add(AddToQueueAction(context!!, item))
+
+			// Menu with more actions
+			add(SecondariesPopupAction(context!!).apply {
+				add(DeleteAction(context!!, item) { activity?.finish() })
```
nielsvanvelzen: This one depends on the permissions of the authenticated user

app/src/main/java/org/jellyfin/androidtv/details/actions/DeleteAction.kt
```diff
@@ -0,0 +1,47 @@
...
+					}
+
+					override fun onError(ex: Exception) {
+						msg.Cancel()
+						TvApp.getApplication().logger.ErrorException("Failed to delete item %s", ex, item.title)
+						Utils.showToast(context, ex.localizedMessage)
+					}
+				})
+			}
+			setNegativeButton(context.getText(R.string.lbl_cancel)) { dialog, which -> Utils.showToast(context, R.string.lbl_item_not_deleted) }
```
nielsvanvelzen: Just use `Toast.makeText(context, text, length).show()` instead of a util function for a one-liner.

app/src/main/java/org/jellyfin/androidtv/details/actions/DeleteAction.kt
```diff
@@ -0,0 +1,47 @@
...
+		AlertDialog.Builder(context).apply {
+			setIcon(R.drawable.ic_trash)
+			setTitle(R.string.lbl_really_delete_item_title)
+			setMessage(context.getString(R.string.lbl_item_deletion_warning, item.title))
+			setPositiveButton(R.string.lbl_delete) { _, _ ->
+				val msg = DelayedMessage(context, 150)
+				TvApp.getApplication().apiClient.DeleteItem(item.id, object : EmptyResponse() {
+					override fun onResponse() {
+						msg.Cancel()
+						Utils.showToast(context, context.getString(R.string.lbl_item_deleted, item.title))
```
nielsvanvelzen: Just use Toast.makeText(context, text, length).show() instead of a util function for a one-liner.

app/src/main/java/org/jellyfin/androidtv/details/actions/DeleteAction.kt
```diff
@@ -0,0 +1,47 @@
...
+						TvApp.getApplication().lastDeletedItemId = item.id
+						onItemDeleted()
+					}
+
+					override fun onError(ex: Exception) {
+						msg.Cancel()
+						TvApp.getApplication().logger.ErrorException("Failed to delete item %s", ex, item.title)
+						Utils.showToast(context, ex.localizedMessage)
```
nielsvanvelzen: Just use Toast.makeText(context, text, length).show() instead of a util function for a one-liner.

app/src/main/java/org/jellyfin/androidtv/details/actions/DeleteAction.kt
```diff
@@ -0,0 +1,47 @@
...
+		icon = context.getDrawable(R.drawable.ic_trash)
+	}
+
+	override fun onClick() {
+		AlertDialog.Builder(context).apply {
+			setIcon(R.drawable.ic_trash)
+			setTitle(R.string.lbl_really_delete_item_title)
+			setMessage(context.getString(R.string.lbl_item_deletion_warning, item.title))
+			setPositiveButton(R.string.lbl_delete) { _, _ ->
+				val msg = DelayedMessage(context, 150)
```
nielsvanvelzen: It doesn't look like this is used?

app/src/main/java/org/jellyfin/androidtv/details/actions/SecondariesPopupAction.kt
```diff
@@ -0,0 +1,37 @@
...
+	}
+
+	fun remove(toRemove: BaseAction) {
+		containedActions.remove(toRemove)
+	}
+
+	override fun onClick() {
+		val menu = PopupMenu(context, anchor)
+		containedActions.forEach { action ->
+			val item = menu.menu.add(action.label1)
```
nielsvanvelzen: Can we add icons to the menu too?

app/src/main/java/org/jellyfin/androidtv/details/actions/DeleteAction.kt
AndreasGB: You would think so, but this inherited piece of premium poetry starts itself in its constructor :D

app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
AndreasGB: Uh, yes

app/src/main/java/org/jellyfin/androidtv/details/actions/SecondariesPopupAction.kt
AndreasGB: Not easily it seems, there is [this](https://stackoverflow.com/questions/15454995/popupmenu-with-icons), but it crashes weirdly although I am using it just as in the example.

app/src/main/java/org/jellyfin/androidtv/details/MovieDetailsFragment.kt
```diff
@@ -50,15 +50,17 @@ class MovieDetailsFragment(item: Movie) : BaseDetailsFragment<Movie>(item) {
...
 			add(ToggleFavoriteAction(context!!, item))
-			add(AddToQueueAction(context!!, item))
+			// TODO: Bring this back once we have a more understandable queue implementation for users
+			//add(AddToQueueAction(context!!, item))
 
 			// Menu with more actions
-			add(SecondariesPopupAction(context!!).apply {
-				add(DeleteAction(context!!, item) { activity?.finish() })
-			})
+			if (TvApp.getApplication().currentUser.policy.enableContentDeletion) {
```
nielsvanvelzen: This if statement should be for the DeleteAction only. I imagine you did it this way to the popup button wouldn't show when there's no items but we need to fix that in another way.

1. Extend action with a `isvisible` property or something similar
2. implement it for secondariespopupaction and check if the amount if children >= 1
2. When showing the actions, look if it's visible
4. that's it

app/src/main/java/org/jellyfin/androidtv/details/actions/DeleteAction.kt
nielsvanvelzen: I looked at the code of `DelayedMessage` and now I understand it. wtf

app/src/main/java/org/jellyfin/androidtv/details/actions/SecondariesPopupAction.kt
nielsvanvelzen: The `.add` method returns a `Menuitem` instance. This class contains a [setIcon](https://developer.android.com/reference/android/view/MenuItem.html#setIcon(android.graphics.drawable.Drawable)) function

AndreasGB: Ah my bad, did not link the correct answer. I used the `.icon` property, but icons are only shown in certain cases normally if you don't force them on. I meant to link to [this answer](https://stackoverflow.com/a/39026923/1915406) instead of just the question, and this one fails with some weird crash during measuring (iirc)

nielsvanvelzen: Hmm weird, let's leave them out for now

app/src/main/java/org/jellyfin/androidtv/details/actions/SecondariesPopupAction.kt
```diff
@@ -9,26 +9,37 @@ class SecondariesPopupAction(context: Context) : BaseAction(ActionID.SECONDARIES
...
 
 	fun remove(toRemove: BaseAction) {
+		if (toRemove.onVisibilityChanged == ::recomputeVisibility) toRemove.onVisibilityChanged = null
 		containedActions.remove(toRemove)
+		recomputeVisibility()
 	}
 
 	override fun onClick() {
 		val menu = PopupMenu(context, anchor)
 		containedActions.forEach { action ->
```
nielsvanvelzen: `containedActions.filter { it.isVisible }.forEach {`

PR #348, opened 2020-02-17
Add ability to annoy users from server

nielsvanvelzen:
Tells the server it allows messages and shows them to the user when an admin decides to annoy the user.


AndreasGB:
Ah yes, very important feature, great implementation, but where is the Snackbar

nielsvanvelzen:
The snackbar is in maaskantje

PR #349, opened 2020-02-18
Remove dependency on apiclient library

nielsvanvelzen:
Because the android package of the library now exposes the library it isn't necessary anymore to declare it in our own gradle file.

It was fixed by [this](https://github.com/jellyfin/jellyfin-apiclient-java/commit/8a93388a6da3c26b514f9c0aa94ef050ca47a892) commit in the apiclient.


thornbill:
It looks like `settings.gradle` will also need to be updated for dependency substitution to work as expected.

nielsvanvelzen:
I somehow missed that, I changed the substitution to replace the android library (as it actually should've done from the start).

nielsvanvelzen:
Fixed the conflict using GitHub's editor

PR #354, opened 2020-02-19
[DRAFT] Add remote trailer support

### LINKED ISSUE #71: Movie / TVShow Trailer link
Not sure if this is an issue or just an enhancement, in the beta 4 apk the option to watch the trailer is not present for Movies or TV-Shows like in the android app or in browser.

AndreasGB:
This PR adds a new Presenter for Trailers which also shows external ones and has special support for YouTube trailers.

Testing on device is recommended, otherwise the YouTube app has to be present on the emulator.

Closes #71


app/src/main/java/org/jellyfin/androidtv/model/itemtypes/LiftedItems.kt
```diff
@@ -3,11 +3,19 @@ package org.jellyfin.androidtv.model.itemtypes
...
 
-val FIELDS_REQUIRED_FOR_LIFT = arrayOf(ItemFields.DateCreated, ItemFields.MediaSources, ItemFields.MediaStreams, ItemFields.People)
+val FIELDS_REQUIRED_FOR_LIFT = arrayOf(
+	ItemFields.DateCreated,
+	ItemFields.MediaSources,
+	ItemFields.MediaStreams,
+	ItemFields.People,
+	ItemFields.Genres,
+	ItemFields.Tags,
+	ItemFields.RemoteTrailers)
```
nielsvanvelzen: This `)` annoys me lol, should be placed on the next line (_although not necessary for this PR_)

app/src/main/java/org/jellyfin/androidtv/details/TrailerPresenter.kt
```diff
@@ -0,0 +1,149 @@
...
+		} else {
+			return null
+		}
+	}
+
+	private fun setupURLCard(cardView: MultiBadgeImageCardView, name: String, uri: Uri) {
+		cardView.apply {
+			titleText = name
+
+			contentText = if (isYoutubeUrl(uri)) "YouTube" else getDomain(uri)
```
nielsvanvelzen: YouTube should be a string resource

app/src/main/java/org/jellyfin/androidtv/details/TrailerPresenter.kt
```diff
@@ -0,0 +1,149 @@
...
+
+	private fun getProviderIcon(uri: Uri): Drawable? {
+		return if (isYoutubeUrl(uri) && !SAFE_BRANDING_COMPLIANCE)
+			context.getDrawable(R.drawable.ic_youtube)
+		else
+			null
+	}
+
+	private fun getThumbnailURL(uri: Uri): String? {
+		if (isYoutubeUrl(uri) && uri.queryParameterNames.contains("v")) {
```
nielsvanvelzen: `isYoutubeUrl` will return true for `youtu.be` urls but they don't use the `v` parameter. Example:
- https://youtu.be/-50NdPawLVY

app/src/main/java/org/jellyfin/androidtv/details/TrailerPresenter.kt
```diff
@@ -0,0 +1,149 @@
...
+import org.jellyfin.androidtv.ui.FavoriteBadge
+import org.jellyfin.androidtv.ui.MultiBadgeImageCardView
+import org.jellyfin.androidtv.ui.WatchedBadge
+import org.jellyfin.androidtv.util.ImageUtils
+import org.jellyfin.androidtv.util.dp
+import org.jellyfin.apiclient.model.entities.MediaUrl
+
+private const val SAFE_BRANDING_COMPLIANCE: Boolean = true
+
+class TrailerPresenter(private val context: Context, private val imageHeight: Int) : Presenter(), IItemClickListener {
```
nielsvanvelzen: At this moment the server only provides youtube trailers but a new plugin could easily add urls for Vevo or other sites/apps. I think we should prepare for this by abstracting the external media.

A separate class could be made for `YoutubeTrailer` that extends a class `ExternalTrailer`. Some magic could convert the info we have to the youtube trailer class and that one contains some functions like `.open()` `.getThumbnail()` `.getBrandingIcon()`. Or something similar.

Next, we do some calculations to get the video-id from youtube urls. I want unit tests for this so it's properly tested.

nielsvanvelzen:
image

Amazing, transparant bars on the sides and black bars on top & bottom! I think we shouldn't assume a 16:9 aspect ratio, the black bars are a YouTube thing though.

app/src/main/java/org/jellyfin/androidtv/details/TrailerPresenter.kt
```diff
@@ -0,0 +1,149 @@
...
+
+	private fun getThumbnailURL(uri: Uri): String? {
+		if (isYoutubeUrl(uri) && uri.queryParameterNames.contains("v")) {
+			return "https://img.youtube.com/vi/%s/0.jpg".format(uri.getQueryParameter("v"))
+		} else {
+			return null
+		}
+	}
+
+	private fun setupURLCard(cardView: MultiBadgeImageCardView, name: String, uri: Uri) {
```
nielsvanvelzen: Add `cardView.setMainImageDimensions((4.0 / 3.0 * imageHeight).toInt(), imageHeight)`
Optionally add the aspect ratio to ImageUtils (4:3)

image

nielsvanvelzen: We had a discussion in Matrix and another solution would be to still use a 16:9 based card but have the image "fit". This would remove the black bars for the 16:9 thumbnails but remove some pixels foor 4:3 thumbs

AndreasGB:
This is not yet fully implemented, but can you please give me feedback whether this architecture is more in line what you expected from it being extensible?

nielsvanvelzen:
I looked into it and I suspect the best solution for this would be to use the [ClassPresenterSelector] for the various trailer types. But this would require changed when adding items to the row. I haven't looked too much into it though. 

What do you think about that? @AndreasGB 

[ClassPresenterSelector]: https://developer.android.com/reference/kotlin/androidx/leanback/widget/ClassPresenterSelector

AndreasGB:
I looked into it now and I think we could also go this route if we want to. 

We would have to build additional lifting logic when lifting `BaseItemDto` that converts a `MediaUrl` Object into a `YouTubeTrailer`, `VevoTrailer`, `GenericTrailer`, that inherit from `ExternalTrailer` and then have one `ClassPresenterSelector` per layer in the inheritance hierarchy. 

The actuall `Presenter`s then could mirror that inheritance structure.

On the one hand, this approach might be a bit cleaner still as we deal with a better object hierarchy, on the other hand I expect the current design to work well enough for the probably four types of external trailer sources the app will have to deal with.

AndreasGB:
Thinking about this some more, we may yet want to go down the way of a third refactor because splitting responsibilities between lifting and filling cards would allow for unit tests of the lifting, which if implemented in the current version would muddy the external interface of the combined classes.

AndreasGB:
Closing this PR, as this attempt did not prove fruitful. Will shortly reopen a draft PR for a new design.

PR #359, opened 2020-02-23
Detail misc

nielsvanvelzen:
- Poster now stays in the same place (there was no animation when switching position and this was really annoying IMO)
- Removed the properties for `Episode` as they are not used at this moment
- Fixed crashes when properties were set to null
  - This can be tested by creating a "garbage" file like `sldfjlksjfkl.mkv`


AndreasGB:
Whoops, forgot to greenlight this yesterday. We should look into adding back a proper animation at some point though.

PR #360, opened 2020-02-24
Add child count to requested fields in queries

thornbill:
Jellyfin adds `ChildCount` to the list of requested fields for particular app names. That is dumb, we should just request the fields we need. I added `ChildCount` to all queries since that seems to be what what was happening server side although it's probably not strictly needed for all of them.

https://github.com/jellyfin/jellyfin/blob/a6f883345f1941bf99db1667e84c1caf0b370479/MediaBrowser.Api/BaseApiService.cs#L157-L171

This will allow us to _finally_ add a space to the app name because that has been annoying me for far too long. :laughing:

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/querying/StdItemQuery.java
+++ b/app/src/main/java/org/jellyfin/androidtv/querying/StdItemQuery.java
@@ -7,13 +7,15 @@ import org.jellyfin.apiclient.model.querying.ItemQuery;
 
 public class StdItemQuery extends ItemQuery {
     public StdItemQuery(ItemFields[] fields) {
-        if (fields == null) fields = new ItemFields[] {
-                ItemFields.PrimaryImageAspectRatio,
-                ItemFields.Overview,
-                ItemFields.ItemCounts,
-                ItemFields.DisplayPreferencesId,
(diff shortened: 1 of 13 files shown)


app/src/main/java/org/jellyfin/androidtv/querying/StdItemQuery.java
```diff
@@ -5,13 +5,15 @@
...
 import org.jellyfin.apiclient.model.querying.ItemQuery;
 
-/**
- * Created by Eric on 12/5/2014.
- */
 public class StdItemQuery extends ItemQuery {
-
     public StdItemQuery(ItemFields[] fields) {
-        if (fields == null) fields = new ItemFields[] {ItemFields.PrimaryImageAspectRatio, ItemFields.Overview, ItemFields.ItemCounts, ItemFields.DisplayPreferencesId};
+        if (fields == null) fields = new ItemFields[] {
```
nielsvanvelzen: I think it would look better to use curly brackets because the assignment is multi-line.

thornbill: Good catch. It's fixed now. (I actually prefer always using curly brackets.)

PR #370, opened 2020-02-29
Remove "end_of_line" rule from editorconfig

nielsvanvelzen:
It caused issues in Android Studio on Windows where it would "detect" changed files but there were none. git will automatically convert line endings on windows so this should not cause any problems.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,4 @@
 [*]
-end_of_line = lf
 insert_final_newline = true
 
 [{*.kts, *.kt}]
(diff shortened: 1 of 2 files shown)


thornbill:
> git will automatically convert line endings on windows

Do we need to add something in the git config to handle this? I have definitely seen files committed with CRLF line endings.

nielsvanvelzen:
The git installer on Windows gives the user the ability to choose but it will default to automatically convert all line endings in the commits (so not on the local file system).

Might be worth adding a linter of some sort for it to make sure everything in the repository uses the correct line endings.

PR #375, opened 2020-03-05
Add books to excluded types in latest row fragment

thornbill:
The Latest Books row would just display a never ending loading element. I don't think we have any plans to support book libraries on TVs so let's just exclude them for now. :smile:


cwegener:
> I don't think we have any plans to support book libraries on TVs
:disappointed:

PR #376, opened 2020-03-05
Remove auto-logoff "feature"

AndreasGB:
I think this should be all the stuff related to it, but I did not delete the translated strings. Unsure about weblates reaction to it.


app/src/main/java/org/jellyfin/androidtv/base/BaseActivity.java
```diff
@@ -126,44 +122,9 @@ protected void onPause() {
 
     @Override
     protected void onDestroy() {
```
nielsvanvelzen: You can remove this function completely

app/src/main/res/values/array.xml
```diff
@@ -85,14 +85,6 @@
         <item>6 Hours</item>
```
thornbill: These entries for `pref_timeout_entries` should be removable also.

PR #379, opened 2020-03-06
Replace leanback details fragment with custom implementation

nielsvanvelzen:
This PR replaces the `DetailsSupportFragment` with the way more simple `RowsSupportFragment` and creates a custom row for the details instead of using `DetailsOverviewRow`.

These changes allow us to add more customization instead of being bound to the way Leanback wants us to do it.

**Screenshots**

_Initial changes that could compile_

_Initial working version with basic layouting_

_Adding poster/backdrop_

_Fixed backdrop+actions_

image
_Converting to a single ConstraintLayout_

image
_Pink theme!_

image
_Final design for now, working on fixing actions_

image
_Changed the padding/marings on the buttons and started on fixing their behavior_

---
What still needs to be done before merging:
- [x] Fix action styling
- [ ] Add back parallax scrolling
- [x] Fix action logic


nielsvanvelzen:
This PR does not include the parallax scrolling and doesn't have behavior to hide the backdrop when pressing the down arrow (to make the description more visible).

Both of those features need to be added before release but will be in another PR as this one is big enough already.

AndreasGB:
I have not yet looked at the code, but if you have a movie which has a lot of rows (in my case chapters, specials, cast, trailers, media info) and you scroll to the bottom and then up, the action buttons are duplicated.

AndreasGB:
Also, I'm not quite so fond of the new usage of space, I prefer the way the leanback version looked.

nielsvanvelzen:
> if you have a movie which has a lot of rows and you scroll to the bottom and then up, the action buttons are duplicated.

This is fixed now, good find!

app/src/main/java/org/jellyfin/androidtv/details/DetailsOverviewPresenter.kt
```diff
@@ -52,17 +68,41 @@ class DetailsDescriptionPresenter : Presenter() {
...
+		row as DetailsOverviewRow
+		val item = row.item
+
+		// banner
+		//todo hide banner view when none found, support multiple banners
+		item.images.backdrops.firstOrNull()?.let {
+			// Android doesn't crop automatically but Glide does
+			// Picasso can also do this but doesn't read the XML attributes of the target view for it
+			// so the way Glide does it is preferred to avoid duplicate settings
+			Glide.with(viewHolder.view.context).load(it.url).into(viewHolder.banner)
```
AndreasGB: We probably should decide on one image handling library before we ship this

app/src/main/java/org/jellyfin/androidtv/details/actions/ActionAdapter.kt
```diff
@@ -1,86 +1,74 @@
...
+		viewHolder.button.setOnClickListener { view ->
+			GlobalScope.launch(Dispatchers.Main) {
+				action.onClick(view)
 			}
 		}
 
-		override fun onUnbindViewHolder(viewHolder: ViewHolder) {
-			viewHolder as ActionViewHolder
+		// Set state so the observers initialize
+		viewHolder.lifecycle.currentState = Lifecycle.State.STARTED
```
AndreasGB: Should this be changed on unbind?

app/src/main/java/org/jellyfin/androidtv/details/DetailsOverviewPresenter.kt
nielsvanvelzen: I made a [separate issue](https://github.com/jellyfin/jellyfin-androidtv/issues/403) for this.

PR #381, opened 2020-03-07
Use system provided light font instead of Roboto-Light

nielsvanvelzen:
Uses the system provided sans-serif-light font (which by default is Roboto Light) instead of hardcoding our own (outdate) version of Roboto Light.

For most devices everything will look exactly the same, devices using a different system font will look a bit different.

When possible the font family declaration is done in XML. Files touched have their author comments removed and imports optimized.


app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -267,18 +266,12 @@ public void onActivityCreated(Bundle savedInstanceState) {
         setupPopupAnimations();
 
         //live guide
-        Typeface roboto = TvApp.getApplication().getDefaultFont();
         mDisplayDate = mActivity.findViewById(R.id.displayDate);
         mGuideTitle = mActivity.findViewById(R.id.guideTitle);
-        mGuideTitle.setTypeface(roboto);
         mGuideCurrentTitle = mActivity.findViewById(R.id.guideCurrentTitle);
-        mGuideCurrentTitle.setTypeface(roboto);
```
AndreasGB: Missing xml adjustment?

app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -267,18 +266,12 @@ public void onActivityCreated(Bundle savedInstanceState) {
...
         mSummary = mActivity.findViewById(R.id.summary);
-        mSummary.setTypeface(roboto);
```
AndreasGB: Missing xml adjustment?

app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
nielsvanvelzen: We search for the view in the `mActivity` which is `PlaybackOverlayActivity`. This inflates the `activity_playback_overlay` layout. This uses the `CustomPlaybackOverlayFragment` fragment (at this point I'm wondering why it doesn't use `this` instead of `mActivity`) which inflates `vlc_player_interface`.

I can't find where it exactly initializes the XML but the **only** file containing an id `guideCurrentTitle` was `overlay_tv_guide` so I adjusted the XML in that file.


TL;DR:: everything should be set for those 2

app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
nielsvanvelzen: See other comment

app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
AndreasGB: I don't see it [there](https://github.com/jellyfin/jellyfin-androidtv/blob/614865563c72b66e7ab9175d85e292af61e9ae44/app/src/main/res/layout/overlay_tv_guide.xml#L196), or am I misunderstanding this?

nielsvanvelzen: You're right, I somehow missed that 2 times lol

AndreasGB: Happens :D

Issue #313, opened 2020-02-07
Unable to Play Video with .ASS Subtitles


anthonylavado:
From u/PixelDew on [Reddit](https://www.reddit.com/r/jellyfin/comments/ezgy3q/new_android_tv_release_v0110_improved_video/fgstc8d/):

> Ever since the update, my devices don't play files on mpv correctly. I also get an error about duration and then it skips the file. Before the update i could use MPV as an external  player for .ASS subtitles and everything would work well! is there a workaround for this?

> On my firestick, after the update, the ui hasn't changed (not sure if its a problem) but i cant play anything using mpv, it gives me the error "The item didn't appear to play as long as its duration." If i attempt to let it automatically choose a player, it says "Player error encountered" a few times, then stops the playback due to too many error.
>
>On my android phone, The new UI is present, but it wont even let me use mpv as an external player, and if i use the default it plays the file alright.
>
>Overall, as i only use my firestick, this update has made most of my library unusable :/

anthonylavado:
From u/linetrimmer (same thread):
> I’m having the same issue with files with .ass subtitles. Used to be able to use mx player, now it says can’t play this link. Exoplayer tried to play it but then says error and retries a couple times and exits. Reverting to 10.2 and everything’s fine

anthonylavado:
From u/PixelDew (same thread):
Yep, after sideloading MPV again, sideloading the latest version of jellyfin, I finally tried the Nov 9th version and everything works again on the firestick! Thanks!

>I hope it's fixed in the upcoming update, because MPV seems like the only player that supports h.264 10 bit with .ASS subtitles. Looking forward to the UI update too! :)

PixelDew121:
I only tested two files, so i can't say if .ASS subtitles are the reason behind this, but both of them work through MPV in the previous version, and the non 10 bit file works with the default player in the old version.
[File 1](https://pastebin.com/9Uc1QcAz)
[File 2](https://pastebin.com/NhXVszjS)

linetrimmer:
It’s pretty much all of my anime files that I’m experiencing this issue with. The issue is present in 10 bit, 8 bit, h264, and hevc files, so the only constant that I’m seeing is the subtitles. 

When MX Player attempts to play the file, it loads for a few seconds before saying, “Can’t play this link.” The same file plays in MX Player with no issue in 10.2 and prior versions.

[...]

PR #395, opened 2020-03-12
Update ProfileHelper.java to embed .ASS subtitles in external players

linetrimmer:
Leaving the subtitle delivery method as encode caused external players to crash with .ass subtitles.


nielsvanvelzen:
Closes #313

PR #396, opened 2020-03-13
Improve subtitle rendering

thornbill:
* Fix whitespace rendering issues in subtitles
* Remove position calculation
* Improve styling of subtitles
  * Normal weight font instead of bold
  * Use a less huge font size
  * Use background color instead of shadow
  * Left align multiple lines


AlexHFOkumura:
please, i would like to adjust the size of subtitles like on kodi, because i need very big subtitles to be able to see, i use mi box with android tv, thanks!

nextlooper42:
Maybe we should take advice from Netflix size implementation of subtitles, i find them somehow just right size on all devices. As default/recommanded size. Size right now is just freakishly big tbh and it refuses to change no matter what.

PR #397, opened 2020-03-13
External trailers 🎬 2

AndreasGB:
After the architecture of the first attempt at implementing external trailers was less than ideal, I'm opening this new PR with a new proposed architecture that should allow for better testability of the individual components.

It's by far not fully implemented and I'm particularly looking for feedback on the general architecture.


app/src/main/java/org/jellyfin/androidtv/model/trailers/lifter/FirstMatchMultiExternalTrailerLifter.kt
```diff
@@ -0,0 +1,31 @@
+package org.jellyfin.androidtv.model.trailers.lifter
+
+import org.jellyfin.androidtv.model.trailers.external.ExternalTrailer
+import org.jellyfin.apiclient.model.entities.MediaUrl
+import java.util.*
+
+class FirstMatchMultiExternalTrailerLifter : ExternalTrailerLifter() {
```
nielsvanvelzen: This name is looooong

AndreasGB:
It's now basically implemented for YouTube, but missing the generic case as well as lacking a little polishing still. 

Can you please have another look, especially on the tests and lifting?

app/src/main/java/org/jellyfin/androidtv/details/trailers/YouTubeTrailerPresenter.kt
```diff
@@ -0,0 +1,78 @@
...
+import com.bumptech.glide.request.RequestListener
+import com.bumptech.glide.request.target.Target
+import kotlinx.android.synthetic.main.multi_badge_image_card_view.view.*
+import org.jellyfin.androidtv.R
+import org.jellyfin.androidtv.base.IItemClickListener
+import org.jellyfin.androidtv.model.trailers.external.YouTubeTrailer
+import org.jellyfin.androidtv.ui.MultiBadgeImageCardView
+import org.jellyfin.androidtv.util.ImageUtils
+
+class YouTubeTrailerPresenter(private val context: Context, private val imageHeight: Int, private val safeBrandingCompliance: Boolean) : Presenter(), IItemClickListener {
```
nielsvanvelzen: When a class (or function for that matter) gets multiple parameters the code might look a bit messy. A way to prevent this is by putting each parameter on it's own line:
```kotlin
class YouTubeTrailerPresenter(
	private val context: Context,
	private val imageHeight: Int,
	private val safeBrandingCompliance: Boolean
) : Presenter(), IItemClickListener {
```

app/src/main/java/org/jellyfin/androidtv/details/trailers/YouTubeTrailerPresenter.kt
```diff
@@ -30,11 +39,29 @@ class YouTubeTrailerPresenter(private val context: Context, private val imageHei
 			contentText = context.getString(R.string.brand_youtube)
 		}
 
-		// TODO: Kick off loading of thumbnail here
-	}
+		if (!safeBrandingCompliance) {
+			Glide.with(context).load(trailer.thumbnailURL).listener(object : RequestListener<String, GlideDrawable> {
```
nielsvanvelzen: IIRC you can use `.get()` instead of the listener when using a coroutine (GlobalScope.launch thingy...) If you want to catch the errors just use `try {} catch {}`

app/src/test/java/org/jellyfin/androidtv/model/trailers/lifter/YouTubeLifterUnsupportedURLsTest.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.junit.runners.Parameterized
+
+
+@RunWith(Parameterized::class)
+class YouTubeLifterUnsupportedURLsTest(private val testURL: String) {
+	private val lifter = YouTubeExternalTrailerLifter()
+
+	@Test
+	fun shouldIgnore() {
+		Assert.assertEquals(
```
nielsvanvelzen: Formatting looks weird for this whole statement

app/src/main/java/org/jellyfin/androidtv/model/trailers/lifter/ExternalTrailerLifter.kt
```diff
@@ -1,15 +1,28 @@
 package org.jellyfin.androidtv.model.trailers.lifter
 
-import android.net.Uri
 import org.jellyfin.androidtv.model.trailers.external.ExternalTrailer
 import org.jellyfin.apiclient.model.entities.MediaUrl
+import java.net.MalformedURLException
+import java.net.URL
 
 abstract class ExternalTrailerLifter {
 	private val domainRegex = Regex("""^(.*\.|)(.+\...+)$""")
```
nielsvanvelzen: I'm not really sure what this regex does, strip subdomains?

app/src/main/java/org/jellyfin/androidtv/model/trailers/lifter/ExternalTrailerLifter.kt
```diff
@@ -1,15 +1,28 @@
...
 
-	protected fun getDomain(uri: Uri): String? = uri.host?.let { host -> domainRegex.find(host)?.groups?.get(2)?.value }
+	protected fun getDomain(url: URL): String? {
+		return url.host?.let { host -> domainRegex.find(host)?.groups?.get(2)?.value }
```
nielsvanvelzen: I've once read that the URI class is preferred above URL because URL will make DNS queries but I can't find a proper source for it right now.

app/src/main/java/org/jellyfin/androidtv/model/trailers/lifter/YouTubeExternalTrailerLifter.kt
```diff
@@ -1,21 +1,58 @@
...
 class YouTubeExternalTrailerLifter : ExternalTrailerLifter() {
+	private val LOG_TAG = "YouTubeLifter"
+
 	private val youtubeDomains = arrayListOf("youtube.com", "youtu.be")
 
-	private fun isYoutubeUrl(uri: Uri) = youtubeDomains.contains(getDomain(uri))
-	private fun isYoutubeUrl(url: MediaUrl) = isYoutubeUrl(Uri.parse(url.url))
+	private fun isYoutubeUrl(url: URL) = youtubeDomains.contains(getDomain(url))
+
+	private val videoParameterRegex = Regex("v=([a-zA-Z0-9]+)")
```
nielsvanvelzen: An YouTube id is always 11 characters so you can make the regex more strict

app/src/main/java/org/jellyfin/androidtv/model/trailers/lifter/YouTubeExternalTrailerLifter.kt
```diff
@@ -1,21 +1,58 @@
...
+				if (matchGroups == null || matchGroups.size != 2) {
+					Log.e(LOG_TAG, "URL $convertedURL is not supported")
+					return null
+				} else {
+					matchGroups[1]!!.value
+				}
+			}
+
+			"youtu.be" -> {
+				convertedURL.path.subSequence(1, convertedURL.path.length)
```
nielsvanvelzen: I'd say use the last 11 characters from the path in the case a url looks like `youtu.be/sample/[id]`

app/src/main/java/org/jellyfin/androidtv/model/trailers/lifter/ExternalTrailerLifter.kt
AndreasGB: That would be really unexpected, but idk. Maybe I should try to wireshark the emulator.

app/src/main/java/org/jellyfin/androidtv/model/trailers/lifter/ExternalTrailerLifter.kt
AndreasGB: That was the idea. Should I write test cases for it as well?

nielsvanvelzen: You can probably do it without the regex

app/src/main/java/org/jellyfin/androidtv/model/trailers/lifter/ExternalTrailerLifter.kt
nielsvanvelzen: Using the Android URI class will give you some additional methods like getting a query parameter too (to replace the `v=xxx` regex):
https://developer.android.com/reference/android/net/Uri

app/src/main/java/org/jellyfin/androidtv/model/trailers/lifter/ExternalTrailerLifter.kt
AndreasGB: Got rid of it

AndreasGB:
As the saying goes: Was lange währt wird endlich gut

I think it's ready now to be reviewed :D

app/src/main/java/org/jellyfin/androidtv/details/trailers/ExternalTrailerPresenter.kt
```diff
@@ -0,0 +1,55 @@
...
+			setMainImageDimensions((ImageUtils.ASPECT_RATIO_16_9 * imageHeight).toInt(), imageHeight)
+			mainImageDrawable = thumbnail
+			titleText = trailer.name
+			contentText = description
+		}
+	}
+
+	override fun onUnbindViewHolder(viewHolder: ViewHolder) {
+		// do nothing
+	}
```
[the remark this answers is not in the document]
AndreasGB: This has to be implemented as the interface requires it.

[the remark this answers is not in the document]
nielsvanvelzen: Right, I wasn't 100% sure about it

AndreasGB:
Lol, should have run the Unit tests before pushing :D

PR #401, opened 2020-03-15
Add theme support

thornbill:


app/src/main/java/org/jellyfin/androidtv/preferences/enums/AppTheme.kt
```diff
@@ -0,0 +1,18 @@
+package org.jellyfin.androidtv.preferences.enums
+
+enum class AppTheme {
```
nielsvanvelzen: Can we rename the themes (or the enums) to have consistent naming?

It doesn't make sense to me like this:
- Dark, Theme_Jellyfin
- Emerald, Theme_Jellyfin_Legacy
- Hotdog, Theme_Jellyfin_HotDogStand

app/src/main/res/xml/preferences.xml
```diff
@@ -26,6 +26,11 @@
     </PreferenceCategory>
 
     <PreferenceCategory android:title="@string/pref_general">
+        <ListPreference
+            android:entries="@array/app_theme_entries"
+            android:entryValues="@array/app_theme_values"
+            android:key="app_theme"
+            android:title="App Theme" />
```
nielsvanvelzen: Missing `@string`

app/src/main/java/org/jellyfin/androidtv/base/AppThemeCallbacks.kt
```diff
@@ -0,0 +1,54 @@
...
+				is PreferencesActivity -> lastPreferencesTheme = it
+				else -> lastTheme = it
+			}
+		}
+	}
+
+	override fun onActivityResumed(activity: Activity) {
+		val lastThemeForActivity = if (activity is PreferencesActivity) lastPreferencesTheme else lastTheme
+		TvApp.getApplication().userPreferences.appTheme.let {
+			if (lastThemeForActivity != null && lastThemeForActivity != it) {
```
nielsvanvelzen: Isn't it possible to use `activity.setTheme()` here?

thornbill: The activity has to be recreated for the theme to be applied.

app/src/main/res/values/array.xml
```diff
@@ -3,6 +3,14 @@
     xmlns:tools="http://schemas.android.com/tools"
     tools:ignore="MissingTranslation">
 
+    <string-array name="app_theme_entries">
+        <item>@string/pref_theme_dark</item>
+        <item>@string/pref_theme_emerald</item>
+    </string-array>
+    <string-array name="app_theme_values">
```
nielsvanvelzen: You forgot to update these with the enum change. (They should be exactly the same as the enums)

app/src/main/res/xml/preferences.xml
```diff
@@ -26,6 +26,12 @@
     </PreferenceCategory>
 
     <PreferenceCategory android:title="@string/pref_general">
+        <ListPreference
+            android:defaultValue="Theme_Jellyfin"
+            android:entries="@array/app_theme_entries"
+            android:entryValues="@array/app_theme_values"
+            android:key="app_theme"
+            android:title="@string/pref_app_theme" />
```
nielsvanvelzen: Adding `app:useSimpleSummaryProvider="true"` will show the currently selected theme as "description".

PR #404, opened 2020-03-22
Show progress on poster

nielsvanvelzen:
image

In the future I'd like to use this design in other places too but the margins were causing issues on the `MyImageCardView`.

It's partially based on the next up progress bar (shape) and the MyImageCardView progress bar (color)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/details/DetailsOverviewPresenter.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/DetailsOverviewPresenter.kt
@@ -7,6 +7,7 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.ImageView
 import android.widget.LinearLayout
+import android.widget.ProgressBar
 import android.widget.TextView
 import androidx.core.text.bold
 import androidx.leanback.widget.RowPresenter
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/jellyfin/androidtv/details/DetailsOverviewPresenter.kt
```diff
@@ -149,6 +151,16 @@ class DetailsOverviewPresenter : RowPresenter() {
 		}
 
 		if (item is PlayableItem) {
+			// (userData.getPlaybackPositionTicks() * 100.0 / itemDto.getRunTimeTicks()
+			val progress = (item.playbackPositionTicks * 100.0) / (item.durationTicks ?: 0)
```
AndreasGB: I don't like dividing by zero on purpose, moving the null-check outside feels cleaner to me.

nielsvanvelzen: It would totally crash the application lol, I didn't' even remove the comment with the java code. Something went wrong when reviewing before commiting.

dkanada:
I'm personally not a fan of the padding or rounded corners, since it doesn't seem to match material design very well. It also might be more granular if you somehow used the whole page length instead, maybe on the banner above, but I have no idea how that would work.

nielsvanvelzen:
> I'm personally not a fan of the padding or rounded corners, 

I actually think it looks quite nice. It uses transparency causing it to blend in a bit with the posters and the padding makes it look as an on-top element (instead of beneath). One thought I had was using the androidx.palette library to calculate the color to use but I didn't try it.

> since it doesn't seem to match material design very well.

It wasn't my intention to follow material design guidelines at all. Personally I'm not aiming to adhere to it.

> It also might be more granular if you somehow used the whole page length instead, maybe on the banner above, but I have no idea how that would work.

I places it inside the poster to stay consistent with other places that show a thumbnail/poster. Although the design of the progress bar is a bit different the location stays the same

PR #407, opened 2020-03-24
Update output filename to published format

thornbill:
This was being renamed in the build scripts, but it probably makes more sense to have in Gradle.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -41,6 +41,16 @@ android {
 			resValue("string", "app_name", "@string/app_name_debug")
 		}
 	}
+
+	applicationVariants.all {
+		val variant = this
+		variant.outputs.all {
+			val output = this as com.android.build.gradle.internal.api.BaseVariantOutputImpl
+			output.outputFileName = output.outputFileName
+				.replace("app-", "jellyfin-androidtv_")
+				.replace(".apk", "_${variant.versionName}.apk")
+		}
+	}
 }
 
 dependencies {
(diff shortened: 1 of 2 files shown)


thornbill:
My initial approach on this was to simply use the following:

```kotlin
outputFileName = "jellyfin-android_${variant.name}_${variant.versionName}.apk"
```

The problem with this approach was that getting the signed status of the release build seemed overly complex vs just replacing the relevant parts of the default output.

nielsvanvelzen:
One easy way to change the file name is to set the `archivesBaseName` property in the `android {}` block: `setProperty("archivesBaseName", "jellyfin-androidtv_${versionName}")` will result in something like `jellyfin-androidtv_0.11.0-debug.apk` or `jellyfin-androidtv_0.11.0-release.aab`.

But when doing some searching I found [this](https://developer.android.com/studio/known-issues?hl=ro#variant_output) part of the Android Studio documentation noting that some Gradle tasks can break when the naming is changed. Therefore I think it's better to keep it in the shell scripts.

thornbill:
> Using the Variant API to manipulate variant outputs is broken with the new plugin. It still works for simple tasks, such as changing the APK name during build time, as shown below:

That sounds like our usage here is fine, but more complex changes to outputs could cause issues? I am pretty strongly in favor of moving this to Gradle to make the CI pipeline as simple as possible as we move towards using it for publishing.

nielsvanvelzen:
I will approve it for now but we should be aware that it might break

PR #427, opened 2020-04-11
Fix Exoplayer Caching

### LINKED ISSUE #417: AndroidTV app complains of "low storage" after live tv playback of several hours
I'm noticing that after a 2-3+ hours of watching live tv, my fire tv stick will complain that it is almost out of storage and suggests that I remove apps.

image

The thing is -- Jellyfin is the ONLY app installed on my Fire TV stick.  And if I select the remove apps button, the only one I can remove (that isn't built-in) is Jellyfin itself!

If I hit "not now", I can start watching live tv again without issue, and the fire tv doesn't appear to have full storage.

I'm wondering if there is some sort of buffer on the fire tv that is getting filled up and causing this issue.  It seems like the moment this pop-up happens, since it interrupts the live tv stream, I'm guessing that Jellyfin app clears whatever is eating up all of the storage?

Any advice?  Let me know if I can help by providing more info.

Thanks in advance.

thornbill:
It turns out Picasso was actually using OkHttp through reflection. It seems that without it, Picasso was setting up some crazy caching to disk behavior on the default HTTP stack. This caused streams played by Exoplayer to be cached to disk when we updated Exoplayer from 2.10.6 to 2.11.3.

what

Fixes #417

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -87,7 +87,7 @@ dependencies {
 
 	// HTTP utility
 	// NOTE: This is used by Picasso through reflection and can cause weird caching issues if removed!
-	val okhttpVersion = "2.7.0"
+	val okhttpVersion = "2.7.5"
 	implementation("com.squareup.okhttp:okhttp:$okhttpVersion")
 	implementation("com.squareup.okhttp:okhttp-urlconnection:$okhttpVersion")
 
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -86,6 +86,7 @@ dependencies {
 	implementation("com.flaviofaria:kenburnsview:1.0.6")
 
 	// HTTP utility
+	// NOTE: This is used by Picasso through reflection and can cause weird caching issues if removed!
 	val okhttpVersion = "2.7.0"
 	implementation("com.squareup.okhttp:okhttp:$okhttpVersion")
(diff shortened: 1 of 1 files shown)


app/build.gradle.kts
```diff
@@ -85,6 +85,12 @@ dependencies {
 	implementation("com.github.bumptech.glide:glide:3.7.0")
 	implementation("com.flaviofaria:kenburnsview:1.0.6")
 
+	// HTTP utility
+	// NOTE: This is used by Picasso through reflection and can cause weird caching issues if removed!
+	val okhttpVersion = "2.7.0"
```
nielsvanvelzen, line 90: Assuming we use 2.x because the libraries _might_ not work with newer versions. The latest 2.x version is [2.7.5](https://square.github.io/okhttp/changelog_2x/) so this needs to be bumped.

thornbill: I just reverted the original removal. I will get this updated later today.

thornbill: Version has been bumped.

samwathegreat:
Hi folks.  Forgive my ignorance, but do you know when this fix will be available in an updated version that I can download from the amazon appstore?

Thanks in advance

nielsvanvelzen:
We're hoping to have this fix out in the next few days @samwathegreat

PR #430, opened 2020-04-11
fix translation erro - pt-BR

tayhr:


Vitorvlv:
Wouldn't it be better to fix this through weblate?

AndreasGB:
Thank you very much for your contribution! 

@Vitorvlv is correct though, translation changes need to be submitted via https://translate.jellyfin.org/, please re-submit it there. I will close this PR :)

PR #432, opened 2020-04-13
Add noisy background for "BrowseHeaders" in home screen

nielsvanvelzen:
_**10% more cool**_

Uses the noise drawable used in the "play next" thingy as background for the sidebar in the home screen. This will make the backdrop visible and make the app look 10% more cool.

Before:  
image  
After:  
image

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -45,4 +45,8 @@
         <item name="android:paddingEnd">16dp</item>
         <item name="android:minHeight">0dp</item>
     </style>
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
It's way better than the grayish color, I tried semi-transparent black first but IMO the grain adds a nice effect to it. Ideally I'd like to use some blur but that's too hard to do in Android :(

AndreasGB:
Yeh, blur seems to be Apple's thing so I doubt Google will adopt it soon

PR #434, opened 2020-04-15
Add androidx.fragment as direct dependency, update androidx.preference & update kotlin/android plugins

nielsvanvelzen:
I noticed we don't depend directly on the androidx.fragment library but we do use it's classes. It's probably exposed by some other androidx library. I now added it so we can be sure it's the latest version. The updated plugins are just patch version so nothing really changed.

Because the androidx.fragment library got bumped to a newer version additional linter checks are automatically enabled. They pass on master but will fail on the details branch, therefore when this PR is merged I will rebase the details branch and submit a PR for fixing the linter.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -65,10 +65,11 @@ dependencies {
 
 	// Android(x)
 	implementation("androidx.core:core-ktx:1.2.0")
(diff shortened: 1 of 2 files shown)


thornbill:
The Code Climate failure is because I enabled the sonar-java plugin. It considers everything it found as a new issue.

PR #443, opened 2020-04-18
Music artist details

nielsvanvelzen:


AndreasGB:
I've discovered a bug:

Steps:
* Open Music -> Artists -> (some artist) -> Shuffle -> music starts playing
* pause playback from some other screen
* Open an artist -> Shuffle -> no music starts playing 😢

nielsvanvelzen:
When something is playing it will show an popup asking to replace the queue or to append to it. Unfortunately this uses the dirty "currentActivity" property in TvApp.java, this is normally done by BaseActivity. I can add a statement to the DetailsActivity to replace the currentActivity value to fix it but it definitely needs to be properly fixed later.

nielsvanvelzen:
Another way would be to always replace the queue for now and think about a better solution in 0.13

AndreasGB:
I think always replacing is the more expected result, as other apps seem to do this as well (e.g. Spotify, unless the user has manually queued songs... :D)

nielsvanvelzen:
I made the change in #447 so it will find its way to the details branch after the next rebase

PR #448, opened 2020-04-20
Startup layout redesign

nielsvanvelzen:
Before  
image  
After  
image

Bonus: Updated the logo to a vector version to get some extra jelly out of 4K televisions


thornbill:
Can we remove the text all together? I didn't care for it in the old loading screen either.

Also yay for vector graphics! :tada:

nielsvanvelzen:
I didn't really care too much about the text (although I did change it) so it's now removed. I can't really fix the strings.xml update now (weblate always removes the line break at the end but my editor is set to always add it because of editorconfig)

PR #449, opened 2020-04-21
Add issue and pull request templates

nielsvanvelzen:
**Changes**
- Add 3 issue templates: bug, feature & playback issue
- Add 1 pull request template (this PR uses it!)
- Add issue template config to provide link to our documentation

**Issues**
- Fixes #436 

Note:
- Added a new label "playback" to the repository for the playback template, this label should be used for issues related to playback

PR #453, opened 2020-04-22
Fix pull request template not working

nielsvanvelzen:
The pull request template didn't work after merging #449. The [documentation](https://help.github.com/en/github/building-a-strong-community/creating-a-pull-request-template-for-your-repository) isn't that clear on how it should work but I think using a template directory is only meant when specifying the template to use in the URL. To use a default template it should just be a file named `pull_request_template.md`.

**Changes**
- Renamed pull request template to make it work

**Issues**
- none


nielsvanvelzen:
I can confirm that it works now 🎉

Issue #416, opened 2020-04-03
After stopping live TV, androidtv often does not tell jellyfin server that playback has stopped


samwathegreat:
....and as a result Jellyfin server will continue to "direct stream" the channel for hours and hours well after the fire tv (android tv) stick is no longer watching something.  This results in wasted CPU cycles, and dozens of gigabytes of space wasted to direct stream in the transcodes folder.

Here is my original report that I mistakenly put under jellyfin server, but actually seems to be an issue with androidtv specifically:

Basically, I'm having an issue where one of my clients (Fire TV stick) can be watching live TV via my HDHomeRun. Most channels are available via direct stream. I've noticed that while I'm direct-streaming live tv, Jellyfin is repacking the stream using the defined "transcodes" folder.

Well what is happening is that even well after jellyfin (on Fire TV) has stopped watching the channel, or even watching something different entirely, ffmpeg continues to re-pack the stream (and keep the hdhomerun tuner locked) taking up gigabytes of space in the transcodes folder.

The only way I know to get rid of them is to restart jellyfin -- then the transcode stops immediately, and jellyfin even deletes the massive files for me automatically.

Also, if it isn't obvious, Jellyfin is still using an additional tuner of the HDHomeRun while it continues to repack the stream for the channel that is no longer being watched.

So basically, it's a big waste of resources and space. Fortunately in my situation, direct stream takes very little CPU, but it sure is using up tons of space in my transcodes folder.

Here you can see that jellyfin has been direct-streaming this channel since around 10PM last night (over 15 hours ago) and has consumed nearly 30 gigabytes of storage.

![image](https://user-images.githubusercontent.com/4813336/78376313-21c30b80-759c-11ea-8333-0f3529970505.png)

I've attached the two most recent logs as well. TBH, I'm having difficulty determining which logs would be useful to you -- especially the directstream and transcode logs. If you need any, let me know. I'm assuming the jellyfin daily log files are what are needed.

Please advise. It gets old having to check the transcodes folder on a daily basis and restart jellyfin each time something is amiss. Additionally, this frequently results in me not being able to watch something I want because all tuners are in use.

Thanks

[jellyfin20200401.log](https://github.com/jellyfin/jellyfin-androidtv/files/4428189/jellyfin20200401.log)
[jellyfin20200402.log](https://github.com/jellyfin/jellyfin-androidtv/files/4428190/jellyfin20200402.log)

samwathegreat:
[...]

PR #454, opened 2020-04-23
Stop progress report if playback was already stopped

Froghut:
It often happened to me that after stopping playback on the android tv client I found the media still shown with progress where I stopped it in the Dashboard. After debugging it seems that sometimes the report progress loop would run one more time after the stop method was called. I'm not sure if there is a better method of preventing that, but seeing as the loop right above this one also checks the mPlaybackState I thought this might be fine.


nielsvanvelzen:
Closes #416

PR #457, opened 2020-04-25
Remove some currentActivity usages

nielsvanvelzen:
**Changes**
- Remove more usages of the currentActivity property in Tvapp
- Replaced the "showMessage" calls with toasts (I don't think I ever saw one of these messages lol)
- Some miscellaneous cleaning (formatting/removing unused functions) in the smaller files I touched

**Issues**

**Notes**
- Unfortunately the usages now left are a lot harder to fix (needing activity instance in views for example)
- I did not touch the usages inside the playback code
- Most changes have their own commits

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/presentation/ChannelCardPresenter.java
+++ b/app/src/main/java/org/jellyfin/androidtv/presentation/ChannelCardPresenter.java
@@ -1,47 +1,29 @@
 package org.jellyfin.androidtv.presentation;
 
-import android.content.Context;
-import androidx.leanback.widget.Presenter;
 import android.view.View;
 import android.view.ViewGroup;
 
-import org.jellyfin.androidtv.TvApp;
-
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/jellyfin/androidtv/presentation/ChannelCardPresenter.java
```diff
@@ -60,12 +42,6 @@ public void onBindViewHolder(Presenter.ViewHolder viewHolder, Object item) {
 
     @Override
     public void onUnbindViewHolder(Presenter.ViewHolder viewHolder) {
-        //Log.d(TAG, "onUnbindViewHolder");
+        // Unused
```
thornbill, line 45: Why keep this override if it is unused?

nielsvanvelzen: It's abstract in the parent so it needs to be implemented

PR #465, opened 2020-04-29
Fix playback activity crashing

nielsvanvelzen:
**Changes**
- Cleaned the cleaned code
- When creating the view of the fragment it will now **not** attach to the root ([as it should be](https://developer.android.com/guide/components/fragments#UI))
  - This caused a new issue
- Moved the code that created the videomanager and pushed it to the playbackcontroller
  - It's now created as soon as the view in the fragment is created ([lifecycle](https://developer.android.com/guide/components/fragments#Creating))
  - This also cleans the activity a bit (which is nice because it will be easier to switch to less activities in the future)

**Issues**
- none!

🤡


nielsvanvelzen:
This fixes an issue caused by #458

PR #476, opened 2020-05-15
More cleaning in TvApp.java

nielsvanvelzen:
In my effort to clean the TvApp class I present to you:

**Changes**
- TvApp.java
  - Removed the "DisplayPriority" functionality
    - It's value was never set (always null), non-configurable
    - It was used in just 1 place
  - Removed "getServerBitrateLimit" 
    - It would always return 100000000, hardcoded that integer at the place it was used
  - Removed serverConfiguration, audioManager & currentPlayingItem
    - Never set or used
  - Removed currentBackgroundGradient
    - This includes the setter for the background "setCurrentBackground"
    - It was used to create a slight gradient as background row rows in certain screens but was barely visible and the implementation was ugly.
  - Removed playingIntros
    - This one was awesome, it was set at multiple places but the only place that used it would always sett he value to false if it was true. That's it. Peak programming was reached
  - Removed "showSearch"
    - Just create the intent at the places you want to show search, it's not that complicated
- Other files
  - Small changes like removal of unnecessary stuff

**Issues**

**Note:** Changes are separated by commit, I suggest reviewing per commit instead of the final diff. This way it's easier to track changes between different files


thornbill:
It looks like you also removed the code that cycles through multiple backdrops for items. Is that right?

nielsvanvelzen:
If you mean the code in FullDetailsActivity.java, it didn't actually update the background image, the picasso target only invoked the gradient thingy. Also, that file will be replaces in the next version so it's not that bad if some secondary functions break

PR #478, opened 2020-05-19
Fix fast forward/rewind not working on Fire Stick

### LINKED ISSUE #351: Fast forward and rewind buttons no longer work with new player overlay
It seems the fast forward and rewind buttons on Fire TV remotes no longer work after the player overlay redesign.

MrLemur:
**Changes**
This adds in some new logic to intercept the fast forward and rewind
keys presses if the popup panel is not in view.

If the media playing is not live TV, it will skip forwards or backwards
the appropriate amount of seconds. It will show the overlay as expected]
and then initiate the fadeout timer.

There is also a filter that has been added to only intercept the `ACTION_DOWN` of the keypress. Before, this would cause a fast forward/rewind to be done twice, effectively doubling the intended seek time.

**Issues**
Fixes #351 

I would appreciate any testing on other Fire Sticks, this has only been tested on a Fire Stick 4K.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
@@ -432,6 +432,7 @@ public class CustomPlaybackOverlayFragment extends Fragment implements IPlayback
     private View.OnKeyListener keyListener = new View.OnKeyListener() {
         @Override
         public boolean onKey(View v, int keyCode, KeyEvent event) {
+            if (event.getAction() == KeyEvent.ACTION_DOWN) {
             leanbackOverlayFragment.setShouldShowOverlay(true);
             if (keyCode == KeyEvent.KEYCODE_MEDIA_STOP && mActivity != null && !mActivity.isFinishing()) {
                 mActivity.finish();
@@ -545,7 +546,7 @@ public class CustomPlaybackOverlayFragment extends Fragment implements IPlayback
                 //and then manage our fade timer
                 if (mFadeEnabled) startFadeTimer();
             }
-
+            }
             return false;
         }
     };
--- a/app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
@@ -501,6 +501,23 @@ public class CustomPlaybackOverlayFragment extends Fragment implements IPlayback
                     }
                 }
 
+                // Control fast forward and rewind if overlay hidden and not showing live TV
+                if (!mPlaybackController.isLiveTv()) {
+                    if (keyCode == KeyEvent.KEYCODE_MEDIA_FAST_FORWARD || keyCode == KeyEvent.KEYCODE_BUTTON_R1 || keyCode ==KeyEvent.KEYCODE_BUTTON_R2) {
+                        mPlaybackController.skip(30000);
+                        if (!mIsVisible) show();
+                        setFadingEnabled(true);
+                        return true;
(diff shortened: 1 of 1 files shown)


MrLemur:
@thornbill @nielsvanvelzen I understand you might have some devices to test with?

app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -432,6 +432,7 @@ public void onItemClicked(Presenter.ViewHolder itemViewHolder, Object item,
     private View.OnKeyListener keyListener = new View.OnKeyListener() {
         @Override
         public boolean onKey(View v, int keyCode, KeyEvent event) {
+            if (event.getAction() == KeyEvent.ACTION_DOWN) {
```
nielsvanvelzen, line 435: The formatting inside this if statement is wrong. Please indent it or change the code to return false when the action is not ACTION_DOWN

app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -501,6 +502,23 @@ public boolean onKey(View v, int keyCode, KeyEvent event) {
...
+
+                    if (keyCode == KeyEvent.KEYCODE_MEDIA_REWIND || keyCode == KeyEvent.KEYCODE_BUTTON_L1 || keyCode == KeyEvent.KEYCODE_BUTTON_L2) {
+                        mPlaybackController.skip(-11000);
+                        if (!mIsVisible) show();
+                        setFadingEnabled(true);
+                        return true;
+                    }
+                }
+
                 if (!mIsVisible) {
```
nielsvanvelzen, line 522: I expect everything inside this statement to probably trigger twice on non-firetv devices

app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
MrLemur: What are the current indent preferences for java files?

app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
MrLemur: Will be good to get a test to see, and if so I can add in the fire TV check.

app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
nielsvanvelzen: It's 4 spaces currently

app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
MrLemur: Testing in Android TV emulator in AVD has the expected behaviour using the emulated d-pad fast forward and rewind controls.

app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
MrLemur: Should be fixed with the most recent commit

app/src/main/java/org/jellyfin/androidtv/playback/CustomPlaybackOverlayFragment.java
nielsvanvelzen: I did some testing and it, indeed, works as expected! 🎉

PR #481, opened 2020-05-20
Add device model to app settings about section

thornbill:
**Changes**
Adds the device model to the "About" section of app settings under the app version. This can be useful information for debugging.

**Issues**
N/A


app/src/main/java/org/jellyfin/androidtv/preferences/ui/UserPreferencesFragment.kt
```diff
@@ -142,6 +143,10 @@ class UserPreferencesFragment : LeanbackSettingsFragmentCompat() {
 				Utils.getVersionString()
 			}
 
+			findPreference<EditTextPreference>("device_model")?.summaryProvider = Preference.SummaryProvider<EditTextPreference> {
+				Build.MODEL
```
nielsvanvelzen, line 147: We might want to add the manufacturer too. Just like the apiclient does:
https://github.com/jellyfin/jellyfin-apiclient-java/blob/master/android/src/main/java/org/jellyfin/apiclient/interaction/AndroidDevice.kt#L24

PR #486, opened 2020-05-27
add settings to remap tv remote buttons

Froghut:
The settings show the keyCode right now, that is not very descriptive but I did not find anything that would make it clearer which button is assigned. Also I'm not sure about the naming and texts of the preferences since english is not my native language.
I'm not an expert in android development so any feedback and criticism is welcome!

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preferences/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preferences/UserPreferences.kt
@@ -1,6 +1,7 @@
 package org.jellyfin.androidtv.preferences
 
 import android.content.Context
+import android.view.KeyEvent
 import androidx.preference.PreferenceManager
 import org.jellyfin.androidtv.preferences.enums.*
 
@@ -144,6 +145,16 @@ class UserPreferences(context: Context) : SharedPreferenceStore(PreferenceManage
 	 * Preferred video player for live TV
 	 */
 	var liveTvVideoPlayer by enumPreference("live_tv_video_player", PreferredVideoPlayer.AUTO)
+
+	/**
+	 * Default keycode for audio language switch button on a tv remote
+	 */
+	var audioLanguageButtonKeyCode by intPreference("audio_language_button_keycode", KeyEvent.KEYCODE_MEDIA_AUDIO_TRACK)
+
+	/**
+	 * Default keycode for subtitle language switch button on a tv remote
+	 */
+	var subtitleLanguageButtonKeyCode by intPreference("subtitle_language_button_keycode", KeyEvent.KEYCODE_CAPTIONS)
 
 	/* ACRA */
 	/**
--- a/app/src/main/java/org/jellyfin/androidtv/preferences/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preferences/UserPreferences.kt
@@ -147,14 +147,14 @@ class UserPreferences(context: Context) : SharedPreferenceStore(PreferenceManage
 	var liveTvVideoPlayer by enumPreference("live_tv_video_player", PreferredVideoPlayer.AUTO)
 
 	/**
-	 * Default keycode for audio language switch button on a tv remote
+	 * Shortcut used for changing the audio track
 	 */
-	var audioLanguageButtonKeyCode by intPreference("audio_language_button_keycode", KeyEvent.KEYCODE_MEDIA_AUDIO_TRACK)
+	var shortcutAudioTrack by intPreference("shortcut_audio_track", KeyEvent.KEYCODE_MEDIA_AUDIO_TRACK)
 
 	/**
-	 * Default keycode for subtitle language switch button on a tv remote
+	 * Shortcut used for changing the subtitle track
 	 */
-	var subtitleLanguageButtonKeyCode by intPreference("subtitle_language_button_keycode", KeyEvent.KEYCODE_CAPTIONS)
+	var shortcutSubtitleTrack by intPreference("shortcut_subtitle_track", KeyEvent.KEYCODE_CAPTIONS)
 
 	/* ACRA */
 	/**
--- a/app/src/main/java/org/jellyfin/androidtv/preferences/ui/ButtonRemapDialogFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preferences/ui/ButtonRemapDialogFragment.kt
@@ -1,0 +1,91 @@
+package org.jellyfin.androidtv.preferences.ui
+
+import android.content.Context
+import android.os.Bundle
+import android.text.TextUtils
+import android.view.*
+import android.widget.TextView
+import androidx.leanback.preference.LeanbackPreferenceDialogFragmentCompat
+
+class ButtonRemapDialogFragment : LeanbackPreferenceDialogFragmentCompat() {
+	private var mDialogTitle: CharSequence? = null
+	private var mDialogMessage: CharSequence? = null
+	private var mKeyCode: Int = 0
+
+	override fun onCreate(savedInstanceState: Bundle?) {
+		super.onCreate(savedInstanceState)
+		if (savedInstanceState == null) {
+			val preference = preference
+			mDialogTitle = preference.dialogTitle
+			mDialogMessage = preference.dialogMessage
+			if (preference is ButtonRemapPreference) {
+				mDialogTitle = preference.getDialogTitle()
+				mDialogMessage = preference.getDialogMessage()
+				mKeyCode = preference.getKeyCode()
+			} else {
+				throw IllegalArgumentException("Preference must be a ButtonRemapPreference")
+			}
+		} else {
+			mDialogTitle = savedInstanceState.getCharSequence(SAVE_STATE_TITLE)
+			mDialogMessage = savedInstanceState.getCharSequence(SAVE_STATE_MESSAGE)
+			mKeyCode = savedInstanceState.getInt(SAVE_STATE_KEYCODE)
+		}
+	}
+
+	override fun onSaveInstanceState(outState: Bundle) {
+		super.onSaveInstanceState(outState)
+		outState.putCharSequence(SAVE_STATE_TITLE, mDialogTitle)
+		outState.putCharSequence(SAVE_STATE_MESSAGE, mDialogMessage)
+		outState.putInt(SAVE_STATE_KEYCODE, mKeyCode)
+	}
+
+	override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?,
+							  savedInstanceState: Bundle?): View? {
+		val theme = org.jellyfin.androidtv.R.style.PreferenceThemeOverlayLeanback
+		val styledContext: Context = ContextThemeWrapper(activity, theme)
+		val styledInflater = inflater.cloneInContext(styledContext)
+		val view: View = styledInflater.inflate(org.jellyfin.androidtv.R.layout.button_remap_preference,
+			container, false)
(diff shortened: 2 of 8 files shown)


Froghut:
not sure what the EndOfSentenceFormat issue means, it would be great if someone could explain the issue.

thornbill:
It looks like it's just expecting punctuation at the end of the comments @Froghut.

We're still tweaking some of the configuration for Codacy. I'm not sure if we need this to be flagged. What do you think @nielsvanvelzen?

nielsvanvelzen:
I think the linter rules for comments are a must for the apiclient but only a nice to have for the app. It does give a better explanation of the code and makes it easier to follow. 

In the case a comment is made I think the [EndOfSentenceFormat](https://arturbosch.github.io/detekt/comments.html#endofsentenceformat) is important though as it checks for proper [KDoc Syntax](https://kotlinlang.org/docs/reference/kotlin-doc.html#kdoc-syntax). Basically what it wants right now is that all sentences end with a dot.

I disabled the rule that forces kdoc for public functions just now.

app/src/main/java/org/jellyfin/androidtv/preferences/UserPreferences.kt
```diff
@@ -145,6 +146,16 @@ class UserPreferences(context: Context) : SharedPreferenceStore(PreferenceManage
 	 */
 	var liveTvVideoPlayer by enumPreference("live_tv_video_player", PreferredVideoPlayer.AUTO)
 
+	/**
+	 * Default keycode for audio language switch button on a tv remote
```
nielsvanvelzen, line 150: "Button used for changing the audio track"

nielsvanvelzen:
I made a screen recording to illustrate how this works. I like the idea of button mapping but I'm not sure if this is the right way to do it. I'll hopefully have my review ready in the next few days.

nielsvanvelzen:
After some thinking I thought about some things to improve the user experience:
- Ignore dpad, ok & back buttons when changing a mapping (blacklisting the common keys)
- Add 3 buttons: Save, Cancel, Reset to default so the mapping will only be saved when the user explicitly tells it to
- Show the name of the button, Android can do this with the `KeyEvent.keyCodetoString` function: (https://developer.android.com/reference/android/view/KeyEvent#keyCodeToString(int)). Maybe even convert it to lowercase, capitalized and underscore replaced with whitespace.

In the future we should look into:
- Check if the pressed button is already assigned to another action
- Add all possible shortcuts to the preferences
- Bulk reset all shortcuts to default
- Add all shortcuts to a separate screen to modify them

Froghut:
@nielsvanvelzen your suggestions make a lot of sense.
Just two questions:
1. I have the feeling it would be best to have a button, or something (not sure what exactly) you have to click on to start listening for the next key press, I think that would "feel" better, what do you think?
2. when using `KeyEvent.keyCodetoString` unknown buttons will return a numerical value (like in my current implementation), should it then just show the number, or perhaps something like "Generic Button 1234"?

nielsvanvelzen:
> I have the feeling it would be best to have a button, or something (not sure what exactly) you have to click on to start listening for the next key press, I think that would "feel" better, what do you think?

What about a "press again to confirm" kind of thing? Making the user press the new key twice to confirm it is the right one.

> when using KeyEvent.keyCodetoString unknown buttons will return a numerical value (like in my current implementation), should it then just show the number, or perhaps something like "Generic Button 1234"?

I'd say something like "Unknown (1337)". This will show the user that we don't know the name but also include the code to make it recognizable.

app/src/main/java/org/jellyfin/androidtv/preferences/ui/ButtonRemapPreference.kt
```diff
@@ -0,0 +1,71 @@
...
+	context: Context,
+	attrs: AttributeSet? = null
+) : DialogPreference(context, attrs) {
+
+	/**
+	 * Saves a KeyCode in this preference.
+	 *
+	 * @param mKeyCode the KeyCode to save
+	 */
+	fun setKeyCode(mKeyCode: Int) {
```
nielsvanvelzen, line 22:
```suggestion
	fun setKeyCode(keyCode: Int) {
```

app/src/main/java/org/jellyfin/androidtv/preferences/ui/ButtonRemapPreference.kt
```diff
@@ -0,0 +1,71 @@
...
+	fun getKeyCode(): Int {
+		return getPersistedInt(-1)
+	}
+
+	override fun onSetInitialValue(defaultValue: Any?) {
+		if (defaultValue != null)
+			persistInt(defaultValue as Int)
+	}
+
+	override fun onGetDefaultValue(a: TypedArray, index: Int): Any {
```
nielsvanvelzen, line 40: `a` is really ambiguous, should be called something more clear

app/src/main/java/org/jellyfin/androidtv/preferences/ui/ButtonRemapPreference.kt
```diff
@@ -0,0 +1,71 @@
...
+		return a.getInt(index, -1)
+	}
+
+	init {
```
nielsvanvelzen, line 44: Move the init block to the top of the class

app/src/main/java/org/jellyfin/androidtv/preferences/ui/ButtonRemapPreference.kt
```diff
@@ -0,0 +1,71 @@
...
+
+	internal class ButtonRemapSummaryProvider private constructor() : SummaryProvider<ButtonRemapPreference> {
+		override fun provideSummary(preference: ButtonRemapPreference): CharSequence {
+			return provideSummary(preference.context, preference.getKeyCode())
+		}
+
+		fun provideSummary(context: Context, keyCode: Int): CharSequence {
+			var keyCodeString = KeyEvent.keyCodeToString(keyCode)
+			if (keyCodeString.startsWith("KEYCODE")) {
+				keyCodeString = keyCodeString.split("_").drop(1).joinToString(" ") { e -> e.toLowerCase(Locale.getDefault()).capitalize() }
```
nielsvanvelzen, line 58:
```suggestion
				keyCodeString = keyCodeString
					.removePrefix("KEYCODE_")
					.toLowerCase()
					.capitalize()
```

app/src/main/java/org/jellyfin/androidtv/preferences/ui/ButtonRemapPreference.kt
```diff
@@ -0,0 +1,71 @@
...
+			} else {
+				keyCodeString = "${context.getString(R.string.lbl_unknown_key)} ($keyCodeString)"
```
nielsvanvelzen, line 60: Use string formatting for the code;

```suggestion
				keyCodeString = context.getString(R.string.lbl_unknown_key, keyCodeString)
```

(don't forget to update the string itself)

app/src/main/java/org/jellyfin/androidtv/preferences/ui/ButtonRemapPreference.kt
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	internal class ButtonRemapSummaryProvider private constructor() : SummaryProvider<ButtonRemapPreference> {
+		override fun provideSummary(preference: ButtonRemapPreference): CharSequence {
+			return provideSummary(preference.context, preference.getKeyCode())
+		}
+
+		fun provideSummary(context: Context, keyCode: Int): CharSequence {
+			var keyCodeString = KeyEvent.keyCodeToString(keyCode)
+			if (keyCodeString.startsWith("KEYCODE")) {
```
nielsvanvelzen, line 57:
```suggestion
			val keyCodeString = KeyEvent.keyCodeToString(keyCode)
			return if (keyCodeString.startsWith("KEYCODE")) {
```

Remove the `keyCodeString = ` parts within the if and else blocks. Remove the return statement after the else block.

app/src/main/res/layout/button_remap_preference.xml
```diff
@@ -0,0 +1,72 @@
...
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toEndOf="parent">
+
+            <Button
+                android:id="@+id/buttonSave"
+                style="@style/Widget.AppCompat.Button.ButtonBar.AlertDialog"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_weight="1"
+                android:text="Save" />
```
nielsvanvelzen, line 59: Missing a string resource

app/src/main/res/layout/button_remap_preference.xml
```diff
@@ -0,0 +1,72 @@
...
+
+            <Button
+                android:id="@+id/buttonReset"
+                style="@style/Widget.AppCompat.Button.ButtonBar.AlertDialog"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_weight="1"
+                android:text="Reset" />
```
nielsvanvelzen, line 67: Missing a string resource

app/src/main/java/org/jellyfin/androidtv/preferences/ui/ButtonRemapDialogFragment.kt
```diff
@@ -0,0 +1,129 @@
...
+			KeyEvent.KEYCODE_ENTER
+		)
+	private var checkKeys: View.OnKeyListener = View.OnKeyListener { _, keyCode, _ ->
+		// ignore navigation buttons
+		if (ignoreKeys.contains(keyCode))
+			false
+		else {
+			this.keyCode = keyCode
+			setKeyCodeText()
+			requireView().findViewById<Button>(R.id.buttonSave).isEnabled = this.keyCode != originalKeyCode
```
nielsvanvelzen, line 34:
```suggestion
			requireView().buttonSave.isEnabled = this.keyCode != originalKeyCode
```

app/src/main/java/org/jellyfin/androidtv/preferences/ui/ButtonRemapDialogFragment.kt
```diff
@@ -0,0 +1,129 @@
...
+
+	override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+		val styledContext = ContextThemeWrapper(activity, R.style.PreferenceThemeOverlayLeanback)
+		val styledInflater = inflater.cloneInContext(styledContext)
+		return styledInflater.inflate(R.layout.button_remap_preference, container, false)
+	}
+
+	override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+		super.onViewCreated(view, savedInstanceState)
+		if (!TextUtils.isEmpty(dialogTitle)) {
```
nielsvanvelzen, line 74:
```suggestion
		if (dialogTitle.isNotBlank()) {
```

app/src/main/java/org/jellyfin/androidtv/preferences/ui/ButtonRemapDialogFragment.kt
```diff
@@ -0,0 +1,129 @@
...
+			decor_title.text = dialogTitle
+		}
+		if (!TextUtils.isEmpty(dialogMessage)) {
```
nielsvanvelzen, line 77:
```suggestion
		if (dialogMessage.isNotBlank()) {
```

PhillyMay:
I really like this shortcut implementation. Is there the possibility, that a few other shortcuts can be implemented the same way. 
For example a similar shortcut to directly spawn the chapter selection overlay would be fantastic.

PR #490, opened 2020-05-30
Removed Picasso to be replaced by Glide.

### LINKED ISSUE #403: Unify the image processor libraries
At the moment we use both [Picasso](https://square.github.io/picasso/) and [Glide](https://bumptech.github.io/glide/). We don't need multiple image processor libraries though. Both libraries have caching techniques for the loaded images but they don't share this cache.
Using a single library will improve the maintainability and performance. 

Steps to be taken:
1. Decide on the library to use
2. Replace all image loading code with the chosen library
4. Add proper configuration for the library (caching settings and such)
8. Ship it 🚢 

---
I assigned this to 0.13.x but we can already use a single library for all new code in 0.12.x

tukilo:
**Changes**

Migration from Glide version 3 to 4.
Removed completly Picasso and OkHttp from the project to be replaced by Glide.
Created new BackgroundManager util to set the background drawable.
Removed unused imports.

**Issues**

Closes #403


app/src/main/java/org/jellyfin/androidtv/util/BackgroundManagerUtils.kt
```diff
@@ -0,0 +1,42 @@
...
+import androidx.leanback.app.BackgroundManager
+import com.bumptech.glide.Glide
+import com.bumptech.glide.load.DataSource
+import com.bumptech.glide.load.engine.GlideException
+import com.bumptech.glide.load.resource.bitmap.BitmapTransformation
+import com.bumptech.glide.load.resource.bitmap.CenterInside
+import com.bumptech.glide.request.RequestListener
+import com.bumptech.glide.request.target.Target
+
+/** Created by Oriol on 30/05/20. **/
```
nielsvanvelzen, line 14: We don't add "created by" comments to files. We only have them in some other files that were inherited from the old Emby code.

app/src/main/java/org/jellyfin/androidtv/util/BackgroundManagerUtils.kt
```diff
@@ -0,0 +1,42 @@
...
+
+fun BackgroundManager.drawable(
+	activity: Activity,
```
nielsvanvelzen, line 17: This could be a context

app/src/main/java/org/jellyfin/androidtv/util/BackgroundManagerUtils.kt
```diff
@@ -0,0 +1,42 @@
...
+		.override(widthPixels, heightPixels)
+		.transform(transformation)
+		.listener(object : RequestListener<Drawable?> {
+			override fun onLoadFailed(e: GlideException?, model: Any, target: Target<Drawable?>, isFirstResource: Boolean): Boolean {
+				return false
+			}
+
+			override fun onResourceReady(resource: Drawable?, model: Any, target: Target<Drawable?>, dataSource: DataSource, isFirstResource: Boolean): Boolean {
+				drawable = resource
+				return false
```
nielsvanvelzen, line 37: We should return `true` here

app/src/main/java/org/jellyfin/androidtv/util/BackgroundManagerUtils.kt
tukilo: indeed, what a mistake

app/src/main/java/org/jellyfin/androidtv/util/BackgroundManagerUtils.kt
```diff
@@ -0,0 +1,36 @@
...
+fun BackgroundManager.drawable(
+	context: Context,
+	url: String,
+	widthPixels: Int,
+	heightPixels: Int
+) {
+
+	Glide.with(context)
+		.load(url)
+		.override(widthPixels, heightPixels)
```
nielsvanvelzen, line 22:
```suggestion
		.override(widthPixels, heightPixels)
		.centerCrop()
```

When testing I noticed some backdrops getting stretched which did not happen before this PR. Adding the centerCrop function here fixed it.

nielsvanvelzen:
I applied the "blocked" tag because our own CI and Codacy are not working at the moment. The PR is approved unless the pipeline fails or Codacy reports errors once working again.

app/src/main/java/org/jellyfin/androidtv/util/BackgroundManagerUtils.kt
tukilo: And it makes sense because I eliminated it without realizing

app/src/main/java/org/jellyfin/androidtv/util/BackgroundManagerUtils.kt
```diff
@@ -0,0 +1,37 @@
...
+import android.graphics.drawable.Drawable
+import androidx.leanback.app.BackgroundManager
+import com.bumptech.glide.Glide
+import com.bumptech.glide.load.DataSource
+import com.bumptech.glide.load.engine.GlideException
+import com.bumptech.glide.request.RequestListener
+import com.bumptech.glide.request.target.Target
+import timber.log.Timber
+
+fun BackgroundManager.drawable(
```
nielsvanvelzen, line 13: As this file only contains a single extension function I'd recommend renaming it to "BackgroundManagerExtensions" to stay consistent with the other file(s)

app/src/main/java/org/jellyfin/androidtv/util/BackgroundManagerUtils.kt
```diff
@@ -0,0 +1,37 @@
...
+	context: Context,
+	url: String,
+	widthPixels: Int,
+	heightPixels: Int
+) {
+
```
nielsvanvelzen, line 19: Unnecessary whitespace

app/src/main/java/org/jellyfin/androidtv/util/BackgroundManagerUtils.kt
```diff
@@ -0,0 +1,37 @@
...
+				return false
+			}
+
+			override fun onResourceReady(resource: Drawable?, model: Any, target: Target<Drawable?>, dataSource: DataSource, isFirstResource: Boolean): Boolean {
+				drawable = resource
+				return true
+			}
+		}).submit()
+
+
```
nielsvanvelzen, line 36: Unnecessary whitespace

PR #492, opened 2020-05-31
Add Sonarqube

nielsvanvelzen:
**Changes**
- Add Sonarqube

**To do**
- [x] Add Azure Pipelines job to report the status to Sonarcloud (can be based on the [server tasks](https://github.com/jellyfin/jellyfin/blob/master/.ci/azure-pipelines-test.yml#L45-L69))
- ~~Add detekt (separate PR)~~
- ~~Add a Java linter (not sure which one, separate PR)~~ SonarJava is enabled by default (this is what Codeclimate uses too)

PR #495, opened 2020-06-01
Add Detekt support

nielsvanvelzen:
Added Detekt as Gradle plugin. Can be executed via `gradlew detekt` or via the lint task.
It will use the default detekt configuration and options can be overwritten using the `detekt.yml` file.

The maximum amount of issues is currently set to `0` which means it will fail if it finds an issue. Depending on how the pipeline will be configured in #492 this number might change.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -31,6 +31,7 @@ subprojects {
 
 	detekt {
 		buildUponDefaultConfig = true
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
Turns out detekt was automatically enabled in the pipeline so I had to disable the failure already!

PR #496, opened 2020-06-01
Make sure all catch statements in the code log their exception

nielsvanvelzen:
Some try-catch statements silently fail. I added logging statements to those. We can never have enough logs.

**Changes**
- Add a logging statement `Timber.e()` for all catch statements

**Issues**


nielsvanvelzen:
Merge #497 before this one. Conflict will happen so I will rebase rebase this one later

PR #497, opened 2020-06-01
Remove unused SubtitleHelper

nielsvanvelzen:
Found a funny md5 function which returned UUID's, did some digging found out it was used in the path of a unused function. Started cleaning, whole class was unused.

Also did some cleaning in the PlaybackController when I removed the instance of SubtitleHelper (which was not used) and removed some unsused properties/functions from Utils class.


app/src/main/java/org/jellyfin/androidtv/playback/PlaybackController.java
```diff
@@ -615,47 +608,18 @@ public void run() {
...
-                @Override
-                public void run() {
-                    if (mSpinner != null) mSpinner.setVisibility(View.VISIBLE);
-                    spinnerOff = false;
-                }
-            });
-
-        }
-
+        spinnerOff = false;
```
thornbill, line 618: Seems like this may not be needed now?

nielsvanvelzen: It is actually used for something else and I did not want to break that code

PR #498, opened 2020-06-02
Item details rewrite for Series

nielsvanvelzen:
**Changes**

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/details/fragments/SeriesDetailsFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/details/fragments/SeriesDetailsFragment.kt
@@ -9,28 +9,29 @@ import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.withContext
 import org.jellyfin.androidtv.TvApp
 import org.jellyfin.androidtv.details.actions.DeleteAction
-import org.jellyfin.androidtv.details.actions.PlayFromBeginningAction
 import org.jellyfin.androidtv.details.actions.SecondariesPopupAction
+import org.jellyfin.androidtv.details.actions.ShuffleAction
 import org.jellyfin.androidtv.details.actions.ToggleFavoriteAction
-import org.jellyfin.androidtv.details.presenters.ChapterInfoPresenter
 import org.jellyfin.androidtv.details.presenters.ItemPresenter
 import org.jellyfin.androidtv.details.presenters.PersonPresenter
+import org.jellyfin.androidtv.details.presenters.SeriesMediaRowPresenter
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/jellyfin/androidtv/details/fragments/SeriesDetailsFragment.kt
```diff
@@ -0,0 +1,82 @@
...
+		// Add rows
+		rowAdapter.apply {
+			add(detailRow)
+			add(mediaRow)
+			addIfNotEmpty(castRow)
+			addIfNotEmpty(relatedItemsRow)
+		}
+	}
+
+	private suspend fun loadAdditionalInformation() = withContext(Dispatchers.IO) {
```
vnidens, line 69: Maybe it makes sense to follow MVVM approach delegating all the work to the ViewModel and leaving a Fragment only for View tasks? Now it doesn't has too much inside but anyway it has business logic at the View level

app/src/main/java/org/jellyfin/androidtv/details/presenters/SeriesMediaRowPresenter.kt
```diff
@@ -0,0 +1,166 @@
...
+		}
+	}
+
+	private fun setSelectedSeason(vh: ViewHolder, item: Season) {
+		item.images.primary?.load(context) {
+			vh.poster.setImageBitmap(it)
+		}
+
+		Toast.makeText(context, item.title, Toast.LENGTH_LONG).show()
+		GlobalScope.launch(Dispatchers.IO) {
```
vnidens, line 77: `GlobalScope` is not recommended (please see this post: https://medium.com/@elizarov/the-reason-to-avoid-globalscope-835337445abc). It's better to use either a "built-in" scope (in Activity, Fragment or ViewModel) or make your own by extending CoroutineScope interface.

app/src/main/java/org/jellyfin/androidtv/details/fragments/SeriesDetailsFragment.kt
nielsvanvelzen: Thanks for both of your comments. Right now the details branch is kind of a playground and far from being done. We'd like to use the MVVM pattern but decided to do that at a later stadium (might be in another branch too).

This branch started out a few months ago and now that we made some progress I've thought of some things we could improve, when I'm done with my work on the apiclient refactoring I'll refactor this branch. I will keep your comments in mind when I do! 😄

nielsvanvelzen:
I'm closing this PR for now, will re-open (or re-create) when I pick this up again.

PR #501, opened 2020-06-02
Removed unneeded code

tukilo:
**Changes**

Take the opportunity to transform that fragment from java to kotlin. Removed variable that was not being used and was always null.

**Issues**


nielsvanvelzen:
We try to keep git history for converted files. In order to do this you need to convert in 2 commits:

1. Rename the file from .java to .kt
2. Update the code to Kotlin code

If you don't do that the history of the file will be lost like what happend now: https://github.com/jellyfin/jellyfin-androidtv/commits/8e752363489f341c77e0eee285c1d1ca93c121b7/app/src/main/java/org/jellyfin/androidtv/startup/SelectUserFragment.kt

tukilo:
> We try to keep git history for converted files. In order to do this you need to convert in 2 commits:
> 
> 1. Rename the file from .java to .kt
> 2. Update the code to Kotlin code
> 
> If you don't do that the history of the file will be lost like what happend now: https://github.com/jellyfin/jellyfin-androidtv/commits/8e752363489f341c77e0eee285c1d1ca93c121b7/app/src/main/java/org/jellyfin/androidtv/startup/SelectUserFragment.kt

Oh, sure, make sense. I Keep that in mind for next time. I'm going to cancel the PR for now and come back later on with a proper git history.

PR #502, opened 2020-06-02
Removed unneeded code

tukilo:
**Changes**

Take the opportunity to transform that fragment from java to kotlin. Removed variable that was not being used and was always null.

**Issues**

Cleaning code


nielsvanvelzen:
Looks like the rename didn't work. One commit creates a Kotlin file and the other one removed the Java one. This removes the history again.

Also a small tip: instead of closing this PR you could force-push to this branch. This way you don't need to recreate the PR itself.

PR #503, opened 2020-06-02
Search improvements

### LINKED ISSUE #461: Search Query Lost after hitting search
when you go to search for content, you can see results in the background. As soon as you leave the keyboard focus, all search results are lost and the clients has to retrieve the results.

nielsvanvelzen:
**Changes**
- Search timeout from 1.5 seconds to 600 milliseconds
- Don't query the server for the same input twice
- Always show search results, even if one or multiple requests failed

**Issues**
Fixes #461

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/search/SearchProvider.java
+++ b/app/src/main/java/org/jellyfin/androidtv/search/SearchProvider.java
@@ -3,19 +3,19 @@ package org.jellyfin.androidtv.search;
 import android.content.Context;
 import android.os.Handler;
 
-import org.jellyfin.androidtv.util.Utils;
-
 import androidx.leanback.app.SearchSupportFragment;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/search/SearchProvider.java
```diff
@@ -54,6 +54,10 @@ private void search(String query, boolean delayed) {
             return;
         }
 
+        // Don't search the same thing twice
+        if (query.equals(lastQuery)) return;
```
thornbill, line 58: May want to add a null check for query here

nielsvanvelzen: There is a check just above this one which uses Utils.isEmpty. That function checks for null values already.

Issue #281, opened 2020-01-31
HD Audio Passthrough on Nvidia Shield


Robert-D:
Hello and sorry if this is already being addressed, but it appears HD audio streams such as DTS-X and TrueHD encoded with Atmos metadata are not being passed properly to receivers/sound bars from the Nvidia Shield. It appears most people are using Kodi as a work around:

[https://www.reddit.com/r/jellyfin/comments/ewf73h/audio_passthrough_direct_on_shield_tv/](https://www.reddit.com/r/jellyfin/comments/ewf73h/audio_passthrough_direct_on_shield_tv/)

If there is anything I can do to help please let me know and sorry again if this is a duplicate.

StereotypicalCat:
Even after #524, i am still unable to get either Atmos or DTS:X to passthrough.

Robert-D:
Thanks for the reminder stale bot, but I would still love to see proper audio passthrough with the native player instead of using VLC.

metronidazole:
Is it just tracks with atmos or also lossless audio? I am able to passthrough TrueHD and DTS-HD MA without issue when building this from source.

roblav96:
More info on this wide-spread problem can be found here:
https://github.com/moneytoo/Player/issues/139

bokysan:
Anybody tried this lately? As of late, I don't have a lot of issues using ExoPlayer, but would need to recheck if this is still an issue with the latest Jellyfin version.

ElegyD:
I've got Jellyfin 10.7.7 with 0.12.3 on my Nvidia Shield. I'm using Exoplayer for Dolby Vision compatibility.
Dolby TrueHD passthrough works, but DTS-HD MA doesn't. Regardless of the "Bitstream DTS audio" toggled on or off.

morkyy:
Same issue here on the shield. Most codecs passthrough just fine (including Dolby TrueHD and Atmos). But DTS seems to be very buggy. A lot of files with DTS:X/Neural X audio play fine but multiple files with DTS-HD MA and sometimes DTS:X MA do not work. Video will play fine but there will be no audio regardless of the "bitstream DTS audio" setting. Since v0.13 some files might not play at all even though they used to play with v0.12 with no sound. 

Receiver detects the correct audio format.

roblav96:
@nielsvanvelzen Here is a diff of adding DTS-HD to ExoPlayer's `MatroskaExtractor.java`:
[...]

PR #524, opened 2020-07-22
Fix VLC audio passthrough

willtrking:
Noticed some issues with VLC audio passthrough on the nvidia shield which caused Atmos to not work correctly. Believe this was due to an overall slight misconfiguration of VLC audio passthrough.

**Changes**
This slightly changes the way audio is configured for libvlc to be more in-line with the way https://github.com/videolan/vlc-android works.

**Issues**
https://github.com/jellyfin/jellyfin-androidtv/issues/281

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/playback/MediaManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/playback/MediaManager.java
@@ -227,8 +227,12 @@ public class MediaManager {
                 mLibVLC = new LibVLC(TvApp.getApplication(), options);
 
                 mVlcPlayer = new org.videolan.libvlc.MediaPlayer(mLibVLC);
-                mVlcPlayer.setAudioOutput(Utils.downMixAudio() ? "opensles_android" : "android_audiotrack");
-                mVlcPlayer.setAudioOutputDevice("hdmi");
+                if(!Utils.downMixAudio()) {
+                    mVlcPlayer.setAudioDigitalOutputEnabled(true);
+                } else {
+                    mVlcPlayer.setAudioOutput("opensles_android");
+                }
+
 
                 mVlcHandler.setOnProgressListener(new PlaybackListener() {
                     @Override
--- a/app/src/main/java/org/jellyfin/androidtv/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/playback/VideoManager.java
@@ -108,6 +108,7 @@ public class VideoManager implements IVLCVout.OnNewVideoLayoutListener {
             }
         }).build();
 
+
         mExoPlayerView = view.findViewById(R.id.exoPlayerView);
         mExoPlayerView.setPlayer(mExoPlayer);
         mExoPlayer.addListener(new Player.EventListener() {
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/playback/MediaManager.java
```diff
@@ -227,8 +227,12 @@ public void onPlayerError(ExoPlaybackException error) {
                 mLibVLC = new LibVLC(TvApp.getApplication(), options);
 
                 mVlcPlayer = new org.videolan.libvlc.MediaPlayer(mLibVLC);
-                mVlcPlayer.setAudioOutput(Utils.downMixAudio() ? "opensles_android" : "android_audiotrack");
-                mVlcPlayer.setAudioOutputDevice("hdmi");
+                if(!Utils.downMixAudio()) {
+                    mVlcPlayer.setAudioDigitalOutputEnabled(true);
```
thornbill, line 231: It looks like this will returns a boolean indicating if it is successful or not, we should probably do something like:
```java
if (!Utils.downMixAudio() || !mVlcPlayer.setAudioDigitalOutputEnabled(true)) {
    mVlcPlayer.setAudioOutput("opensles_android");
}
```
So it will fallback to downmixing if it fails.

app/src/main/java/org/jellyfin/androidtv/playback/MediaManager.java
```diff
@@ -227,8 +227,12 @@ public void onPlayerError(ExoPlaybackException error) {
...
+                } else {
+                    mVlcPlayer.setAudioOutput("opensles_android");
```
thornbill, line 233: I'm not sure if it's necessary or not, but it seems to be inconsistent on whether `mVlcPlayer.setAudioOutputDevice("hdmi");` is called or not. Maybe it should be still be called here?

app/src/main/java/org/jellyfin/androidtv/playback/VideoManager.java
```diff
@@ -442,16 +443,24 @@ public long getAudioDelay() {
...
+            setVlcAudioOptions();
+        }
+    }
+
+    private void setVlcAudioOptions() {
+
+        if(!Utils.downMixAudio()) {
+            mVlcPlayer.setAudioDigitalOutputEnabled(true);
+        } else {
+            mVlcPlayer.setAudioOutput("opensles_android");
```
thornbill, line 463: Same notes as above apply here.

app/src/main/java/org/jellyfin/androidtv/playback/MediaManager.java
willtrking: Should be addressed

app/src/main/java/org/jellyfin/androidtv/playback/MediaManager.java
willtrking: Not an expert on VLC, but are we sure in the case that `setAudioDigitalOutputEnabled` fails, `opensles_android` is preferrable? FWIW audio was outputting previously WITHOUT `setAudioDigitalOutputEnabled` being called AND without downmixing, however it was not properly outputting Atmos.

Seems like if it is not always preferrable, might be less confusing for users to have to manually set this in preferences.

app/src/main/java/org/jellyfin/androidtv/playback/VideoManager.java
willtrking: Should be addressed

app/src/main/java/org/jellyfin/androidtv/playback/MediaManager.java
thornbill: Ah I think I follow now. This seems fine.

Should we be calling ‘mVlcPlayer.setAudioOutput("android_audiotrack");‘ here like you do below?

bbalz:
I updated to 11.5 and the automatic player on my Shield is only giving me PCM 2.0 stereo for TrueHD Atmos tracks now.  It is direct playing it, versus transcoding...  If I use an external player like VLC or use Kodi + Jellyfin plugin, I get the TrueHD Atmos playing directly but I believe that worked before anyways.  

There's no transcoding going on, it's direct playing no matter what but it's odd that I'm now only getting PCM 2.0 stereo on any media file that is TrueHD Atmos.  Audio is set to passthrough in Jellyfin.

Wondering if anyone has been able to test this out too and found similar results?  Or perhaps there is some setting I am completely overlooking for the automatic player of Jellyfin.

PR #534, opened 2020-07-30
`Enter manually` button position

### LINKED ISSUE #236: Move Enter Manually for User Name

vnidens:
**Changes**
Fixes position of the "Enter manually" button on the user selection screen.
Minor refactoring of the `ItemRowAdapter`.

**Preview**
image

**Issues**
Fixes #236


vnidens:
Hi Jellyfin team,
How do you usually manage codeclimate complaints if they are mostly caused by original code?

thornbill:
Hey @vnidens! I can review those and approve it. I think we’ll be disabling it sometime since we have added codacy.

PR #540, opened 2020-08-08
Upgrade to apiclient 0.7.0

nielsvanvelzen:
This PR upgrades the code to be compatible with the new apiclient. Most of the changes are simple except for the credential manager changes.

Right now the login flow is dialog-only. No servers or credentials are stored anymore. Only if a automatic login was configured before this change will it work. I could spend a lot of time making that work but considering those things will be rewritten for the new auth flow @thornbill is working on that is scheduled for the next release I did not bother.

# TODO
- [ ] Change the version in build.gradle


app/src/main/java/org/jellyfin/androidtv/TvApp.java
```diff
@@ -94,7 +100,14 @@ protected void attachBaseContext(Context base) {
     public void onCreate() {
         super.onCreate();
         app = (TvApp) getApplicationContext();
-        playbackManager = new PlaybackManager(new AndroidDevice(this), new AndroidLogger("PlaybackManager"));
+        JellyfinOptions.Builder options = new JellyfinOptions.Builder();
+        JellyfinAndroidKt.android(options, this);
+        options.setLogger(new AndroidLogger());
```
dkanada, line 105: Why bother keeping the logger if we're removing the credential management?

nielsvanvelzen: Some other places currently use the logger

nielsvanvelzen:
@thornbill this one is ready now, when merged the app is usable but it sucks to use it as it requests server+user info each start. It sort of makes your new authentication flow a dependency.

nielsvanvelzen:
I updated the repositories because we're now in JCenter! Also removed the videolan one because that one is in JCenter too.

PR #438, opened 2020-04-16
Fix NPE when finished playing media

nielsvanvelzen:
should fix
```
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.get(ArrayList.java:437)
	at org.jellyfin.androidtv.playback.PlaybackController.getCurrentlyPlayingItem(PlaybackController.java:129)
	at org.jellyfin.androidtv.playback.PlaybackController$15.run(PlaybackController.java:962)
	at android.os.Handler.handleCallback(Handler.java:789)
	at android.os.Handler.dispatchMessage(Handler.java:98)
	at android.os.Looper.loop(Looper.java:164)
	at android.app.ActivityThread.main(ActivityThread.java:6541)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
```

PR #408, opened 2020-03-25
Update Azure builds

thornbill:

PR #542, opened 2020-08-11
Release 0.11.4

thornbill:
**Changes**
This branch includes the following changes for the 0.11.4 release

* #459 
* #349 
* #466 
* #499 
* #539 
* #438 
* #454 
* #525 
* #533 
* #408 
* #407 

**Issues**
N/A

### DIFF
--- a/.ci/azure-pipelines.yml
+++ b/.ci/azure-pipelines.yml
@@ -17,10 +17,36 @@ pool:
 
 steps:
 - task: Gradle@2
+  displayName: 'Build Debug'
   inputs:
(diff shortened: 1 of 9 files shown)


nielsvanvelzen:
I was wondering if we want to add the latest translation updates. I'm not sure how easy that would be to add though (build may complain about unused strings)

nielsvanvelzen:
We should probably cherrypick c7c52096b3141b2b1b2c7f7077f131568a7c7d9c to master so we don't forget what the highest version code is.

PR #545, opened 2020-08-18
Fix corrupt images crashing NextUpActivity

thornbill:
**Changes**
This fixes an issue where corrupt images being loaded in the NextUpActivity would cause the app to crash.

**Issues**
N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/playback/nextup/NextUpActivity.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/playback/nextup/NextUpActivity.kt
@@ -49,12 +49,10 @@ class NextUpActivity : FragmentActivity() {
 		}
 	}
 
-	private fun safelyLoadBitmap(url: String): Bitmap? {
-		return try {
-			Glide.with(this@NextUpActivity).asBitmap().load(url).submit().get()
-		} catch (e: Exception) {
-			null
-		}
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/playback/nextup/NextUpActivity.kt
```diff
@@ -50,6 +49,14 @@ class NextUpActivity : FragmentActivity() {
 		}
 	}
 
+	private fun safelyLoadBitmap(url: String): Bitmap? {
+		return try {
+			Glide.with(this@NextUpActivity).asBitmap().load(url).submit().get()
```
nielsvanvelzen, line 54: You might not need the `@NextUpActivity` part here because `this` didn't change.

app/src/main/java/org/jellyfin/androidtv/playback/nextup/NextUpActivity.kt
```diff
@@ -50,6 +49,14 @@ class NextUpActivity : FragmentActivity() {
 		}
 	}
 
+	private fun safelyLoadBitmap(url: String): Bitmap? {
+		return try {
```
nielsvanvelzen, line 53:
```suggestion
	private fun safelyLoadBitmap(url: String) = try {
```
(also, one of the closing brackets should be removed)

PR #556, opened 2020-09-11
Cleanup app structure

thornbill:
**Changes**
Cleans up the overall structure of the app to (hopefully) simplify things a bit. Feedback is welcome!

```
├── constant -- Constants/Enums
├── data
│   ├── compat -- Classes ported from old apiclient to maintain compatibility in the app (Deprecated should be replaced/removed)
│   ├── eventhandling -- API webservice event handling
│   ├── model -- Various data models
│   ├── querying -- Extensions to the querying package in the apiclient (Should probably be replaced/removed)
│   └── repository -- Data repositories for shared access
├── integration -- Android TV homescreen channel integrations
├── preference -- Interface for Android shared preferences
├── ui
│   ├── browsing -- Views for browsing items (rows, grids, etc.)
│   ├── home -- Home screen views
│   ├── itemdetail -- Item detail views
│   ├── itemhandling -- BaseItem views
│   ├── livetv -- Live TV views
│   ├── playback -- Media player views
│   ├── preference -- App preferences/settings views
│   ├── presentation -- Presenters from MVP architecture
│   ├── search -- Search views
│   ├── shared -- Shared code for UI classes
│   └── startup -- Authentication views
└── util -- Various utilities
```

**Issues**
N/A


nielsvanvelzen:
While we're at it. I think it's a good idea to make everything singular or plural. Right now we have some packages that are singular "util", "ui" and a bunch that are plural "preference**s**" "constant**s**".

nielsvanvelzen:
I also think the "channels" package should be renamed to "integration" so we can put others things in there like integrating with Android TV search or a Google Cast receiver (if we ever add both)

nielsvanvelzen:
Also, we can't rename the MainActivity because it would clear launcher shortcuts in Android (and I'm quite sure the same behavior applies to Android TV / Fire TV). We could maybe add a `typealias` for that though.

thornbill:
Changes have been applied!

nielsvanvelzen (commented):
Two more:

`org.jellyfin.androidtv.ui.preference.preference` **=>** `??`  
Package name looks ugly because preference is repated, I'm not sure what to rename it to though

`org.jellyfin.androidtv.ui.base` **=>** `org.jellyfin.androidtv.ui.shared`  
"base" is quite abstract, I think the term "shared" is more descriptive.

thornbill:
Yeah I agree on the preference thing but couldn’t think of a better name for it either.

nielsvanvelzen:
Maybe option or implementation?

thornbill:
I landed on "custom" for the preference package since it contains custom views for different preference types.

nielsvanvelzen (approved):
LGTM!

PR #557, opened 2020-09-12
Remove the remaining "Created by" comments

thornbill:
**Changes**
These are useless. Attribution is handled by the license and contributors files. We have already removed most of these, just a few stragglers remained.

**Issues**
N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/presentation/MyImageCardView.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/presentation/MyImageCardView.java
@@ -1,10 +1,5 @@
 package org.jellyfin.androidtv.ui.presentation;
 
-/**
- * Created by Eric on 12/29/2014.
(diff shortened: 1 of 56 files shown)


nielsvanvelzen:
Can you rebase from master? :)

app/src/main/java/org/jellyfin/androidtv/ui/presentation/MyImageCardView.java
```diff
@@ -27,6 +22,7 @@
 import androidx.leanback.widget.BaseCardView;
 
 /**
+ * Modified ImageCard with no fade on the badge
  * A card view with an {@link ImageView} as its main region.
```
nielsvanvelzen, line 26: I don't think this comment is 100% accurate but it's a messy class anyway

nielsvanvelzen (approved):

PR #560, opened 2020-09-14
Add Koin for dependency injection

thornbill:
**Changes**
Adds Koin for dependency injection and migrates some global app state to using it instead of referencing TvApp.

**Issues**
N/A


nielsvanvelzen:
I've started reviewing but it will take some time since there's a lot of changes.

nielsvanvelzen (changes requested):
LGTM. I noticed a few places where `get()` was called multiple times instead of using a property for it but it was all in ugly Java classes so I don't care too much about those.

thornbill:
My general approach was to use `inject` if there were more than 1-2 usages of a dependency in a file except for when it was being used in a static public method. That seemed like a reasonable approach to me, but I’m new to Koin lol. 🤷‍♂️

nielsvanvelzen:
I don't see a problem with that, although I prefer using inject via properties everywhere (but I think you did that already in the Kotlin code)

nielsvanvelzen (approved):

PR #566, opened 2020-09-19
Fixed misleading login error message

okan35:
When I enter a server address of 192.xxx.xx.xxx:8096, I get wrong username or pass error while they are corect because when you check the logcat, it is actually bad url, no protocol error from Volley library. This is misleading and causes people to search for solutions. A simple message output such "Please put an http protocol to beginning of address" can easily save people from searching for solution.

I just made a pull request for this as it is not a huge change, please let me know if something needs a change.
I am using jellyfin all the time and looking to contribute more.

**Changes**
Changed address hint on user login.
Added a more clear error message for wrong url or missing protocol error.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/apiclient/AuthenticationHelper.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/apiclient/AuthenticationHelper.java
@@ -22,6 +22,8 @@ import org.jellyfin.apiclient.serialization.GsonJsonSerializer;
 
 import java.io.IOException;
 import java.io.OutputStream;
+import java.io.PrintWriter;
+import java.io.StringWriter;
 
 import timber.log.Timber;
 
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
Did you get this error on the stable app or the current master branch? Because the stable release should work, only the master branch doesn't because we're actively refactoring the authentication code.

okan35:
I got this from master branch. But also on the app, login screen is acting strangely. On the stable app when you write address such as 192.xxx.xx.xx:8096 it doesn't work, just nothing happens, if you write 192.xxx.xx.xx:8096/jellyfin then it works. So maybe we can change the hint for address or add this jellyfin part ? Dont know how it works for others.

Also is there any other branch that I should develop from when I want to work on something  ?

nielsvanvelzen:
The apps should call the [addressCandidates](https://github.com/jellyfin/jellyfin-apiclient-java/blob/master/library/src/main/java/org/jellyfin/apiclient/DiscoveryService.kt#L24) function from the apiclient which will add the protocol (https/http). So inputting just "192.xxx.xx.xx:8096" should work. AFAIK a similar function is used in the current stable app (0.11.4).

And for development, you can branch from and target the master branch.

okan35:
Ok, then I guess we can delete this pull request. I would suggest if you can check writing server addresses with and without /jellyfin at the end and see if you experience something strange.

I experience this on xiaomi mi android tv box, android version is 9.

nielsvanvelzen:
I want to add a bunch of extra "candidates" by adding a port number and/or baseurl to make it easier to connect. Knowing just the ip adress / host name should be enough in the future when using common values like port 8096 or a "/jellyfin" baseurl. Right now it only does the part that adds http/https though.

Feel free to drop by our Android matrix channel if you have further questions. I'll close this PR for now, thanks for your efforts :)

PR #568, opened 2020-09-21
Add channel name to info row in LiveTV guide

MrChip53:
**Changes**
- Add the name of the channel in the info row for any channels that display "\<No Program Data\>"

Can this get merged and pushed to Play Store ASAP please? It's a huge QoL improvement for channels that have no EPG data in the Program Guide.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
@@ -937,6 +937,12 @@ public class LiveTvGuideActivity extends BaseActivity implements ILiveTvGuide {
             }
         } else {
             mInfoRow.removeAllViews();
+
+            TextView channelName = new TextView(mActivity);
+            channelName.setTextSize(16);
+            channelName.setText(TvManager.getChannel(TvManager.getAllChannelsIndex(mSelectedProgram.getChannelId())).getName());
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
@thornbill I don't have Live TV setup so unfortunately I can't test this.

@MrChip53 Thanks for you pull request. I understand you'd like to have these changes in a release as soon as possible but the current state of the app is not stable. We might be able to add this change to a patch release if we need to do any immediate bugfixing but I consider that unlikely.

nielsvanvelzen (approved):
Code looks fine

MrChip53:
Code wise I went for the most simple implementation. If you all would prefer i can move the view creation into the helper function and rework the detailUpdate flow.

On the topic of 0.12.0, what issues are needing resolved?

thornbill (changes requested):
I think it would be better if we could handle this in `InfoLayoutHelper.addInfoRow()` since that _should_ handle adding the channel name in the same location as when EPG data is available.

That would require some changes to the check for `mSelectedProgram.getId() != null` to be able to handle this specific case and changes to `InfoLayoutHelper.addInfoRow()` to ensure it doesn't introduce `NullPointerException`s.

MrChip53:
I do agree. Im going to delete this pull request and start a new one in a feature branch though. Accidentally, did this in my master branch. Start adding some other tweaks and open a new one as a draft.

Issue #284, opened 2020-02-02
LiveTV Guide Update Tick


Artiume:
The Guide doesn't automatically update to the new half hour block. This typically is not noticed after finishing a show, exiting it to find a new show and it leaves you in the original time slot.

nielsvanvelzen:
I'm not that familiar with LiveTV but I'm assuming this is a feature request / enhancement to add dynamic updating to a certain screen.

Artiume:
Can we move this to the jellyfin-web repository? I think it should be implemented there rather than here.

thornbill:
Nope this is definitely an issue in the Android TV app. If it happens in both, then we should have separate tickets for each implementation. :smiley:

Artiume:
Dang. I was hoping it fixing it upstream would fix it here as well

MrChip53:
Is this still an issue? When working on the live tv code I definitely saw code blocks that SHOULD do exactly what you are wanting. It would force the live version of the guide to refresh upon opening every 30 minutes.

MrChip53:
Just tested and this issue seems to be resolved.

MrChip53:
@nielsvanvelzen @thornbill This was fixed in PR #569 and can be closed now.

PR #569, opened 2020-09-23
Live TV improvements

MrChip53:
### Changes
- Added channel name to info bar for programs with no data.
- Ignore programs that fill time slots that have already been filled by a previous program.
- Change live tv guide from 6 hours to 9 hours of programming to be consistent with non live guide.
- Channel surf using channel headers in live guide
- Long/short press in non-live guide and live guide
- Added full back button handling properly
- Split no program data into 30 minute slots for easier side scrolling
- Fixed favorite programs not sorting to the top when the preference is checked

### Known Issues
- ~~When watching live TV the live guide is set to force refresh every 30 minutes, it appears when it is supposed to refresh the loading spinner will not disappear.~~

### Possible Issues(I can't reproduce)
- None

### To do (hopefully):
- [x] Address issue #284 
~~1. Look into channel sorting~~ **My guide data is so bad I can't get good results no matter what here**
~~2. Lazy load channels~~ **Will work on in different commit**
~~3. If program is in a time slot that conflicted with another program then highlight the background to signify we don't know if this is what is actually playing~~ **Scrapped for now**

### Finished:
- [x] Replace \<No Prorgam Data\> with resource string
- [x] Skip program data if another program has filled the time slot
- [x] Channel headers can be used to surf in guide
- [x] Short press to play channel, long press for detail screen
- [x] Split no program data in 30 minute slots

**Maybe implement in the far off future**
- Do not refresh entire player when switching channels by channel header, this appears to only happen if channel doesn't work with the player(libVlc or exoplayer) currently being used
- Split \<No Program Data\> shows into 30 minute segments for the filler slot

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/GuideChannelHeader.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/GuideChannelHeader.java
@@ -1,6 +1,7 @@
 package org.jellyfin.androidtv.ui;
 
 import android.content.Context;
+import android.graphics.Rect;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.widget.AbsListView;
@@ -13,6 +14,7 @@ import com.bumptech.glide.Glide;
 import org.jellyfin.androidtv.R;
 import org.jellyfin.androidtv.TvApp;
 import org.jellyfin.androidtv.ui.livetv.LiveTvGuideActivity;
+import org.jellyfin.androidtv.ui.playback.CustomPlaybackOverlayFragment;
 import org.jellyfin.androidtv.util.ImageUtils;
 import org.jellyfin.androidtv.util.Utils;
 import org.jellyfin.apiclient.interaction.ApiClient;
@@ -27,15 +29,16 @@ public class GuideChannelHeader extends RelativeLayout {
 
     private ImageView mChannelImage;
     private ChannelInfoDto mChannel;
-    private Context mActivity;
+    private Context mContext;
+    private CustomPlaybackOverlayFragment mActivity;
 
-    public GuideChannelHeader(Context context, ChannelInfoDto channel) {
+    public GuideChannelHeader(Context context, CustomPlaybackOverlayFragment fragment, ChannelInfoDto channel, boolean focusable) {
         super(context);
-        initComponent(context, channel);
+        initComponent(context, fragment, channel, focusable);
     }
 
-    private void initComponent(Context context, ChannelInfoDto channel) {
-        mActivity = context;
+    private void initComponent(Context context, CustomPlaybackOverlayFragment fragment, ChannelInfoDto channel, boolean focusable) {
+        mContext = context;
         mChannel = channel;
         LayoutInflater inflater = LayoutInflater.from(context);
         View v = inflater.inflate(R.layout.channel_header, this, false);
@@ -45,14 +48,39 @@ public class GuideChannelHeader extends RelativeLayout {
         ((TextView) findViewById(R.id.channelName)).setText(channel.getName());
         ((TextView) findViewById(R.id.channelNumber)).setText(channel.getNumber());
         mChannelImage = (ImageView) findViewById(R.id.channelImage);
+
+        if (fragment != null) {
+            mActivity = fragment;
+            this.setFocusable(focusable);
+            setOnClickListener(new OnClickListener() {
+                @Override
+                public void onClick(View v) {
+                    fragment.switchChannelKeepGuide(channel.getId());
+                }
+            });
+        }
     }
 
     public void loadImage() {
-        Glide.with(mActivity)
+        Glide.with(mContext)
                 .load(ImageUtils.getPrimaryImageUrl(mChannel, get(ApiClient.class)))
                 .override(IMAGE_WIDTH, IMAGE_HEIGHT)
                 .centerInside()
                 .into(mChannelImage);
+    }
+
+    @Override
+    protected void onFocusChanged(boolean gainFocus, int direction, Rect previouslyFocusedRect) {
+        super.onFocusChanged(gainFocus, direction, previouslyFocusedRect);
+
+        if (gainFocus) {
+            setBackgroundColor(Utils.getThemeColor(getContext(), android.R.attr.colorAccent));
+
+            mActivity.setSelectedProgram(this);
+        } else {
+            setBackground(getResources().getDrawable(R.drawable.light_border));
+        }
+//        TvApp.getApplication().getLogger().Debug("Focus on " + mProgram.getName() + " was " + (gainFocus ? "gained" : "lost"));
     }
 
 }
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/GuideChannelHeader.java
```diff
@@ -45,14 +48,39 @@ private void initComponent(Context context, ChannelInfoDto channel) {
...
+        super.onFocusChanged(gainFocus, direction, previouslyFocusedRect);
+
+        if (gainFocus) {
+            setBackgroundColor(Utils.getThemeColor(getContext(), android.R.attr.colorAccent));
+
+            mActivity.setSelectedProgram(this);
+        } else {
+            setBackground(getResources().getDrawable(R.drawable.light_border));
+        }
+//        TvApp.getApplication().getLogger().Debug("Focus on " + mProgram.getName() + " was " + (gainFocus ? "gained" : "lost"));
```
nielsvanvelzen, line 83: Either remove, uncomment or add a comment explaining why this line is commented to avoid confusion for other devs

MrChip53: I'll pull it. That line was already there. I didn't know why myself.

app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
```diff
@@ -767,26 +797,36 @@ protected void onPostExecute(Void aVoid) {
...
-            cell.setLast();
-            cell.setFirst();
+            int minutes = ((Long)((mCurrentLocalGuideEnd - mCurrentLocalGuideStart) / 60000)).intValue();
+            int slot = 0;
+            do {
+                BaseItemDto empty = new BaseItemDto();
+                empty.setName("  " + getString(R.string.no_program_data));
+                empty.setChannelId(channelId);
+                empty.setStartDate(TimeUtils.convertToUtcDate(new Date(mCurrentLocalGuideStart + ((30*slot) * 60000))));
+                empty.setEndDate(TimeUtils.convertToUtcDate(new Date(mCurrentLocalGuideStart + ((30+(slot+1)) * 60000))));
```
thornbill, line 818: Shouldn’t this be `30*(slot+1)`?

app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
```diff
@@ -767,26 +797,36 @@ protected void onPostExecute(Void aVoid) {
...
+                cell.setId(currentCellId++);
+                cell.setLayoutParams(new ViewGroup.LayoutParams(30 * PIXELS_PER_MINUTE, ROW_HEIGHT));
+                cell.setFocusable(true);
+                programRow.addView(cell);
+                if (slot == 0)
+                    cell.setFirst();
+                if (slot == (minutes / 30) - 1)
+                    cell.setLast();
+                slot++;
+            } while((30*slot)*60000 < minutes*60000);
```
thornbill, line 829: It shouldn’t be necessary to multiply both sides by 60000 here.

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -842,19 +910,35 @@ protected void onPostExecute(Void aVoid) {
...
+
+            int minutes = ((Long)((mCurrentLocalGuideEnd - mCurrentLocalGuideStart) / 60000)).intValue();
+            int slot = 0;
+
+            do {
+                BaseItemDto empty = new BaseItemDto();
+                empty.setName("  " + getString(R.string.no_program_data));
+                empty.setChannelId(channelId);
+                empty.setStartDate(TimeUtils.convertToUtcDate(new Date(mCurrentLocalGuideStart + ((30*slot) * 60000))));
+                empty.setEndDate(TimeUtils.convertToUtcDate(new Date(mCurrentLocalGuideStart + ((30+(slot+1)) * 60000))));
```
thornbill, line 929: Same here

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -842,19 +910,35 @@ protected void onPostExecute(Void aVoid) {
...
+                cell.setId(currentCellId++);
+                cell.setLayoutParams(new ViewGroup.LayoutParams(30 * PIXELS_PER_MINUTE, LiveTvGuideActivity.ROW_HEIGHT));
+                cell.setFocusable(true);
+                programRow.addView(cell);
+                if (slot == 0)
+                    cell.setFirst();
+                if (slot == (minutes / 30) - 1)
+                    cell.setLast();
+                slot++;
+            } while((30*slot)*60000 < minutes*60000);
```
thornbill, line 940: Same here

thornbill (changes requested):
Looks good overall! I noticed a couple small things, but still want to test it out.

app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
MrChip53: Yes. Fixed.

app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
MrChip53: Agree. Didn't realize the redundancy. Fixed.

app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
```diff
@@ -260,7 +260,10 @@ protected void onResume() {
     }
 
     protected void doLoad() {
-        if (TvManager.shouldForceReload() || System.currentTimeMillis() > mLastLoad + 3600000) {
+        Calendar curTime = Calendar.getInstance();
+        curTime.setTime(new Date());
```
thornbill, line 264: Why did you change this from using `System.currentTimeMillis()`?

thornbill (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
MrChip53: I changed a lot of these to use Calendar. For some reason, it appears that the calendar.getTimeMs fucntion adds some sort of zero padding and so Calendars current time would not ever equal Systems current time for me. I can look into this further and see if it was just my bad code or a real issue. The mCurrentGuide longs are set based off of Calendar variables.

thornbill (approved):

Issue #561, opened 2020-09-14
Automatically choose function for player


okan35:
Basically I would like to know how automatically choose function works for players. Because I have some video files that only work on libvlc and mxplayer.

When a video plays om mxplayer I get aac3 not supported error so no sound but when I play it with libvlc it works or another video file which is 4k doesn't work on libvlc but works on mxplayer.

So it would be good  if we could add additional functionality on auto choose player where it tries to play the video if it doesn't, it could try the other existing player.

Or where we have play button, we could put additional button which says play with and user could choose additional player which is suitable for my situation.

okan35:
This can be closed it was implemented with "Feature playWith pull request".

PR #571, opened 2020-09-24
Feature play with button

okan35:
Feature to choose player from the menu where we have play button. I created this because some files play fine on vlc and some play fine on external app. It is better to have such button than going to settings and changing there.

**Changes**
Added play with button next to play button. This button will only be visible if preferred player is set external app.

**Issues**
#561

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -174,12 +174,6 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Include system logs in crash reports
 		 */
 		var acraIncludeSystemLogs = Preference.boolean(ACRA.PREF_ENABLE_SYSTEM_LOGS, true)
-
-		/**
-		 * chosen player for play with button - changes every time user chooses a player with "play with" button
-		 */
-
-		var chosenPlayer = Preference.enum("chosen_player",PreferredVideoPlayer.VLC)
 	}
 
 	init {
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -174,6 +174,12 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Include system logs in crash reports
 		 */
 		var acraIncludeSystemLogs = Preference.boolean(ACRA.PREF_ENABLE_SYSTEM_LOGS, true)
+
+		/**
+		 * chosen player for play with button - changes every time user chooses a player with "play with" button
+		 */
+
+		var chosenPlayer = Preference.enum("chosen_player",PreferredVideoPlayer.VLC)
 	}
 
 	init {
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
@@ -841,7 +841,7 @@ public class FullDetailsActivity extends BaseActivity implements IRecordingIndic
     private TextUnderButton queueButton = null;
     private TextUnderButton deleteButton = null;
     private TextUnderButton moreButton;
-    private TextUnderButton playWith = null;
+    private TextUnderButton playWithButton = null;
 
     private void addButtons(int buttonSize) {
         String buttonLabel;
@@ -939,7 +939,7 @@ public class FullDetailsActivity extends BaseActivity implements IRecordingIndic
         }
 
 
-        mDetailsOverviewRow.addAction(playWith);
+
         if (mBaseItem.getLocalTrailerCount() != null && mBaseItem.getLocalTrailerCount() > 0) {
             TextUnderButton trailer = new TextUnderButton(this, R.drawable.ic_trailer, buttonSize, getString(R.string.lbl_play_trailers), new View.OnClickListener() {
                 @Override
@@ -1267,16 +1267,20 @@ public class FullDetailsActivity extends BaseActivity implements IRecordingIndic
                 more.show();
             }
         });
+        if (userPreferences.getValue().get(UserPreferences.Companion.getVideoPlayer()) == PreferredVideoPlayer.EXTERNAL) {
+            playWithButton = new TextUnderButton(this, R.drawable.ic_add, buttonSize, 3, "Play with", new View.OnClickListener() {
+                @Override
+                public void onClick(View view) {
+                    Toast.makeText(mApplication, "Play With Button", Toast.LENGTH_SHORT).show();
+                    PopupMenu more = new PopupMenu(mActivity, view);
+                    more.inflate(R.menu.menu_details_play_with);
+                    more.setOnMenuItemClickListener(playWithMenuListener);
+                    more.show();
+                }
+            });
+            mDetailsOverviewRow.addAction(playWithButton);
+        }
 
-        playWith = new TextUnderButton(this, R.drawable.ic_add, buttonSize, 3, "Play with", new View.OnClickListener() {
-            @Override
-            public void onClick(View view) {
-
-                PopupMenu more = new PopupMenu(mActivity, view);
-                more.inflate(R.menu.menu_details_play_with);
-                more.setOnMenuItemClickListener(playWithMenuListener);
-            }
-        });
         moreButton.setVisibility(View.GONE);
         mDetailsOverviewRow.addAction(moreButton);
         if (mBaseItem.getBaseItemType() != BaseItemType.Episode) showMoreButtonIfNeeded();  //Episodes check for previous and then call this above
@@ -1349,12 +1353,15 @@ public class FullDetailsActivity extends BaseActivity implements IRecordingIndic
 
                 case R.id.play_with_vlc:
                         userPreferences.getValue().set(UserPreferences.Companion.getVideoPlayer(), PreferredVideoPlayer.VLC);
+                    play(mBaseItem, 0, false);
                     return true;
                 case R.id.play_with_exo:
                     userPreferences.getValue().set(UserPreferences.Companion.getVideoPlayer(), PreferredVideoPlayer.EXOPLAYER);
+                    play(mBaseItem, 0, false);
                     return true;
                 case R.id.play_with_external:
                     userPreferences.getValue().set(UserPreferences.Companion.getVideoPlayer(), PreferredVideoPlayer.EXTERNAL);
+                    play(mBaseItem, 0, false);
                     return true;
 
             }
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
@@ -14,6 +14,7 @@ import android.view.KeyEvent;
 import android.view.MenuItem;
 import android.view.View;
 import android.widget.PopupMenu;
+import android.widget.Toast;
 
 import androidx.leanback.app.BackgroundManager;
 import androidx.leanback.app.RowsSupportFragment;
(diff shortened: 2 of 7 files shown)


app/src/main/res/menu/menu_details_play_with.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu
+    xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <item android:title="Play with Vlc"
```
nielsvanvelzen, line 5: All titles should use strings so they can be translated.

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -1260,6 +1267,19 @@ public void onClick(View v) {
                 more.show();
             }
         });
+        if (userPreferences.getValue().get(UserPreferences.Companion.getVideoPlayer()) == PreferredVideoPlayer.EXTERNAL) {
+            playWithButton = new TextUnderButton(this, R.drawable.ic_add, buttonSize, 3, "Play with", new View.OnClickListener() {
```
nielsvanvelzen, line 1271: "Play with" should use a string so it can be translated.

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -1260,6 +1267,19 @@ public void onClick(View v) {
...
+                @Override
+                public void onClick(View view) {
+                    Toast.makeText(mApplication, "Play With Button", Toast.LENGTH_SHORT).show();
```
nielsvanvelzen, line 1274: This toast looks like it was intended for debugigng purposes, should probably remove it or replace it with a logger debug message.

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -1326,6 +1346,29 @@ public boolean onMenuItemClick(MenuItem item) {
         }
     };
 
+    PopupMenu.OnMenuItemClickListener playWithMenuListener = new PopupMenu.OnMenuItemClickListener() {
+        @Override
+        public boolean onMenuItemClick(MenuItem item) {
+            switch (item.getItemId()) {
+
+                case R.id.play_with_vlc:
+                        userPreferences.getValue().set(UserPreferences.Companion.getVideoPlayer(), PreferredVideoPlayer.VLC);
```
nielsvanvelzen, line 1355: User preferences should **never** be changed without the user explicitly telling the application to do so. In this case you should find a way to set the player for the next play call without changing the setting.

nielsvanvelzen (changes requested):
Thanks for your PR. I did an initial review on the changed code and would like to request some changes to your implementation:
- Right now the playWithButton is added when the preferred player is "External". This should be changed to a new player type called "Choose" (string: "Always ask")
- In adition, the playWithButton should replace the play button
- When a player is chosen for playback the user preferences should **not** be changed, instead only the next playing file should use the chosen player.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -174,6 +174,12 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Include system logs in crash reports
 		 */
 		var acraIncludeSystemLogs = Preference.boolean(ACRA.PREF_ENABLE_SYSTEM_LOGS, true)
+
+		/**
+		 * chosen player for play with button - changes every time user chooses a player with "play with" button
+		 */
+
+		var chosenPlayer = Preference.enum("chosen_player",PreferredVideoPlayer.VLC)
```
nielsvanvelzen, line 182: Should use SystemPreferences

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -888,18 +893,25 @@ public void onError(Exception exception) {
...
+                    }
+                });
+                mDetailsOverviewRow.addAction(play);    
+            }
+            
             if (resumeButtonVisible) {
                 mResumeButton.requestFocus();
             } else {
-                play.requestFocus();
+                if (!isPlayerPrefChoose) {
```
nielsvanvelzen, line 911: Use `else if`

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -1260,6 +1276,18 @@ public void onClick(View v) {
                 more.show();
             }
         });
+        if (userPreferences.getValue().get(UserPreferences.Companion.getVideoPlayer()) == PreferredVideoPlayer.CHOOSE) {
```
nielsvanvelzen, line 1279: Probably easier to add an `else` statement when checking if the play button should be shown

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -1260,6 +1276,18 @@ public void onClick(View v) {
...
+            playWithButton = new TextUnderButton(this, R.drawable.ic_add, buttonSize, 3, getString(R.string.play_with), new View.OnClickListener() {
```
nielsvanvelzen, line 1280: Since either the play or play with button is shown you can probably reuse the `playButton` property

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -478,6 +491,16 @@ public void onResponse(StreamInfo internalResponse) {
...
                                         (!DeviceUtils.isFireTvStick() ||
                                                 (vlcResponse.getMediaSource().getVideoStream() != null && vlcResponse.getMediaSource().getVideoStream().getWidth() < 1000));
+                            } else if (preferredVideoPlayer == PreferredVideoPlayer.CHOOSE) {
+                                PreferredVideoPlayer preferredVideoPlayerByPlayWith = userPreferences.getValue().get(UserPreferences.Companion.getChosenPlayer());
+                                if (preferredVideoPlayerByPlayWith == PreferredVideoPlayer.VLC) {
+                                    useVlc = true;
+                                } else if (preferredVideoPlayerByPlayWith == PreferredVideoPlayer.EXOPLAYER) {
+                                    // Make sure to not use VLC
+                                    useVlc = false;
+                                }
```
nielsvanvelzen, line 501:
```suggestion
                                useVlc = preferredVideoPlayerByPlayWith == PreferredVideoPlayer.VLC;
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -478,6 +491,16 @@ public void onResponse(StreamInfo internalResponse) {
...
+
+                                System.out.println("PREFERRED PLAYER " + preferredVideoPlayerByPlayWith.name());
```
nielsvanvelzen, line 503: Use `Timber` for logging

app/src/main/res/values/strings.xml
```diff
@@ -403,4 +404,8 @@
     <string name="not_deleted">Item NOT Deleted</string>
     <string name="turn_off">Turn this option off</string>
     <string name="just_one">Just this one</string>
+    <string name="play_with_vlc">Play with Vlc</string>
```
nielsvanvelzen, line 407:
```suggestion
    <string name="play_with_vlc">Play with libVLC</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -403,4 +404,8 @@
...
+    <string name="play_with_exo_player">Play with Exo Player</string>
```
nielsvanvelzen, line 408:
```suggestion
    <string name="play_with_exo_player">Play with ExoPlayer</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -403,4 +404,8 @@
...
+    <string name="play_with_external_app">Play with External App</string>
```
nielsvanvelzen, line 409:
```suggestion
    <string name="play_with_external_app">Play with external app</string>
```

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -1325,6 +1353,42 @@ public boolean onMenuItemClick(MenuItem item) {
             return false;
         }
     };
+    PlaybackController playbackController= new  PlaybackController(MediaManager.getCurrentVideoQueue());
```
nielsvanvelzen, line 1356: I think this is unused

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -1325,6 +1353,42 @@ public boolean onMenuItemClick(MenuItem item) {
...
+
+    PopupMenu.OnMenuItemClickListener playWithMenuListener = new PopupMenu.OnMenuItemClickListener() {
```
nielsvanvelzen, line 1358: Mark private

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -112,6 +112,19 @@ public PlaybackController(List<BaseItemDto> items, IPlaybackOverlayFragment frag
         useVlc = userPreferences.getValue().get(UserPreferences.Companion.getVideoPlayer()) == PreferredVideoPlayer.VLC;
     }
 
+    public PlaybackController(List<BaseItemDto> items) {
```
nielsvanvelzen, line 115: I think this is unused too

nielsvanvelzen (changes requested):
Some files have weird formatting or too many line breaks. Please check those and fix where necessary, but  try to avoid reformatting the whole file.

nielsvanvelzen (changes requested):
- The "choose" option shows up for LiveTV. But I don't have LiveTV setup so I can't test if that works
- The "choose" button shows up on the artist page but music shouldn't do that

okan35:
I haven't even seen anything related to music, So I have to check this. I also don't listen to music on jellyfin either but I believe if the page is music we don't show play with button, right ?

nielsvanvelzen:
The FullDetailsActivity is shared for pretty much everything that exists and the play with button did show up for an artist for me.

okan35:
So basically if the current page is for music, we don't show play with button, right ?

nielsvanvelzen:
It happens now but it shouldn't

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -884,55 +889,71 @@ public void onError(Exception exception) {
...
 
-        if (BaseItemUtils.canPlay(mBaseItem)) {
-            mDetailsOverviewRow.addAction(mResumeButton);
-            boolean resumeButtonVisible = (mBaseItem.getBaseItemType() == BaseItemType.Series && !mBaseItem.getUserData().getPlayed()) || (mBaseItem.getCanResume());
-            mResumeButton.setVisibility(resumeButtonVisible ? View.VISIBLE : View.GONE);
-
-            TextUnderButton play = new TextUnderButton(this, R.drawable.ic_play, buttonSize, 2, getString(BaseItemUtils.isLiveTv(mBaseItem) ? R.string.lbl_tune_to_channel : mBaseItem.getIsFolderItem() ? R.string.lbl_play_all : R.string.lbl_play), new View.OnClickListener() {
+        //playButton becomes playWith button
+        if (userPreferences.getValue().get(UserPreferences.Companion.getVideoPlayer()) == PreferredVideoPlayer.CHOOSE) {
+            playButton = new TextUnderButton(this, R.drawable.ic_add, buttonSize, 3, getString(R.string.play_with), new View.OnClickListener() {
```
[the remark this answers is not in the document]
nielsvanvelzen: This one isn't fixed yet. We should use the same icon.

nielsvanvelzen:
I just tested it and it doesn't work for episodes

okan35:
I will check it now, what exactly is not working ?

nielsvanvelzen:
The normal play button showed up in the episode page instead of the play with button

nielsvanvelzen (approved):

PR #576, opened 2020-09-29
Fix keyboard not closing when focus is lost

### LINKED ISSUE #167: User not identified + can't enter user manually
Jellyfin server: 10.4.0.
nVidia Shield AndroidTV Jellyfin version: 0.9.1g
I only have nVidia remote to navigate.

1) When I open up Jellyfin I'm presented with a "Select a user", that doesn't display any users. The server seems to be correctly identified.
2) Under "Other options" I select "Enter Manually".
3) I get a "Enter user name" and a keyboard. I enter the user I've created in the server.
4) When pressing next it still reads "Enter user name" on the top but the "blue line" becomes same 
color as the text. Also the keyboard change to something containing all characters (including special characters, numbers etc.). I'm assuming I'm going to enter a password here, which I do, and then click the Enter sign.
5) The app wraps back to state 1 above, and I'm not logged in or anything.

2019-10-14_21-38-18_screenshot

thornbill:
**Changes**
This fixes an issue where Android does not close the soft keyboard when the text input view loses focus.

**Issues**
Fixes #167

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/shared/KeyboardFocusChangeListener.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/shared/KeyboardFocusChangeListener.kt
@@ -5,12 +5,10 @@ import android.view.View
 import android.view.inputmethod.InputMethodManager
 
 class KeyboardFocusChangeListener : View.OnFocusChangeListener {
-	override fun onFocusChange(v : View?, hasFocus : Boolean) {
-		if (v != null) {
-			(v.context.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager).apply {
-				if (!hasFocus) hideSoftInputFromWindow(v.windowToken, 0)
-				else showSoftInput(v, 0)
-			}
+	override fun onFocusChange(v : View, hasFocus : Boolean) {
+		(v.context.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager).apply {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/shared/KeyboardFocusChangeListener.kt
```diff
@@ -0,0 +1,16 @@
+package org.jellyfin.androidtv.ui.shared
+
+import android.content.Context
+import android.view.View
+import android.view.inputmethod.InputMethodManager
+
+class KeyboardFocusChangeListener : View.OnFocusChangeListener {
+	override fun onFocusChange(v : View?, hasFocus : Boolean) {
+		if (v != null) {
```
nielsvanvelzen, line 9:
```suggestion
	override fun onFocusChange(v : View, hasFocus : Boolean) {
```

If the docs are clear that it won't be null you can just remove the nullability.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/shared/KeyboardFocusChangeListener.kt
thornbill: I blame Android Studio for adding the `?` there. :wink:

nielsvanvelzen (approved):

nielsvanvelzen (approved):

PR #577, opened 2020-09-30
Fix Live TV settings

MrChip53:
**Changes**
Fixes saving and loading of Live TV settings

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/TvApp.java
+++ b/app/src/main/java/org/jellyfin/androidtv/TvApp.java
@@ -56,6 +56,7 @@ import static org.koin.java.KoinJavaComponent.inject;
 )
 @AcraLimiter
 public class TvApp extends Application {
+    public static final String DISPLAY_PREFS_APP_NAME = "ATV";
     public static final String CREDENTIALS_PATH = "org.jellyfin.androidtv.login.json";
 
     public static final int LIVE_TV_GUIDE_OPTION_ID = 1000;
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
I think this feature will break in 10.7 since the move to EF Core made the server devs remove this feature and implementing a "backwards-compatible" implementation that allows just a few keys. Need to check though. Relevant file is here: https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Api/Controllers/DisplayPreferencesController.cs

MrChip53:
So should this get moved to client side preferences then?

nielsvanvelzen:
Yes, I think it a good solution is to move all of it to `SystemPreferences.kt`. What do you think @thornbill ?

MrChip53:
Also, can web save client side preferences reliably? If not, would web have server devs come up with a solution for them?

nielsvanvelzen:
All the properties the webui saves in the displaypreferences have been added to the server so nothing will break there.

MrChip53:
Web saves these settings the exact same way we do. In 10.6 if we wanted to we could sync these with web settings to have the same experience across the two clients. This change will break the webui live tv settings unless I'm understanding wrong.

Should an issue be opened on server to address this?

Relevant files:

https://github.com/jellyfin/jellyfin-web/blob/25cddbcce3ef1c603e0a1576f7bd5e300553795a/src/components/guide/guide-settings.js

https://github.com/jellyfin/jellyfin-web/blob/e2aff662038bdf0fb730f9811975a02bb13a52be/src/scripts/settings/userSettings.js

thornbill:
I think we will need to convince the server team to fix support for this for 10.7.

MrChip53:
You want me to make an issue for the server?

thornbill:
The server team seems on board with adding back this functionality, but it would probably be good to add an issue there to track it.

thornbill (approved):
:+1: LGTM and seems to fix the issue!

PR #579, opened 2020-10-01
Adds the ability to favorite live tv channels

MrChip53:
**Changes**
Lets the user favorite live tv channels.
- Adds a heart in the channel header for channels marked as favorite
- Adds favorite button in `showProgramOptions` popup
- Long press on channel headers to quick toggle favorite

**Issues**
None

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/LiveProgramDetailPopup.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/LiveProgramDetailPopup.java
@@ -37,7 +37,6 @@ import timber.log.Timber;
 import static org.koin.java.KoinJavaComponent.inject;
 
 public class LiveProgramDetailPopup {
-    private final int MOVIE_HEIGHT = Utils.convertDpToPixel(TvApp.getApplication(), 540);
     private final int NORMAL_HEIGHT = Utils.convertDpToPixel(TvApp.getApplication(), 400);
 
     private PopupWindow mPopup;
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/LiveProgramDetailPopup.java
```diff
@@ -337,6 +344,37 @@ public void onClick(View v) {
         return tune;
     }
 
+    public android.widget.ImageButton createFavoriteButton() {
```
nielsvanvelzen, line 347: Why not just "ImageButton"?

MrChip53: I looked into this more and initially it was because I was lazy and didn't look into the `org.jellyfin.androidtv.ui.ImageButton` class. After experimenting with that class the default android ImageButton is what should be used here. The other buttons('Tune to Channel', 'Record', etc) are default android Buttons so the styling matches. In this file ImageButton defaults to the jellyfin image button so the full package name has to be specified for the android version.

So if you are asking why I got specific with the package it's becuase `ImageButton` maps to the jellyfin specific class. If you are asking why I don't use the jellyfin ImageButton, it does not match the style of the rest of the popup.

nielsvanvelzen: I didn't realize we had our own ImageButton in the same package. It makes sense now to me!

thornbill (approved):

PR #583, opened 2020-10-01
Remove color vomit card backgrounds

thornbill:
**Changes**
Removes the functionality that pulls the background color of cards from the image being displayed in the card for a more uniform appearance.

**Before**
color vomit cards

**After**
grey cards

**Issues**
N/A


nielsvanvelzen (commented):
Do we still need the palette library?

thornbill:
Good call. It doesn't look like it's used anywhere else.

nielsvanvelzen (approved):

PR #586, opened 2020-10-03
Fix unlabeled image buttons for accessibility

### LINKED ISSUE #418: Unlabeled buttons in application's interface and player activity
Hey there,
I have been testing this application in my android TV and I just can say this is absolutely amazing. previously I have tried to use the web player (context: I am a blind user and rely in accessibility services to navigate through the interface) and a native application works way better than the web interface. I already switched to Jellyfin from Plex and the Android TV app.

However I just would like to ask, if possible, could you add labels to buttons in the application interface? Basically when opening a TV show, where you see options for playing, shuffle, playing next and stuff, all of those buttons are unlabeled, which makes my accessibility service to report all of those as "image button". Also, in the player activity the situation is about the same, all of those buttons are reported as "unlabeled" and that's basically all.

If you need more info, [Here is an article about adding information for accessibility services in android.](https://support.google.com/accessibility/android/answer/7158690?hl=en)
If labeled correctly, that would be basically the best thing we could have as an accessible media center software, as basically all others have some parts incorrectly labeled or unlabeled at all.
Thank you very much for the great project.

jemlule:
**Changes**

- Add `setContentDescription` to the buttons unlabeled: those to sort the movies/musics, those before the play/resume (put to favorite, mark the media viewed…) and those in the playback overlay.
- Add 3 strings to translate, *lbl_pause*, *lbl_audio_track* and *lbl_subtitle_track*, because I didn't find any string that would have matched.

**Issues**

Fixes #418

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
@@ -91,9 +91,6 @@ public class LiveTvGuideActivity extends BaseActivity implements ILiveTvGuide {
     private HorizontalScrollView mTimelineScroller;
     private View mSpinner;
     private View mResetButton;
-    private View mFilterButton;
-    private View mOptionsButton;
-    private View mDateButton;
 
     private BaseItemDto mSelectedProgram;
     private ProgramGridCell mSelectedProgramView;
@@ -139,7 +136,7 @@ public class LiveTvGuideActivity extends BaseActivity implements ILiveTvGuide {
         mSpinner = findViewById(R.id.spinner);
         mSpinner.setVisibility(View.VISIBLE);
 
-        mFilterButton = findViewById(R.id.filterButton);
+        View mFilterButton = findViewById(R.id.filterButton);
         mFilterButton.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View v) {
@@ -148,7 +145,7 @@ public class LiveTvGuideActivity extends BaseActivity implements ILiveTvGuide {
         });
         mFilterButton.setContentDescription(getString(R.string.lbl_filters));
 
-        mOptionsButton = findViewById(R.id.optionsButton);
+        View mOptionsButton = findViewById(R.id.optionsButton);
         mOptionsButton.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View v) {
@@ -157,7 +154,7 @@ public class LiveTvGuideActivity extends BaseActivity implements ILiveTvGuide {
         });
         mOptionsButton.setContentDescription(getString(R.string.lbl_other_options));
 
-        mDateButton = findViewById(R.id.dateButton);
+        View mDateButton = findViewById(R.id.dateButton);
         mDateButton.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View v) {
--- a/app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
@@ -91,6 +91,9 @@ public class LiveTvGuideActivity extends BaseActivity implements ILiveTvGuide {
     private HorizontalScrollView mTimelineScroller;
     private View mSpinner;
     private View mResetButton;
+    private View mFilterButton;
+    private View mOptionsButton;
+    private View mDateButton;
 
     private BaseItemDto mSelectedProgram;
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
```diff
@@ -142,20 +142,23 @@ public void onClick(View v) {
                 showFilterOptions();
             }
         });
+        findViewById(R.id.filterButton).setContentDescription(getString(R.string.lbl_filters));
```
nielsvanvelzen, line 145: `findViewById` is an expensive operation so it's better to assign the found view to a variable before using it to avoid finding a view multiple times. (same for the next 2)

app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingActivity.java
```diff
@@ -235,8 +243,14 @@ public void onClick(View v) {
         mPlayPauseButton.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View v) {
-                if (MediaManager.isPlayingAudio()) MediaManager.pauseAudio();
-                else MediaManager.resumeAudio();
+                if (MediaManager.isPlayingAudio()) {
+                    MediaManager.pauseAudio();
+                    mPlayPauseButton.setContentDescription(getString(R.string.lbl_play));
```
nielsvanvelzen, line 248: You need to add this somewhere around line 441

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
```diff
@@ -63,15 +65,24 @@ private void initActions(Context context) {
         fastForwardAction = new PlaybackControlsRow.FastForwardAction(context);
         skipNextAction = new PlaybackControlsRow.SkipNextAction(context);
         selectAudioAction = new SelectAudioAction(context, this);
+        selectAudioAction.setLabels(new String[]{TvApp.getApplication().getString(R.string.lbl_audio_track)});
```
nielsvanvelzen, line 68: Use `context.getString()`

nielsvanvelzen (changes requested):

jemlule:
I think I just made a mistake by fetching and rebasing the upstream commits. Should I git rebase and drop the upstream commits?

nielsvanvelzen:
Yes, please do

app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
```diff
@@ -136,26 +139,32 @@ public void onCreate(Bundle savedInstanceState) {
         mSpinner = findViewById(R.id.spinner);
         mSpinner.setVisibility(View.VISIBLE);
 
-        findViewById(R.id.filterButton).setOnClickListener(new View.OnClickListener() {
+        mFilterButton = findViewById(R.id.filterButton);
```
nielsvanvelzen, line 142: You can add the property in the function like so:

```suggestion
        View mFilterButton = findViewById(R.id.filterButton);
```

And then remove the `private View xx` declarations.

app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingActivity.java
```diff
@@ -424,7 +438,14 @@ private void updateButtons(final boolean playing) {
             @Override
             public void run() {
                 mPoster.setKeepScreenOn(playing);
-                mPlayPauseButton.setState(!playing ? ImageButton.STATE_PRIMARY : ImageButton.STATE_SECONDARY);
+                if (!playing) {
+                    mPlayPauseButton.setState(ImageButton.STATE_PRIMARY);
+                    mPlayPauseButton.setContentDescription(getString(R.string.lbl_play));
+                }
+                else {
```
nielsvanvelzen, line 445:
```suggestion
                } else {
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingActivity.java
nielsvanvelzen: And you can now remove it from this line and line 252

nielsvanvelzen (changes requested):

nielsvanvelzen:
Hi @jemlule, are you still interested in finishing this pull request?

jemlule:
Yes I am! Sorry for being inactive.

nielsvanvelzen (approved):

Issue #103, opened 2019-05-14
Feature request: navigate with number keys in live TV


godkarmachine:
I'd love to see the ability to navigate live TV using the number keys.  I have a remote I use with my Shield that behaves like a USB keyboard, and it has a 10-key number pad on it.  I'd love to be able to change channels just by inputting a number.

Specific behavior I'd like to see:
* When viewing a live TV channel, inputting a number (up to 4 digits) changes to the selected channel.
* When viewing the guide, inputting a number moves to that channel in the guide.

Additionally, it'd be nice to have the Page Up/Page Down keys function as channel up/channel down keys, to go to the next higher/lower channel number while viewing live TV.

PR #587, opened 2020-10-05
Live TV button navigations

MrChip53:
**Changes**
Change channels with number buttons on remote
Jump/Last button goes to previous channel
Channel up/down work as they should also

**Todo**
- [x] Channel up and down buttons
- [x] Last channel button
- [x] Sort channels by channel number if they arent already

**Issues**

Relevant issue: #103

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
@@ -517,7 +517,7 @@ public class CustomPlaybackOverlayFragment extends Fragment implements IPlayback
             if (event.getAction() == KeyEvent.ACTION_DOWN) {
 
                 if (mPlaybackController.isLiveTv()) {
-                    if (mChannelNumberVisible && KeyUtilsKt.isDPADCenter(keyCode)) {
+                    if (mChannelNumberVisible && KeyUtilsKt.isSelect(keyCode)) {
                         mHandler.removeCallbacks(mHideNumberTask);
                         switchChannelByNumber(mChannelNumber);
                         hideChannelNumberView();
@@ -525,7 +525,7 @@ public class CustomPlaybackOverlayFragment extends Fragment implements IPlayback
                     }
 
                     if (!mGuideVisible) {
-                        if (keyCode >= KeyEvent.KEYCODE_0 && keyCode <= KeyEvent.KEYCODE_9) {
+                        if (KeyUtilsKt.isNumber(keyCode)) {
                             leanbackOverlayFragment.setShouldShowOverlay(false);
                             if (mPopupPanelVisible)
                                 hidePopupPanel();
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -620,6 +688,75 @@ public boolean onKey(View v, int keyCode, KeyEvent event) {
         }
     };
 
+    private void switchChannelByNumber(String number) {
+        List<ChannelInfoDto> channels = TvManager.getAllChannels();
+        for (ChannelInfoDto channel:
+                channels) {
+            if (channel.getNumber().equals(number)) {
```
thornbill, line 695: A NullPointerException is thrown here if an invalid channel number is entered.

app/src/main/java/org/jellyfin/androidtv/util/KeyUtils.kt
```diff
@@ -0,0 +1,24 @@
...
+object KeyUtils {
+	@JvmStatic
+	fun isSelect(code: Int) = code == KeyEvent.KEYCODE_DPAD_CENTER || code == KeyEvent.KEYCODE_ENTER
+
+	@JvmStatic
+	fun isNumber(code: Int) = code >= KeyEvent.KEYCODE_0 && code <= KeyEvent.KEYCODE_9
+
+	@JvmStatic
+	fun isChannelSurfUp(code: Int) = code == KeyEvent.KEYCODE_PAGE_UP || code == KeyEvent.KEYCODE_CHANNEL_UP
+	@JvmStatic
```
thornbill, line 14: Spacing is inconsistent in this file. I would prefer a blank line before each annotated function like was done above.

thornbill (changes requested):
Looks good overall. I couldn't seem to find a way to emulate the paging behavior in the emulator. I did notice entering an incorrect channel number caused an NPE. This also don't seem to be allowing "." to be entered for digital OTA channels like "3.1". I'm not sure how we should account for those channels since remotes may not have a "." button either.

MrChip53:
Okay im going to move this back to draft and try some things out to address the 3.1, 3.2, etc issue. I think it may require a decent rework of the channel up/down logic.

PrplHaz4:
> 
> 
> Looks good overall. I couldn't seem to find a way to emulate the paging behavior in the emulator. I did notice entering an incorrect channel number caused an NPE. This also don't seem to be allowing "." to be entered for digital OTA channels like "3.1". I'm not sure how we should account for those channels since remotes may not have a "." button either.

Maybe this could be handled like a currency input with an applied `.` that always has one decimal place (when the channel list includes digital OTA channels).

So for channel 3 you'd enter 3-0 = 3.0, 3.1 you'd enter 3-1 = 3.1...

MrChip53:
I think what i will look at doing is when the user inputs 3 it will have just 3 in the corner but then a drop down list with any .x channels if they exist. That way if a remote doesnt have a . button those channels will still be accessible.

MrChip53:
Closing this for now

PR #578, opened 2020-09-30
Login redesign

thornbill:
**Changes**
The long awaited redesign for the login workflow.
* Moves login workflow to MVVM architecture
* Uses a single activity for startup and login workflow

To Do:
* [x] Finalize data structure for users and servers
* [x] Error handling and input validation
* [x] Prevent duplicate server listings
* [x] Remove old login code

Follow Up Tasks:
* [ ] Saving servers and previously logged in users (current implementation only saves most recent login)
* [ ] Finish ui/ux of server listing screen (currently relies on existing `GridButton` implementation)
* [ ] Finish ui/ux of login screen
* [ ] PIN support
* [ ] Managing saved servers and users (option to remove/forget)
* [ ] Use kotlinx.serialization instead of gson
* [ ] Quick connect support

**Issues**

PR #588, opened 2020-10-05
Update string lbl_ip_hint

tinect:
Update string lbl_ip_hint

**Changes**
Adjust string `lbl_ip_hint` to show valid url.

**Issues**
It is not possible to login via ip-adresse cause it results in `com.android.volley.VolleyError invalid Url`


nielsvanvelzen:
The string shown should work in the current and next release. Only on the current master it doesn't because we're actively reworking that (See #578)

Issue #575, opened 2020-09-29
Aspect ratio for the Live TV library tiles on the app homescreen are wrong


Nickbert7:
**Describe the bug**
Aspect ratio for the Live TV library  tiles on the app homescreen are wrong, they thould be the same as the other cards.

**Screenshots**

![image](https://user-images.githubusercontent.com/14938497/94559662-dc894900-0261-11eb-8bd2-c471a9f7422a.png)

**System (please complete the following information):**
 - Android, Android TV and Fire TV 
 - Device manufacturer: Amazon, Samsung
 - Device model: Fire TV Stick (2nd Gen), Samsung Phone/Tablet
 - Jellyfin server version: 10.6.4

**Additional context**
As mentioned here I created a new seperate issue https://github.com/jellyfin/jellyfin-androidtv/issues/197

PR #589, opened 2020-10-06
Fix Live TV tiles aspect ratio in CardPresenter.java

MrChip53:
**Changes**
Fix Live TV tile aspect ratio.

**Issues**
#575 

...you be the judge 😅

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
@@ -17,7 +17,6 @@ import org.jellyfin.androidtv.util.ImageUtils;
 import org.jellyfin.androidtv.util.TimeUtils;
 import org.jellyfin.androidtv.util.Utils;
 import org.jellyfin.apiclient.model.dto.BaseItemDto;
-import org.jellyfin.apiclient.model.dto.BaseItemType;
 import org.jellyfin.apiclient.model.dto.UserItemDataDto;
 import org.jellyfin.apiclient.model.entities.LocationType;
 import org.jellyfin.apiclient.model.livetv.ChannelInfoDto;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
```diff
@@ -90,7 +91,10 @@ public void setItem(BaseRowItem m, String imageType, int lHeight, int pHeight, i
                     } else if (imageType.equals(ImageType.THUMB)) {
                         aspect = ImageUtils.ASPECT_RATIO_16_9;
                     } else {
-                        aspect = Utils.getSafeValue(ImageUtils.getImageAspectRatio(itemDto, m.getPreferParentThumb()), ImageUtils.ASPECT_RATIO_7_9);
+                        if (itemDto.getBaseItemType() == BaseItemType.UserView)
```
nielsvanvelzen, line 94: Make this an `else if`

nielsvanvelzen (changes requested):

thornbill:
This stretches out the default card images used in the app. We should probably have different images for these that use the proper aspect ratio.

MrChip53:
Maybe we could find another fix that changes aspect ratio only if a custom image is defined. I'll look into it later.

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
MrChip53: Completely changed the method used due to the issue raised below.

MrChip53:
@thornbill @nielsvanvelzen I believe this fix is more proper and addresses the stretched image issue

thornbill (approved):
:+1: This seems like a good approach to me.

nielsvanvelzen (approved):

PR #591, opened 2020-10-06
General cleanup

thornbill:
**Changes**
* Removes unused classes
* Removes unused utility methods
* Removes unused imports
* Remove unneeded SDK version checks
* Replace SDK version ints with constants

**Issues**
N/A


app/src/main/java/org/jellyfin/androidtv/TvApp.java
```diff
@@ -197,11 +199,7 @@ public boolean canManageRecordings() {
     }
 
     public Drawable getDrawableCompat(int id) {
```
nielsvanvelzen, line 201: Better option is to just remove this function from the TvApp class

thornbill: Agreed. Wasn't sure if it belonged in this PR though since I started out just removing stuff and this is used _everywhere_. :rofl:

nielsvanvelzen: Right, I'll merge this one then

nielsvanvelzen (approved):

PR #597, opened 2020-10-14
Bump kotlin version

thornbill:
**Changes**
Updates the Kotlin version to current. No migrations were needed.

**Issues**
N/A

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -5,8 +5,9 @@ buildscript {
 	}
 
 	dependencies {
+		val kotlinVersion: String by project
 		classpath("com.android.tools.build:gradle:4.0.1")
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -5,8 +5,9 @@ buildscript {
 	}
 
 	dependencies {
+		val kotlinVersion : String by project
```
nielsvanvelzen, line 8:
```suggestion
		val kotlinVersion: String by project
```

nielsvanvelzen (commented):
I'd prefer adding it to the Dependencies.kt in the buildSrc but realized we don't have that yet for this repo! So this approach seems fine to me :)

nielsvanvelzen (approved):

PR #615, opened 2020-11-06
Re-add "Ends at" time to playback control overlay

MrChip53:
**Changes**
Add the "Ends At: ####" back to the control overlay

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlueHelper.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlueHelper.java
@@ -1,0 +1,460 @@
+/*
+ * Copyright (C) 2016 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlueHelper.java
```diff
@@ -0,0 +1,460 @@
...
+                };
+
+        PlaybackTransportRowPresenter rowPresenter = new PlaybackTransportRowPresenter() {
+            @Override
+            protected void onBindRowViewHolder(RowPresenter.ViewHolder vh, Object item) {
+                super.onBindRowViewHolder(vh, item);
+                vh.setOnKeyListener(CustomPlaybackTransportControlGlueHelper.this);
+                endsAt = new TextView(getContext());
+                endsAt.setText("Test");
+                LinearLayout view = (LinearLayout)vh.view;
```
MrChip53, line 168: Here is the start of the custom code. I store the top row of buttons to a view, create a RelativeLayout, add the buttons to the relative layout along with the "ends at" text then add the relative layout back to the playback control view.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlueHelper.java
```diff
@@ -0,0 +1,460 @@
...
+        if (action != null) {
+            if (event.getAction() == KeyEvent.ACTION_DOWN) {
+                dispatchAction(action, event);
+            }
+            return true;
+        }
+        return false;
+    }
+
+    public void setEndTime(String endTime) {
```
MrChip53, line 255: this function is just used to update the text

PR #616, opened 2020-11-06
Re-add "Ends at" time to playback control overlay

MrChip53:
**Changes**
Add the "Ends At: ####" back to the control overlay

**Issues**


MrChip53:
Just realized I didn't do a proper string resource but before I worry about fixing let me know if you all approve of this method and then if the colon should be in the ends at string

nielsvanvelzen:
Test notes:

- [ ] When playback starts it takes a second or so before the "ends at" shows up
- [ ] The format is always in 12-hour clock notation
- [ ] The text style is different from the time/duration text

For the second you can use the [DateFormat](https://developer.android.com/reference/java/text/DateFormat.html) class which should follow system preferences.
`DateFormat.getTimeInstance(DateFormat.SHORT).format(date)`

MrChip53:
> Test notes:
> 
> * [x]  When playback starts it takes a second or so before the "ends at" shows up
> * [x]  The format is always in 12-hour clock notation
> * [x]  The text style is different from the time/duration text
> 
> For the second you can use the [DateFormat](https://developer.android.com/reference/java/text/DateFormat.html) class which should follow system preferences.
> `DateFormat.getTimeInstance(DateFormat.SHORT).format(date)`

I can't check these off but I'm pretty sure I addressed all of them. I also added some null checks but I don't know if they are really needed. The delay on the text showing up is due to when `updatePlaybackState` gets called. I just made it show 'Loading...' by default. I also added the string resource for 'Ends at' and removed the colon so it now displays 'Ends at 9:00 PM' for example.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
```diff
@@ -59,6 +70,67 @@
...
+
+        PlaybackTransportRowPresenter rowPresenter = new PlaybackTransportRowPresenter() {
+            @Override
+            protected RowPresenter.ViewHolder createRowViewHolder(ViewGroup parent) {
+                RowPresenter.ViewHolder vh = super.createRowViewHolder(parent);
+                Context context = getContext();
+
+                if (context != null) {
+                    mEndsText = new TextView(context);
+                    mEndsText.setTextAppearance(context, R.style.PlaybackControlsText);
```
nielsvanvelzen, line 94: Can we use `playbackControlsTimeStyle` here?

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:leanback/leanback/src/main/res/layout/lb_playback_transport_controls_row.xml;l=113

nielsvanvelzen (changes requested):
>  I just made it show 'Loading...' by default.

I noticed the video duration shows up earlier so it should be possible to show the "ends at" text earlier too. Just need to find the right place to add it.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
MrChip53: Apparently, yes. I did not know that. Although it's not using the attr but instead the style the attr value points to.

MrChip53:
> > I just made it show 'Loading...' by default.
> 
> I noticed the video duration shows up earlier so it should be possible to show the "ends at" text earlier too. Just need to find the right place to add it.

I addressed both the issues. Should we save the context to a variable and replace all calls to `getContext()` with the variable?

MrChip53:
I over complicated the solution. I'm going to move the textview creation back inside the onCreateViewHolder function and use the setEndsText function in there to remove the need for the null pointer check. But still, are we okay with storing the context received on class creation in a variable and reusing or are the calls to get context not taxing enough to worry about that?

nielsvanvelzen:
The parent class stores the context and `getContext` just returns it, so no need to store ourselves.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
```diff
@@ -197,6 +273,15 @@ public void onCustomActionClicked(Action action, View view) {
         }
     }
 
+    public void setEndTime() {
```
nielsvanvelzen, line 276: I think the implementation of this method could be added to the existing `updatePlayState` method.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
```diff
@@ -197,6 +273,15 @@ public void onCustomActionClicked(Action action, View view) {
...
+        long msLeft = playerAdapter.getDuration() - playerAdapter.getCurrentPosition();
+        Calendar ends = Calendar.getInstance();
+        ends.setTimeInMillis(ends.getTimeInMillis() + msLeft);
+        Context context = getContext();
+        if (context != null && mEndsText != null)
```
nielsvanvelzen, line 281: These null checks are unnecessary

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
```diff
@@ -51,6 +62,9 @@
     private ArrayObjectAdapter primaryActionsAdapter;
     private ArrayObjectAdapter secondaryActionsAdapter;
 
+    //Custom views
```
nielsvanvelzen, line 65:
```suggestion
    // Injected views
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
```diff
@@ -59,6 +73,68 @@
...
+                        viewHolder.getTitle().setText(glue.getTitle());
+                        viewHolder.getSubtitle().setText(glue.getSubtitle());
+                    }
+                };
+
+        PlaybackTransportRowPresenter rowPresenter = new PlaybackTransportRowPresenter() {
+            @Override
+            protected RowPresenter.ViewHolder createRowViewHolder(ViewGroup parent) {
+                RowPresenter.ViewHolder vh = super.createRowViewHolder(parent);
+                Context context = getContext();
```
nielsvanvelzen, line 93:
```suggestion
                Context context = parent.getContext();
```

That one should always be not-null

nielsvanvelzen (changes requested):
When you pause a video for multiple minutes the "ends at" won't update. It will update when playback starts again but this is slightly confusing. We might want to start a handler or similar to update the contents every minute/few seconds while a video is paused - make sure to destroy it when playback starts again or the view is destroyed.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
MrChip53: I was thinking to leave it as is so if we want to call it somewhere else we don't need duplicate code.

MrChip53: Actually do you mean the function call or the contents of the function itself? Do we want the setEndTime function to be private? For now I will make it private and create the handler inside the Glue class so the TextView can't be modified externally.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
MrChip53: mEndsText does need to be null checked as it appears the play state gets updated before we hook into the view holder function

MrChip53:
I also found an issue where the TextView would not hide when scrolling on the seek bar. I fixed this in the handler but in order to have it be snappy I had to make the handler repeat faster. The handler also has to run 24/7 now when playing as it polls the other view for its visibility state. Let me know what you think about that and if you may have a better method that may not be so taxing, if this method even is taxing on the system.

nielsvanvelzen:
You might not need to keep the handler loop active when you use addOnLayoutChangeListener. I'm not sure if that listener is always called though.

MrChip53:
I'll try that out. I did also find another spot where I can catch the views being set invisible instantly so from there I could start a handler to check if they are visible again and then kill it so it won't need to run 24/7.

MrChip53:
Forgot to change the commit message on the first one. 🙈 Whoops lol. Anyways, I ended up doing the second approach. It seems to work nicely. The ends at time also updates while paused. I believe that address all issues.

MrChip53:
There is a live tv bug that is causing crashes either when changing the channel when a channel is already playing or when the hour/half hour happens and the program name and time is updated.

nielsvanvelzen (approved):

nielsvanvelzen:
For the current leanback player this works fine but we should definitely use another approach when rewriting the playback UI.

MrChip53:
> For the current leanback player this works fine but we should definitely use another approach when rewriting the playback UI.

I agree. I honestly don't think it would be too hard to make our own playback controls and still use it with the Glue/Host template.

nielsvanvelzen (approved):

PR #623, opened 2020-11-13
Update dependencies (November edition)

nielsvanvelzen:
**Changes**
- Compile / Target Android API 30
- Update Gradle to 6.7
- Update Android Build Tools to 4.1
- Initialize Timber before Koin (in case an injected class uses logging in constructor)
- Update Koin to 2.2
  - Enabled logger again
  - Fixed a bunch of classes using `KoinComponent`, which is discouraged
  - Added  the fragment extension
  - Switched to androidx version for viewmodel
- Updated all other dependencies

**Issues**


nielsvanvelzen:
I meant to add the link to the Koin 2.2 blog post but forgot to do so:
https://medium.com/koin-developers/whats-next-with-koin-2-2-3-0-releases-6c5464ae5e3d

thornbill:
I get the following error when I try to run with these changes in Android Studio.

> Unable to determine application id: com.android.tools.idea.run.ApkProvisionException: No outputs for the main artifact of variant: debug

nielsvanvelzen:
Is your Android studio installation updated to the latest version and did you sync the gradle configuration?

nielsvanvelzen:
I just experienced the same issue when I switched from this branch to another branch (based on current master) and a Gradle Sync fixed it.

thornbill:
Weird I had tried a Gradle sync, but restarting Android Studio fixed it for me.

app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentLiveTVRow.kt
```diff
@@ -21,12 +15,11 @@ import org.jellyfin.androidtv.ui.presentation.CardPresenter
...
 import org.jellyfin.apiclient.serialization.GsonJsonSerializer
-import org.koin.core.KoinComponent
-import org.koin.core.inject
-
-class HomeFragmentLiveTVRow(val activity: Activity) : HomeFragmentRow(), OnItemViewClickedListener, KoinComponent {
-	private val serializer: GsonJsonSerializer by inject()
 
+class HomeFragmentLiveTVRow(
+	private val activity: Activity,
+	private val serializer: GsonJsonSerializer
```
thornbill, line 21: The sample I saw seemed to suggest that we should be using `by inject()` here as the default value and would only need to override for special conditions like tests. Although their docs do not seem to reflect the suggestions in their post so :man_shrugging:

nielsvanvelzen: we can only use the `inject` and `get` functions when the current context is Koin aware. This can be done by extending the KoinComponent class or is automatically done for Activities, Fragments and Services with the Android module.
For other places passing dependencies via constructor is recommended. The apiclient uses that approach in the openapi-generator.

> use constructor style injection as much as possible.

- Blog post above

thornbill (approved):

PR #625, opened 2020-11-13
Fix some crashes when trying to play series

nielsvanvelzen:
**Changes**
- Fix crash when using the "Play all" button
- Fix crash when using the "Play Next Up" button

Both are "patches" to make it work, I suspect the underlying issue is a race condition or some other weird spaghetti monster. I've experienced both crashes quite a lot but only now found a way to fix it.

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -42,6 +42,7 @@ import org.jellyfin.apiclient.model.livetv.ChannelInfoDto;
 import org.jellyfin.apiclient.model.mediainfo.SubtitleTrackInfo;
 import org.jellyfin.apiclient.model.session.PlayMethod;
 
+import java.util.ArrayList;
 import java.util.List;
 
 import kotlin.Lazy;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -143,7 +144,17 @@ public void setPlaybackMethod(PlayMethod value) {
...
     }
-    public MediaSourceInfo getCurrentMediaSource() { return mCurrentStreamInfo != null && mCurrentStreamInfo.getMediaSource() != null ? mCurrentStreamInfo.getMediaSource() : getCurrentlyPlayingItem().getMediaSources().get(0);}
+
+    public MediaSourceInfo getCurrentMediaSource() {
+        if (mCurrentStreamInfo != null && mCurrentStreamInfo.getMediaSource() != null) {
+            return mCurrentStreamInfo.getMediaSource();
+        } else {
+            ArrayList<MediaSourceInfo> mediaSources = getCurrentlyPlayingItem().getMediaSources();
+            if (mediaSources == null || mediaSources.isEmpty()) return null;
+            else return mediaSources.get(0);
```
thornbill, line 154: Brackets please :pray:

nielsvanvelzen: Only because it's Java

thornbill (approved):

PR #626, opened 2020-11-14
Replace bitmaps with vectors

nielsvanvelzen:
**Changes**
  - Removed some unused drawables
  - Moved the moviebg.jpg from nodpi to dpi aware drawables
  - Added vector icons from:
    - item-details-rewrite branch (partially custom made icons)
    - material design icon set (built-in Android Studio feature)
    - materialdesignicons.com
  - Added vector tiles based on the above icons
  - A few renames for consistency

Some of the new icons look slightly different in comparison to the old ones. 

**Notes**

Because a bunch of graphics are cached by Glide you need to stop the app, clear the cache and start it again for the new tiles to show up.

PR #628, opened 2020-11-14
Make tiles themable

nielsvanvelzen:
Depends on #626 

**Changes**
  - Make GridButton a Kotlin class and properly tag the image as `@DrawableRes`
  - Update GridButtonPresenter to not use Glide so the drawables are theme aware
  - Each theme has it's own file now
  - Use attributes for all tile drawables
    - Colors are now set per theme
    - Emerald theme updated to use green for everything

**Issues**


nielsvanvelzen:
If a tile is loaded via Glide or not in a tree with the Jellyfin theme it will have a transparent background. I think it works everywhere because as far as I know all tiles use the GridButtonPresenter but I'll need to validate it some more.

thornbill (approved):

Issue #584, opened 2020-10-02
Main Menu Movie Cards IMDB Ratings Instead of Rotten


tomaswl1:
Hi guys just a little suggestion don't no who agrees but on the main title cards for movies you have a rotten/fresh tomato image noticed it doesn't pull the rotten rating for most movies i have added but going into a movie i have noticed every single movie as a IMDB rating could you maybe get rid of the rotten rating on main movie menu cards and maybe add the IMDB rating instead with a little IMDB logo at side of rating no worries if not and thanks for all the hard work u have put into the app already its coming along nicely

nextlooper42:
Not to say that Rotter is accurate for US only most times, and is missing anything not english. Completely irrelevant in EU, or global scale. Maybe option to pick provider would be much better.

Or at least give us option to hide rotten rating.

thornbill:
Setting added in https://github.com/jellyfin/jellyfin-androidtv/pull/630

PR #630, opened 2020-11-15
Add the ability to change the rating type shown on cards

MrChip53:
**Changes**

- Adds an option under settings to toggle between Rotten Tomatoes or IMDb for what is shown on the ImageCards.

**Issues**
#584

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
@@ -16,6 +16,8 @@ import com.bumptech.glide.Glide;
 
 import org.jellyfin.androidtv.R;
 import org.jellyfin.androidtv.constant.ImageType;
+import org.jellyfin.androidtv.preference.UserPreferences;
+import org.jellyfin.androidtv.preference.constant.RatingType;
 import org.jellyfin.androidtv.ui.itemhandling.BaseRowItem;
 import org.jellyfin.androidtv.util.ImageUtils;
 import org.jellyfin.androidtv.util.TimeUtils;
@@ -28,6 +30,8 @@ import org.jellyfin.apiclient.model.livetv.ChannelInfoDto;
 import java.util.Date;
 
 import timber.log.Timber;
+
+import static org.koin.java.KoinJavaComponent.get;
 
 public class CardPresenter extends Presenter {
     private static final String TAG = "CardPresenter";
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
```diff
@@ -394,9 +398,17 @@ public void onBindViewHolder(Presenter.ViewHolder viewHolder, Object item) {
             holder.mCardView.setPlayingIndicator(true);
         } else {
             holder.mCardView.setPlayingIndicator(false);
-            Drawable badge = rowItem.getBadgeImage();
-            if (badge != null) {
-                ((ViewHolder) viewHolder).mCardView.setBadgeImage(badge);
+
+            if (get(UserPreferences.class).get(UserPreferences.Companion.getDefaultRatingProvider()) == RatingProvider.RATING_RT) {
+                Drawable badge = rowItem.getBadgeImage();
+                ((ViewHolder) viewHolder).mCardView.setRating(null);
```
thornbill, line 404: Looks like you could use `holder` here instead of casting repeatedly 😉 .

MrChip53: ah yes. it appears I started with some copy pasta and rolled with it 🤦‍♂️

nielsvanvelzen:
With the metadata plugins I use the "stars" are not IMDb ratings. This can happen with the rotten tomatoes too as nowhere in the code did we imply it's their rating. We should therefor use more abstract name for this option.

app/src/main/java/org/jellyfin/androidtv/preference/constant/RatingProvider.kt
```diff
@@ -0,0 +1,18 @@
+package org.jellyfin.androidtv.preference.constant
+
+import org.jellyfin.androidtv.R
+import org.jellyfin.androidtv.ui.preference.dsl.EnumDisplayOptions
+
+enum class RatingProvider {
+	/**
+	 * Sets default rating provider to Rotten tomatoes.
```
nielsvanvelzen, line 8:
```suggestion
	 * Sets default rating provider to tomatoes.
```

app/src/main/java/org/jellyfin/androidtv/preference/constant/RatingProvider.kt
```diff
@@ -0,0 +1,18 @@
...
+	 */
+	@EnumDisplayOptions(R.string.lbl_tomatoes)
+	RATING_TOMATOES,
+
+	/**
+	 * Sets the default rating provider to IMDb.
```
nielsvanvelzen, line 14:
```suggestion
	 * Sets the default rating provider to stars.
```

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
```diff
@@ -394,13 +398,21 @@ public void onBindViewHolder(Presenter.ViewHolder viewHolder, Object item) {
             holder.mCardView.setPlayingIndicator(true);
         } else {
             holder.mCardView.setPlayingIndicator(false);
-            Drawable badge = rowItem.getBadgeImage();
-            if (badge != null) {
-                ((ViewHolder) viewHolder).mCardView.setBadgeImage(badge);
+
+            if (get(UserPreferences.class).get(UserPreferences.Companion.getDefaultRatingProvider()) == RatingProvider.RATING_TOMATOES) {
```
nielsvanvelzen, line 402: I think the code will be more readable if the result of the `get(UserPreferences.class).get(UserPreferences.Companion.getDefaultRatingProvider())` part is moved to a variable before the if statement.

app/src/main/res/values/strings.xml
```diff
@@ -418,4 +418,7 @@
     <string name="lbl_manual_login">Manual Login</string>
     <string name="lbl_playback_control_ends">Ends at %1$s</string>
     <string name="watch_count_overflow">99+</string>
+    <string name="lbl_stars">Stars</string>
+    <string name="lbl_tomatoes">Tomatoes</string>
+    <string name="pref_default_rating">Default Ratings Scale</string>
```
nielsvanvelzen, line 423:
```suggestion
    <string name="pref_default_rating">Default rating type</string>
```

Could apply this consistently in the code too

nielsvanvelzen (changes requested):
Works good while testing, few suggestions for the code.

app/src/main/res/values/strings.xml
MrChip53: Changed all references of "RatingProvider" or any variation of "RatingProvider" to "RatingType". I assume that is what you meant here.

nielsvanvelzen (approved):

Issue #605, opened 2020-10-23
[Feature Request] Add ability to turn off Clock


Artiume:
**Describe the feature you'd like**	
	
Make a toggle-able option to disable the clock on the home screen.

htpcBeginner:
@Artiume thanks for opening this request. Bit of additional information, Plex has this. Emby does not. Really helpful for people who do not want to see time at night due to sleep issues. 

I am going to open a similar request fo the Roku client.

nielsvanvelzen:
Added in #631

PR #631, opened 2020-11-15
Add a setting toggle to change the clocks display behavior

MrChip53:
**Changes**
Adds a setting toggle that allows the user to pick when the clock is shown.
Options are:
- Always
- In Menus
- In Video
- Never

**Issues**
#605

...shown. 🤦‍♂️

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -174,6 +174,16 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Include system logs in crash reports
 		 */
 		var acraIncludeSystemLogs = Preference.boolean(ACRA.PREF_ENABLE_SYSTEM_LOGS, true)
+
+		/**
+		 * When to show the clock.
+		 */
+		var clockBehavior = Preference.enum("pref_clock_behavior", ClockBehavior.ALWAYS)
+
+		/**
+		 * Set which ratings provider should show on MyImageCardViews
+		 */
+		var defaultRatingType = Preference.enum("pref_rating_type", RatingType.RATING_TOMATOES)
 	}
 
 	init {
--- a/app/src/main/java/org/jellyfin/androidtv/ui/ClockUserView.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/ClockUserView.java
@@ -1,17 +1,23 @@
 package org.jellyfin.androidtv.ui;
 
+import android.app.Activity;
 import android.content.Context;
+import android.content.ContextWrapper;
 import android.util.AttributeSet;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.widget.ImageView;
 import android.widget.RelativeLayout;
+import android.widget.TextClock;
 import android.widget.TextView;
 
 import com.bumptech.glide.Glide;
 
 import org.jellyfin.androidtv.R;
 import org.jellyfin.androidtv.TvApp;
+import org.jellyfin.androidtv.preference.UserPreferences;
+import org.jellyfin.androidtv.preference.constant.ClockBehavior;
+import org.jellyfin.androidtv.ui.playback.PlaybackOverlayActivity;
 import org.jellyfin.androidtv.util.ImageUtils;
 import org.jellyfin.apiclient.interaction.ApiClient;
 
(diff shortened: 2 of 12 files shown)


thornbill:
This should probably apply to the “ends at” time also.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -174,6 +174,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Include system logs in crash reports
 		 */
 		var acraIncludeSystemLogs = Preference.boolean(ACRA.PREF_ENABLE_SYSTEM_LOGS, true)
+
+		/**
+		 * When to show the clock.
+		 */
+		var showClock = Preference.enum("pref_show_clock", ClockBehavior.ALWAYS)
```
nielsvanvelzen, line 181:
```suggestion
		var clockBehavior = Preference.enum("pref_clock_behavior", ClockBehavior.ALWAYS)
```

app/src/main/res/values/strings.xml
```diff
@@ -418,4 +418,8 @@
     <string name="lbl_manual_login">Manual Login</string>
     <string name="lbl_playback_control_ends">Ends at %1$s</string>
     <string name="watch_count_overflow">99+</string>
+    <string name="pref_clock_display">Show Clock</string>
+    <string name="lbl_always">Always</string>
+    <string name="pref_clock_display_menus">In Menus</string>
+    <string name="pref_clock_display_video">In Video</string>
```
nielsvanvelzen, line 424:
```suggestion
    <string name="pref_clock_display">Show clock</string>
    <string name="lbl_always">Always</string>
    <string name="pref_clock_display_menus">While browsing</string>
    <string name="pref_clock_display_video">During playback</string>
```

app/src/main/java/org/jellyfin/androidtv/ui/ClockUserView.java
```diff
@@ -32,6 +38,33 @@ private void init(Context context) {
         LayoutInflater inflater = LayoutInflater.from(context);
         View v = inflater.inflate(R.layout.clock_user_bug, null, false);
         this.addView(v);
+        TextClock tc = v.findViewById(R.id.clock);
+        ClockBehavior showClock = get(UserPreferences.class).get(UserPreferences.Companion.getShowClock());
+        if (showClock != ClockBehavior.ALWAYS) {
```
nielsvanvelzen, line 43: The combination of this if-statement and the switch-statement for all the `setVisibility` can be simplified for better readability.

nielsvanvelzen (changes requested):
- The clock still shows on the login screen
- The logo on the home screen looks weird without a clock

MrChip53:
Addressed reviews and made the login clock respond to the setting. I don't know what you really mean by the logo looking weird. The user avatar? I think it looks okay but any suggestions on how it could look better let me know. Otherwise assuming it passes tests it should be good to go.

app/src/main/java/org/jellyfin/androidtv/ui/shared/ToolbarView.kt
```diff
@@ -7,10 +7,17 @@ import android.view.ViewGroup
...
 import org.jellyfin.androidtv.R
+import org.jellyfin.androidtv.preference.UserPreferences
+import org.jellyfin.androidtv.preference.UserPreferences.Companion.clockBehavior
+import org.jellyfin.androidtv.preference.constant.ClockBehavior
+import org.koin.java.KoinJavaComponent.get
 
 class ToolbarView(context: Context, attrs: AttributeSet) : FrameLayout(context, attrs) {
 	init {
 		inflate(context, R.layout.view_toolbar, this)
+		val clockBehavior = get(UserPreferences::class.java)[clockBehavior]
```
nielsvanvelzen, line 18:
```suggestion
		val clockBehavior = get(UserPreferences::class.java)[UserPreferences.Companion.clockBehavior]
```

Although the `UserPreferences.Companion.` doesn't look that great, it's easier to read since your variable has the exact same name. Could also just rename the variable.

nielsvanvelzen (changes requested):

nielsvanvelzen:
>  I don't know what you really mean by the logo looking weird.

I'm not sure either, looks fine to me. Past me should've been more descriptive haha.

nielsvanvelzen (approved):

Issue #612, opened 2020-11-05
Login Redesign


thornbill:
A lot of progress has been made now and the original list is now mostly finished, with some features being dropped or considered for later releases. A new list of smaller tasks related to the login UI below:

- [ ] After adding a new server the server list screens become out of sync - issue in viewpager1 needs migration
- [ ] Error reporting in the add server dialog could be more detailed, For example SSL errors happen often with our users but right now it shows the name of the SSL error exception..
- [ ] Error reporting in the add user dialog could be more detailed
- [x] "logout" should be renamed to "switch user" or something similar - #816
- [x] Normalize server address input so the user can forget to add a protocol etc. - #818
- [x] Removing a server from the server management screen doesn't update the server management screen - #815

Those are not for the login redesign but I needed to put it somewhere - @nielsvanvelzen
- [ ] Remove unused strings (avoid conflict with weblate please)

__Original below__
------

**Describe the feature you'd like**

The login workflow is being completely reimplemented. The initial work for this was handled in #578, but the following items need addressed as follow up tasks:

* Saving servers and previously logged in users (current implementation only saves most recent login)
  * [x] Saving previously logged in users and servers
  * [x] Add new server without restart  #730 
  * [x] After restarting the app my user (that I just signed in) is gone and I'm not automatically authenticated
* [x] Finish ui/ux of server listing screen
    * Currently relies on existing `GridButton` implementation
    * "Add Server" button is using default styling and may need some tweaking
    * [x] User avatars are not displayed #701
    * Round buttons?
    * ~~Go directly to add server screen if no servers are found~~
    - [x] The login screen when no server is added should show something like "no servers here yet" or maybe even show a "welcome" screen for new users. (https://f.ndat.nl/1604672131.png) #730
* [x] Finish ui/ux of add server screen
    - [x] The "enter server address" screen should submit the url when pressing the OK/ENTER button on a remote to avoid dpad navigation to the "OK" button #727
* [x] Finish ui/ux of login screen
    * Should we display user avatars when possible?
    - [x] Pressing OK/ENTER should signin instead of focussing on the OK button #726
* [ ] PIN support
* [ ] Quick connect support
* [x] Managing saved servers and users (option to remove/forget) #713 
* [ ] Support re-ordering servers?
* [ ] Use kotlinx.serialization instead of gson
* [ ] Add error messaging for invalid server or user credentials
[...]

PR #644, opened 2020-11-29
Add initial authentication storage

nielsvanvelzen:
**Changes**

- Store credentials
  - Uses Android's AccountManager to store access tokens and the ability to remove accounts from within the system settings
- Display all servers from: storage, server discovery, legacy credentials
- Display all users from: public user endpoint, storage, legacy credentials
- Display errors when adding a server/user (rough implementation)
- Sign in
- Backup Jellyfin app settings and credentials (without access tokens)

**Future pull requests**

- Cache & show profile pictures
- Remember last user

**TODO**

- [x] Remove accidently committed dev note files
- [ ] Squash all commits
- [x] Rewrite legacy credential migration code
- [x] Clean up all code

**Issues**
Part of #612 

**Dev notes**

The AlertFragment uses callbacks passed via the constructor. This is not a good idea for 2 reasons
1) Fragment's should not have custom constructors per Android guidelines, [this StackOverflow article](https://stackoverflow.com/questions/10450348/do-fragments-really-need-an-empty-constructor) explains it a bit.
2) In this case the callbacks came from another fragment which was unloaded at the time the callback executed. This caused issues were the viewmodel was unavailable and crashed the app.

We should aim for fragments that are not dependent on other fragments/activies. All UI should only interact via viewmodels with repositories/services/etc. and data should be passed via Bundles.

**Other**

I decided to use UUID everywhere since this is what the new apiclient does too. In order to make things work I added a few helper methods to convert string -> uuid that support the "simple" format (which doesn't use hyphens). Because those UUID's differ from the strings in the old credentials I needed to convert strings -> uuid -> string a bunch of times. This will all be solved when we remove the legacy credential support and upgrade to the new apiclient.

I've also decided to use Kotlin's flow internally and convert it to Android's LiveData at the end of the pipeline because a flow is much easier to work with and provides a more rich API. I added a helper method for flows too to collect them to a `LiveData<List<T>>`.


nielsvanvelzen:
This PR became quite massive and I'm a bit lost in what still needs to happen etc. so I did some final cleaning and marking it ready or review.

I think there's still some small bugs which I'd say are easier to fix in future PR's instead of this one, but let me know what you think about that @thornbill.

thornbill:
This is crashing on launch for me on a Fire Stick:

```
java.lang.IllegalStateException: Flow invariant is violated:
		Flow was collected in [StandaloneCoroutine{Active}@c69d5fa, Main [immediate]],
		but emission happened in [DispatchedCoroutine{Active}@19e77ab, LimitingDispatcher@32313e08[dispatcher = DefaultDispatcher]].
		Please refer to 'flow' documentation or use 'flowOn' instead
	at kotlinx.coroutines.flow.internal.SafeCollector_commonKt.checkContext(SafeCollector.common.kt:84)
	at kotlinx.coroutines.flow.internal.SafeCollector.checkContext(SafeCollector.kt:82)
	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:68)
	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:54)
	at org.jellyfin.androidtv.auth.ServerRepositoryImpl$getDiscoveryServers$1$1$invokeSuspend$$inlined$map$1$2.emit(Collect.kt:139)
	at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke(SafeCollector.kt:14)
	at kotlinx.coroutines.flow.internal.SafeCollectorKt$emitFun$1.invoke(SafeCollector.kt)
	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:71)
	at kotlinx.coroutines.flow.internal.SafeCollector.emit(SafeCollector.kt:54)
	at org.jellyfin.apiclient.discovery.ServerDiscovery$discover$1.invokeSuspend(ServerDiscovery.kt:111)
	at org.jellyfin.apiclient.discovery.ServerDiscovery$discover$1.invoke(ServerDiscovery.kt)
	at kotlinx.coroutines.flow.SafeFlow.collect(Builders.kt:56)
	at org.jellyfin.androidtv.auth.ServerRepositoryImpl$getDiscoveryServers$1$1$invokeSuspend$$inlined$map$1.collect(SafeCollector.common.kt:114)
	at org.jellyfin.androidtv.auth.ServerRepositoryImpl$getDiscoveryServers$1$1.invokeSuspend(ServerRepository.kt:104)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:738)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

nielsvanvelzen:
Discovery crash should be fixed now.

Used [this](https://elizarov.medium.com/execution-context-of-kotlin-flows-b8c151c9309b) source (which explains flows and coroutines quite well). Basically I did an illegal operation by using `withContext` in the flow. I should've applied the scope to the flow by using `flowOn` instead.

nielsvanvelzen:
Bug found:
- Login using auto discovery doesn't work - when authenticating the server does not exists yet and thus the address can not be found.
  - Fix for now is to manually add the server

thornbill:
I've found a couple things so far:

1. Discovered server is displayed before manually added server
2. Saved user listed twice if it's also a saved user
3. Passwordless login no longer bypasses the login form

app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
```diff
@@ -0,0 +1,138 @@
...
+			accessToken = user.accessToken
+		})
+
+		try {
+			val userDto = callApi<UserDto?> { callback -> apiClient.GetUserAsync(user.id.toString(), callback) }
+			if (userDto != null) {
+				application.currentUser = userDto
+				return true
+			}
+		} catch (err: Exception) {
```
thornbill, line 69: We should suppress these warnings.

thornbill:
The tests are failing due to a reference to some old code

```
e: /home/vsts/work/1/s/app/src/test/java/org/jellyfin/androidtv/util/apiclient/ModelExtensionsTest.kt: (45, 5): Unresolved reference: toUser
```

nielsvanvelzen:
I realized that I accidently pushed some changes I was working on for the CardPresenter to optimize it. But since those changes were ready for review I'll leave them in.

thornbill:
Just tested this again in an emulator and noticed the following issues:
1. Passwordless login of public users no longer bypasses the login form
2. Adding multiple servers seems to not work only my saved server is ever displayed.

nielsvanvelzen:
> Passwordless login of public users no longer bypasses the login form

I just realized it only works for users not saved. I'll fix that.

> Adding multiple servers seems to not work only my saved server is ever displayed.

Did you restart the app? There is a bug where new servers only show up after restart. The fix for that one will be in another PR because I expect it to be quite large.

nielsvanvelzen:
_copied from commit message_

We now remember if a user has a password (using a boolean) in the offline data set. If a user fails to authenticate because a password was added it will prompt for a password. If it's the other way around the user can just proceed in the password-prompt with an empty password and the boolean will be updated for future logins.

thornbill:
I still seem to be randomly prompted for a password when logging in and out with a passwordless user. This can probably be addressed later though. Let's get this in once the conflicts are resolved. :+1:

thornbill (approved):

thornbill (approved):

PR #652, opened 2020-12-11
Add all current translations to 0.11.5

nielsvanvelzen:
<font color=red><marquee><blink><h1>Targets the release branch</h1></blink></marquee></font>

I downloaded the latest translations from Weblate (https://translate.jellyfin.org/download/jellyfin/jellyfin-androidtv/?format=zip) and added them on top of the 0.11 release branch. I'm not sure if Android will build with the unused strings so I'll add a second commit  only if that's required.


nielsvanvelzen:
Oh snap, we removed exactly one string: pref_auto_logoff_title. I don't feel like adding it back to all the languages it existed in though 🤣

nielsvanvelzen:
Dropping this one as it's not worth the effort.

PR #659, opened 2020-12-20
Preferences 2.1

nielsvanvelzen:
**Changes**
- Added new preference type "link"
  - Can open other fragments
  - Uses icons!
- Rewrite the way preferences are loaded to support linking to other screens
  - We can now use separate screens instead of categories to organize the preferences
- Fix titles in dialogs missing

I added one new screen for now; authentication. This will add all authentication settings in their own screen. We can use this in the authentication rewrite to show the authentication settings when not signed in.

This also creates the possibility to use the preferences code for library settings etc. In the future (possible before 0.12) I want to revisit the settings and group them. The initial "settings" screen should be links only instead of one giant list of options.


nielsvanvelzen:
[HD version in Matrix](https://matrix.to/#/!qVMSlLFGYTLOFolpZT:matrix.org/$QVnVTwI1lFOI9KKckAIK-V0ahTLRTOqnayEczY3kBxo?via=bonifacelabs.ca&via=matrix.org&via=matrix.possumlodge.me)

nielsvanvelzen:
image

Icon support

thornbill (approved):

PR #662, opened 2020-12-24
Fix a bunch of "unused" warnings

nielsvanvelzen:
Fixes a few small linter issues

**Changes**
- Remove unused imports
- Remove unused class (CustomBrowseFragment.java - was used by old login UI)
- Suppress a few false-positive unused warnings
- Resolve a few other unused warnings

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -639,7 +639,6 @@ private void changeSurfaceLayout(int videoWidth, int videoHeight, int videoVisib
         double ar, vw;
```
thornbill, line 639: The declaration of `vw` could be moved to line 645 since it is only used to compute the `ar` value there.

thornbill (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
nielsvanvelzen: Fixed, I've also removed some more junk code from this file.
Some properties were set but never read, other were read but never set.

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -488,10 +452,7 @@ public void destroy() {
     }
 
     private void createPlayer(int buffer, boolean isInterlaced) {
-        if (mVlcPlayer != null && mIsInterlaced == isInterlaced && mCurrentBuffer == buffer) return; // don't need to re-create
```
nielsvanvelzen, line 491: For this check; `mCurrentBuffer` was never set so this if-statement would never execute the return.

thornbill (approved):

PR #664, opened 2020-12-24
Convert ChannelManager to work manager

nielsvanvelzen:
The ChannelManager is an asynchronous background job to update information in the OS. The current implementation relied on the app running and calling the `.update()` function (in the home activity). This PR changes it to use the androidx work manager library so we can start updating the channels when the app is not running. This can hopefully be finished after the new auto-sign in code is merged so it actually works with the app in the background. Since we use Koin for dependency injection I used that as backend for the workers.

**References**
- [Koin release notes explaining workmanager in DI](https://medium.com/koin-developers/whats-next-with-koin-2-2-3-0-releases-6c5464ae5e3d)
- [WorkManager class reference](https://developer.android.com/reference/androidx/work/WorkManager)
- [Work manager docs](https://developer.android.com/topic/libraries/architecture/workmanager/advanced/custom-configuration)
- [Work manager tutorial](https://medium.com/androiddevelopers/workmanager-basics-beba51e94048)

**Draft state**

I want to test this change on my own television before marking it as ready.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
@@ -43,8 +43,8 @@ class LeanbackChannelWorker(
 		 */
 		private const val TICKS_IN_MILLISECOND = 10000
 
-		const val SINGLE_UPDATE_REQUEST_NAME = "channelUpdateRequest"
-		const val PERIODIC_UPDATE_REQUEST_NAME = "channelUpdateRequest"
+		const val SINGLE_UPDATE_REQUEST_NAME = "LeanbackChannelSingleUpdateRequest"
+		const val PERIODIC_UPDATE_REQUEST_NAME = "LeanbackChannelPeriodicUpdateRequest"
 	}
 
 	private val application = TvApp.getApplication()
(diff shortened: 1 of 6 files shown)


nielsvanvelzen:
I can confirm it works perfectly fine after some days of testing

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -32,14 +32,19 @@ import org.jellyfin.apiclient.model.querying.NextUpQuery
...
-	private companion object {
+) : CoroutineWorker(context, workerParams) {
+	companion object {
 		/**
 		 * Amount of ticks found in a millisecond, used for calculation
 		 */
 		private const val TICKS_IN_MILLISECOND = 10000
+
+		const val SINGLE_UPDATE_REQUEST_NAME = "channelUpdateRequest"
+		const val PERIODIC_UPDATE_REQUEST_NAME = "channelUpdateRequest"
```
thornbill, line 47: Why are these names the same?

nielsvanvelzen: Since they are different request types it doesn't matter too much, I should've used different names for readability though.

thornbill (approved):

thornbill:
You have a conflict on this one now. :sweat_smile:

nielsvanvelzen:
Fixed 😉

Issue #49, opened 2019-02-20
Android TV App doesn't support HDR


fuzzynet:
The Android TV app on my Shield TV will direct play my 4k movies, but doesn't enable HDR.
The same movie will enable HDR on the emby app.

nextlooper42:
This is not enhancement, is a bug. Basically every 4k video has HDR and this makes those videos unwatchable, eg. washed out...

sparky8251:
Since HDR support requires backend changes in addition to major front end changes to work, it *is* an enhancement and not a bug.

HDR support isn't magicaly supported without us having to put in effort. It's a deliberate feature addition since currently, almost none of the JF stack is HDR capable (with 10.4.0 we have put one of a few required backend changes for HDR into place).

nextlooper42:
I get that. Just saying that JF should be developed with HDR and possible future "enhancements" in mind, that includes newer/current ffmpeg (nightly is now running on 4.0.2, which is quite old) to support features that matter in that time, not to be stuck in past.

Stop thinking about Emby (i hate that name), JF should be better than that :D

anthonylavado:
@nextlooper42 The good news is - we've updated our core server code to use either 4.0.x or 4.1, and work is ongoing for 4.2. We are thinking about it - it's just a question of time.

sparky8251:
If I'm not mistaken, the nightly docker uses ffmpeg 4.2. But yeah, we have started with HDR support at ffmpeg.

If I'm not mistaken, @vitorsemeano has been working on HDR playback support (by moving to exoplayer) in web and/or android, but it's not going to land in 10.4.z since it still needs work.

We are getting there, slowly but surely. It's just a LOT of work to add in HDR support :)

viggy96:
Just wanted to check in here since Jellyfin uses ffmpeg 4.2.1 now. Hopefully this comment will stop more people from asking constantly.

nielsvanvelzen:
HDR support _should_ work after #244 is merged as exoplayer 2 [supports](https://exoplayer.dev/supported-formats.html#hdr-video-playback) it.

viggy96:
I built and installed the Android TV app yesterday, and while it plays HDR content, it does not properly display the HDR colours. This is most likely due to the server detecting that the video codec is not compatible, and the server is transcoding the file, when its not necessary. The file in question is HEVC, Main 10 profile, yuv420p10le pixel format.

Ebrithil95:
[...]

PR #671, opened 2021-01-02
Enable hevc support universally

thornbill:
**Changes**
This enables support for HEVC universally. According to [this](https://developer.android.com/guide/topics/media/media-formats#video-codecs) it should be supported on anything running Android 5+ which is the minimum version we support in the app. I did notice that some of the first gen Amazon Fire devices do not support it, so we should probably add some specific exclusions for those.

**Issues**
Should help with #49


nielsvanvelzen (approved):
Looks good

PR #672, opened 2021-01-02
Remove constructors from fragments

nielsvanvelzen:
Depends on #669

**Changes**
- Removed constructor from NextUpFragment
  - Refactored the code to use a view model
- Remove constructor from StartupToolbarFragment
  - Simple approach by calling a function in the activity

**Todo**
- [ ] Remove constructor from AlertFragment
- [ ] Remove constructor from UserLoginFragment
- [ ] Remove constructor from AddServerFragment

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpActivity.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpActivity.kt
@@ -1,82 +1,44 @@
 package org.jellyfin.androidtv.ui.playback.nextup
 
-import android.graphics.Bitmap
+import android.content.Intent
 import android.os.Bundle
 import androidx.fragment.app.FragmentActivity
 import androidx.leanback.app.BackgroundManager
-import com.bumptech.glide.Glide
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
-import org.jellyfin.androidtv.util.apiclient.getItem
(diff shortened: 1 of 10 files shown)


nielsvanvelzen:
I skipped the AlertFragment and subclasses because I expect it to be a big change, so it's easier to do in it's own PR.

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpActivity.kt
```diff
@@ -1,82 +1,44 @@
...
+			when (state) {
+				// Open next item
+				NextUpState.PLAY_NEXT -> {
+					startActivity(Intent(this, PlaybackOverlayActivity::class.java))
+					finish()
+				}
+				// Close activity
+				NextUpState.CLOSE -> finish()
+				// Unknown state
+				else -> Unit
```
thornbill, line 27: Should we log this?

nielsvanvelzen: The other possibilities are:
- NO_DATA, which is an indicator that we don't know the details of the next item, everything will work as normal but no episode title or image is shown
- INITIALIZED - Which is the default value and persists when data is correctly loaded

PLAY_NEXT and CLOSE are "actions" which this listener is meant to listen for, I therefore don't think logging is required.

thornbill (approved):

thornbill:
Should we add an issue to track removing the other instances?

nielsvanvelzen:
I'll probably get to those before 0.12 releases, we could probably add it to #612 though.

PR #673, opened 2021-01-03
Remove NextUpActivity from displaying after intro videos or trailers

MrChip53:
**Changes**
Removes the next up activity from showing up after the intro and trailer videos.

This works by setting intro videos manually to the trailer type and immediately playing the next item in the queue if the current item is a trailer.

This is the default behavior of v0.11

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -1048,7 +1048,9 @@ public class PlaybackController {
         if (nextItem != null) {
             Timber.d("Moving to next queue item. Index: " + (mCurrentIndex + 1));
 
-            if (userPreferences.getValue().get(UserPreferences.Companion.getNextUpEnabled())) {
+            BaseItemDto curItem = getCurrentlyPlayingItem();
+
+            if (userPreferences.getValue().get(UserPreferences.Companion.getNextUpEnabled()) && curItem.getBaseItemType() != BaseItemType.Trailer) {
                 // Show "Next Up" fragment
                 spinnerOff = false;
                 mFragment.showNextUp(nextItem.getId());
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -1048,7 +1048,9 @@ private void itemComplete() {
         if (nextItem != null) {
             Timber.d("Moving to next queue item. Index: " + (mCurrentIndex + 1));
 
-            if (userPreferences.getValue().get(UserPreferences.Companion.getNextUpEnabled())) {
+            BaseItemDto curItem = getCurrentlyPlayingItem();
```
thornbill, line 1051: It looks like `getCurrentlyPlayingItem()` can return `null` so we need to handle that properly.

app/src/main/java/org/jellyfin/androidtv/util/apiclient/PlaybackHelper.java
```diff
@@ -217,7 +217,11 @@ public void onResponse(ChannelInfoDto response) {
                         @Override
                         public void onResponse(ItemsResult response) {
                             if (response.getTotalRecordCount() > 0){
-                                Collections.addAll(items, response.getItems());
+
+                                for(BaseItemDto intro : response.getItems()){
```
nielsvanvelzen, line 221:
```suggestion
                                for (BaseItemDto intro : response.getItems()) {
```

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
MrChip53: I believe this is now resolved.

nielsvanvelzen (approved):
Linter task failure is unrelated

thornbill (approved):

PR #678, opened 2021-01-05
Remove more unnecessary popups

nielsvanvelzen:
**Changes**
- You can now mark shows as watched without a stupid popup
  - You can just unmark it again if it was not intended, but why would you press the button without intention anyway!?
- Shuffle music is now 10% easier (removed a stupid popup)
- Marking played/unplayed in that hidden menu that nobody knows about was improved too
  - I removed the popup

TL;DR: Searched for all "AlertDialog.Builder" instances in the code and picked a bunch that I 100% think should be removed.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
@@ -1,7 +1,5 @@
 package org.jellyfin.androidtv.util;
 
-import android.app.AlertDialog;
-import android.content.DialogInterface;
 import android.content.Intent;
 import android.view.Gravity;
 import android.view.KeyEvent;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
```diff
@@ -420,47 +418,8 @@ public void onError(Exception exception) {
...
-                            public void onClick(DialogInterface dialog, int which) {
-                                    markPlayed();
-                            }
-                        }).show();
-
-                    } else {
-                        markPlayed();
-                    }
-                    return true;
                 case MENU_UNMARK_PLAYED:
```
thornbill, line 421: Should this case not be calling `markUnplayed()`?

nielsvanvelzen: Yup, I misread it.

thornbill (approved):

PR #688, opened 2021-01-10
Remove live tv mode 

nielsvanvelzen:
Depends on #684

**Changes**
Removed live tv mode. I think we managed to break it in the auth rewrite as it crashed the app after signing in. I think this feature is useless because it wasn't thought out. You couldn't go back to the home view.

**Issues**


MrChip53:
Is live tv mode just the setting to jump straight to live tv?

nielsvanvelzen:
I will rebase this one after the homesections PR is merged

MrChip53 (approved):

thornbill (approved):

PR #689, opened 2021-01-10
Add mutable object adapter

nielsvanvelzen:
**Changes**

The default ObjectAdapter's from Leanback are:
1. Not type safe (no generic used, returns "Any" / "Object")
2. Not null-safe
3. Incompatible with Kotlin collections

To fix this I made a new adapter called the "MutableObjectAdapter". It uses a generic to set the element type, this fixes 1 and 2. It then implements Iterable for 3.
To illustrate the abilities of this new adapter I made some changes in the ListServerFragment that uses it, and while I was at it I changes the behavior of the server list to not reset itself but instead only append.
Items in the adapter now have a id, the Row class in leanback likes to use a Long for that so I convert uuid->long now. I might create a custom row class so we can set the id using generics or something later.

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/presentation/MutableObjectAdapter.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/presentation/MutableObjectAdapter.kt
@@ -1,0 +1,30 @@
+package org.jellyfin.androidtv.ui.presentation
+
+import androidx.leanback.widget.ObjectAdapter
+import androidx.leanback.widget.Presenter
+import androidx.leanback.widget.PresenterSelector
+
+/**
+ * A leanback ObjectAdapter using a Kotlin list as backend. Implements Iterable to allow collection
+ * operations. And uses generics for strong typing. Uses a MutableList as internal stucture.
+ */
+class MutableObjectAdapter<T : Any> : ObjectAdapter, Iterable<T> {
+	private val data = mutableListOf<T>()
+
+	// Constructors
+	constructor(presenterSelector: PresenterSelector) : super(presenterSelector)
+	constructor(presenter: Presenter) : super(presenter)
+	constructor() : super()
+
+	// ObjectAdapter
+	override fun size(): Int = data.size
+	override fun get(index: Int): T? = data.getOrNull(index)
+
+	// Iterable
+	override fun iterator(): Iterator<T> = data.iterator()
+
+	// Custom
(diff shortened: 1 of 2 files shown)


MrChip53 (approved):
LGTM 💰

app/src/main/java/org/jellyfin/androidtv/ui/presentation/MutableObjectAdapter.kt
```diff
@@ -0,0 +1,55 @@
...
+		data.set(index, element)
+		notifyItemRangeChanged(index, 1)
+	}
+
+	fun clear() {
+		val size = data.size
+		if (size == 0) return
+
+		notifyItemRangeRemoved(0, size)
+		data.clear()
```
thornbill, line 47: I don’t imagine this should cause an issue, but should the data be cleared before notifying here?

app/src/main/java/org/jellyfin/androidtv/ui/startup/ListServerFragment.kt
```diff
@@ -76,26 +82,30 @@ class ListServerFragment : RowsSupportFragment() {
 	}
 
 	private fun buildRows(servers: Map<Server, Set<User>>) {
-		val rowAdapter = ArrayObjectAdapter(CustomListRowPresenter())
-
 		servers.forEach { (server, users) ->
-			Timber.d("Adding server row %s", server.name)
+			val exists = rowAdapter.any { it.id == server.id.mostSignificantBits and Long.MAX_VALUE }
```
thornbill, line 86: Also it maybe worth adding a comment to explain what `server.id.mostSignificantBits and Long.MAX_VALUE` is doing here as it’s not immediately clear.

app/src/main/java/org/jellyfin/androidtv/ui/presentation/MutableObjectAdapter.kt
nielsvanvelzen: The hook on `notifyItemRangeRemoved` _could_ try to read the items so to be safe I clear the data after notifying of the removal.

app/src/main/java/org/jellyfin/androidtv/ui/startup/ListServerFragment.kt
nielsvanvelzen: Done

thornbill (approved):

PR #701, opened 2021-01-28
Add user images to login screen

thornbill:
**Changes**
Adds support for user images to the login screen

**Issues**
N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/auth/model/AuthenticationStoreUser.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/auth/model/AuthenticationStoreUser.kt
@@ -11,5 +11,6 @@ import java.util.*
 data class AuthenticationStoreUser(
 	val name: String,
 	@SerialName("last_used") val lastUsed: Long = Date().time,
-	@SerialName("require_password") val requirePassword: Boolean = true
+	@SerialName("require_password") val requirePassword: Boolean = true,
+	@SerialName("primary_image_tag") val primaryImageTag: String? = null
 )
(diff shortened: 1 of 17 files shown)


app/src/main/java/org/jellyfin/androidtv/auth/model/AuthenticationStoreUser.kt
```diff
@@ -11,5 +11,6 @@ import java.util.*
 data class AuthenticationStoreUser(
 	val name: String,
 	@SerialName("last_used") val lastUsed: Long = Date().time,
-	@SerialName("require_password") val requirePassword: Boolean = true
+	@SerialName("require_password") val requirePassword: Boolean = true,
+	@SerialName("primary_image_tag") val primaryImageTag: String? = null
```
nielsvanvelzen, line 15:
```suggestion
	@SerialName("image_tag") val imageTag: String? = null
```

It's not like we're going to add more types of images so this makes more sense to me

nielsvanvelzen (changes requested):

nielsvanvelzen:
Turns out you don't need an image tag at all for profile pictures. The URL generated looks something like this:
`https://[jellyfin]/Users/[user-id]/Images/Primary?EnableImageEnhancers=true&MaxHeight=370`

No tag in the URL at all, so we don't need to store the tag.

thornbill:
I updated the naming of the image tag variable. As we discussed in chat, it probably makes sense to keep the tag as it's used for cache busting when the image is changed.

nielsvanvelzen (approved):

Issue #702, opened 2021-01-29
Search page uses wrong card design and is missing backgrounds


nielsvanvelzen:
**Describe the bug**
- The new transparent cards are enabled but they still show shadow
- The page doesn't use backgrounds like pretty much all other pages

PR #703, opened 2021-01-29
Rewrite background code

### LINKED ISSUE #674: delayed change in Background image in full details screen movie or tv series
Hi as the title says when you go into a movie and the full details screen shows theres a background image for that movie but if you then go into another movie full detail screen the background from the previous movie still shows for a few seconds before changing like its delayed

nielsvanvelzen:
**Changes**
- Move from leanback BackgroundManager to custom written BackgroundService
  - Uses slideshows for items with multiple backgrounds
  - Better animations
  - More easy to use
  - Adds a semi-transparent black overlay to the backgrounds to make text easier to read
- Move all background code from activities/fragments to the new service
- Add AbstractActivityLifecycleCallbacks that implements the interface so we don't have all the empty functions

Uses androidx.window alpha because there is no stable version yet.

**Issues**  
Fixes #674  
See #702


nielsvanvelzen:
I need to make some changes to fix the linter issues. Apparently I've used quite a lot of functions not available in older Android versions.

MrChip53:
I have found a bug where when going back out of an activity it keeps the last loaded activity.

Example: go to a TV or Movie library -> all items, scroll over an item to load its background -> push back, background is kept as the previously hovered item.

MrChip53 (approved):
LGTM but I'm not very well versed in Kotlin yet so @thornbill give BackgroundService.kt and BackgroundServiceFragment.kt a good second look. Rest was pretty straight forward converting old background management to new background management.

thornbill (approved):

PR #704, opened 2021-01-29
Disable LeakCanary by default and add option to gradle properties to enable

nielsvanvelzen:
Simple change, will disable LeakCanary by default, we should probably enable it by default after the playback rewrite since most of the leaks come from there.


MrChip53 (approved):
Merge now please 🤣

thornbill (approved):

PR #710, opened 2021-01-30
Use Timber everywhere

nielsvanvelzen:
> Since Timber is included in the project, it is likely that calls to Log should instead be going to Timber.

> Since Timber handles String#format() automatically, use this instead of String concatenation.


MrChip53 (approved):

thornbill (approved):

thornbill:
This one has conflicts now :wink:

PR #713, opened 2021-01-31
Manage Servers menu

MrChip53:
**Changes**
Adds a preferences style menu to manage servers and their respective users.

Known issues:
~~The back stack.~~

To-do in different PRs:
- Return more friendly strings for last used, something like, a few seconds ago, a few minutes ago, 7 minutes ago, 2 hours ago, 10 days ago, over a month ago.
- Add user images?
- Add the add server button into preferences?

**Issues**


nielsvanvelzen:
I don't think the confirmation dialog is necessary, we've removed these kinds of dialogs from other places in the app too. When a user presses the "remove user/server" button just remove it. The "manage servers" screen could use the server address as description, additionally an "add server" button could be added. The "users screen  could use an "add user" button too, we could use the last active time as description for those users, it would also be cool if we could how profile pictures instead of the icons here. You could wrap the users in a `category {}` so they are separated from the delete server action.

app/src/main/java/org/jellyfin/androidtv/ui/preference/dsl/OptionsLink.kt
```diff
@@ -18,9 +20,16 @@ class OptionsLink(
...
+	var extras: Bundle = bundleOf()
+	var clickListener: Preference.OnPreferenceClickListener? = null
 
-	inline fun <reified T : OptionsFragment> withFragment() {
+	inline fun <reified T : OptionsFragment> withFragment(extraBundle: Bundle = bundleOf()) {
 		fragment = T::class
+		extras = extraBundle
+	}
+
+	fun setListener(preferenceClickListener: Preference.OnPreferenceClickListener) {
```
nielsvanvelzen, line 31: I don't like the clicklistener in the `link` item. The link item is meant to link to other fragments. I think adding a new type called an "action" would be better.
You can copy the OptionsLink file to a new OptionsActions file and remove the fragment code from it and include the click listener.

app/src/main/java/org/jellyfin/androidtv/ui/startup/StartupToolbarFragment.kt
```diff
@@ -19,6 +19,11 @@ class StartupToolbarFragment : Fragment() {
 			if (activity is StartupActivity) activity.addServer()
 		}
 
+		binding.manageServers.setOnClickListener {
+			val activity = requireActivity()
+			if (activity is StartupActivity) activity.manageServers()
```
nielsvanvelzen, line 24:
```suggestion
			val settingsIntent = Intent(requireActivity(), PreferencesActivity::class.java)
			settingsIntent.putExtra(PreferencesActivity.EXTRA_SCREEN, ManageServersScreen::class.qualifiedName)
			startActivity(settingsIntent)
```

The fragment shouldn't depend on a specific activity (we should fix this later for the `addServer` function too)

app/src/main/res/values/strings.xml
```diff
@@ -433,4 +433,9 @@
     <string name="login_server_unavailable">Unable to connect to server</string>
     <string name="login_username_field_empty">Username cannot be empty</string>
     <string name="pref_authentication_link">Authentication options</string>
-</resources>
\ No newline at end of file
+    <string name="lbl_manage_servers">Manage Servers</string>
```
nielsvanvelzen, line 436:
```suggestion
    <string name="lbl_manage_servers">Manage servers</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -433,4 +433,9 @@
...
+    <string name="lbl_manage_users">Users for %1$s</string>
```
nielsvanvelzen, line 437:
```suggestion
    <string name="lbl_manage_users">%1$s users</string>
```

Maybe even just use the server name, not sure.

app/src/main/java/org/jellyfin/androidtv/ui/startup/ManageServersScreen.kt
```diff
@@ -0,0 +1,29 @@
...
+
+	override val screen by optionsScreen {
+		val authenticationStore = get<AuthenticationStore>()
+
+		val serverMap = authenticationStore.getServers()
+
+		setTitle(R.string.lbl_manage_servers)
+
+		for (entry in serverMap) {
+			link {
```
nielsvanvelzen, line 20:
```suggestion
		serverMap.forEach(entry -> link {
```

Also more readable to call it "server" instead of "entry".

app/src/main/java/org/jellyfin/androidtv/ui/startup/ManageServersScreen.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.jellyfin.androidtv.ui.preference.dsl.OptionsFragment
+import org.jellyfin.androidtv.ui.preference.dsl.optionsScreen
+import org.koin.android.ext.android.get
+
+class ManageServersScreen : OptionsFragment() {
+
+	override val screen by optionsScreen {
+		val authenticationStore = get<AuthenticationStore>()
+
+		val serverMap = authenticationStore.getServers()
```
nielsvanvelzen, line 15:
```suggestion
		val servers = authenticationStore.getServers()
```

app/src/main/java/org/jellyfin/androidtv/ui/startup/EditServerScreen.kt
```diff
@@ -0,0 +1,51 @@
...
+				withFragment<RemoveServerUserScreen>(bundleOf(
+						RemoveServerUserScreen.ARG_REMOVE_USER to true,
+						RemoveServerUserScreen.ARG_SERVER_UUID to serverUUID,
+						RemoveServerUserScreen.ARG_USER_UUID to entry.key
+				))
+			}
+		}
+
+		link {
+			setTitle(R.string.lbl_remove_server)
```
nielsvanvelzen, line 38: I'd use a trash icon here. Also this one could maybe use a description like "This also removes all users"

nielsvanvelzen (changes requested):

MrChip53:
> I don't think the confirmation dialog is necessary, we've removed these kinds of dialogs from other places in the app too. When a user presses the "remove user/server" button just remove it. The "manage servers" screen could use the server address as description, additionally an "add server" button could be added. The "users screen could use an "add user" button too, we could use the last active time as description for those users, it would also be cool if we could how profile pictures instead of the icons here. You could wrap the users in a `category {}` so they are separated from the delete server action.

I addressed everything I believe except the user images because at this time there is no way to load anything except a resource id and the add server button which I think may be big enough to justify it own PR, either by adding an edittext pref to the Options system or making the preferences be fragments on startup actvity so we can just push the addserver frag to the top when the button is clicked

app/src/main/java/org/jellyfin/androidtv/ui/preference/dsl/OptionsAction.kt
```diff
@@ -0,0 +1,73 @@
...
+import androidx.core.os.bundleOf
+import androidx.preference.Preference
+import androidx.preference.PreferenceCategory
+import java.util.*
+import kotlin.reflect.KClass
+
+/**
+ * Link to a different fragment.
+ */
+class OptionsAction(
```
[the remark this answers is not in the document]
nielsvanvelzen: also kdoc is not correct now

nielsvanvelzen (changes requested):
I'm wondering if we should add another menu when clicking on users to allow a few actions:

- Sign out
- Remove

We could add some more in the future like:
- Set default
- Set as service user

app/src/main/java/org/jellyfin/androidtv/ui/startup/EditUserScreen.kt
```diff
@@ -0,0 +1,66 @@
...
+		val authenticationStore = get<AuthenticationStore>()
+		val server = authenticationStore.getServer(serverUUID) ?: return@optionsScreen
+
+		val user = authenticationStore.getUser(serverUUID, userUUID) ?: return@optionsScreen
+		title = context?.getString(R.string.lbl_user_server, user.name, server.name)
+
+		val account = accountManagerHelper.getAccount(userUUID)
+
+		when {
+			account?.accessToken != null -> {
```
nielsvanvelzen, line 31: Should be disabled instead of hidden when access token is not set

nielsvanvelzen, line 31: Should use an icon/description

app/src/main/java/org/jellyfin/androidtv/ui/startup/EditUserScreen.kt
```diff
@@ -0,0 +1,66 @@
...
+						(pref.parent as PreferenceCategory).removePreference(pref)
+
+						return@setListener true
+					}
+				}
+			}
+		}
+
+		action {
+			setTitle(R.string.lbl_remove)
```
nielsvanvelzen, line 47: Should use an icon/description

app/src/main/java/org/jellyfin/androidtv/ui/startup/EditUserScreen.kt
```diff
@@ -0,0 +1,66 @@
...
+
+			setListener { _ ->
+				authenticationStore.removeUser(serverUUID, userUUID)
+
+				//Pop the back stack manually
+				//Do it twice because the EditServerScreen doesn't update to remove the listed user we just deleted
+				requireActivity().supportFragmentManager.fragments[0].childFragmentManager.popBackStackImmediate()
```
nielsvanvelzen, line 54: How can we be certain that the first element in `fragments` is the fragment we want to pop?

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/preference/dsl/OptionsCategory.kt
```diff
@@ -19,6 +19,16 @@ class OptionsCategory(
...
+	@OptionsDSL
+	fun link(init: OptionsLink.() -> Unit) {
+		this += OptionsLink(context).apply { init() }
+	}
+
+	@OptionsDSL
+	fun action(init: OptionsAction.() -> Unit) {
+		this += OptionsAction(context).apply { init() }
+	}
+
```
nielsvanvelzen, line 31: These should be extension functions in their own files (OptionsLink.kt,OptionsAction.kt). Adding _all_ functions in this file will make the code too long.

app/src/main/java/org/jellyfin/androidtv/ui/preference/dsl/OptionsAction.kt
```diff
@@ -0,0 +1,51 @@
...
+	private val context: Context
+) : OptionsItemMutable<Unit>() {
+	@DrawableRes
+	var icon: Int? = null
+	var content: String? = null
+	var clickListener: Preference.OnPreferenceClickListener? = null
+
+	fun setListener(preferenceClickListener: Preference.OnPreferenceClickListener) {
```
nielsvanvelzen, line 21: Add an type alias for `Preference.OnPreferenceClickListener` so we don't leak androidx.preference classes from the DSL. Not sure if it's possible to add the typealias in the class. Extending the listener might work too if it's not final.

See https://kotlinlang.org/docs/type-aliases.html

app/src/main/java/org/jellyfin/androidtv/ui/preference/PreferencesActivity.kt
```diff
@@ -20,13 +20,14 @@ class PreferencesActivity : FragmentActivity() {
...
 				)
-			})
+			}, MAIN_PREF_TAG)
 			.commit()
 	}
 
 	companion object {
 		const val EXTRA_SCREEN = "screen"
 		const val EXTRA_SCREEN_ARGS = "screen_args"
+		const val MAIN_PREF_TAG = "PreferencesActivity"
```
nielsvanvelzen, line 30:
```suggestion
		const val FRAGMENT_TAG = "PreferencesActivity"
```

app/src/main/java/org/jellyfin/androidtv/ui/preference/dsl/OptionsLink.kt
```diff
@@ -18,9 +20,11 @@ class OptionsLink(
 	var icon: Int? = null
 	var content: String? = null
 	var fragment: KClass<out OptionsFragment>? = null
+	var extras: Bundle = bundleOf()
```
nielsvanvelzen, line 23:
```suggestion
	var extras: Bundle? = null
```

app/src/main/java/org/jellyfin/androidtv/ui/preference/dsl/OptionsLink.kt
```diff
@@ -18,9 +20,11 @@ class OptionsLink(
...
 
-	inline fun <reified T : OptionsFragment> withFragment() {
+	inline fun <reified T : OptionsFragment> withFragment(extraBundle: Bundle = bundleOf()) {
```
nielsvanvelzen, line 25:
```suggestion
	inline fun <reified T : OptionsFragment> withFragment(extraBundle: Bundle? = null) {
```

app/src/main/java/org/jellyfin/androidtv/ui/preference/dsl/OptionsLink.kt
```diff
@@ -36,6 +40,7 @@ class OptionsLink(
 			it.isPersistent = false
 			it.key = UUID.randomUUID().toString()
 			it.fragment = fragment?.qualifiedName
+			it.extras.putAll(extras)
```
nielsvanvelzen, line 43:
```suggestion
			extras?.let { extras -> it.extras.putAll(extras) }
```

app/src/main/java/org/jellyfin/androidtv/ui/preference/dsl/OptionsLink.kt
```diff
@@ -49,8 +54,3 @@ class OptionsLink(
 		}
 	}
 }
-
-@OptionsDSL
-fun OptionsCategory.link(init: OptionsLink.() -> Unit) {
-	this += OptionsLink(context).apply { init() }
-}
```
nielsvanvelzen, line 56: Put this one back (see other comment in OptionsCategory.kt)

app/src/main/res/values/strings.xml
```diff
@@ -433,4 +433,16 @@
     <string name="login_server_unavailable">Unable to connect to server</string>
     <string name="login_username_field_empty">Username cannot be empty</string>
     <string name="pref_authentication_link">Authentication options</string>
-</resources>
\ No newline at end of file
+    <string name="lbl_manage_servers">Manage servers</string>
+    <string name="lbl_remove_users">This will also remove all associated users</string>
+    <string name="lbl_remove_user_content">Remove this account</string>
```
nielsvanvelzen, line 438:
```suggestion
    <string name="lbl_remove_user_content">Remove account</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -433,4 +433,16 @@
...
+    <string name="lbl_sign_out_content">Sign this user out</string>
```
nielsvanvelzen, line 439:
```suggestion
    <string name="lbl_sign_out_content">Sign this account out.</string>
```

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

MrChip53 (commented):
Your changes LGTM @nielsvanvelzen

nielsvanvelzen (approved):

thornbill (changes requested):
I tested this a bit and it looks great!

I did notice an issue though where I was unable to remove a server. I'm guessing this was the server that I last logged into. It seems to keep a reference to the last logged in user even after removing the server, so it doesn't actually go away.

nielsvanvelzen:
I think that behavior _should_ be fixed with #730

thornbill:
Whoops there's another conflict here now.

thornbill (approved):

PR #714, opened 2021-02-01
Add episode full name as subtitle while playing TV episode

MrChip53:
**Changes**
Adds the "Series Name S1 E1" text as the subtitle while playing a TV episode. Useful for a quick glimpse at what season and episode you are on without having to back out of your video.

**Issues**


MrChip53:
Untitled

nielsvanvelzen (approved):

PR #717, opened 2021-02-03
Disable headers on home screen

nielsvanvelzen:
Finally found out why it didn't work before. The StdBrowseFragment used onActivityCreated because it needs to tweak UI in the activity (lol). Changed it to use the onCreate function for most code (except clock initialization) and now we can disable the headers perfectly fine. I decided to not add a preference to enable/disable the headers because it can't show more than 6 anyway.

**Changes**
- Disable headers on all screens
- Enabling them is not possible when extending StdBrowseFragment

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdBrowseFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdBrowseFragment.java
@@ -1,17 +1,3 @@
-/*
- * Copyright (C) 2014 The Android Open Source Project
- *
- * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
- * in compliance with the License. You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software distributed under the License
- * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
- * or implied. See the License for the specific language governing permissions and limitations under
- * the License.
- */
-
 package org.jellyfin.androidtv.ui.browsing;
 
 import android.content.Intent;
@@ -26,6 +12,7 @@ import android.widget.ImageView;
 import android.widget.TextClock;
 import android.widget.TextView;
 
+import androidx.annotation.Nullable;
 import androidx.core.content.ContextCompat;
 import androidx.leanback.app.BrowseSupportFragment;
 import androidx.leanback.widget.ArrayObjectAdapter;
@@ -87,8 +74,8 @@ public class StdBrowseFragment extends BrowseSupportFragment implements IRowLoad
     private Lazy<BackgroundService> backgroundService = inject(BackgroundService.class);
 
     @Override
-    public void onActivityCreated(Bundle savedInstanceState) {
-        super.onActivityCreated(savedInstanceState);
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
 
         if (getActivity() instanceof BaseActivity) mActivity = (BaseActivity)getActivity();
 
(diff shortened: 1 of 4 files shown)


MrChip53:
Oh my. This looks so good. I think we should remove headers from every screen. Your opinion?

MrChip53 (approved):

nielsvanvelzen:
The other screens using headers are genres/byletter/someother screens which use the headers to quickly navigate to a section. I think they could stay there for now.

MrChip53:
I don't think there is much of a true speed difference between navigating with headers or directly on the rows though. Either method is still just one click up or down.

nielsvanvelzen:
Yeah you do have a point. @thornbill what do you think? Should I remove the headers everywhere?

MrChip53:
I also think it gives a bit of that "streamlined netflix" look people like even though it's not the true netflix experience. Maybe a reddit poll could be done on whether to remove completely or make it into a toggle able preference app wide. I definitely think removing the headers makes the app look a lot more polished though.

thornbill:
I think removing them everywhere is fine. Really the "faster navigation" argument only holds up if a section has multiple rows which afaik none of ours do.

nielsvanvelzen:
Updates the PR to remove headers everywhere, it's not possible anymore to enable them when using the StdBrowseFragment now.

Rebasing now to fix conflicts

MrChip53 (approved):

app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdBrowseFragment.java
```diff
@@ -1,17 +1,3 @@
-/*
- * Copyright (C) 2014 The Android Open Source Project
```
thornbill, line 2: I believe this license header needs to be kept. I believe that since the project as a whole is GPL v2 we cannot relicense this file from Apache v2.

thornbill (changes requested):
Looks great! Only question I have is regarding licensing.

app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdBrowseFragment.java
nielsvanvelzen: I wasn't sure about removing it either but I'd argue that pretty much none of the code in this file is from the Android OSS project now (except for the function names we inherit from the base class).

nielsvanvelzen: This is the iniial version of the file:
https://github.com/jellyfin/jellyfin-androidtv/blob/96831e66be089b4fe2cd87cd6ed83f24a599caa9/MediaBrowserTv/app/src/main/java/tv/emby/embyatv/browsing/StdGridFragment.java

And I'm quite sure it is based on this file:
https://github.com/googlearchive/leanback-showcase/blob/master/app/src/main/java/android/support/v17/leanback/supportleanbackshowcase/app/grid/GridExampleFragment.java

thornbill: Given that the effective license of this project is GPLv3 due to the inclusion of licensed code that is incompatible with GPLv2 prior to our fork, and since the GPLv3 is compatible with the Apache 2 license in that Apache 2 code can be relicensed as GPLv3, I believe this removal is fine.

dankell66:
Hello Support Team,
Why on gods green earth would you have removed the side bar feature in v0.12.0?
Very bad move indeed ...removing this feature also removes MANY further features which include Genre search to name a few.
Why would you do such a BAD move....when the majority welcomed the side bar.
Please could you make sure to reencorporate this feature back ASAP Please.
I am sure you can add a setting for the minority whom do not like this feature.
You have broken a perfectly Great app for a very small minority .

Best Regards,
Dankell

PR #718, opened 2021-02-03
Rewrite TextButton in Kotlin and center content.

nielsvanvelzen:
I don't really use this feature but when I looked at the text alignment I wanted to fix it.

**Changes**
Rewrite of TextButton in Kotlin and fixes the centering of the text. The button is only used in the JumpList that helps navigating in the browsing view.

**Before**
image

**After**
image

**Issues**


MrChip53 (approved):

app/src/main/java/org/jellyfin/androidtv/ui/TextButton.kt
```diff
@@ -0,0 +1,39 @@
...
+	listener: OnClickListener?
+) : AppCompatButton(context) {
+	init {
+		// Set box size
+		layoutParams = Utils.convertDpToPixel(context, size).let { pixels ->
+			ViewGroup.LayoutParams(pixels * 2 + 15, pixels + 40)
+		}
+
+		setBackgroundColor(0)
+		textSize = size
```
thornbill, line 28: It looks like the old code had a check that would set `size` to 16 if it was 0. Maybe we should continue to default that?

thornbill (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/TextButton.kt
nielsvanvelzen: I looked at the call sides and it never passes `0` so I removed that check.

thornbill (approved):

PR #721, opened 2021-02-03
Remove audio behavior warning

nielsvanvelzen:
**Changes**
Removed the warning to enable audio downmixing in the home screen. It only appeared once for new installs. In the future we should auto-detect capabilities (with the playback rewrite) so users don't need to do anything, for now going to the preferences and changing the behavior enum is the way to go.

**Issues**


nielsvanvelzen:
I kept the strings because I don't want to break weblate

MrChip53 (approved):

thornbill:
You should be able to remove the English strings and the weblate should clean up the translations.

nielsvanvelzen:
I've removed the strings that became unused from my changes, using the "unused resources" inspection in Android Studio tells me we have a lot more unused resources, we should clean those before releasing 0.12.

MrChip53 (approved):

thornbill (approved):

PR #725, opened 2021-02-07
Fix next up bug

MrChip53:
**Changes**
Fixes the next up bug, also a NPE with the ends at text

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -122,6 +122,14 @@ public class PlaybackController {
         useVlc = userPreferences.getValue().get(UserPreferences.Companion.getVideoPlayer()) == PreferredVideoPlayer.VLC;
     }
 
+    public void clearFragment() {
+        mFragment = null;
+    }
+
+    public boolean hasFragment() {
+        return mFragment != null ? true : false;
+    }
+
     public void init(VideoManager mgr) {
         mVideoManager = mgr;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -122,6 +122,14 @@ public PlaybackController(List<BaseItemDto> items, IPlaybackOverlayFragment frag
         useVlc = userPreferences.getValue().get(UserPreferences.Companion.getVideoPlayer()) == PreferredVideoPlayer.VLC;
     }
 
+    public void clearFragment() {
+        mFragment = null;
```
nielsvanvelzen, line 126: Is this meant to "destroy" the fragment? Because if that's the case you need to destroy it via the fragmentmanager, otherwise the fragment will still be bound to the activity and cause memory issues.

nielsvanvelzen (approved):
Next up still works the same for me so I guess it's fine!

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
MrChip53: Kind of. I mean it was supposed to but it didn't so I added the `hasFragment` function so we only run the `onDestroy` code if we haven't cleared the fragment. The fragment is still getting destroyed though beacause it's onDestroy function is still getting called. It's just getting called after we want it to. I was going to look for a better way to do all this but then figured we will rewrite it eventually AND this guarantees things are done when we want them to be.

PR #728, opened 2021-02-08
Refresh servers so restart isn't required

MrChip53:
**Changes**
Manually refresh servers when ListServerFragment is focused

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/shared/IFocusListener.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/shared/IFocusListener.kt
@@ -1,0 +1,5 @@
+package org.jellyfin.androidtv.ui.shared
+
+interface IFocusListener {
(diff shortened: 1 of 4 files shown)


nielsvanvelzen (changes requested):
I understand your thinking but this is _not_ how it should work when using LiveData. What we should do is always mutate the server storage via the repository, and when a mutation happens update the livedata automatically. Any existing observers will then trigger and update the UI automatically.

The reason it doesn't work like that right now is because I made some mistakes. I'm already working on some UI changes (the ones I shared on Matrix) and with those some modifications need to happen in the repository that actually makes it easier to work with the LiveData as intended, since we don't need to merge multiple flows into one.

I'd recommend reading a bit about Google's recommended Android architecture to understand it better as I'm not too good in explaining it :p

- https://developer.android.com/jetpack/guide
- https://developer.android.com/topic/libraries/architecture/livedata
- https://developer.android.com/topic/libraries/architecture/

PR #737, opened 2021-02-23
Add merge conflict labeler and gradle update and validate tasks

nielsvanvelzen:
**Changes**
- Added new task: gradle update
  automatically updates the wrapper for us 🎉 
- Changed task: gradle wrapper validate
  only run when the wrapper changes, consistent with android/apiclient repositories now
- Changed task: conflict labeler
  added schedule to run each night (for me at least, thanks timezones)
  now consistent with android/apiclient repositories

**Issues**


.github/workflows/gradlew-update.yaml
```diff
@@ -0,0 +1,17 @@
+name: Gradle update
+
+on:
+  schedule:
+    - cron: '0 4 * * *'
+
+jobs:
+  update:
+    runs-on: ubuntu-20.04
```
thornbill, line 9: imo using a specific version here instead of `ubuntu-latest` will require more maintenance in the future than dealing with breakages when they occur. I think whichever direction we go on this should be uniform across the org. Currently we seem to be using `ubuntu-latest` everywhere, but we should have a discussion about changing if you feel it's worthwhile.

.github/workflows/merge-conflict-labeler.yml
```diff
@@ -1,15 +1,16 @@
-name: "Merge Conflicts"
+name: Merge conflicts labeler
```
thornbill, line 1: "labeler" seems odd here perhaps just "label" would be better. It doesn't really match the format of the other action names of "{thing} {verb}". :man_shrugging:

thornbill (changes requested):

.github/workflows/gradlew-update.yaml
nielsvanvelzen: I believe using a specific version is the right way but I agree that we should be consistent. Changed it for now.

thornbill (approved):

PR #741, opened 2021-02-26
Add the ability to hide watched/unwatched indicators on cards

linetrimmer:
**Changes**

Adds the option 'Show watched indicators' to Settings with the following choices:
- 'Always': behaves like it does now
- 'Hide unwatched count': hides the unwatched episode count from cards, keeps the watched check mark
- 'On episodes only': hides the unwatched episode count from cards, hides the watched check mark unless the item is an episode
- 'Never': hides the unwatched episode count and the watched check mark on all items


thornbill:
There are conflicts in the strings file now. :smile:

linetrimmer:
Should be good now!

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
```diff
@@ -191,7 +192,18 @@ public void setItem(BaseRowItem m, String imageType, int lHeight, int pHeight, i
...
+                                if (itemDto.getBaseItemType() == BaseItemType.Episode) {
+                                    mCardView.setUnwatchedCount(0);
+                                }
+                                else {
+                                    mCardView.setUnwatchedCount(-1);
+                                }
+                            }
+                            else {
+                                mCardView.setUnwatchedCount(0);
+                            }
```
nielsvanvelzen, line 206:
```suggestion
                            if (showIndicator == WatchedIndicatorBehavior.EPISODES_ONLY && itemDto.getBaseItemType() != BaseItemType.Episode) {
                                    mCardView.setUnwatchedCount(-1);
                             } else {
                                    mCardView.setUnwatchedCount(0);
                             }
```

Could be simplified to something like this

app/src/main/java/org/jellyfin/androidtv/ui/presentation/MyImageCardView.java
```diff
@@ -230,15 +234,26 @@ public void clearBanner() {
     }
 
     public void setUnwatchedCount(int count) {
+        WatchedIndicatorBehavior showIndicator = get(UserPreferences.class).get(UserPreferences.Companion.getWatchedIndicatorBehavior());
```
nielsvanvelzen, line 237: I'd rather keep this class "dumb" and not lookup settings in here, would be better to add that logic to the calling side.

nielsvanvelzen (changes requested):

app/src/main/res/values/strings.xml
```diff
@@ -461,4 +461,7 @@
     <string name="login_help_title">Need help?</string>
     <string name="login_help_description">Jellyfin requires a server to connect with. Visit our documentation at docs.jellyfin.org to get started with Jellyfin.</string>
     <string name="connect_title">Connect</string>
-</resources>
+    <string name="pref_watched_indicator">Show watched indicators</string>
+    <string name="lbl_hide_unwatched_count">Hide unwatched count</string>
+    <string name="lbl_episode_marks">On episodes only</string>
+</resources>
```
thornbill, line 467: You lost the trailing newline here.

thornbill (approved):

PR #705, opened 2021-01-29
Improve search activity

MrChip53:
**Changes**
- [x] Use CustomListRowPresenter in SearchProvider.java
- [x] Fix background in search activity
- [ ] ~~Add rating to search cards?~~ no, no and no

Requires #703 

**Issues**

#702

PR #742, opened 2021-02-27
Use FragmentActivity in SearchActivity

nielsvanvelzen:
**Changes**
- Rewrite the file in Kotlin
- Use FragmentActivity as base instead of the deprecated (custom) BaseActivity

Tested with #705 and works fine!

**Issues**


thornbill (approved):
:+1: lgtm

PR #743, opened 2021-02-27
Remove more code from BaseActivity

nielsvanvelzen:
BaseActivity bad

**Changes**
- Migrate more activities to use FragmentActivity instead of BaseActivity
- Remove a bunch of code from BaseActivity
  - Moved the "currentActivity" trash to a callback, still deprecated and should not be used though
- Removed the "onSearchRequested" function
  - Not sure if it was even used anywhere, replaced with a search intent filter for now
- Fixed a few small issues with refreshing rows after I changed all of the above

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdBrowseFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdBrowseFragment.java
@@ -126,7 +126,7 @@ public class StdBrowseFragment extends BrowseSupportFragment implements IRowLoad
         new Handler().postDelayed(new Runnable() {
             @Override
             public void run() {
-                if (mActivity != null && mActivity.isFinishing()) return;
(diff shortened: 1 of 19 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdBrowseFragment.java
```diff
@@ -126,10 +126,10 @@ protected void refreshRows() {
         new Handler().postDelayed(new Runnable() {
             @Override
             public void run() {
-                if (mActivity.isFinishing()) return;
+                if (mActivity != null && mActivity.isFinishing()) return;
```
thornbill, line 129: This should probably be `mActivity == null || mActivity.isFinishing()` to avoid NPEs.

thornbill (changes requested):

thornbill (approved):

PR #749, opened 2021-03-01
Leanback channel changes

linetrimmer:
**Changes**
- Add Next Up channel
- Use the series thumbnail when preferred by the user
- Fix for every episode showing the "Resume watching" message in the Play Next channel

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
@@ -27,6 +27,7 @@ import org.jellyfin.apiclient.model.dto.BaseItemDto
 import org.jellyfin.apiclient.model.dto.ImageOptions
 import org.jellyfin.apiclient.model.entities.ImageType
 import org.jellyfin.apiclient.model.querying.ItemFields
+import org.jellyfin.apiclient.model.querying.ItemsResult
 import org.jellyfin.apiclient.model.querying.NextUpQuery
 import org.koin.java.KoinJavaComponent.inject
 
@@ -50,17 +51,13 @@ class LeanbackChannelWorker(
 		const val PERIODIC_UPDATE_REQUEST_NAME = "LeanbackChannelPeriodicUpdateRequest"
 	}
 
+	private val userPreferences by inject(UserPreferences::class.java)
+
 	/**
 	 * Check if the app can use Leanback features and is API level 26 or higher
 	 */
 	private val isSupported = Build.VERSION.SDK_INT >= Build.VERSION_CODES.O
 		&& context.packageManager.hasSystemFeature("android.software.leanback")
-
-	/**
-	 * Check if "Prefer series thumbnails" is enabled for the user
-	 */
-	private val userPreferences by inject(UserPreferences::class.java)
-	private val preferParentThumb = userPreferences[UserPreferences.seriesThumbnailsEnabled]
 
 	/**
 	 * Update all channels for the currently authenticated user
@@ -71,13 +68,16 @@ class LeanbackChannelWorker(
 		// Retry later if no authenticated user is found
 		TvApp.getApplication().currentUser == null -> Result.retry()
 		else -> {
-			// Delete current items
+			// Get next up episodes
+			val nextUpItems = getNextUpItems()
+
+			// Delete current items from the My Media and Next Up channels
 			context.contentResolver.delete(TvContractCompat.PreviewPrograms.CONTENT_URI, null, null)
 
 			// Update various channels
 			updateMyMedia()
-			updateNextUp()
-			updateWatchNext()
+			updateNextUp(nextUpItems)
+			updateWatchNext(nextUpItems)
 
 			// Success!
 			Result.success()
@@ -149,9 +149,48 @@ class LeanbackChannelWorker(
 	}
 
 	/**
-	 * Updates the "next up" row with current episodes
+	 * Gets the poster art for an [item]
+	 * Uses the [preferParentThumb] parameter to fetch the series image when preferred
 	 */
-	private suspend fun updateNextUp() {
+	private fun getPosterArtImageUrl(preferParentThumb: Boolean, item: BaseItemDto): Uri? {
+		return if (preferParentThumb && item.parentThumbItemId != null) {
+			Uri.parse(apiClient.GetImageUrl(item.seriesId, ImageOptions().apply {
+				format = ImageFormat.Png
+				height = 288
+				width = 512
+				imageType = ImageType.Thumb
+			}))
+		} else {
+			Uri.parse(apiClient.GetImageUrl(item, ImageOptions().apply {
+				format = ImageFormat.Png
+				height = 288
+				width = 512
+			}))
+		}
+	}
+
+	/**
+	 * Gets the next up episodes or returns null
+	 */
+	private suspend fun getNextUpItems(): ItemsResult? = withContext(Dispatchers.Default) {
+		// Get user or return if no user is found (not authenticated)
+		val user = TvApp.getApplication().currentUser ?: return@withContext null
+
+		return@withContext apiClient.getNextUpEpisodes(NextUpQuery().apply {
+			userId = user?.id
+			imageTypeLimit = 1
+			limit = 10
+			fields = arrayOf(ItemFields.DateCreated)
+		})
+	}
+
+	/**
+	 * Updates the "next up" row with current episodes
+	 * Uses the [nextUpItems] parameter to store items returned by a NextUpQuery()
+	 */
+	private suspend fun updateNextUp(nextUpItems: ItemsResult?) = withContext(Dispatchers.Default) {
+		val preferParentThumb = userPreferences[UserPreferences.seriesThumbnailsEnabled]
+
 		// Get channel
 		val channelUri = getChannelUri("next_up", Channel.Builder()
 			.setType(TvContractCompat.Channels.TYPE_PREVIEW)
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -154,14 +153,14 @@ class LeanbackChannelWorker(
 			userId = user.id
 			imageTypeLimit = 1
 			limit = 10
-			fields = arrayOf(ItemFields.DateCreated)
```
nielsvanvelzen, line 157: Field is used on line 197

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -154,14 +153,14 @@ class LeanbackChannelWorker(
...
 		})
 
 		// Add new items
-		response?.items?.let { items ->
-			context.contentResolver.bulkInsert(
-				WatchNextPrograms.CONTENT_URI,
-				items.map { item -> getBaseItemAsWatchNextProgram(item).toContentValues() }.toTypedArray()
+		response?.items?.reverse()
```
nielsvanvelzen, line 159: We should specify the sorting in the `getNextUpEpisodes()` call. The sorting is already correct for me though.

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -178,11 +177,21 @@ class LeanbackChannelWorker(
...
+				imageType = org.jellyfin.apiclient.model.entities.ImageType.Thumb
+			})))
+		}
+		else {
+			setPosterArtUri(Uri.parse(apiClient.GetImageUrl(item, ImageOptions().apply {
+				format = ImageFormat.Png
+				height = 288
+				width = 512
+			})))
+		}
```
nielsvanvelzen, line 194: The "next up" row displays singular entries, not shows. It should use the thumbnail instead of poster.

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -192,10 +201,9 @@ class LeanbackChannelWorker(
 			item.canResume -> {
 				setWatchNextType(WatchNextPrograms.WATCH_NEXT_TYPE_CONTINUE)
 				setLastPlaybackPositionMillis((item.resumePositionTicks / TICKS_IN_MILLISECOND).toInt())
-			}
-			// Episode runtime has been determined
-			item.runTimeTicks != null -> {
-				setDurationMillis((item.runTimeTicks / TICKS_IN_MILLISECOND).toInt())
+				if (item.runTimeTicks != null) {
+					setDurationMillis((item.runTimeTicks / TICKS_IN_MILLISECOND).toInt())
+				}
```
nielsvanvelzen, line 206: I'm not 100% sure if it's used for the NEW and NEXT types but maybe this should be moved outside of the `when {}` statement,

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
linetrimmer: For me the ten episodes that load into the Play Next channel are the same as the first ten episodes in my Next Up row, but instead of sorting by most recently watched like inside the app, the channel sorts by the episode’s Date Added value. e.g. my first entry in the Next Up row inside the app is the eighth entry in the Play Next channel

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
linetrimmer: Right now the Next Up row inside the app displays the thumbnail for the series if there is one, and the thumbnail for the episode if there isn’t. This would be to match that behavior

nielsvanvelzen: I agree that we should make this behavior consistent but it's something that needs to change in the home screen, not the next up row.

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
nielsvanvelzen: I just checked this and you're correct. I think this can be fixed by changing the sort order in that case.

linetrimmer: It looks like the Play Next channel order is being determined by `setLastEngagementTimeUtcMillis(item.dateCreated?.time ?: System.currentTimeMillis())`, which is why removing ItemFields.DateCreated from the NextUpQuery caused the order to change (it made setLastEngagement fall back to System.currentTime). 

I tried `setLastEngagementTimeUtcMillis(apiClient.getItem(item.seriesId)?.userData?.lastPlayedDate?.time ?: System.currentTimeMillis())` but it just falls back to System.currentTime. Do you have any ideas?

nielsvanvelzen:
Hey @linetrimmer is this PR ready now? It's still marked as draft.

linetrimmer:
The thumbnails and message reporting (‘Resume Watching/Play Next’) are good, but I’m not sure how to handle the episode order. The order is determined by `setLastEngagementTime`. 

The two ways I’ve tested it:
1) Reversing the order of the items returned by the NextUpQuery, not using bulkInsert, and using only the current system time for setLastEngagementTime
- Pro: the episodes are displayed in the correct order
- Con: if other apps are putting items in the channel, the 10 Jellyfin items will jump to the front of the channel each time the app is opened since the engagement time is the current system time

2) Keeping it the way it is now (last engagement time is set to Date Created or system time as a fallback)
- Pro: the Jellyfin items will be mixed in with items from other apps
- Con: the episode order will be wrong since it’s going by the date added
- Con: depending on how long ago the episode was added to the library and how many items other apps are adding to the channel, the Jellyfin items may be pushed down to the end of the channel since the date added could be weeks/months earlier

I’m not sure which way would be better. Ideally, setLastEngagementTime would use the last played time for the most recently completed episode, but I haven’t been able to figure out how to access that 😓

nielsvanvelzen:
After thinking a bit about it I think we should not try to change the order of the items. Just give the information as-is to leanback and have the launcher sort it itself. This will give a consistent behavior across multiple apps.

We could however add a custom row specifically for the Jellyfin "next up" section later.

linetrimmer:
Is there something we could use for setLastEngagementTime other than dateCreated? Items scanned into my library since updating to 10.7 aren’t appearing in the play next channel, but they do if dateCreated isn’t used for the last engagement time.

linetrimmer:
I added a Next Up channel and marked the PR as ready for review. The dateCreated field is causing items scanned in after updating to 10.7 to not appear in Play Next, so I tested using premiereDate instead, but that also caused the items to be missing from the channel. I'm not sure if it's just my server being weird about those episodes, though.

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -138,6 +144,75 @@ class LeanbackChannelWorker(
...
+		val preferParentThumb = get(UserPreferences::class.java)[UserPreferences.seriesThumbnailsEnabled]
+
+		// Add new items
+		response?.items?.map { item ->
+			val imageUri = if (preferParentThumb && item.parentThumbItemId != null) {
+				Uri.parse(apiClient.GetImageUrl(item.seriesId, ImageOptions().apply {
+					format = ImageFormat.Png
+					height = 288
+					width = 512
+					imageType = org.jellyfin.apiclient.model.entities.ImageType.Thumb
```
nielsvanvelzen, line 176:
```suggestion
					imageType = ImageType.Thumb
```

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -138,6 +144,75 @@ class LeanbackChannelWorker(
...
+					height = 288
+					width = 512
+				}))
+			}
+
+			val seasonString = if (item.parentIndexNumber != null) {
+				"${item.parentIndexNumber}"
+			} else {
+				""
+			}
```
nielsvanvelzen, line 190:
```suggestion
			val seasonString = item.parentIndexNumber?.toString().orEmpty()

File: app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt [hunk 7]
```diff
@@ -138,6 +144,75 @@ class LeanbackChannelWorker(
...
+
+			val episodeString = if (item.indexNumber != null && item.indexNumberEnd != null) {
+				"${item.indexNumber}-${item.indexNumberEnd}"
+			} else if (item.indexNumber != null) {
+				"${item.indexNumber}"
+			} else {
+				""
+			}
```
nielsvanvelzen, line 198: I think a `when` statement will look cleaner here

nielsvanvelzen (changes requested):
Initial code review, might have some additional suggestions in the next

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -171,18 +246,31 @@ class LeanbackChannelWorker(
 	 *
 	 * Assumes the item type is "episode"
 	 */
-	private fun getBaseItemAsWatchNextProgram(item: BaseItemDto) = WatchNextProgram.Builder().apply {
+	private suspend fun getBaseItemAsWatchNextProgram(item: BaseItemDto) = WatchNextProgram.Builder().apply {
+		val preferParentThumb = get(UserPreferences::class.java)[UserPreferences.seriesThumbnailsEnabled]
```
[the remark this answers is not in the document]
nielsvanvelzen: I wasn't clear enough, I only meant the userPreferences should be a class property. The actual retrieval of the preference can be moved inline. (line 63 right now)

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -138,6 +148,69 @@ class LeanbackChannelWorker(
...
+
+			val seasonString = item.parentIndexNumber?.toString().orEmpty()
+
+			val episodeString = when {
+				item.indexNumberEnd != null && item.indexNumber != null -> {
+					"${item.indexNumber}-${item.indexNumberEnd}"
+				} else -> {
+					item.indexNumber?.toString().orEmpty()
+				}
+			}
```
nielsvanvelzen, line 196:
```suggestion
			val episodeString = when {
				item.indexNumberEnd != null && item.indexNumber != null -> 
 "${item.indexNumber}-${item.indexNumberEnd}"
				else -> item.indexNumber?.toString().orEmpty()
			}
```

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -138,6 +148,69 @@ class LeanbackChannelWorker(
...
+					width = 512
+					imageType = ImageType.Thumb
+				}))
+			} else {
+				Uri.parse(apiClient.GetImageUrl(item, ImageOptions().apply {
+					format = ImageFormat.Png
+					height = 288
+					width = 512
+				}))
+			}
```
nielsvanvelzen, line 186: Might be worth moving this to an extension function in this class since the same code is used in `getBaseItemAsWatchNextProgram`.

Something like `private fun BaseItemDto.getPosterArtImageUrl(preferParentThumb: Boolean) { ... }`

nielsvanvelzen (approved):

PR #759, opened 2021-03-08
Fix EnhancedBrowseFragment crash

nielsvanvelzen:
The fragment uses the messaging trash in BaseActivity.

**Changes**
- Add check in EnhancedBrowseFragment if the parent activity is a BaseActivity
- Use BaseActivity for all activities (indirectly) using the EnhancedBrowseFragment

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/EnhancedBrowseFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/EnhancedBrowseFragment.java
@@ -14,6 +14,7 @@ import android.widget.Toast;
 
 import androidx.annotation.Nullable;
 import androidx.fragment.app.Fragment;
+import androidx.fragment.app.FragmentActivity;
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/browsing/EnhancedBrowseFragment.java
```diff
@@ -65,7 +66,7 @@
 import static org.koin.java.KoinJavaComponent.inject;
 
 public class EnhancedBrowseFragment extends Fragment implements IRowLoader {
-    BaseActivity mActivity;
+    FragmentActivity mActivity;
```
thornbill, line 69: Do you want to add a scope here to appease Codacy?

nielsvanvelzen: Added it to all properties

thornbill (approved):

PR #761, opened 2021-03-10
Fix semitransparent row headers

thornbill:
**Changes**
Removes the 50% transparency from row headers

**Issues**
N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/presentation/CustomRowHeaderPresenter.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/presentation/CustomRowHeaderPresenter.kt
@@ -3,7 +3,7 @@ package org.jellyfin.androidtv.ui.presentation
 import androidx.leanback.widget.RowHeaderPresenter
 
 class CustomRowHeaderPresenter : RowHeaderPresenter() {
-	override fun onSelectLevelChanged(holder: ViewHolder?) {
+	override fun onSelectLevelChanged(holder: ViewHolder) {
 		// Do nothing - this keeps headers opaque
 	}
 }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/presentation/CustomRowHeaderPresenter.kt
```diff
@@ -0,0 +1,9 @@
+package org.jellyfin.androidtv.ui.presentation
+
+import androidx.leanback.widget.RowHeaderPresenter
+
+class CustomRowHeaderPresenter : RowHeaderPresenter() {
+	override fun onSelectLevelChanged(holder: ViewHolder?) {
```
nielsvanvelzen, line 6:
```suggestion
	override fun onSelectLevelChanged(holder: ViewHolder) {
```

I'm quite sure the parameter is not-null. Documentation doesn't explicitly state it though.
https://developer.android.com/reference/androidx/leanback/widget/RowHeaderPresenter#onSelectLevelChanged(androidx.leanback.widget.RowHeaderPresenter.ViewHolder)

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CustomRowHeaderPresenter.kt
thornbill: 👍 okay. I just went with whatever autocomplete added lol.

nielsvanvelzen: Autocomplete almost always makes it nullable so it's good to double check

nielsvanvelzen (approved):

PR #767, opened 2021-03-12
Add new autologin preferences

nielsvanvelzen:
**Changes**

- Merge the current "manage servers" screen and the "authentication category" to a single screen
  - Can be opened from the both the settings screen and startup activity
- Remove old authentication preferences
  - Includes a migration that wouldn't work now 
  - Includes their preferences UI
- Added new authentication preferences, in a separate class "AuthenticationPreferences"
  - Consists of two pairs (behavior+userid) for auto-login and service-user
- Created new dsl to easily add user-pickers
- Used new user picker dsl to implement new authentication preferences ui

This PR **does not** implement the actual auto-signin. It only provides the new preferences.

**TODO**

There are two things I'd like to get feedback on before this PR is "ready":

- ~~Sorting - Right now the servers/users shown in the UserPicker are not sorted. This also doesn't happen in the server management preferences. I'm not sure if we should sort both by last use (just like the  startup activity)~~
- Strings - I think some of the text could be improved

**Issues**

Part of #612


nielsvanvelzen:
> For strings, I don't understand what a Service User is for

This is how the preferences look right now, does it explain the service user enough for you?

image image

PrplHaz4:
> 
> 
> > For strings, I don't understand what a Service User is for
> 
> This is how the preferences look right now, does it explain the service user enough for you?
> 

It doesn't but I was actually going to guess it should be the most recent user.  Sorry for the string of questions, don't feel obligated to answer, I'll use the app either way :)

Why wouldn't the logged in user be used for system integrations?  Does the app need to be able to search the Jellyfin library before someone logs in?  Is the pick list populated by users that have previously logged in?  Once someone is logged in, shouldn't all operations be performed as that user?

nielsvanvelzen:
The system integrations are active when the app isn't. For example; when you search in Android TV or when the "next up" row in leanback updates. This is what the service user will be used for. A sane default is to always use the last active user for that but I can imagine some users not wanting that so they could entirely disable it or select a specific user.

fsaris:
> A sane default is to always use the last active user for that 

I would vote for this behavior. Similar to the Netflix app

nielsvanvelzen:
I've added sorting to the AuthenticationRepository and made sure all UI that shows users/servers now uses the correct sorting.

PrplHaz4:
> 
> 
> The system integrations are active when the app isn't. For example; when you search in Android TV or when the "next up" row in leanback updates. This is what the service user will be used for. A sane default is to always use the last active user for that but I can imagine some users not wanting that so they could entirely disable it or select a specific user.

Maybe something like this would be more clear:

**Service User**
Selected user will be used for AndroidTV platform integrations like search or Next Up

thornbill (commented):
Looks good overall!

String suggestions:
* Automatically sign in => Automatic sign in
* Feature will be disabled => Always show the sign in screen
* Uses the user that was last active in the app => Sign in as the last active user
* Service user => System user
* Uses the user that was last active in the app => Use the last active user

We should also standardize on either "login" or "sign in" right now we use "manual login" and "logout" in other places in the app. Web also mixes the two. :man_shrugging: 

I would avoid any specific mentions of "AndroidTV" since it may be confusing for Fire TV users.

thornbill (approved):

PR #768, opened 2021-03-12
Disable folders views because types are missing in the apiclient

thornbill:
**Changes**
The necessary types for folders are missing in `BaseItemType` in the current apiclient. This causes the app ui to break due to an exception, so best to disable for now.

**Issues**
N/A


nielsvanvelzen (approved):
I don't think this breaks anything

xxxsskxxx:
Any updates on folder view coming back for Android TV?  It works fine on my Andoir Mobile App and the web, but really would like it for Android TV on my Nvidia Shield. Thanks.

PR #770, opened 2021-03-13
Remove a bunch of static

nielsvanvelzen:
This is what happens when I'm waiting for other PR's to be merged

**Changes**

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
@@ -9,6 +9,7 @@ import android.content.Intent;
 import android.media.AudioManager;
 import android.net.Uri;
 import android.os.Handler;
+import android.os.Looper;
 import android.text.InputType;
 import android.widget.EditText;
 import android.widget.Toast;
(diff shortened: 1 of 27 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
```diff
@@ -305,40 +306,40 @@ public void onAudioFocusChange(int focusChange) {
...
         mCurrentAudioQueue = new ItemRowAdapter(items, new CardPresenter(true, Utils.convertDpToPixel(TvApp.getApplication(), 140)), null, QueryType.StaticAudioQueueItems);
         mCurrentAudioQueue.Retrieve();
         mManagedAudioQueue = null;
         fireQueueStatusChange();
     }
 
-    private static int TYPE_AUDIO = 0;
-    private static int TYPE_VIDEO = 1;
+    private int TYPE_AUDIO = 0;
+    private int TYPE_VIDEO = 1;
```
thornbill, line 332: These should probably be `static` and `final`.

thornbill (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
nielsvanvelzen: Good catch

thornbill (approved):

PR #771, opened 2021-03-13
Fix ItemLauncher crashing due to unsafe casting

nielsvanvelzen:
**Changes**

When activating an item in the "now playing" section on the home screen the app crashed because the menu did some unsafe casting.

**Issues**


MrChip53 (approved):

thornbill (approved):
I wasn't able to replicate the crash in an emulator, but the changes look good.

PR #775, opened 2021-03-14
Allow music videos to be searchable

MrChip53:
**Changes**
Again, maybe. Server won't pull in a test music video so can't test.

**Issues**


MrChip53:
Doesnt work

PR #783, opened 2021-03-20
Show stored users instantly and update row when public users are retrieved

nielsvanvelzen:
**Changes**
- Emit users in 2 steps from the ServerRepository; once for stored users and once for stored+public users

Stored users now show up instantly on the server fragment, while previously they only showed up after public users were retrieved (which could take a few seconds depending on the network quality).

It does not cause any flickering or focus changes since the MutableObjectAdapter is used.

**Issues**


thornbill (changes requested):
I found an interesting bug with these changes. If you select manually enter a use and then cancel, the server rows become duplicated.

thornbill (approved):

PR #785, opened 2021-03-22
Fix BackgroundService fragment crash and default background not restoring

nielsvanvelzen:
**Changes**

- The default background didn't restore properly because a bitmap of the backgroundservice background was saved (whoops)
- The fragment could crash because it didn't have an empty constructor (as is required for fragments)

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
@@ -1,7 +1,6 @@
 package org.jellyfin.androidtv.data.service
 
 import android.animation.ValueAnimator
-import android.app.Activity
 import android.content.Context
 import android.graphics.Color
 import android.graphics.drawable.ColorDrawable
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
```diff
@@ -92,31 +92,27 @@ class BackgroundService(
...
-		windowBackground = current?.copy() ?: ColorDrawable(Color.BLACK)
+		windowBackground = if (current !is LayerDrawable) current.copy() else ColorDrawable(Color.BLACK)
 		backgroundDrawable.setDrawableByLayerId(R.id.background_static, windowBackground)
 
 		// Store size of window manager for this activity
 		windowSize = WindowManager(activity).currentWindowMetrics.bounds.let {
 			Size(it.right, it.bottom)
 		}
 
-		activity.window.decorView.background = backgroundDrawable
```
nielsvanvelzen, line 106: It might look weird that this line is removed but it actually makes sense. The BackgroundServiceFragment now manages  the lifecycle of the set background, the `onResume` function sets (and restores) the drawable now.

thornbill (approved):

PR #787, opened 2021-03-22
Fix for HDR always being activated

### LINKED ISSUE #782: All content triggers HDR on display, even if it’s not HDR content
**Describe the bug**
All videos are triggering HDR on my tv. This happens with any video regardless of if it’s HDR or not.

**To Reproduce**

1. Play any video
2. Check your display to see that HDR is triggered

**Expected behavior**
It’s expected that HDR is only triggered on my TV when playing media with HDR. SDR content should not trigger HDR on my tv.

**Logs**

**Screenshots**

 - Android TV - Shield version 8.2.2
 - Device manufacturer: NVidia
 - Device model: Shield TV
 - Jellyfin server version: 10.7

**Additional context**

linetrimmer:
**Changes**
I think this might be a Shield-specific issue, so I couldn't reproduce it (I don't have a Shield). This seems to have fixed it, though.

Edit: ~~Agh, this breaks Dolby Vision in mkv. Dolby Vision in mp4 still works.~~ I take that back, the file was transcoding instead of direct playing. HDR behavior seems the same on my Sony TV and Chromecast with Google TV.

**Issues**
Fixes #782

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -90,7 +90,6 @@
 
         <activity
             android:name=".ui.playback.PlaybackOverlayActivity"
-            android:colorMode="wideColorGamut"
             android:label="PlaybackOverlayActivity"
             android:screenOrientation="landscape" />
 


thornbill:
Requesting @nielsvanvelzen to review since he added this as to fix an issue on some TVs in https://github.com/jellyfin/jellyfin-androidtv/pull/369.

nielsvanvelzen (approved):

PR #792, opened 2021-03-27
Remove flexbox dependency and move GenreButton uses to a simple TextView

nielsvanvelzen:
**Changes**
Removed some flex

The TextView actually looks the same as the complicated list with TextViews except for the font (I didn't copy that over because it looks better with the default font IMO).

**Issues**


nielsvanvelzen:
Draft until I bother to fix the linter issue (Can't use String.join because it's a Java 8>= feature)

thornbill (approved):

PR #801, opened 2021-04-05
Use blurhashes while loading card images

linetrimmer:
**Changes**
Uses blurhashes when loading cards. 

Tested and working correctly on my server for movie and tv series posters/thumbnails/banners, episode thumbnails, season posters, and album covers. 

~~'My Media' library cards are the only cards on my server that do not load a blurhash when they should.~~  Got the 'My Media' cards to work

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/BaseRowItem.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/BaseRowItem.java
@@ -1,7 +1,6 @@
 package org.jellyfin.androidtv.ui.itemhandling;
 
 import android.content.Context;
-import android.graphics.Bitmap;
 import android.graphics.drawable.Drawable;
 import android.text.format.DateUtils;
 
@@ -11,7 +10,6 @@ import org.jellyfin.androidtv.R;
 import org.jellyfin.androidtv.TvApp;
 import org.jellyfin.androidtv.data.model.ChapterItemInfo;
 import org.jellyfin.androidtv.ui.GridButton;
-import org.jellyfin.androidtv.util.BlurHashDecoder;
 import org.jellyfin.androidtv.util.ImageUtils;
 import org.jellyfin.androidtv.util.TimeUtils;
 import org.jellyfin.androidtv.util.Utils;
@@ -32,7 +30,6 @@ import org.jellyfin.apiclient.model.search.SearchHint;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
 import java.util.Date;
-import java.util.HashMap;
 
 import kotlin.Lazy;
 
@@ -546,28 +543,6 @@ public class BaseRowItem {
         }
 
         return ContextCompat.getDrawable(TvApp.getApplication(), R.drawable.blank10x10);
-    }
-
-    public Bitmap getBlurHashBitmap(double aspect) {
-        final double ASPECT_RATIO_BANNER = 5.414;
-
-        Bitmap bitmap = null;
-        if (baseItem != null && baseItem.getImageBlurHashes() != null) {
-            HashMap<String, String> blurHashMap;
-            if (aspect == ASPECT_RATIO_BANNER) {
-                blurHashMap = baseItem.getImageBlurHashes().get(ImageType.Banner);
-            } else if (aspect == ImageUtils.ASPECT_RATIO_16_9) {
-                blurHashMap = baseItem.getImageBlurHashes().get(ImageType.Thumb);
-            } else {
-                blurHashMap = baseItem.getImageBlurHashes().get(ImageType.Primary);
-            }
-
-            if (blurHashMap != null) {
-                String blurHash = (String) blurHashMap.values().toArray()[0];
-                bitmap = BlurHashDecoder.INSTANCE.decode(blurHash, 32, 32, 1, true);
-            }
-        }
-        return bitmap;
     }
 
     public void refresh(final EmptyResponse outerResponse) {
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/BaseRowItem.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/BaseRowItem.java
@@ -1,10 +1,13 @@
 package org.jellyfin.androidtv.ui.itemhandling;
 
 import android.content.Context;
+import android.graphics.Bitmap;
 import android.graphics.drawable.Drawable;
 import android.text.format.DateUtils;
 
 import androidx.core.content.ContextCompat;
+
+import com.wolt.blurhashkt.BlurHashDecoder;
 
 import org.jellyfin.androidtv.R;
 import org.jellyfin.androidtv.TvApp;
@@ -30,6 +33,7 @@ import org.jellyfin.apiclient.model.search.SearchHint;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
 import java.util.Date;
+import java.util.HashMap;
 
 import kotlin.Lazy;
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/BaseRowItem.java
```diff
@@ -545,6 +549,28 @@ public Drawable getBadgeImage() {
...
 
+    public Bitmap getBlurHashBitmap(double aspect) {
+        final double ASPECT_RATIO_BANNER = 5.414;
+
+        Bitmap bitmap = null;
+        if (baseItem != null && baseItem.getImageBlurHashes() != null) {
+            HashMap<String, String> blurHashMap;
+            if (aspect == ASPECT_RATIO_BANNER) {
+                blurHashMap = baseItem.getImageBlurHashes().get(org.jellyfin.apiclient.model.entities.ImageType.Banner);
+            } else if (aspect == ImageUtils.ASPECT_RATIO_16_9) {
```
linetrimmer, line 560:
```suggestion
            } else if (aspect == ImageUtils.ASPECT_RATIO_16_9 && this.getBaseItemType() != BaseItemType.UserView) {
```
I tried this since the My Media library images are `ImageType.Primary`, but it didn't work. The blurhash loads when I ignore aspect ratio and force `ImageType.Primary` for everything, so they are there—it's just trying for the nonexistent thumb blurhash

build.gradle.kts
```diff
@@ -17,6 +17,7 @@ allprojects {
 	repositories {
 		jcenter()
 		google()
+		maven { setUrl("https://jitpack.io") }
```
nielsvanvelzen, line 20: Jitpack is a PITA to support, it often breaks and is slow. I'm glad we finally removed it when I put in the effort to migrate the (now deprecated) apiclient to JCenter. You'll need to find an alternative to this cause I don't want to depend on jitpack again.

nielsvanvelzen: One solution for this is to copy the BlurHashDecoder class to our codebase and adding a license header, not the best solution but it doesn't look like the BlurHash lib is published anywhere

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
```diff
@@ -410,7 +427,25 @@ public void onBindViewHolder(Presenter.ViewHolder viewHolder, Object item) {
...
+            if (aspect == ASPECT_RATIO_BANNER) {
+                blurHashMap = rowItem.getBaseItem().getImageBlurHashes().get(org.jellyfin.apiclient.model.entities.ImageType.Banner);
+            } else if (aspect == ImageUtils.ASPECT_RATIO_16_9 && !isUserView) {
+                blurHashMap = rowItem.getBaseItem().getImageBlurHashes().get(org.jellyfin.apiclient.model.entities.ImageType.Thumb);
+            } else {
+                blurHashMap = rowItem.getBaseItem().getImageBlurHashes().get(org.jellyfin.apiclient.model.entities.ImageType.Primary);
+            }
+
+            if (blurHashMap != null) {
+                String blurHash = (String) blurHashMap.values().toArray()[0];
```
nielsvanvelzen, line 442: I'm not sure if just taking the first value is the way to go. AFAIK you need to get the blurhash for the current image id. You also want to verify if the hashmap is not empty because the server sometimes doesn't include blurhashes.

app/src/main/java/org/jellyfin/androidtv/util/BlurHashDecoder.kt
```diff
@@ -0,0 +1,197 @@
+package org.jellyfin.androidtv.util
+
```
nielsvanvelzen, line 2: Can you add a link to the original source (include the commit hash in it please)

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
```diff
@@ -143,6 +150,7 @@ public void setItem(BaseRowItem m, String imageType, int lHeight, int pHeight, i
                                     break;
                             }
                             showProgress = true;
+                            isUserView = false;
```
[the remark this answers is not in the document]
linetrimmer: Without setting it to false here, the blurhash for the episode thumbnail loads when it should be loading the blurhash for the parent thumbnail. It was working correctly (blurhash for episode when prefer series thumbnail is off, blurhash for series when prefer series thumbnail is on) before I started using the isUserView boolean, so I added this for episodes and it fixed the problem. I'm not sure why it ends up being true without this, though.

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
```diff
@@ -410,7 +427,25 @@ public void onBindViewHolder(Presenter.ViewHolder viewHolder, Object item) {
             }
         }
 
-        holder.updateCardViewImage(rowItem.getImageUrl(holder.mCardView.getContext(), mImageType, holder.getCardHeight()));
+        BitmapDrawable blurHashDrawable = null;
+        if (rowItem.getBaseItem() != null && rowItem.getBaseItem().getImageBlurHashes() != null) {
+            HashMap<String, String> blurHashMap;
+            if (aspect == ASPECT_RATIO_BANNER) {
+                blurHashMap = rowItem.getBaseItem().getImageBlurHashes().get(org.jellyfin.apiclient.model.entities.ImageType.Banner);
```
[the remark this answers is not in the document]
linetrimmer: Removing this causes an error for me in the IDE since `org.jellyfin.androidtv.constant.ImageType` was already imported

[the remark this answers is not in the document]
nielsvanvelzen: I wasn't aware of that, would it be possible to replace the usages of `org.jellyfin.androidtv.constant.ImageType` with `org.jellyfin.apiclient.model.entities.ImageType`? If not you can keep it like it is now.

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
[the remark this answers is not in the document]
nielsvanvelzen: Hmm weird, maybe resetting it to false at the top of the `setItem` function is a better approach then.

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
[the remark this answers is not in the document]
linetrimmer: `org.jellyfin.androidtv.constant.ImageType` is used in the CardPresenter constructor so it'd also require changes to StdGridFragment (and maybe others but that was the first to pop up). I'll keep it like it is now

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
linetrimmer: You're right, episode thumbs and season posters have arrays with two values—one is the blurhash for the image and the other is the blurhash for the parent image. I didn't notice because a lot of times the parent blurhash is second, but sometimes they're flipped, so this doesn't work. Using image tags works.

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
```diff
@@ -410,7 +421,29 @@ public void onBindViewHolder(Presenter.ViewHolder viewHolder, Object item) {
...
+                blurHashMap = rowItem.getBaseItem().getImageBlurHashes().get(org.jellyfin.apiclient.model.entities.ImageType.Thumb);
+                imageTag = (rowItem.getPreferParentThumb()) ? rowItem.getBaseItem().getParentThumbImageTag() : rowItem.getBaseItem().getImageTags().get(org.jellyfin.apiclient.model.entities.ImageType.Thumb);
+            } else {
+                blurHashMap = rowItem.getBaseItem().getImageBlurHashes().get(org.jellyfin.apiclient.model.entities.ImageType.Primary);
+                imageTag = rowItem.getBaseItem().getImageTags().get(org.jellyfin.apiclient.model.entities.ImageType.Primary);
+            }
+
+            if (blurHashMap != null && !blurHashMap.isEmpty() && imageTag != null) {
+                String blurHash = blurHashMap.get(imageTag);
+                Bitmap blurHashBitmap = BlurHashDecoder.INSTANCE.decode(blurHash, 32, 32, 1, true);
```
nielsvanvelzen, line 441: I'm wondering if creating a bitmap with the correct aspect ratio would make more sense. Right now you're creating a 32x32 bitmap but if you multiply the width with the aspect ratio it would be proportional to the actual image.

nielsvanvelzen (commented):
Looks good! I tested it and it really gives a luxurious feel to the app now. Got one last question for now.

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
linetrimmer: Definitely looks better using the correct aspect ratio. I pushed a commit that keeps one dimension at 32 and adjusts the other by the aspect ratio. I'm not sure if there's potential for a performance hit with banner images since they end up being 173x32. Posters are 32x48, and thumbs are 57x32. I can flip it so 32 is the maximum for a dimension if you think performance might be an issue.

nielsvanvelzen (approved):

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
nielsvanvelzen: I think it's fine, the decoding shouldn't impact performance that much

PR #808, opened 2021-04-07
Cancel setCurrentSession call if the UserDto cannot be retrieved

nielsvanvelzen:
**Changes**
- Fix infinite loading issue when the apiBinder fails to set the currentUser property in TvApp.

It's kind of a hack until we fully migrate to the Kotlin SDK and remove the currentUser property entirely.

**Issues**

- https://github.com/jellyfin/jellyfin-androidtv/pull/784#issuecomment-813879511


thornbill (approved):

linetrimmer:
Ahh 20 seconds late -- I can't get past the splash screen with a fresh install on my emulator, but it does open if there was already a user signed in before updating the app.

PR #810, opened 2021-04-08
Fix null session causing app to hang

thornbill:
**Changes**
Fixes an issue that slipped through in #808. I think it makes sense to return true here since the value of `currentUser` is updated. :man_shrugging: 

**Issues**
N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/auth/ApiBinder.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/auth/ApiBinder.kt
@@ -24,7 +24,7 @@ class ApiBinder(
 	fun updateSession(session: Session?, resultCallback: (Boolean) -> Unit) {
 		if (session == null) {
 			application.currentUser = null
-			resultCallback(false)
+			resultCallback(true)
 			return
 		}
 


app/src/main/java/org/jellyfin/androidtv/auth/ApiBinder.kt
```diff
@@ -23,8 +23,9 @@ class ApiBinder(
 ) {
 	fun updateSession(session: Session?, resultCallback: (Boolean) -> Unit) {
 		if (session == null) {
+			Timber.d("Session is null, resetting current user.")
 			application.currentUser = null
-			resultCallback(false)
+			resultCallback(true)
```
nielsvanvelzen, line 28: Doing this could cause the StartupActivity to wait for a not-null currentUser here:

https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/ui/startup/StartupActivity.kt#L78-L82

Instead, we should probably check in the callback if the currentUser is set instead of using this boolean:

https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/auth/SessionRepository.kt#L92

thornbill: Would it ever hit this condition since the session is `null` though?

https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/ui/startup/StartupActivity.kt#L71

nielsvanvelzen: No actually it shouldn't be able to hit that condition :p

I took another look at the code (I already forgot how it worked...) and this is indeed the fix to the issue, I shouldn't write code or review when I'm half asleep 😂 .

nielsvanvelzen: The logging statement doesn't seem necessary to me though

nielsvanvelzen (approved):

PR #811, opened 2021-04-08
Remove TextSearchFragment and use LeanbackSearchFragment only

nielsvanvelzen:
**Changes**
- Updated leanback library to 1.1.0-rc1 which should fix search on Fire TV devices
- Remove TextSearchFragment and use LeanbackSearchFragment only

This PR requires testing on an Fire TV device to see if it works as intended.

**Issues**


nielsvanvelzen:
Thanks for testing. I guess we can keep using the dual-search screens for now. We could probably at voice recognition to the custom fragment at some point and drop leanback.

nielsvanvelzen:
I remember Amazon not allowing it because of their global search. What I meant was adding voice recognition for ATV only.

PR #817, opened 2021-04-10
Remove activity labels

nielsvanvelzen:
**Changes**
Removed labels for activities in the manifest. According to the Android documentation the app name will be used if it's not set.
https://developer.android.com/guide/topics/manifest/activity-element#label

**Issues**


nielsvanvelzen:
I kept the string for now to avoid conflicts with Weblate. I'm planning to do a cleanup removing all unused strings before 0.12.

thornbill (approved):

PR #818, opened 2021-04-11
Use the AddressCandidateHelper of the Kotlin SDK to normalize the inputted server address

nielsvanvelzen:
**Changes**

- Add the Kotlin SDK as dependency
- Use the AddressCandidateHelper from the SDK to find candidates for the inputted server address and try them one-by-one. The first one that works is chosen. If none of the candidates work we show the error of the latest attempt.

Function-wise this works great, I can now just input "demo.jellyfin.org/stable" and it will automatically add  the protocol. UI-wise it's terrible right now, error reporting wasn't great before but it's garbage now.

The `// TODO` is not something that can be fixed right now - we don't have a `Jellyfin()` instance of the SDK at this moment.

**Issues**

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -69,6 +69,7 @@ android {
 dependencies {
 	// Jellyfin
 	implementation("org.jellyfin.apiclient:android:0.7.9")
+	implementation("org.jellyfin.sdk:jellyfin-platform-android:1.0.0-beta.3")
 
 	// Kotlin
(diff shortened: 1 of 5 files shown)


nielsvanvelzen:
linter issue is an issue with a translation, should be fixed in weblate after merge

thornbill (approved):
Confirmed this worked by adding a server using only the hostname. I could see it progress through multiple attempts and resolved to using http://hostname.

PR #821, opened 2021-04-14
Adjust card and column sizes for vertical grid

linetrimmer:
**Changes**
For the vertical grid:
- Adjust column size for thumb and banner views
- Decrease heights of medium posters (175 -> 171 dp) and large posters (210 -> 202 dp) to fit another item in each row
- Setting the image size preference to Default (used to be named Auto) now chooses the correct column size for the image size that gets used


nielsvanvelzen (approved):
I think this pretty much lifts the vertical grids out of the "experimental" stage. The sizing of the cards was the reason I marked it as experimental.

PR #826, opened 2021-04-14
Switch to android DateFormat in addDate function in InfoLayoutHelper

nielsvanvelzen:
**Changes**
- Switch to android DateFormat in addDate function in InfoLayoutHelper

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
@@ -3,6 +3,7 @@ package org.jellyfin.androidtv.util;
 import android.app.Activity;
 import android.graphics.Color;
 import android.text.TextUtils;
+import android.text.format.DateFormat;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.TextView;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
```diff
@@ -274,11 +274,11 @@ private static void addDate(@NonNull  Activity activity, BaseItemDto item, Linea
                 StringBuilder sb = new StringBuilder();
                 if (item.getPremiereDate() != null) {
                     sb.append(activity.getString(R.string.lbl_born));
-                    sb.append(new SimpleDateFormat("d MMM y").format(TimeUtils.convertToLocalDate(item.getPremiereDate())));
+                    sb.append(DateFormat.getDateFormat(activity).format(TimeUtils.convertToLocalDate(item.getPremiereDate())));
```
thornbill, line 277: Maybe we should use `getMediumDateFormat` for these? It seems like it should match closer to the previous format. https://developer.android.com/reference/android/text/format/DateFormat#getMediumDateFormat(android.content.Context)

nielsvanvelzen: `getMediumDateFormat` does give better output, I've changed it. I've also changed the function to use a context instead of an activity which is preferred in android dev (although an activity is an context).

thornbill (approved):

PR #830, opened 2021-04-19
Move grid direction option to DisplayPreferencesScreen

linetrimmer:
**Changes**

- Move the grid direction option from the developer options to DisplayPreferencesScreen so the grid direction can be set on a per-library basis
- ~~Enable the vertical grid by default~~
- Remove the empty space above the first row of the vertical grid
- Add sizes for libraries with square images as the primary image. I'm not sure of all the library types that use square images for the primary image—I know music does and from the demo server it looks like playlists also do.


nielsvanvelzen:
I'm not sure if we should make it the default, after using the vertical grids for a bit I noticed that photo libraries in vertical mode don't work properly, you can scroll both vertically and horizontally.

Moving it to display preferences is a great idea though.

app/src/main/java/org/jellyfin/androidtv/ui/browsing/DisplayPreferencesScreen.kt
```diff
@@ -46,6 +46,19 @@ class DisplayPreferencesScreen : OptionsFragment() {
...
+				entries = mapOf(
+					"0" to requireContext().getString(R.string.grid_direction_vertical),
+					"1" to requireContext().getString(R.string.grid_direction_horizontal),
+				)
+
+				bind {
+					get { displayPreferences.getCustomPrefs().get("GridDirection") ?: "0" }
+					set { displayPreferences.getCustomPrefs().set("GridDirection", it) }
+					default { "0" }
+				}
```
nielsvanvelzen, line 60: I'm guessing you changed from an enum to integers because of this binding part. I'd rather use enums in the code because it's type safe. You can then store the enum names in the displaypreferences.

linetrimmer:
> I'm not sure if we should make it the default, after using the vertical grids for a bit I noticed that photo libraries in vertical mode don't work properly, you can scroll both vertically and horizontally.
> 
> Moving it to display preferences is a great idea though.

Do photo libraries use a square aspect ratio? I only have tv, movies, collections, and a small music library, so those were the only ones I could properly test. The way I did it requires libraries with square cards to be explicitly stated, I just wasn’t sure which to include. Also, if there are any libraries that use a mix of different aspect ratios for the default image, we could hide the grid direction option for that library type, but I’m not sure which, if any, libraries do that.

nielsvanvelzen:
> Do photo libraries use a square aspect ratio?

Nope, the width changes based on the aspect ratio of the image. For vertical grids this should be changed to change the height instead.

> Also, if there are any libraries that use a mix of different aspect ratios for the default image, we could hide the grid direction option for that library type, but I’m not sure which, if any, libraries do that.

I'm not sure either what sizes all types of libraries use. That's why I think it's better to leave the default option to horizontal for now and when we're confident all libraries use correct sizes we _could_ change it.

app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
```diff
@@ -135,10 +139,14 @@ private void setGridSizes() {
                 default:
                     if (mCardHeight == SMALL_VERTICAL_POSTER) {
                         size = 10;
-                    } else if (mCardHeight == MED_VERTICAL_POSTER) {
+                    } else if (mCardHeight == MED_VERTICAL_POSTER || mCardHeight == SMALL_VERTICAL_SQUARE) {
                         size = 7;
-                    } else {
+                    } else if (mCardHeight == LARGE_VERTICAL_POSTER){
```
nielsvanvelzen, line 144:
```suggestion
                    } else if (mCardHeight == LARGE_VERTICAL_POSTER) {
```

app/src/main/java/org/jellyfin/androidtv/ui/browsing/DisplayPreferencesScreen.kt
```diff
@@ -46,6 +47,19 @@ class DisplayPreferencesScreen : OptionsFragment() {
...
+					GridDirection.HORIZONTAL.name to requireContext().getString(R.string.grid_direction_horizontal),
+					GridDirection.VERTICAL.name to requireContext().getString(R.string.grid_direction_vertical),
+				)
+
+				bind {
+					get { displayPreferences.getCustomPrefs().get("GridDirection") ?: GridDirection.HORIZONTAL.name }
+					set { displayPreferences.getCustomPrefs().set("GridDirection", it) }
+					default { GridDirection.HORIZONTAL.name }
+				}
+			}
```
nielsvanvelzen, line 62: There is an `enum` function that uses a list in the back. Something like this should work, although this sample can cause a crash if the value is not null and an bad value.

```suggestion
			enum<GridDirection> {
				setTitle(R.string.grid_direction)

				bind {
					get { displayPreferences.getCustomPrefs().get("GridDirection")?.let(GridDirection::valueOf) ?: GridDirection.HORIZONTAL }
					set { displayPreferences.getCustomPrefs().set("GridDirection", it.name) }
					default { GridDirection.HORIZONTAL }
				}
			}
```

nielsvanvelzen (changes requested):

nielsvanvelzen:
Btw, merge conflict was caused by #824 but should be easy to fix with a rebase.

nielsvanvelzen (approved):

PR #831, opened 2021-04-19
Add link to system subtitle settings in playback preferences

nielsvanvelzen:
**Changes**

In emulator/Android TV looks like this:

image

image

AFAIK ExoPlayer uses the chosen style.

Link is disabled when setting not supported by device.

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/preference/category/playback.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/preference/category/playback.kt
@@ -2,6 +2,8 @@ package org.jellyfin.androidtv.ui.preference.category
 
 import android.app.Activity
 import android.app.AlertDialog
+import android.content.Intent
+import android.provider.Settings
 import org.jellyfin.androidtv.R
 import org.jellyfin.androidtv.preference.UserPreferences
 import org.jellyfin.androidtv.preference.constant.AudioBehavior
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
I definitely did not create this PR after some idiot on Reddit complaining.

<details>

> Jellyfin android TV app seems like the bottom of their list when it should be the first one to maintain and develop properly.

> Lol imagine not having this feature, weird at most. But whatever.

</details>

ManuLinares:
I guess not supported in my "Firestick v1" because it's greyed out. Also, there is no way to get rid of subtitle background?

I used to love v11.5 subtitles, only text no background.

nielsvanvelzen:
I don't have any amazon devices to test it on but the subtitle styling should be the same as 11.5 after this PR.

thornbill:
Does this actually work? afaik we hide the default rendered subtitles and render them manually.

nielsvanvelzen:
I think it did use the system styling for at least some subtitles. But I'll admit I have no idea how the subtitle rendering currently works.

nielsvanvelzen:
Actually you're right, we disable ExoPlayer's rendering here:
https://github.com/jellyfin/jellyfin-androidtv/blob/5192d63ff9184767d16f74f15db7e8d59cba9ca6/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java#L101-L106

nielsvanvelzen:
I think it was the mobile android app that uses the system settings 🤦

thornbill:
I really don't know _why_ we don't use the player renderers unless it's just to normalize the display... possibly something we should revisit in the playback rewrite though

ManuLinares:
> Actually you're right, we disable ExoPlayer's rendering here:
> 
> https://github.com/jellyfin/jellyfin-androidtv/blob/5192d63ff9184767d16f74f15db7e8d59cba9ca6/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java#L101-L106

Do you know in the code, where is the subtitle appearance. I'd like to remove the background and compile my own and build the apk.

PR #837, opened 2021-04-24
Fix height of first episode of next up row in the main item details page

linetrimmer:
**Changes**
Sets the height of the first episode of the next up row of the main item details page to 260 to match the rest of the episodes in the row.

**Issues**
The first episode was 300 while the rest were 260, causing the episode to appear larger and the row height to change as the episode was scrolled off screen.


nielsvanvelzen (approved):
Weirdly enough I did not notice this issue before but I can confirm it existed :p

Issue #839, opened 2021-04-26
Migrate from JCenter to Maven Central


nielsvanvelzen:
This is the Android TV version of the SDK issue. We need to migrate from JCenter/Bintray to 100% Maven Central + Google Maven.

## Copied from jellyfin/jellyfin-sdk-kotlin#176:

Unfortunately JCenter is closing in a few months. We need to publish the apiclient somewhere else before this happens. Additionally all our Java based projects (the Android apps) need to switch to different repositories for all dependencies.

I think our best bet is to wait a bit and see what other projects in the java ecosystem do before we act.

See the jfrog blog article for some more information:

https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

nielsvanvelzen:
Removing the JCenter repository in the current master branch gives the following issues:

- Jellyfin apiclient - The SDK is on Maven Central, the old apiclient is not. To fix: migrate or (temporary) use jitpack
- Koin - Next version of Koin is already available on Maven Cental as beta, the old one is not. To fix: update Koin.
- ExoPlayer Amazon fork - The project is mostly dead and unmaintained, it's not likely it will be pushed to Maven Central as Amazon needs to change the package for it. To fix: switch to official Google ExoPlayer.

It might be possible to fix these issues for the 0.12 release but it's a relatively low priority for now.

nielsvanvelzen:
Apparently Koin 2.x is available on Maven Central too, it wasn't the last time I checked. I will switch Koin asap.

ManuLinares:
> Removing the JCenter repository in the current master branch gives the following issues:
> 
>     * Jellyfin apiclient - The SDK is on Maven Central, the old apiclient is not. To fix: migrate or (temporary) use jitpack
> 
>     * ~Koin - Next version of Koin is already available on Maven Cental as beta, the old one is not. To fix: update Koin.~ #840
> 
>     * ~ExoPlayer Amazon fork - The project is mostly dead and unmaintained, it's not likely it will be pushed to Maven Central as Amazon needs to change the package for it. To fix: switch to official Google ExoPlayer.~ #843
> 
> 
> It might be possible to fix these issues for the 0.12 release but it's a relatively low priority for now.

jitpack can't be used, https://jitpack.io/com/github/jellyfin/jellyfin-sdk-kotlin/v0.7.9/build.log

nielsvanvelzen:
[...]

PR #843, opened 2021-04-28
Update ExoPlayer to 2.14.0

nielsvanvelzen:
**Changes**

- Update ExoPlayer to 2.14.0 (was 2.11.3)
- Switch from Amazon fork to official Google version

  Amazon doesn't maintain their fork anymore and I'm not a fan of using an outdated version. There's lots of complains on Reddit about it too, some users even manually replace the dependency and build their own version of the app. Moving away from the fork should not impact too many users, afaik the patches in the fork are mostly for older generations of Fire OS devices. Those users should update their device or force LibVLC as videoplayer.

**Issues**

Part of #839

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -108,7 +108,7 @@ dependencies {
 	implementation("com.google.code.gson:gson:2.8.6")
 
 	// Media players
-	implementation("com.amazon.android:exoplayer:2.11.3")
+	implementation("com.google.android.exoplayer:exoplayer:2.14.0")
 	implementation("org.videolan.android:libvlc-all:3.3.14")
 
 	// Image utility


thornbill:
> Amazon doesn't maintain their fork anymore

Have a reference for this?

anthonylavado:
> Have a reference for this?

@thornbill This seems to be the place: https://github.com/amzn/exoplayer-amazon-port/issues/107

ManuLinares:
https://github.com/amzn/exoplayer-amazon-port/issues/107#issuecomment-707945344

upgrading to google.exoplayer will deprecate my device.

nielsvanvelzen:
(quoted from the jellyfin-android room on Matrix)

> I'd say the amazon fork is unmaintained, they have been ignoring the project for months now - no updates in the code, no responses in issues etc.
when you look at the commits in any of the branches you can see the patches they apply:
https://github.com/amzn/exoplayer-amazon-port/commits/amazon/r2.12.1
>
> As far as I can tell only 2/3 of them are fixes, others are for debugging issues. If I had to choose between supporting a few patches for a small subset of devices or running the latest exoplayer with way more fixes for way more devices, I'd choose the latter

@ManuLinares I understand this is not an ideal solution for you but like as I said in the above quote, I'd rather update to a newer ExoPlayer version and help more users then keep supporting a small subset of devices with less users. Fortunately we support multiple playback backends (LibVLC/ExoPlayer) so using LibVLC should work fine.

ManuLinares:
> (quoted from the jellyfin-android room on Matrix)
> 
> > I'd say the amazon fork is unmaintained, they have been ignoring the project for months now - no updates in the code, no responses in issues etc.
> > when you look at the commits in any of the branches you can see the patches they apply:
> > https://github.com/amzn/exoplayer-amazon-port/commits/amazon/r2.12.1
> > As far as I can tell only 2/3 of them are fixes, others are for debugging issues. If I had to choose between supporting a few patches for a small subset of devices or running the latest exoplayer with way more fixes for way more devices, I'd choose the latter
> 
> @ManuLinares I understand this is not an ideal solution for you but like as I said in the above quote, I'd rather update to a newer ExoPlayer version and help more users then keep supporting a small subset of devices with less users. Fortunately we support multiple playback backends (LibVLC/ExoPlayer) so using LibVLC should work fine.

I would agree, only if libVLC worked fine but when using libVLC and fast forwarding, or jumping to a chapter, etc, the playtime resets to zero, so the subtitles do not work as they get de-synchronized

dani3l3:
I have a gazillion users on the oldest FireTV versions from several years ago... I am worried they would be impacted and I only have newer devices myself to be able to validate... :-(

nielsvanvelzen:
Here is the list of patches applied in the Amazon 2.12.1 branch, they would be missing if we migrate to Google ExoPlayer. I don't think they are significant enough to not migrate. And to be honest, I don't get why Amazon doesn't work with Google to get the patches into Google ExoPlayer...

- FTVPI-1553: Dolby passthrough on Gen 1 FireTV 
- FTVPI-1552: Gen4 Tablet audio latency issues
- FTVPI-1559: [EXO2] Logging support
- FTVPI-2218: Support for disabling snapping video frame release times to VSYNC added
- FTVPI-2396: Toshiba Fire TV edition switch to BT while playing Dolby doesn't work
- MMSW-9180: Disable Dummy Surface workaround for Amazon Devices
- MMSW-6599: Add set state logging to ExoPlayer v2
- MMSW-5642: Add modified version string for Amazon Exoplayer port 

_3 of the patches are for logging/version and do not change behavior at all._

ManuLinares:
> Here is the list of patches applied in the Amazon 2.12.1 branch, they would be missing if we migrate to Google ExoPlayer. I don't think they are significant enough to not migrate. And to be honest, I don't get why Amazon doesn't work with Google to get the patches into Google ExoPlayer...

Amazon is shit, that's not news. You are in favor for deprecating thousands of users, and your argument is:
"Those users should update their device or force LibVLC as videoplayer."

1. Updating the device is not an option for thousands.
2. if libVLC worked as intended I would agree to make it the default for older devices, but it doesn't (subtitles not synchronized, timeline resets to zero, etc).

So, the alternatives as I see them are fixing libVLC in jellyfin, or keep using amazon fork.

btw, I just tested amazon exoplayer 12.1 and it doesn't work.
```
04-30 16:33:12.157 7372-8357/org.jellyfin.androidtv.debug E/ExoPlayerImplInternal: Playback error
      com.google.android.exoplayer2.ExoPlaybackException: Source error
        at com.google.android.exoplayer2.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:556)
        at android.os.Handler.dispatchMessage(Handler.java:98)
        at android.os.Looper.loop(Looper.java:135)
        at android.os.HandlerThread.run(HandlerThread.java:61)
     Caused by: com.google.android.exoplayer2.source.UnrecognizedInputFormatException: None of the available extractors (MatroskaExtractor, FlvExtractor, FlacExtractor, WavExtractor, FragmentedMp4Extractor, Mp4Extractor, AmrExtractor, PsExtractor, OggExtractor, TsExtractor, AdtsExtractor, Ac3Extractor, Ac4Extractor, Mp3Extractor) could read the stream.
        at com.google.android.exoplayer2.source.BundledExtractorsAdapter.init(BundledExtractorsAdapter.java:92)
        at com.google.android.exoplayer2.source.ProgressiveMediaPeriod$ExtractingLoadable.load(ProgressiveMediaPeriod.java:1024)
        at com.google.android.exoplayer2.upstream.Loader$LoadTask.run(Loader.java:415)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
        at java.lang.Thread.run(Thread.java:818)
04-30 16:33:12.167 7372-7372/org.jellyfin.androidtv.debug E/VideoManager: ***** Got error from player
04-30 16:33:12.167 7372-7372/org.jellyfin.androidtv.debug E/PlaybackController: Playback error - Failed to load video
```

NebelNidas:
Would it be too much work to just include both? So Google ExoPlayer for normal users, and the outdated Amazon fork for the ones who absolutely need it? Just fixing libVLC would of course be the better option, but this might take quite a while 🤔

nielsvanvelzen:
It's not possible to include both ExoPlayer versions because Amazon didn't change the package. So only one could be included at a given time. We could go back to using separate build flavors but this results in longer build times and we still need to keep the dependencies in sync by using the same version, so we can't update Google ExoPlayer until Amazon gets their shit together. It also doesn't help with the "move away from JCenter" effort.

glaslos:
> It's not possible to include both ExoPlayer versions because Amazon didn't change the package

Could this be solved by forking Amazon's package?

nielsvanvelzen:
Yes it should be possible to fork for someone willing to maintain it. But I have no intention to do so.

svemonix:
If people want a continued Amazon device support, they should fork jellyfin-androidtv to something like jellyfin-fireos and maintain it rather than forking the amazon exoplayer itself. After all, it is not like the app published on the Play store has to match the one on Amazon, isn't it?

I don't see why Amazon's android fork should dictate the development on all other legit Android TV devices.

nielsvanvelzen:
> Is there somewhere where I could grab the apk with the latest Exoplayer to help testing?

In the build output from this PR: https://dev.azure.com/jellyfin-project/jellyfin/_build/results?buildId=28830&view=artifacts&pathAsName=false&type=publishedArtifacts

svemonix:
Thanks that didn’t turn out the way I hoped. 
Couldn’t install the release one and I couldn’t log in with the debug one.

svemonix:
Thanks and sorry I’m not too familiar with Android development.

I’ve tested the new exoplayer on one of my Android TV device and the results seemed overall quite good. 
I noticed that on my hardware the output is EAC3 for both EAC3 and AC3 content which is a little odd but works. Before it was switching.
The output now switches to DTS for DTS files contrary to the official release. The only issue is that it seems to try to output the full DTS stream even if the device is not capable. E.g. I believe it tries to output 7.1 DTS HD MA even if the Android TV device can’t do it. It results in very bad audio but this is an exoplayer issue. VLC and Kodi can keep only the 5.1 DTS core when needed.

nielsvanvelzen:
Updated the PR to target ExoPlayer 2.14 instead of 2.13.3.

svemonix:
Good news, apparently the DTS issue described above should be fixed with 2.14.0:
_Allow fall back from DTS-HD to DTS when playing via passthrough._

thornbill (approved):

PR #848, opened 2021-05-02
Replace item details end time when clock is hidden and add an option to hide the chapters row

linetrimmer:
**Changes**

- Adds an option to hide the chapters row on the item details screen. Since I don't extract chapter images and I never use the row, it just ends up being a rather unattractive row of teal.
- Replaces the "Ends" time with the episode/movie writer or series network if the clock is hidden on the screen

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
@@ -43,6 +43,7 @@ import org.jellyfin.androidtv.data.querying.TrailersQuery;
 import org.jellyfin.androidtv.data.service.BackgroundService;
 import org.jellyfin.androidtv.preference.SystemPreferences;
 import org.jellyfin.androidtv.preference.UserPreferences;
+import org.jellyfin.androidtv.preference.constant.ClockBehavior;
 import org.jellyfin.androidtv.preference.constant.PreferredVideoPlayer;
 import org.jellyfin.androidtv.ui.IRecordingIndicatorView;
 import org.jellyfin.androidtv.ui.RecordPopup;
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -523,6 +530,7 @@ protected void addItemRow(ArrayObjectAdapter parent, ItemRowAdapter row, int ind
 
     protected void addAdditionalRows(ArrayObjectAdapter adapter) {
         Timber.d("Item type: %s", mBaseItem.getBaseItemType().toString());
+        boolean chaptersRowEnabled;
```
nielsvanvelzen, line 533: You can just assign it here
```suggestion
        boolean chaptersRowEnabled = userPreferences.getValue().get(UserPreferences.Companion.getChaptersRowEnabled());
```

app/src/main/res/values/strings.xml
```diff
@@ -474,4 +474,6 @@
     <string name="user_picker_last_user_title">Most recently used</string>
     <string name="user_picker_last_user_summary">Use the last active user</string>
     <string name="lbl_switch_user">Switch user</string>
+    <string name="lbl_show_chapters_row">Show chapters row</string>
```
nielsvanvelzen, line 477:
```suggestion
    <string name="lbl_show_chapters_row">Show video chapters</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -474,4 +474,6 @@
...
+    <string name="lbl_show_chapters_row_description">"Display the video's chapters on the item screen"</string>
```
nielsvanvelzen, line 478:
```suggestion
    <string name="lbl_show_chapters_row_description">Display the video's chapters on the item details</string>
```

nielsvanvelzen (changes requested):

linetrimmer:
Hiding the end time got slightly more complicated than I initially thought it would be since I think just hiding the time leaves too much empty space. Rather than hiding the end time, it's now replaced with the writer or series network

nielsvanvelzen:
>  it's now replaced with the writer or series network

I don't like this approach, it means that instead of hiding the time (which is expected from the preference) it actually changes behavior and shows different information. What if I want that information together with the time? I think only hiding it is the solution. But...

> I think just hiding the time leaves too much empty space.

Can you explain this or share a screenshot? I don't understand what the problem with the empty space would be.

linetrimmer:
Screenshot_1620408915
Screenshot_1620408976

I think that in general a group of 3 items looks more complete than a group of 2 items, and in this case the three items frame the description section better since they connect the top of the description to the start of the buttons. This is information that's displayed on web but not on ATV, and I think displaying it makes better use of the space made available by hiding the end time. 

> What if I want that information together with the time?

There's currently not room to display the writer/network along with the end time, but there could be an option to display the information in place of the end time even if the clock is visible. Personally I lean towards there always being 3 items in that section, though.

thornbill:
I'm not sure I see the value in adding an option to hide chapters unless it is to prevent spoilers. I think that could be combined with the changes in https://github.com/jellyfin/jellyfin-androidtv/pull/850 into a single generic option to "avoid spoilers" that could be expanded to include things like descriptions in the future.

linetrimmer:
Closing in favor of #997

PR #849, opened 2021-05-02
Allow themes to change rounding values/indicator shape, add Muted Purple theme

linetrimmer:
**Changes**

- Allows themes to change the indicator shape and the rounding values for cards/buttons
- Adds a Muted Purple theme that uses an accent color from the inner triangle of the Jellyfin logo, translucent buttons, and a rounded square watched indicator
- Increases the rounding value for btn_focus and btn_non_focus from 2dp to 4dp (default theme)
- Raises the brightness of the non-focus grid items (all themes)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/constant/AppTheme.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/constant/AppTheme.kt
@@ -19,7 +19,7 @@ enum class AppTheme {
 	/**
 	 * A theme with a more muted accent color, inspired by CTalvio's Monochromic CSS theme for Jellyfin Web
 	 */
-	@EnumDisplayOptions(R.string.pref_theme_muted)
+	@EnumDisplayOptions(R.string.pref_theme_muted_purple)
 	MUTED_PURPLE,
 
 	/**
--- a/app/src/main/java/org/jellyfin/androidtv/preference/constant/AppTheme.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/constant/AppTheme.kt
@@ -17,6 +17,12 @@ enum class AppTheme {
 	EMERALD,
 
 	/**
+	 * A theme with a more muted accent color, inspired by CTalvio's Monochromic CSS theme for Jellyfin Web
+	 */
+	@EnumDisplayOptions(R.string.pref_theme_muted)
+	MUTED_PURPLE,
+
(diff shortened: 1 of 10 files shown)


app/src/main/res/values/theme_mutedpurple.xml
```diff
@@ -0,0 +1,21 @@
...
+    <style name="Theme.Jellyfin.MutedPurple" parent="Theme.Jellyfin">
+        <!-- Colors -->
+        <item name="android:colorAccent">@color/theme_muted_purple_accent</item>
+        <item name="defaultSearchColor">@color/theme_muted_purple_accent</item>
+        <item name="progressPrimary">@color/theme_muted_purple_accent</item>
+        <item name="android:colorButtonNormal">@color/light_grey_transparent</item>
+
+        <!-- Rounding Values -->
+        <item name="cardRounding">5dp</item>
+        <item name="buttonRounding">7dp</item>
```
nielsvanvelzen, line 16: I think this might be a bit too much rounding:

image

app/src/main/java/org/jellyfin/androidtv/preference/constant/AppTheme.kt
```diff
@@ -16,6 +16,12 @@ enum class AppTheme {
 	@EnumDisplayOptions(R.string.pref_theme_emerald)
 	EMERALD,
 
+	/**
+	 * A theme with a more muted accent color, inspired by CTalvio's Monochromic CSS theme for Jellyfin Web
+	 */
+	@EnumDisplayOptions(R.string.pref_theme_muted)
```
nielsvanvelzen, line 22:
```suggestion
	@EnumDisplayOptions(R.string.pref_theme_muted_purple)
```

nielsvanvelzen (changes requested):

app/src/main/res/values/theme_mutedpurple.xml
nielsvanvelzen: Looked at it again and I actually don't mind it now :p

linetrimmer: Ah cool :) I've been testing 5dp and 6dp on my tv the past couple days and I do still prefer 7dp

nielsvanvelzen (approved):

thornbill (approved):

PR #850, opened 2021-05-02
Adjust layout of next up screen, add option to hide thumbnail

linetrimmer:
**Changes**

- Adjusts the layout of the next up screen
- Removes the item description
- Adds an option to hide the thumbnail for the more Spoiler Sensitive
- Checks if the timer has already started once before so it doesn't restart if the device went into/came back out of a screensaver
- Includes the episode end number for files that contain more than one episode
- Formats the title as"S{x}:E{x} - Title" or "Special - Title"


ManuLinares:
> **Changes**
> 
>     * Adds an option to hide the thumbnail and description on the next up screen for the more Spoiler Sensitive

Tested, works. Love the no spoiler on Next Up screen option.

thornbill (changes requested):
Could we change the visibility of those specific views instead of using a second layout for this?

linetrimmer:
I think there’s too much empty space to just hide them without repositioning things

thornbill (changes requested):
There are some code style warnings here that look like they could be addressed here.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -65,6 +65,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 */
 		var nextUpTimeout = Preference.int("next_up_timeout", 1000 * 7)
 
+		/**
+		 * Display the next up item's thumbnail on the next up screen
+		 */
+		var nextUpThumbnailEnabled = Preference.boolean("next_up_thumbnail_enabled", true)
```
thornbill, line 71: I'm still in favor of having a generic option for hiding spoilers to be used for these types of features instead of individual options.

nielsvanvelzen: I too think that's a better approach.

linetrimmer: Regarding this specific option (hiding thumbnails on the next up screen), I don't think it should be included in a hide spoilers feature. Hiding spoilers should apply only to unwatched items, whereas this option hides the thumbnail whether the episode is watched or not. Since nothing appears on the next up screen until everything is loaded, loading the thumbnail makes the episode title and logo take longer to appear. In my experience, it sometimes takes 5+ seconds for the thumbnail to load, meaning the next up screen is blank for almost the entire duration (or the entire duration) the screen is set to be shown before starting the episode. Without having to load the thumbnail, the title and logo appear instantly since the logo image is cached after watching one episode of the show. Plus, in my opinion, the layout looks better without the thumbnail. 

Regarding options to hide spoilers, I'm not sure I understand why it would be better to group the options together. There could be spoilers in thumbnails, descriptions, episode titles, cast information, and video chapters. These things make up most of the metadata shown in the app, and I don't think it's safe to assume that if someone wants one of these things hidden they want all of them hidden. 

What if someone wants thumbnails hidden, but they like having episode titles, movie descriptions, and cast information? Their only option would be to hide everything.

Episode descriptions contain spoilers more often than movie descriptions or series descriptions; someone might want episode descriptions to be hidden since they're already watching the show but still want movie descriptions or series descriptions to be shown when they're deciding on something new to watch. 

The only reason I can think of to group the options into one generic option would be to prevent clutter in the settings menu, which is why I think it would be good to include them in a submenu like how the 'Developer options' are implemented—this way the options would only take up one slot in the main settings menu.

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpFragment.kt
```diff
@@ -43,7 +41,11 @@ class NextUpFragment : Fragment() {
 	override fun onResume() {
 		super.onResume()
 
-		binding.fragmentNextUpButtons.startTimer()
+		if (!timerStarted) {
+			binding.fragmentNextUpButtons.startTimer()
+
+			timerStarted = true
+		}
```
nielsvanvelzen, line 48: Why? The implementation of the startTimer function already restarts the timer when necessary, no need to add this check.

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
```diff
@@ -23,6 +26,7 @@ class NextUpViewModel(
 	val item: LiveData<NextUpItemData?> = _item
 	private val _state = MutableLiveData(NextUpState.INITIALIZED)
 	val state: LiveData<NextUpState> = _state
+	private val userPreferences by KoinJavaComponent.inject(UserPreferences::class.java)
```
nielsvanvelzen, line 29:
```suggestion
	private val userPreferences by inject<UserPreferences>()
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
```diff
@@ -52,22 +56,34 @@ class NextUpViewModel(
...
+			(item.parentIndexNumber != null && item.name != null) ->
+				"${context.getString(R.string.lbl_season_number, item.parentIndexNumber)} - ${item.name}"
+			(item.indexNumber != null && item.indexNumberEnd != null && item.name != null) ->
+				"${context.getString(R.string.lbl_episode_number, item.indexNumber)}-${item.indexNumberEnd} - ${item.name}"
+			(item.indexNumber != null && item.name != null) ->
+				"${context.getString(R.string.lbl_episode_number, item.indexNumber)} - ${item.name}"
+			(item.name != null) ->
+				item.name
+			else -> ""
+		}
```
nielsvanvelzen, line 81: This is really difficult to read/understand. You should find another way to create the strings. (and do we really need this many?!)

nielsvanvelzen (changes requested):
Setting aside the discussion about the spoiler preference, I reviewed everything else.
When looking at the new UI I have some additional notes:
- The background is too bright for white text. You should add a shadow or background to the text or darken the background
- The thumbnail is kinda big now in comparison with the text, I'd downsize it a bit and maybe add some rounding
- The position of  the text feels off, I think the top of the text should be aligned with the top of the thumbnail.
  - Maybe we should increase the margins of all components?

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
nielsvanvelzen: What we _could_ do is to change the "enable next up screen" option to an enum with these options:
- Disabled
- Enabled, minimal (no thumb)
- Enabled, extended (yes thumb)

That way the user can chose how it looks without adding an additional preference.

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpFragment.kt
linetrimmer: This is to prevent the timer from restarting after the device comes out of a screensaver or if the device is a tv and the tv had been turned off while on the next up screen. I’m not sure it’s expected that the timer start again in those cases (or at least it wasn’t expected to me) since the user isn’t necessarily ready to start watching immediately after coming out of a screensaver/turning on the tv.

nielsvanvelzen: Fair point, I think in that case you can just rename the `onResume` function to `onStart` so it only starts once.

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
linetrimmer: When I tried this it gave an error:

> No value passed for parameter 'clazz'

Removing `KoinJavaComponent.` works though:
`private val userPreferences by inject(UserPreferences::class.java)`

nielsvanvelzen: That error means the import is wrong, try removing the import and have android studio import it again

linetrimmer: Android studio seems to only want to import the KoinJavaComponent one. I tried removing that import and manually adding `import org.koin.android.ext.android.inject` and `import org.koin.core.component.inject` but it won't use either of them :/

> Unresolved reference. 
> None of the following candidates is applicable because of receiver type mismatch:
> - public inline fun <reified T : Any> ComponentCallbacks.inject(qualifier: Qualifier? = ..., mode: LazyThreadSafetyMode = ..., noinline parameters: ParametersDefinition? /* = (() → DefinitionParameters)? */ = ...): Lazy<TypeVariable(T)> defined in org.koin.android.ext.android
> - public inline fun <reified T : Any> KoinComponent.inject(qualifier: Qualifier? = ..., mode: LazyThreadSafetyMode = ..., noinline parameters: ParametersDefinition? /* = (() → DefinitionParameters)? */ = ...): Lazy<TypeVariable(T)> defined in org.koin.core.component

linetrimmer:
Will rebase and change the "enable next up screen" option to an enum after #932 is merged since that adds another use of the option as a boolean.

Visual changes since review:

- Added shadows to the text
- Decreased the thumbnail height to 145dp
- Rounded the corners using the theme rounding value
- Increased the margins a little
- Added a width constraint to the title text. It'll extend onto a second line with an ellipsis if necessary
- Replaced hyphens (-) with an en dash (–) for episode ranges and an em dash (—) for the separator

> The position of the text feels off, I think the top of the text should be aligned with the top of the thumbnail.

Aligning the top of the text with the top of the thumbnail leaves a lot of space between the text and the buttons that looks weird :/

Screenshots:
<details>
<summary>Single episode</summary>

single episode
</details>

<details>
<summary>Thumbnail hidden</summary>

no thumbnail
</details>

<details>
<summary>Episode range</summary>

episode-range
</details>

<details>
<summary>Special</summary>

special
</details>

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
linetrimmer: The string is now built as 3 pieces joined together—should be a lot easier to read than this mess

nielsvanvelzen (changes requested):
Don't touch translation files - only remove the key from the `/values/strings.xml` file. Weblate will fix the other files for us.

edit: make sure to remove/change the actual commit so the changes to those files are not in the git history

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
```diff
@@ -54,20 +59,35 @@ class NextUpViewModel(
 	private suspend fun loadItemData(id: String) = withContext(Dispatchers.IO) {
 		val item = apiClient.getItem(id) ?: return@withContext null
 
-		val thumbnail = apiClient.GetImageUrl(item, ImageOptions())
+		val thumbnail = when (userPreferences[UserPreferences.nextUpBehavior] == NextUpBehavior.EXTENDED) {
+			true -> apiClient.GetImageUrl(item, ImageOptions())
+			false -> null
+		}
```
nielsvanvelzen, line 65:
```suggestion
		val thumbnail = when (userPreferences[UserPreferences.nextUpBehavior]) {
			NextUpBehavior.EXTENDED -> apiClient.GetImageUrl(item, ImageOptions())
			else -> null
		}
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
```diff
@@ -54,20 +59,35 @@ class NextUpViewModel(
...
-		else if (item.name != null)
-			item.name
-		else ""
+		val seasonNumber = when (item.parentIndexNumber != null && item.parentIndexNumber != 0) {
+			true -> context.getString(R.string.lbl_season_number, item.parentIndexNumber)
+			false -> null
+		}
+		val episodeNumber = when {
+			(item.parentIndexNumber == 0) ->
+				context.getString(R.string.lbl_special)
```
nielsvanvelzen, line 74: I'm not sure if all items with a parentIndexNumber set to 0 are specials, Did you look at other places to determine this behavior?

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
```diff
@@ -54,20 +59,35 @@ class NextUpViewModel(
...
+			(item.indexNumber != null) ->
+				listOfNotNull(
+					context.getString(R.string.lbl_episode_number, item.indexNumber),
+					item.indexNumberEnd?.toString()
+				).joinToString("–")
```
nielsvanvelzen, line 79: It might be better to use 2 string resources. something like "lbl_episode_number" and "lbl_episode_range". That way translators can adapt it to their language. (some languages might use a different order etc.)

app/src/main/res/values/strings.xml
```diff
@@ -411,4 +410,10 @@
     <string name="pref_theme_muted_purple">Muted Purple</string>
     <string name="no_network_permissions">Jellyfin requires network permissions to function</string>
     <string name="press_to_close">Press to close</string>
+    <string name="lbl_special">Special</string>
+    <string name="lbl_season_number">S%1$d</string>
+    <string name="lbl_episode_number">E%1$d</string>
+    <string name="lbl_next_up_extended">Enabled, extended</string>
+    <string name="lbl_next_up_minimal">Enabled, minimal</string>
```
nielsvanvelzen, line 417: We might want to use different names to better explain the difference between extended and minimal. I'm not sure how to name it though.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
linetrimmer: This is how the web client handles it: https://github.com/jellyfin/jellyfin-web/blob/5c09077a2f12ae44e7b7d356ea8d9f35862b0b60/src/components/itemHelper.js#L21

nielsvanvelzen: Thanks, in that case you should check if the item is an episode because the next up screen can show up for movies in collections/playlists.

app/src/main/res/values/strings.xml
linetrimmer: Pushed a commit:

Show Next Up screen
- With thumbnail
- Without thumbnail
- Never

Alternatively, still using the 'Enable' wording:

Enable Next Up screen
- Extended (with thumbnail)
- Minimal (without thumbnail)
- Disabled

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
linetrimmer: If the item's not an episode, the string will now be formatted as something like "12. Title". 

I couldn't figure out how to test this, though: using the playlists in the demo server, the movies always played in the music player rather than a video player, and I couldn't get collections to queue subsequent movies. I'm also not sure how to actually view/start playing from the queue that you can manually create using the "Add to Queue" button in a movie or episode's item details.

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpFragment.kt
linetrimmer: `onStart` is called when coming out of the screensaver/turning back (was able to reproduce this in the emulator), so I went back to using this check unless you have another suggestion

app/src/main/res/values/strings.xml
```diff
@@ -294,8 +294,7 @@
     <string name="pref_next_up_timeout_summary">How long to wait before playing the next item</string>
     <string name="pref_next_up_timeout_title">Next Up timer duration</string>
     <string name="watch_now">Watch now</string>
-    <string name="pref_next_up_enabled_summary">Previews the next queued item before playing</string>
-    <string name="pref_next_up_enabled_title">Enable Next Up screen</string>
+    <string name="pref_next_up_behavior_title">Show Next Up screen</string>
```
nielsvanvelzen, line 297:
```suggestion
    <string name="pref_next_up_behavior_title">Show next video info</string>
```

Stolen from jellyfin-web:

image

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
```diff
@@ -54,20 +60,42 @@ class NextUpViewModel(
...
-			"${item.indexNumber}. ${item.name}"
-		else if (item.name != null)
-			item.name
-		else ""
+		val seasonNumber = when (item.baseItemType == BaseItemType.Episode && item.parentIndexNumber != null
+			&& item.parentIndexNumber != 0) {
+			true -> context.getString(R.string.lbl_season_number, item.parentIndexNumber)
+			false -> null
+		}
+		val episodeNumber = when {
```
nielsvanvelzen, line 74: No need for the parentheses in this when statement

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
```diff
@@ -54,20 +60,42 @@ class NextUpViewModel(
...
-		val title = if (item.indexNumber != null && item.name != null)
-			"${item.indexNumber}. ${item.name}"
-		else if (item.name != null)
-			item.name
-		else ""
+		val seasonNumber = when (item.baseItemType == BaseItemType.Episode && item.parentIndexNumber != null
+			&& item.parentIndexNumber != 0) {
+			true -> context.getString(R.string.lbl_season_number, item.parentIndexNumber)
+			false -> null
+		}
```
nielsvanvelzen, line 73: splitting into multiple lines and using the check as rule in the when statement is more readable

```suggestion
		val seasonNumber = when {
			item.baseItemType == BaseItemType.Episode
			&& item.parentIndexNumber != null
			&& item.parentIndexNumber != 0 ->
				context.getString(R.string.lbl_season_number, item.parentIndexNumber)
			else -> null
		}
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
```diff
@@ -54,20 +60,42 @@ class NextUpViewModel(
 	private suspend fun loadItemData(id: String) = withContext(Dispatchers.IO) {
 		val item = apiClient.getItem(id) ?: return@withContext null
 
-		val thumbnail = apiClient.GetImageUrl(item, ImageOptions())
+		val thumbnail = when (userPreferences[UserPreferences.nextUpBehavior]) {
+			NextUpBehavior.EXTENDED -> apiClient.GetImageUrl(item, ImageOptions())
+			else -> null
+		}
```
nielsvanvelzen, line 66: I'd prefer just sending the thumbnail but also sending the preference value and determine to show/hide in the actual fragment.

nielsvanvelzen: Actually the preference should be it's own value/function in the viewmodel

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
nielsvanvelzen: You're right, the first branch takes care of that.

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
```diff
@@ -54,20 +60,42 @@ class NextUpViewModel(
...
+			(item.baseItemType != BaseItemType.Episode) ->
+				item.indexNumber?.toString()
+			(item.parentIndexNumber == 0) ->
+				context.getString(R.string.lbl_special)
+			(item.indexNumber != null && item.indexNumberEnd != null) ->
+				context.getString(R.string.lbl_episode_range, item.indexNumber, item.indexNumberEnd)
+			(item.indexNumber != null) ->
+				context.getString(R.string.lbl_episode_number, item.indexNumber)
+			else -> null
+		}
```
nielsvanvelzen, line 84:
```suggestion
		val episodeNumber = when {
			item.baseItemType != BaseItemType.Episode -> item.indexNumber?.toString()
			item.parentIndexNumber == 0 -> context.getString(R.string.lbl_special)
			item.indexNumber != null -> when {
				item.indexNumberEnd != null -> context.getString(R.string.lbl_episode_range, item.indexNumber, item.indexNumberEnd)
				else -> context.getString(R.string.lbl_episode_number, item.indexNumber)
			}
			else -> null
		}
```
Same code but with the values on the same line and a secondary when statement for the indexnumberend/no end

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
```diff
@@ -54,20 +60,42 @@ class NextUpViewModel(
...
+		val seasonEpisodeNumbers = listOfNotNull(seasonNumber, episodeNumber).joinToString(":")
+
+		val nameSeparator = when (item.baseItemType) {
+			BaseItemType.Episode -> " — "
+			else -> ". "
+		}
+		val title = listOfNotNull(seasonEpisodeNumbers, item.name)
+			.filter { it.isNotEmpty() }
+			.joinToString(nameSeparator)
```
nielsvanvelzen, line 93: Please move all code to determine the title to a separate (private) function.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpFragment.kt
nielsvanvelzen: Ok then this will do for now. Although one problem that could arise is that the activity would be destroyed in the background and is missing the value when starting again

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
linetrimmer: If I’m understanding correctly, wouldn’t this mean time is spent downloading the thumbnail even though it’s not shown?

nielsvanvelzen: No the used function only creates a url here

app/src/main/res/values/strings.xml
linetrimmer: Planning for #922, would something like this make sense?

Show next video info

- After playback, with thumbnail
- After playback, without thumbnail
- During playback
- Never

With whichever one ends up being the default value as the first option listed (not sure if the plan is to go back to during playback as default). The timer duration could also stay as positive with these strings.

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
linetrimmer: I think I'm missing something here. 😅 The thumbnail is just a url here, but it's sent to the fragment as a bitmap in `NextUpItemData`: 
```
NextUpItemData(
	...
	thumbnail?.let { safelyLoadBitmap(it) },
	logo?.let { safelyLoadBitmap(it) }
)
```
So it's already loaded the bitmap by the time the preference would get used in the fragment

app/src/main/res/values/strings.xml
nielsvanvelzen: The title would just be "Show next video info" so it doesn't indicate when it shows up.

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
nielsvanvelzen: Whoops I forgot that part, you can leave it like it is then

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
```diff
@@ -51,23 +58,49 @@ class NextUpViewModel(
 		null
 	}
 
+	private fun BaseItemDto.getTitle(): String {
```
nielsvanvelzen, line 61: No need for `this` everywhere

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpViewModel.kt
```diff
@@ -23,6 +29,7 @@ class NextUpViewModel(
 	val item: LiveData<NextUpItemData?> = _item
 	private val _state = MutableLiveData(NextUpState.INITIALIZED)
 	val state: LiveData<NextUpState> = _state
+	private val userPreferences by inject(UserPreferences::class.java)
```
nielsvanvelzen, line 32: Move to the constructor instead and update di/AppModule.kt (add extra `get()`)

nielsvanvelzen (approved):

nielsvanvelzen (approved):

PR #851, opened 2021-05-02
Include specials when queuing subsequent episodes

linetrimmer:
**Changes**

- Include specials when queuing subsequent episodes if the server is set to display specials within the seasons they aired in
- Continues playing the next season after finishing the last episode of a season

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/apiclient/PlaybackHelper.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/apiclient/PlaybackHelper.java
@@ -34,6 +34,8 @@ import timber.log.Timber;
 import static org.koin.java.KoinJavaComponent.get;
 
 public class PlaybackHelper {
+    private static final int limit = 150; // guard against too many items - account for episodes shorter than 5 min
+
     public static void getItemsToPlay(final BaseItemDto mainItem, boolean allowIntros, final boolean shuffle, final Response<List<BaseItemDto>> outerResponse) {
         final List<BaseItemDto> items = new ArrayList<>();
         ItemQuery query = new ItemQuery();
@@ -66,7 +68,6 @@ public class PlaybackHelper {
                                     // TODO: Finding the main item should be possible in the query using StartItemId, but it is not currently supported.
                                     // With StartItemId added, the limit could also be included in the query.
                                     boolean foundMainItem = false;
-                                    final int limit = 150; // guard against too many items - account for episodes shorter than 5 min
                                     int numAdded = 0;
                                     for (BaseItemDto item : response.getItems()) {
                                         if (foundMainItem) {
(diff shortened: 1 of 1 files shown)


ManuLinares:
Interesting, gonna test this later.

nielsvanvelzen (commented):
Everything seems to be working as expected. One thing I'm worried about though is that the next up screen does not show the season name. This can be confusing when going from a season 1 -> season 2.

linetrimmer:
It could be something like 
S1E9. The Title Titles
S3E19-20. The Title Returns
Special - The Titling

I can add these to the next up screen PR if you think they're good

app/src/main/java/org/jellyfin/androidtv/util/apiclient/PlaybackHelper.java
```diff
@@ -58,27 +59,29 @@ public static void getItemsToPlay(final BaseItemDto mainItem, boolean allowIntro
...
-                                        if (item.getIndexNumber() > mainItem.getIndexNumber()) {
+                                        if (foundMainItem) {
                                             if (!LocationType.Virtual.equals(item.getLocationType())) {
                                                 items.add(item);
                                             } else {
                                                 //stop adding when we hit a missing one
                                                 break;
                                             }
+                                        } else if (item.getId() != null && item.getId().equals(mainItem.getId())) {
+                                            foundMainItem = true;
```
thornbill, line 76: This _should_ be doable with the api by setting the `StartItemId` but it seems that is not supported in the old apiclient. It maybe a good idea to add a TODO or FIXME comment here about it though.

thornbill (changes requested):

linetrimmer:
Added the TODO for StartItemId, a check for the current user being null, and a limit for how many episodes get queued.

Went with "S{x}:E{x} - Title" and "Special - Title" for the next up screen in #850. Even without queuing subsequent seasons, it can be confusing when playing episodes by Shuffle All since the queued episodes can jump from season to season.

app/src/main/java/org/jellyfin/androidtv/util/apiclient/PlaybackHelper.java
```diff
@@ -63,13 +63,18 @@ public static void getItemsToPlay(final BaseItemDto mainItem, boolean allowIntro
                             @Override
                             public void onResponse(ItemsResult response) {
                                 if (response.getTotalRecordCount() > 0) {
+                                    // TODO: Finding the main item should be possible in the query using StartItemId, but it is not currently supported.
+                                    // With StartItemId added, the limit could also be included in the query.
                                     boolean foundMainItem = false;
+                                    final int limit = 150; // guard against too many items - account for episodes shorter than 5 min
```
thornbill, line 69: This limit should probably be set as a class level constant.

thornbill (changes requested):

thornbill (approved):

PR #853, opened 2021-05-04
Refactor StartupActivity with Kotlin extensions and new permission contracts

nielsvanvelzen:
I'm still figuring out why the app is sometimes on the splashscreen when my tv is off for multiple hours. It looks like the session in the sessionrepository is emtpy but that shouldn't be possible... 

**Changes**

This PR modernizes the StartupActivity a bit by using the new `registerForActivityResult` to check for the network permissions and changing a bunch of things to Kotlin DSL etc. behavior should be the same as before.

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/startup/StartupActivity.kt
```diff
@@ -92,64 +93,51 @@ class StartupActivity : FragmentActivity() {
...
-					override fun onError(exception: Exception) {
-						// go straight into last connection
-						val intent = Intent(application, MainActivity::class.java)
-						startActivity(intent)
-						finish()
-					}
-				})
+				if (item != null) {
+					ItemLauncher.launchUserView(item, this, true)
+					finish()
```
nielsvanvelzen, line 107: I want to use `finishAfterTransition` here but the itemlauncher does async magic so that's not possible right now.

thornbill (approved):
:+1: looks good as seems to be working fine!

app/src/main/java/org/jellyfin/androidtv/ui/startup/StartupActivity.kt
```diff
@@ -18,68 +20,67 @@ import org.jellyfin.androidtv.ui.browsing.MainActivity
...
 		var isLoaded = false
+
 		sessionRepository.currentSession.observe(this) { session ->
 			if (session != null) {
 				Timber.i("Found a session in the session repository, waiting for the currentUser in the application class.")
 
-				supportFragmentManager.beginTransaction()
-					.replace(R.id.content_view, SplashFragment())
-					.commit()
+				showSplash()
```
koying, line 76: @nielsvanvelzen  Not sure if this `showsplash` is actually necessary, but it breaks the functionality of `EXTRA_HIDE_SPLASH`

app/src/main/java/org/jellyfin/androidtv/ui/startup/StartupActivity.kt
```diff
@@ -18,68 +20,67 @@ import org.jellyfin.androidtv.ui.browsing.MainActivity
...
+		networkPermissionsRequester.launch(arrayOf(Manifest.permission.INTERNET, Manifest.permission.ACCESS_NETWORK_STATE))
 	}
 
-	private fun start() {
+	private fun observeSession() {
 		var isLoaded = false
+
 		sessionRepository.currentSession.observe(this) { session ->
 			if (session != null) {
 				Timber.i("Found a session in the session repository, waiting for the currentUser in the application class.")
```
koying, line 74: That log should be debug, at best ,I guess.
You get it every time you launch the app (and you're not actually waiting for anything, unless mistaken)

app/src/main/java/org/jellyfin/androidtv/ui/startup/StartupActivity.kt
nielsvanvelzen: The extra is used to disable the initial splash screen when starting the activity. If we're loading a session it should still show up so this one is necessary.

app/src/main/java/org/jellyfin/androidtv/ui/startup/StartupActivity.kt
nielsvanvelzen: We're waiting for the UserDto to be set in TvApp after this log message, it sometimes fails (although it shouldn't) so that's why the log message is info.

app/src/main/java/org/jellyfin/androidtv/ui/startup/StartupActivity.kt
koying: Obviously a bit nebulous to me ;)
We already have a session because `session != null`, haven't we?

nielsvanvelzen: Yes but we're still waiting for the apibinder to retrieve user info, which can take a few seconds depending on the network.

nielsvanvelzen: Sure for both you and me it is pretty much instant, but this is the first api call to the server when autologin is enabled and some people have weird setups that warm up the server on the first connection, we also need to account for server timeouts here, which is set to 6 seconds by default. Not showing anything in this timeframe is bad.

PR #855, opened 2021-05-04
Update README to 2021

nielsvanvelzen:
**Changes**

Update the README to be more like the jellyfin-android readme.

- Removed `utm_source` parameters from urls
- `alt` before `src`
- Update jellyfin.media to jellyfin.org
- Build process (copied from jf-android)

**Issues**

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <h1 align="center">Jellyfin Android TV</h1>
-<h3 align="center">Part of the <a href="https://jellyfin.media">Jellyfin Project</a></h3>
+<h3 align="center">Part of the <a href="https://jellyfin.org">Jellyfin Project</a></h3>
 
 ---
 
@@ -12,8 +12,8 @@
 <a href="https://github.com/jellyfin/jellyfin-androidtv/releases">
 <img alt="Current Release" src="https://img.shields.io/github/release/jellyfin/jellyfin-androidtv.svg"/>
 </a>
-<a href="https://translate.jellyfin.org/projects/jellyfin/jellyfin-androidtv/?utm_source=widget">
-<img src="https://translate.jellyfin.org/widgets/jellyfin/-/jellyfin-androidtv/svg-badge.svg" alt="Translation status" />
+<a href="https://translate.jellyfin.org/projects/jellyfin/jellyfin-androidtv/">
+<img alt="Translation Status" src="https://translate.jellyfin.org/widgets/jellyfin/-/jellyfin-androidtv/svg-badge.svg"/>
 </a>
 <br/>
 <a href="https://opencollective.com/jellyfin">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,18 +30,53 @@
...
 <br/>
 <a href="https://repo.jellyfin.org/releases/client/androidtv/">Download archive</a>
 </p>
 
-Jellyfin Android TV is a Jellyfin client for Android TV, Nvidia Shield, and Amazon Fire TV devices. We welcome all contributions and pull requests! If you have a larger feature in mind please open an issue so we can discuss the implementation before you start. Translations can be improved very easily from our <a href="https://translate.jellyfin.org/projects/jellyfin/jellyfin-androidtv">Weblate</a> instance. Look through the following graphic to see if your native language could use some work!
+Jellyfin Android TV is a Jellyfin client for Android TV, Nvidia Shield, and Amazon Fire TV devices.
+We welcome all contributions and pull requests! If you have a larger feature in mind please open an
+issue so we can discuss the implementation before you start. 
+
+## Translating
```
dkanada, line 46:
```suggestion
## Translating

```

README.md
```diff
@@ -30,18 +30,54 @@
...
+
+<a href="https://translate.jellyfin.org/engage/jellyfin/">
+<img alt="Detailed Translation Status" src="https://translate.jellyfin.org/widgets/jellyfin/-/jellyfin-androidtv/multi-auto.svg"/>
 </a>
+
+## Build Process
+
+### Dependencies
+
+- Android SDK
```
thornbill, line 60: What about the JDK?

nielsvanvelzen: JDK is a dependency of the Android SDK

thornbill: Right would it be better to be explicit here? I know it would be bundled in an Android Studio installation, but afaik that is not the case if you only install the SDK right? I'm actually having a difficult time finding documentation on installing _only_ the SDK and not AS. :smile:

nielsvanvelzen: Android Studio is the SDK, it's not that easy to install the SDK without Android Studio these days.

thornbill: Then we should probably just say Android Studio... because you _could_ install the command line tools and install SDKs manually or who knows what other weird path.

README.md
```diff
@@ -30,18 +30,53 @@
...
+
+   ```sh
+   git clone https://github.com/jellyfin/jellyfin-androidtv.git
+   cd jellyfin-androidtv
+   ```
+
+2. Open the project in Android Studio and run it from there or build an APK directly through Gradle:
+
+   ```sh
+   ./gradlew assembleDebug
```
ManuLinares, line 74: shouldn't this be 

ANDROID_SDK_ROOT={your_sdkPath} ./gradlew assembleDebug
```suggestion
   ANDROID_SDK_ROOT={AndroidSDK_path} ./gradlew assembleDebug
```

ManuLinares (changes requested):

README.md
nielsvanvelzen: Agreed, changed it

README.md
nielsvanvelzen: I added a note that you need to add the Android SDK to your PATH env.

thornbill (approved):

PR #870, opened 2021-05-10
Fix null media url crash

thornbill:
**Changes**
Fixes the app crashing when the media url is null. Specifically addresses an issue where this condition is hit on 10.7.5 when transcoding and remuxing is disabled.

**Issues**
N/A


nielsvanvelzen (approved):

nielsvanvelzen:
Looks good, do we need to backport this to the 0.11.z branch?

PR #874, opened 2021-05-11
Fix Fire OS focus issues on login

thornbill:
**Changes**
* Prevents the on-screen keyboard from being activated automatically on Fire OS
* Fixes navigating down on alert fragments
* Ensures focus is cleared when confirm is selected
* Fixes username field receiving focus when disabled

**Issues**
N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/startup/AddServerAlertFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/startup/AddServerAlertFragment.kt
@@ -42,6 +42,7 @@ class AddServerAlertFragment : AlertFragment() {
 		with(binding.address) {
 			onFocusChangeListener = KeyboardFocusChangeListener()
 			nextFocusForwardId = parentBinding.confirm.id
+			nextFocusDownId = parentBinding.confirm.id
 			imeOptions = EditorInfo.IME_ACTION_DONE
 			setOnEditorActionListener { _, actionId, _ ->
 				if (actionId == EditorInfo.IME_ACTION_DONE) parentBinding.confirm.performClick()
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/startup/AddServerAlertFragment.kt
```diff
@@ -42,9 +42,14 @@ class AddServerAlertFragment : AlertFragment() {
...
 			onFocusChangeListener = KeyboardFocusChangeListener()
 			nextFocusForwardId = parentBinding.confirm.id
+			nextFocusDownId = parentBinding.confirm.id
 			imeOptions = EditorInfo.IME_ACTION_DONE
 			setOnEditorActionListener { _, actionId, _ ->
-				if (actionId == EditorInfo.IME_ACTION_DONE) parentBinding.confirm.performClick()
+				if (actionId == EditorInfo.IME_ACTION_DONE) {
+					clearFocus()
+					parentBinding.confirm.performClick()
+					return@setOnEditorActionListener true
```
nielsvanvelzen, line 51: The return statement is unnecessary, we return the returned value of the `performClick()` (also a boolean)

```suggestion
```

app/src/main/java/org/jellyfin/androidtv/ui/startup/AddServerAlertFragment.kt
```diff
@@ -42,9 +42,14 @@ class AddServerAlertFragment : AlertFragment() {
...
+				}
 				else false
```
nielsvanvelzen, line 53:
```suggestion
				} else false
```
or
```suggestion
				} else {
					false
				}
```

app/src/main/java/org/jellyfin/androidtv/ui/startup/AddServerAlertFragment.kt
nielsvanvelzen: The performClick() function always returns true because it contains an eventlistener. But returning it explicitly might be easier to understand. In that case you should use:

```suggestion
					true
```

Because the last statement of this function is the returned value

thornbill: Fixed now... I'm still not used to Kotlin's implicit return stuff. :smile:

nielsvanvelzen (approved):

Issue #732, opened 2021-02-16
Text hard to read on app home screen


viggy96:
**Describe the bug**
Text associated with each card on screen is very hard to read with the background image enabled.

**To Reproduce**

1. Open Jellyfin app
2. Profit

**Expected behavior**
All text should be clear to read.

**Screenshots**
![image](https://user-images.githubusercontent.com/9837626/108127179-dc56e200-7078-11eb-89f6-6cee7348bda8.png)

**System (please complete the following information):**
 - Android TV or Fire TV version: Android 9
 - Device manufacturer: Sony
 - Device model: X900F
 - Jellyfin server version: 10.6.4

**Additional context**
The darkening of the background image in #703 helps _slightly_ but I think the text needs to be adjusted as well to improve readability.

viggy96:
I think this is obvious in the above screenshot, but this issue also applies to the clock and username text in the top right corner of the app. That stuff is also hard to read when its over a bright image for a show/movie.

viggy96:
@thornbill Should there be a new bug for time/username being difficult to read,? I mentioned it in my last comment.
A drop shadow would help with that.

thornbill:
@viggy96 I don't think a separate issue is necessary. I changed the project status to be "In Progress" since that part still needs to be worked on.

viggy96:
Sorry to open this again, but this applies again to the episode title text during playback. While you guys could do a drop shadow again for it, maybe it would be better to add a dark gradient at the top of the screen, and possibly also for the bottom controls area?

PR #877, opened 2021-05-12
Remove text dimming on inactive cards

thornbill:
**Changes**
Removes the dimming on card text when not focused.

**Issues**
#732


nielsvanvelzen (commented):
Can you remove `lb_basic_card_title_text_color_inactive` from colors.xml? It's unused now.

nielsvanvelzen (approved):

PR #878, opened 2021-05-12
Move worker initialization and auto bitrate test to after session is created

nielsvanvelzen:
**Changes**

The app crashed on start because the WorkManager was not initialized in the applications onCreate function. This might have caused the Koin issues before,....
Move worker initialization and auto bitrate test to after session is created.

I also moved the autobitrate thingy, it takes around 250ms for me. But when I look at the references it looks like it isn't even used. the only reference that uses it is here:

https://github.com/jellyfin/jellyfin-androidtv/blob/71e52b1bb25fa3c57daddca9bc73ce817fab2d44/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java#L566-L567

I'm considering removing the autobitrate stuff for now, it didn't seem to work anyway.

**Issues**


thornbill (approved):
:rocket:

PR #889, opened 2021-05-14
Set subtitle background to Transparent

ManuLinares:
Cosmetic.

Previously in **v0.11.5**, default was transparent but changed in Master.

* This sets back to transparent subtitle background.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
@@ -1494,7 +1494,7 @@ public class CustomPlaybackOverlayFragment extends Fragment implements IPlayback
 
             SpannableString span = new SpannableString(TextUtilsKt.toHtmlSpanned(htmlText));
             span.setSpan(new ForegroundColorSpan(Color.WHITE), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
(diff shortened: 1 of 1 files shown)


thornbill:
I added the background color because it helps with legibility imo. Most other services default to having a similar style background for their subtitles also. That said I wouldn't be opposed to having it be an option.

ManuLinares:
> ...That said I wouldn't be opposed to having it be an option.
Same.

Still, until an option menu is added, I prefer transparent for legibility imho, and reverts visually to *official* release.

_PD: I also set in my custom build, layout_marginBottom to 10dp in the **[vlc_player_interface.xml](https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/res/layout/vlc_player_interface.xml#L53)**. I asked around, anybody that uses subtitles doesn't want/needs the subs on top of the video. I also lower textSize._

nielsvanvelzen:
It might be possible to change the subtitle style based on the subtitle style preferences from the web client, depends on if they are stored server side or not. But until there is an option in either the app or we can use those from jf-web the current styling is what we're going with.

thornbill:
Closing based on the feedback above.

ivooc:
> It might be possible to change the subtitle style based on the subtitle style preferences from the web client, depends on if they are stored server side or not. But until there is an option in either the app or we can use those from jf-web the current styling is what we're going with.

From my experience with the jellyfin clients (web desktop, mobile and android app) all subtitle preferences are stored on client side (you have to setup your preferences everytime you use a new client, new browser, etc), so I don't see why the androidtv app can't have it's own subtitles preferences too.

PR #891, opened 2021-05-16
FIX: [vlc] "setAudioOutput" disables passthrough

### LINKED ISSUE #786: 5.1 passthrough issues with libvlc
**Describe the bug**
On my Android TV the 5.1 audio passthrough does not work with libvlc, instesd only stereo is output. When selecting exoplayer, 5.1 passthrough works without issues. 

Also with kodi app it works perfectly. 

**To Reproduce**
Set libvlc as player in the options and play a mkv video with 5.1 ac3 sound.

**Expected behavior**
5.1 Sound should be passend through, not stereo.

koying:
Fix VLC player passthrough

**Changes**
As per the vlc javadoc, using `setAudioOutput`  "disable the encoding detection", and that seems to intefere with PT.
As  `android_audiotrack` is the default value (and I don't think you can switch the output during a playback), plainly removing the lines fixes the issue.

**Issues**
Fixes #786

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
@@ -234,7 +234,6 @@ public class MediaManager {
 
                 mVlcPlayer = new org.videolan.libvlc.MediaPlayer(mLibVLC);
                 if(!Utils.downMixAudio()) {
-                    mVlcPlayer.setAudioOutput("android_audiotrack");
                     mVlcPlayer.setAudioDigitalOutputEnabled(true);
                 } else {
                     mVlcPlayer.setAudioOutput("opensles_android");
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
@@ -422,9 +422,7 @@ public class VideoManager implements IVLCVout.OnNewVideoLayoutListener {
     }
 
     private void setVlcAudioOptions() {
-
         if(!Utils.downMixAudio()) {
-            mVlcPlayer.setAudioOutput("android_audiotrack");
             mVlcPlayer.setAudioDigitalOutputEnabled(true);
         } else {
             setCompatibleAudio();


nielsvanvelzen:
This PR would revert #524

koying:
Not really. There was some refactor, but before and after that PR, `setAudioOutput` was used, and it's definitely not needed.

Having been the Kodi Android maintainer for some years, I suspect the key to having TruHD and DTS/X working is the number of channels.

nielsvanvelzen:
You're right - I didn't read it properly and thought it was added in that PR. Reading the documentation from libvlc about the `setAudioOutput` function it makes sense to remove the call when using `android_audiotrack`.

thornbill (approved):
:+1: Makes sense to me.

svemonix:
I'm not a professional developer but isn't this fix going to force the use of passthrough with the direct output setting? If so, what about users who want/need a multichannel PCM output?

Also, will the libVLC passthrough determine properly the actual capabilities? What happens if the device/receiver doesn't support all codecs (e.g. lossless)? What happens if the HDMI connection has limitations (e.g. ARC connection supporting only AC3 & DTS)? If the client assumes everything can be passed through it could prevent some necessary audio transcoding.

koying:
Shouldn't make.any difference re multi pcm.
VLC will detect the audio capabilities of your device.

If you want to force multi pcm while you have HDMI with, eg, ac3 support, that's another story.

svemonix:
How do you mean multi PCM doesn't make any difference?

Some use cases:
1. The device/receiver doesn't support a specific codec (e.g. TrueHD, DTS-HD MA) => having a multi PCM output would effectively allow lossless audio playback as well as direct play on HW that couldn't otherwise.
2. Multichannel AAC cannot be passed through => having multi PCM would solve the issue. Side note: currently, the Jellyfin server doesn't transcode multi-channel AAC to AC3 so it would typically result in a stereo output.
3. AFAIK, the Jellyfin server assumes the libVLC client can direct play everything but it won't always be the case with a passthrough. So if you can only passthrough AC3 or DTS, there will be a problem to playback EAC3, TrueHD and DTS-HD MA since no transcoding will occur on the server side.

In my experience, the VLC app doesn't detect passthrough capabilities very well. VLC passthrough doesn't work at all on my TV (2.0 PCM not matter the audio source) while it works with exoplayer. It "works" with mixed results on another android TV device connected directly to the receiver.

koying:
I meant the _change_ shouldn't make any difference ;)

Does the vlc app works for you, with passthrough enabled? If so, it will behave the same in Jellyfin.

svemonix:
Gotcha! 

Then in a nutshell:
1. There has never been an intention to support multi PCM output - Actually, I thought that was one of the reasons for libVLC vs exoplayer.
2. The output setting called "direct" should rather be called "passthrough". It would be much more self explanatory to users and consistent with other apps.
3. This PR fixes the libVLC passthrough itself but will never leverage the Jellyfin server audio transcoding. Since the Android-VLC profile is predefined regardless of actual passthrough capabilities, playback will be broken in many cases. This defeats the purpose of having a media server in the first place.

I see 2 options going forward:
1. Passing audio through in which case the libVLC implementation would need the same kind of dynamic profile fine tuning exoplayer has. This would ensure proper server integration and correct playback in all cases.
2. Outputting multi PCM which would make more sense with the rest of the current implementation. Playback would be pretty independent from the server like Jellyfin for Kodi. Easier to implement but not supported by all devices and prevents Atmos/X.

I just wanted to raise this point as I've seen on reddit that the plan could be to only rely on libVLC in the future. For now, the libVLC option clearly can't suit everyone.

PR #897, opened 2021-05-17
CHG: Be smarter when switching reso

koying:
**Changes**
Currently, the switch resolution logic is a tad dumb.
By instance, it ignores resolution, which lead to, e.g. a 4K movie being played in 720p because it happens to be the first reso matching the refresh rate.

This:
- Takes fractional rates into account (e.g. 23.97)
- Ignores resolution lower than the media one
- Takes into account 2:3 rates
- Prefers higher rates
- Rates being equal, prefers resolution closest to the media

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -244,16 +244,32 @@ public class PlaybackController {
     }
 
     @TargetApi(23)
-    private Display.Mode findBestDisplayMode(Float refreshRate) {
-        if (mDisplayModes == null || refreshRate == null) return null;
+    private Display.Mode findBestDisplayMode(MediaStream videoStream) {
+        if (mDisplayModes == null || videoStream.getRealFrameRate() == null) return null;
 
-        int sourceRate = Math.round(refreshRate);
-        for (Display.Mode mode : mDisplayModes){
-            int rate = Math.round(mode.getRefreshRate());
-            if (rate == sourceRate || rate == sourceRate * 2) return mode;
+        int curWeight = 0;
+        Display.Mode bestMode = null;
+        int sourceRate = Math.round(videoStream.getRealFrameRate() * 100);
+        for (Display.Mode mode : mDisplayModes) {
+            // Skip unwanted display modes
+            if (mode.getPhysicalWidth() < 1280 || mode.getPhysicalHeight() < 720)  // Skip non-HD
+                continue;
+
+            if (mode.getPhysicalWidth() < videoStream.getWidth() || mode.getPhysicalHeight() < videoStream.getHeight())  // Disallow reso downgrade
+                continue;
+
+            int rate = Math.round(mode.getRefreshRate() * 100);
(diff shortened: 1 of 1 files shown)


koying:
Checks failed because of  `NPath complexity of 219, current threshold is 200`.
Yep, it's a semi-complex function, for sure. But it's a semi-complex functionality as well ;)

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -244,16 +244,32 @@ private void getDisplayModes() {
...
+    private Display.Mode findBestDisplayMode(MediaStream videoStream) {
+        if (mDisplayModes == null || videoStream.getRealFrameRate() == null) return null;
+
+        int curWeight = 0;
+        Display.Mode bestMode = null;
+        int sourceRate = Math.round(videoStream.getRealFrameRate() * 100);
+        for (Display.Mode mode : mDisplayModes) {
+            // Skip unwanted display modes
+            if (mode.getPhysicalWidth() < 1280 || mode.getPhysicalHeight() < 720)  // Skip non-HD
+                continue;
```
nielsvanvelzen, line 256: Does it make sense to disallow low resolution? What if my video is 480p? The next if statement already disallows display resolutions lower then the video resolution.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -244,16 +244,32 @@ private void getDisplayModes() {
...
-        int sourceRate = Math.round(refreshRate);
-        for (Display.Mode mode : mDisplayModes){
-            int rate = Math.round(mode.getRefreshRate());
-            if (rate == sourceRate || rate == sourceRate * 2) return mode;
+    private Display.Mode findBestDisplayMode(MediaStream videoStream) {
+        if (mDisplayModes == null || videoStream.getRealFrameRate() == null) return null;
+
+        int curWeight = 0;
+        Display.Mode bestMode = null;
+        int sourceRate = Math.round(videoStream.getRealFrameRate() * 100);
```
nielsvanvelzen, line 252: Why multiplied by 100?

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
koying: Not per-se mandatory, but in my experience, things get murky when going to non-HD resolutions, so this is just to keep it safe.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
koying: I go to Int to avoid float comparison issues.
And I multiply by 100 to get, e.g., 23.97 (24 * 1000 / 1001, common movie frame rate)-> 2397.

nielsvanvelzen (approved):

thornbill:
This looks fine and seems to work fine in an emulator, but I want to test on a 4k Fire Stick just to be thorough. Apologies for the delay.

thornbill (approved):

PR #899, opened 2021-05-17
Refactor player profiles

thornbill:
**Changes**
* Moves device profile classes to a new `profile` package
* Moves each player profile to an individual class
* Refactors `ProfileHelper` in Kotlin

**Issues**
N/A


app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+	private fun findTranscodingProfile(deviceProfile: DeviceProfile, container: String) = (
+		deviceProfile.transcodingProfiles.find { it.container == container }
+	)
+
+	internal fun subtitleProfile(format: String, method: SubtitleDeliveryMethod) = (
+		SubtitleProfile().apply {
+			this.format = format
+			this.method = method
+		}
+	)
```
nielsvanvelzen, line 80:
```suggestion
	internal fun subtitleProfile(
		format: String,
		method: SubtitleDeliveryMethod,
	) = SubtitleProfile().apply {
		this.format = format
		this.method = method
	}
```

the new lines for parameters is optional but I find it more readable to use new lines when it becomes long.

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+				"${CodecTypes.AC3},${mkvProfile.audioCodec}"
+			} else {
+				"${mkvProfile.audioCodec},${CodecTypes.AC3}"
+			}
+		}
+	}
+
+	private fun findTranscodingProfile(deviceProfile: DeviceProfile, container: String) = (
+		deviceProfile.transcodingProfiles.find { it.container == container }
+	)
```
nielsvanvelzen, line 73:
```suggestion
	private fun findTranscodingProfile(
		deviceProfile: DeviceProfile,
		container: String,
	) = deviceProfile.transcodingProfiles.find { it.container == container }
```

I reviewed this file from bottom to top, so see my next comment.

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+				ProfileCondition(
+					ProfileConditionType.NotEquals,
+					ProfileConditionValue.VideoProfile,
+					"none"
+				)
+			)
+		}
+	}
+
+	@JvmStatic
```
nielsvanvelzen, line 56: I don't think this is required since ProfileHelper is an object?

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+		if (Utils.downMixAudio()) return
+
+		val mkvProfile = findTranscodingProfile(profile, ContainerTypes.MKV)
+		if (mkvProfile != null) {
+			Timber.i("*** Adding AC3 as supported transcoded audio")
+			mkvProfile.audioCodec = if (primary) {
+				"${CodecTypes.AC3},${mkvProfile.audioCodec}"
+			} else {
+				"${mkvProfile.audioCodec},${CodecTypes.AC3}"
+			}
```
nielsvanvelzen, line 67:
```suggestion
			mkvProfile.audioCodec = when(primary) {
				true -> "${CodecTypes.AC3},${mkvProfile.audioCodec}"
				false ->"${mkvProfile.audioCodec},${CodecTypes.AC3}"
			}
```

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+			)
+		}
+	}
+
+	@JvmStatic
+	fun addAc3Streaming(profile: DeviceProfile, primary: Boolean) {
+		if (Utils.downMixAudio()) return
+
+		val mkvProfile = findTranscodingProfile(profile, ContainerTypes.MKV)
+		if (mkvProfile != null) {
```
nielsvanvelzen, line 61:
```suggestion
		val mkvProfile = findTranscodingProfile(profile, ContainerTypes.MKV) ?: return
```

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+import org.jellyfin.apiclient.model.dlna.ProfileCondition
+import org.jellyfin.apiclient.model.dlna.ProfileConditionType
+import org.jellyfin.apiclient.model.dlna.ProfileConditionValue
+import org.jellyfin.apiclient.model.dlna.SubtitleDeliveryMethod
+import org.jellyfin.apiclient.model.dlna.SubtitleProfile
+import org.jellyfin.apiclient.model.dlna.TranscodingProfile
+import timber.log.Timber
+
+object ProfileHelper {
+	private val MediaTest = MediaCodecCapabilitiesTest()
```
nielsvanvelzen, line 18:
```suggestion
	private val MediaTest by lazy { MediaCodecCapabilitiesTest() }
```

So it is initiated on first use

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+
+	val deviceHevcCodecProfile = CodecProfile().apply {
```
nielsvanvelzen, line 20:
```suggestion
	val deviceHevcCodecProfile by lazy {
		CodecProfile().apply {
```

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+		type = CodecType.Video
+		codec = CodecTypes.HEVC
+
+		conditions = if (!MediaTest.supportsHevc()) {
```
nielsvanvelzen, line 24:
```suggestion
		conditions = when {
			!MediaTest.supportsHevc() -> {
```

add all cases

nielsvanvelzen: (the else statement is `else -> {}`

nielsvanvelzen (changes requested):
I did not review ExoPlayerProfile, ExternalPlayerProfile and LibVlcProfile yet, but those are probably fine

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
thornbill: I didn't think so either, but it didn't compile without it :man_shrugging:

app/src/main/java/org/jellyfin/androidtv/util/profile/LibVlcProfile.kt
```diff
@@ -0,0 +1,146 @@
...
+			deviceHevcCodecProfile,
+			// H264 profile
+			CodecProfile().apply {
+				type = CodecType.Video
+				codec = CodecTypes.H264
+				conditions = arrayOf(
+					ProfileCondition(
+						ProfileConditionType.EqualsAny,
+						ProfileConditionValue.VideoProfile,
+						"high|main|baseline|constrained baseline"
```
nielsvanvelzen, line 104: `arrayOf().joinToString()`

app/src/main/java/org/jellyfin/androidtv/util/profile/LibVlcProfile.kt
```diff
@@ -0,0 +1,146 @@
...
+					),
+					ProfileCondition(
+						ProfileConditionType.LessThanEqual,
+						ProfileConditionValue.VideoLevel,
+						if (DeviceUtils.isFireTvStick()) "41" else "51"
```
nielsvanvelzen, line 109: Magic numbers - what do they mean? Probably best to create constants for those.

app/src/main/java/org/jellyfin/androidtv/util/profile/LibVlcProfile.kt
```diff
@@ -0,0 +1,146 @@
...
+					ContainerTypes.OGG,
+					ContainerTypes.OGA,
+					ContainerTypes.WEBMA,
+					CodecTypes.APE
+				).joinToString(",")
+			},
+			// Photo direct play
+			DirectPlayProfile().apply {
+				type = DlnaProfileType.Photo
+				container = "jpg,jpeg,png,gif"
```
nielsvanvelzen, line 89: `arrayOf().joinToString()`

also could probably add `webp` since that works on android too, do we even use libvlc for photos though? lol

app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
```diff
@@ -0,0 +1,189 @@
...
+				conditions = arrayOf(
+					ProfileCondition(
+						ProfileConditionType.EqualsAny,
+						ProfileConditionValue.VideoProfile,
+						"high|main|baseline|constrained baseline"
+					),
+					ProfileCondition(
+						ProfileConditionType.LessThanEqual,
+						ProfileConditionValue.VideoLevel,
+						if (DeviceUtils.isFireTvStick()) "41" else "51"
```
nielsvanvelzen, line 118: Same values used in vlc profile, might be useful to share it using a constant

nielsvanvelzen: actually, I think the complete profile is the same

app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
```diff
@@ -0,0 +1,189 @@
...
+		codecProfiles = arrayOf(
+			// H264 profile
+			CodecProfile().apply {
+				type = CodecType.Video
+				codec = CodecTypes.H264
+				conditions = arrayOf(
+					ProfileCondition(
+						ProfileConditionType.EqualsAny,
+						ProfileConditionValue.VideoProfile,
+						"high|main|baseline|constrained baseline"
```
nielsvanvelzen, line 113: `arrayOf().joinToString()`

app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
```diff
@@ -0,0 +1,189 @@
...
+					ContainerTypes.OGA,
+					ContainerTypes.WEBMA,
+					CodecTypes.APE,
+					CodecTypes.OPUS
+				).joinToString(",")
+			},
+			// Photo direct play
+			DirectPlayProfile().apply {
+				type = DlnaProfileType.Photo
+				container = "jpg,jpeg,png,gif"
```
nielsvanvelzen, line 100: `arrayOf().joinToString()`

add webp bla bla do we show photos? bla bla see libvlc comment on the same part

app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
```diff
@@ -0,0 +1,189 @@
...
+
+class ExoPlayerProfile(
+	isLiveTV: Boolean = false,
+	isLiveTVDirectPlayEnabled: Boolean = false,
+	isDtsEnabled: Boolean = true
+) : BaseProfile() {
+	init {
+		name = "AndroidTV-ExoPlayer"
+
+		directPlayProfiles = listOfNotNull(
```
nielsvanvelzen, line 26: I think using `buildList {}` might be more readable here. Could also split the different profiles to functions/vals to make it easier.

app/src/main/java/org/jellyfin/androidtv/util/profile/LibVlcProfile.kt
thornbill: Is it worth using `joinToString()` here given the space in there? I feel like that would be harder to read.

app/src/main/java/org/jellyfin/androidtv/util/profile/LibVlcProfile.kt
thornbill: I have no idea... just copied from the current profiles :rofl:

app/src/main/java/org/jellyfin/androidtv/util/profile/LibVlcProfile.kt
nielsvanvelzen: I'd say it would be easier to read where the split happens, because "constrained baseline" is a single value, which is not obvious at first glance.

app/src/main/java/org/jellyfin/androidtv/util/profile/LibVlcProfile.kt
nielsvanvelzen: Fair, probably best to keep it like it is then :p

app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
thornbill: The ref frames part is handled differently, but I did move the shared parts out to ProfileHelper

app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
```diff
@@ -0,0 +1,171 @@
...
+@OptIn(ExperimentalStdlibApi::class)
+class ExoPlayerProfile(
+	isLiveTV: Boolean = false,
+	isLiveTVDirectPlayEnabled: Boolean = false,
+	isDtsEnabled: Boolean = true
+) : BaseProfile() {
+	init {
+		name = "AndroidTV-ExoPlayer"
+
+		directPlayProfiles = buildList() {
```
nielsvanvelzen, line 34:
```suggestion
		directPlayProfiles = buildList {
```

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -0,0 +1,145 @@
...
+		}
+
+	fun maxAudioChannelsCodecProfile(channels: Int) = CodecProfile()
+		.apply {
+			type = CodecType.VideoAudio
+			conditions = arrayOf(
+				ProfileCondition(
+					ProfileConditionType.LessThanEqual,
+					ProfileConditionValue.AudioChannels,
+					"$channels"
```
nielsvanvelzen, line 115:
```suggestion
					channels.toString()
```

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

PR #905, opened 2021-05-19
Decode blur hashes in background thread

thornbill:
**Changes**

* Moves blur hash decoding to a background thread via `BlurHashService` and an extension to Glide
* Adds a background color to card images that is shown while the blur hash is being decoded

> This was heavily inspired by this MIT licensed library, but adapted to meet our specific needs and technology stack https://github.com/KingsMentor/BlurHashExt

**Issues**
N/A


app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
```diff
@@ -428,7 +437,7 @@ public void onBindViewHolder(Presenter.ViewHolder viewHolder, Object item) {
             }
         }
 
-        BitmapDrawable blurHashDrawable = null;
+        String blurHash = "";
```
nielsvanvelzen, line 440: Wouldn't it make more sense to use a null value here and allow the null value in the `blurHashPlaceholder` extension function?

app/src/main/java/org/jellyfin/androidtv/data/service/BlurHashService.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import org.jellyfin.androidtv.util.BlurHashDecoder
+
+class BlurHashService(
+	private val context: Context
+) {
+	private val scope = MainScope()
+
+	fun decodeBlurHashDrawable(
```
nielsvanvelzen, line 16: I'd suggest making this function suspended (so no need for the custom `scope`) and pass the viewlifecycle to the `blurHashPlaceholder` function to call the coroutine, this would auto-cancel the loading when the view is destroyed while loading.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
thornbill: Probably. I started down that route initially and then changed to avoid needing null handling on the Kotlin side... I've changed it back now. :smile:

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
```diff
@@ -393,7 +398,7 @@ public ViewHolder onCreateViewHolder(ViewGroup parent) {
         @ColorInt int color = typedValue.data;
         cardView.setBackgroundColor(color);
 
-        return new ViewHolder(cardView);
+        return new ViewHolder(cardView, ViewTreeLifecycleOwner.get(parent));
```
nielsvanvelzen, line 401: Because our custom view is not connected to a parent we can't read the lifecycle from it, fortunately we know the parent  here so we can retrieve it from there.

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
```diff
@@ -344,25 +348,26 @@ protected void updateCardViewImage(@Nullable String url, @Nullable String blurHa
...
                     Glide.with(mCardView.getContext())
-                            .load(mDefaultCardImage)
-                            .diskCacheStrategy(DiskCacheStrategy.RESOURCE)
-                            .into(mCardView.getMainImageView());
+                        .load(mDefaultCardImage)
+                        .diskCacheStrategy(DiskCacheStrategy.RESOURCE)
+                        .into(mCardView.getMainImageView());
                 } else {
-                    GlideUtils.blurHashPlaceholder(
+                    BlurHashUtils.createBlurHashDrawable(
```
nielsvanvelzen, line 355: I created a new function that accepts the lifecycleowner, blurhash, size and a return function. No context or glide object needed now

app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
```diff
@@ -344,25 +348,26 @@ protected void updateCardViewImage(@Nullable String url, @Nullable String blurHa
...
+                        lifecycleOwner,
+                        blurHash,
+                        (aspect > 1) ? (int) Math.round(32 * aspect) : 32,
+                        (aspect >= 1) ? 32 : (int) Math.round(32 / aspect),
+                        bitmap -> {
```
nielsvanvelzen, line 360: When a result is returned (nullable, always called once) we start loading with glide, the bitmap is wrapped in a bitmapdrawable here instead of in the function.

app/src/main/java/org/jellyfin/androidtv/util/BlurHashUtils.kt
```diff
@@ -0,0 +1,27 @@
...
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+fun createBlurHashDrawable(
+	lifecycleOwner: LifecycleOwner,
+	blurHash: String?,
+	width: Int,
+	height: Int,
+	callback: (bitmap: Bitmap?) -> Unit,
+) = lifecycleOwner.lifecycleScope.launch {
```
nielsvanvelzen, line 18: Could really simplify this part now (replaced the GlideExtensions + BlurHashService code)

thornbill:
:+1: lgtm and works well in emulator and on a Fire Stick Gen 2. I can't approve since I am the original author though.

nielsvanvelzen (approved):

PR #907, opened 2021-05-19
Disable AC3 for Fire Stick Gen 1 devices

thornbill:
**Changes**
* Disables support for AC3 for Fire Stick Gen 1 devices as a follow up to #843

**Issues**
N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -198,5 +198,10 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		migration(toVersion = 4) {
 			putInt("libvlc_audio_delay", it.getLong("libvlc_audio_delay", 0).toInt())
 		}
+
+		// Disable AC3 (Dolby Digital) on Fire Stick Gen 1 devices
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -104,7 +105,7 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		/**
 		 * Enable AC3
 		 */
-		var ac3Enabled = Preference.boolean("pref_bitstream_ac3", true)
+		var ac3Enabled = Preference.boolean("pref_bitstream_ac3", !DeviceUtils.isFireTvStickGen1())
```
nielsvanvelzen, line 108: Do we want a migration to automatically set this (once) for all fire tv gen 1 devices?

ManuLinares:
Want me to test it?

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
thornbill: :+1: migration has been added.

nielsvanvelzen (approved):

PR #912, opened 2021-05-21
Add SDK ApiClient and migrate LeanbackChannelWorker

nielsvanvelzen:
**Changes**

- Add KtorClient instances for user and system and apply sessions in SessionRepository
  - This allows the use of the new SDK while keeping the old apiclient working
- Migrate LeanbackChannelWorker to SDK with the system user

Because of an issue in the current SDK beta the `date created` field is not returned from the server. This causes the order of the "next up" row to be incorrect. It's already fixed with jellyfin/jellyfin-sdk-kotlin#262 and confirmed working when using the snapshot release (#913). We can either wait for the SDK developers to release a new beta or just merge this now. The version will definitely be bumped before releasing anyway (SDK will promote to non-beta).

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/auth/SessionRepository.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/auth/SessionRepository.kt
@@ -7,6 +7,7 @@ import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
 import org.jellyfin.androidtv.preference.AuthenticationPreferences
 import org.jellyfin.androidtv.preference.constant.UserSelectBehavior.*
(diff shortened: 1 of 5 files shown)


nielsvanvelzen:
I talked with the amazing SDK team, really nice people, and they published a new beta release. They told me they needed to release that beta version for the mobile app anyway ¯\\_(ツ)_/¯. I've updated the version in #913 so there won't be any merge conflicts.

thornbill (changes requested):
You have a couple line length warnings that should be easy to fix. :smiley:

thornbill (approved):

PR #914, opened 2021-05-22
Rewrite ExpandableTextView to use a popup instead of activity

nielsvanvelzen:
I'm going to use the ExpandableTextView for the new authentication UI when showing the login disclaimer. Unfortunately the implementation was quite messy and used an activity for no real reason. Rewrote it to use an popup instead.

**Changes**
- Renamed all animation styles to start with "WindowAnimation."
- Added new "WindowAnimation.Fade" style
- Reformatted existing animation XML files
- Refactor ExpandableTextView to use a popup
  - Popup content is the old activity layout repurposed
- Removed ExpandedTextActivity
- Added new transparent black color with alpha of 80% 

**Screenshot**

image

This is how it looks in my authentication UI refresh branch. Item details page is the same.

**Issues**


MrChip53:
I think it would look more clean with a "close" button or something similar more like a dialog would have. What do you all think?

nielsvanvelzen:
> I tested this on a Fire Stick on the item details screen and I am unable to exit the popup once it is opened.

You can close it by pressing the back button or pressing the OK button. Do both not work?

MrChip53:
> > I tested this on a Fire Stick on the item details screen and I am unable to exit the popup once it is opened.
> 
> You can close it by pressing the back button or pressing the OK button. Do both not work?
> 
> 
> > I think it would look more clean with a "close" button or something similar more like a dialog would have. What do you all think?
> > 
> > If this possible without too much extra work.
> 
> Where would the button be placed though?

Centered under text. I would need to run this myself to see if it's still a valid idea. The old implementation had two different opacities going on if I remember right and sort of overall looked bad so a button is something I thought of to even it out.

nielsvanvelzen:
current master:

image

pr:

image

MrChip53:
> current master:
> 
> image
> 
> pr:
> 
> image
> 

Ah. No I had it backwards. Yes the old one looks like a fullscreen dialog with no close button. It's odd. I like the new one and I'm sure without a button it fits the android "look and feel" more.

nielsvanvelzen:
image

Something like this?

thornbill:
> > I tested this on a Fire Stick on the item details screen and I am unable to exit the popup once it is opened.
> 
> You can close it by pressing the back button or pressing the OK button. Do both not work?

I tried pushing all the buttons and nothing happened :wink:

nielsvanvelzen:
Can you try it again @thornbill ? I changed the focusing a bit

MrChip53:
> image
> 
> Something like this?

Yeah something like that but I was confused. A button would look better on the old dialog imo. The new one looks fine without a button. Still not sure if a button would be better or not. I think in my head I was picturing a dialog that doesn't opaque the whole screen, only where the text is and has a button inside the opaque region right below the text. Without a button I think looks good though.

nielsvanvelzen:
image

How about adding a hint on the bottom left? (don't mind the colors, creating the svg already took me an hour)

nielsvanvelzen:
Default theme
image
Emerald theme
image
Purple theme
image

thornbill:
> Can you try it again @thornbill ? I changed the focusing a bit

:tada: It works now!

A couple other notes, I think the background could be a bit darker and the font size and line height increased.

nielsvanvelzen:
image

Slightly darker (90% instead of 80%) and larger text (16sp instead of 14sp)

nielsvanvelzen:
In case someone needs it in the future. This is the SVG I created for the dpad before converting it to an Android drawable. Uploaded as a zip because svg files are not allowed by GitHub.

[dpad.zip](https://github.com/jellyfin/jellyfin-androidtv/files/6593774/dpad.zip)

thornbill (approved):

PR #918, opened 2021-05-25
Add PreferenceStore interface

nielsvanvelzen:
**Changes**

Added a new interface that is (right now) only implemented by the SharedPreferencesStore. I want to explore using this for display preferences (stored server side) and possibly others. This PR creates the basis for that.

**Issues**


dkanada:
If this is useful it might also be useful in the SDK whenever it stabilizes.

thornbill (approved):

nielsvanvelzen:
> If this is useful it might also be useful in the SDK whenever it stabilizes.

No this is definitely out of scope for the SDK. Adding some features to more easily read/write display preferences could be a thing but not the preference framework from the ATV app.

Issue #512, opened 2020-06-18
Option to disable menu items


dpickles:
Is it possible to get an option to disable menu items which are not in use? 

Currently I don't use the "Live TV" feature and I doubt I will in the future. I would love to be able to hide this so that it's not an option and create a more person flow to the ui.

Thanks

nielsvanvelzen:
You can re-order what sections appear on your homepage using the web version of Jellyfin. It's currently not possible to change these sections inside of the tv app.

dpickles:
Oh great. I tested this last night and it seems to solve my problem.
Would be great to see this included within the app at a later date. 

Thanks

PR #919, opened 2021-05-25
Hide live TV section when unavailable

thornbill:
**Changes**
* Hides the Live TV rows from the home screen when unavailable on a server. This mimics the logic in web although it's kind of crappy.

**Issues**
Kind of fixes the issue for #512 while not adding the requested feature

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragment.kt
@@ -96,7 +96,7 @@ class HomeFragment : StdBrowseFragment(), AudioEventListener {
 			HomeSectionType.RESUME_AUDIO -> rows.add(helper.loadResumeAudio())
 			HomeSectionType.ACTIVE_RECORDINGS -> rows.add(helper.loadLatestLiveTvRecordings())
 			HomeSectionType.NEXT_UP -> rows.add(helper.loadNextUp())
-			HomeSectionType.LIVE_TV -> if (TvApp.getApplication().currentUser!!.policy.enableLiveTvAccess && includeLiveTvRows) {
+			HomeSectionType.LIVE_TV -> if (includeLiveTvRows) {
 				rows.add(liveTVRow)
 				rows.add(helper.loadOnNow())
 			}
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragment.kt
```diff
@@ -106,6 +109,25 @@ class HomeFragment : StdBrowseFragment(), AudioEventListener {
 			// Update the views before creating rows
 			views = callApi<ItemsResult> { apiClient.GetUserViews(TvApp.getApplication().currentUser!!.id, it) }
 
+			// This is kind of ugly, but it mirrors how web handles the live TV rows on the home screen
+			// If we can retrieve one live TV recommendation, then we should display the rows
+			callApi<ItemsResult> {
```
nielsvanvelzen, line 114: Right now it will always make this additional API call which slows down home section loading. The live tv access might be disabled already so we'd better check that first. You could use `TvApp.getApplication().currentUser!!.policy.enableLiveTvAccess` here instead of in the `addSection` function to set `includeLiveTvRows` to false before running the call.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragment.kt
thornbill: Good call! I've updated it now.

app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragment.kt
```diff
@@ -106,6 +109,28 @@ class HomeFragment : StdBrowseFragment(), AudioEventListener {
...
+				// If we can retrieve one live TV recommendation, then we should display the rows
+				callApi<ItemsResult> {
+					apiClient.GetRecommendedLiveTvProgramsAsync(
+						RecommendedProgramQuery().apply {
+							userId = TvApp.getApplication().currentUser!!.id
+							enableTotalRecordCount = false
+							fields = arrayOf(
+								ItemFields.ChannelInfo,
+								ItemFields.PrimaryImageAspectRatio
+							)
```
nielsvanvelzen, line 124: Don't need the fields since it is only to check if there is any results

app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragment.kt
```diff
@@ -106,6 +109,28 @@ class HomeFragment : StdBrowseFragment(), AudioEventListener {
...
 			views = callApi<ItemsResult> { apiClient.GetUserViews(TvApp.getApplication().currentUser!!.id, it) }
 
+			// Check for live TV support
+			if (TvApp.getApplication().currentUser!!.policy.enableLiveTvAccess) {
+				// This is kind of ugly, but it mirrors how web handles the live TV rows on the home screen
+				// If we can retrieve one live TV recommendation, then we should display the rows
+				callApi<ItemsResult> {
+					apiClient.GetRecommendedLiveTvProgramsAsync(
+						RecommendedProgramQuery().apply {
+							userId = TvApp.getApplication().currentUser!!.id
```
nielsvanvelzen, line 119: Now that  3 places in this function use the currfentUser it might be worth adding it to a variable so we don't need to call the not-null operator each time.

`val currentUser = TvApp.getApplication().currentUser!!`

app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragment.kt
```diff
@@ -106,6 +109,28 @@ class HomeFragment : StdBrowseFragment(), AudioEventListener {
 			// Update the views before creating rows
 			views = callApi<ItemsResult> { apiClient.GetUserViews(TvApp.getApplication().currentUser!!.id, it) }
 
+			// Check for live TV support
+			if (TvApp.getApplication().currentUser!!.policy.enableLiveTvAccess) {
```
nielsvanvelzen, line 113: Not really necessary but you could check if there is at least 1 home section for live tv before doing the api call.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragment.kt
thornbill: I wasn't sure why web included them, but it doesn't seem to make any difference in my testing so I have removed them.

nielsvanvelzen (approved):

PR #920, opened 2021-05-26
Refactor clock view

### LINKED ISSUE #732: Text hard to read on app home screen

thornbill:
**Changes**
* Refactors the clock view in Kotlin
* Removes alpha from clock view
* Adds shadows to the text in the clock view
* Fixes null user crash

**Issues**
Fixes #732 maybe? I don't think adding a shadow helps too much in practice :man_shrugging:


nielsvanvelzen:
> Fixes null user crash

While this does fix the npe in the clockview it will just generate a new one. This crash happens after the app recovers from a crash and opens the last activity. The app doesn't have authentication info stored at that moment and thus crashes on the first part that uses the user id. This was the clock view and the problem will now move to the next usage.

thornbill:
I realize this will likely just push the crash down somewhere else in the stack, but still seems worth fixing the nullability issue here. :man_shrugging:

app/src/main/res/layout/clock_user_bug.xml
```diff
@@ -1,39 +1,48 @@
...
+    android:layout_height="match_parent">
 
     <TextClock
+        android:id="@+id/clock"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:id="@+id/clock"
         android:layout_alignParentTop="true"
         android:layout_centerHorizontal="true"
+        android:fontFamily="sans-serif-light"
```
nielsvanvelzen, line 13: Can we rely in system font or do we specifically need this one?

app/src/main/res/layout/clock_user_bug.xml
```diff
@@ -1,39 +1,48 @@
...
         android:format12Hour="h:mm"
         android:format24Hour="k:mm"
+        android:shadowColor="@color/black"
+        android:shadowRadius="4"
+        android:textColor="@color/white"
         android:textSize="40sp"
-        android:fontFamily="sans-serif-light" />
+        tools:text="13:37" />
 
     <LinearLayout
```
nielsvanvelzen, line 22: The linter warning on this line basically says: use a <TextView> with `drawableStart`. Not sure how loading with glide works in that case though.

app/src/main/res/layout/clock_user_bug.xml
```diff
@@ -1,39 +1,48 @@
...
             android:src="@drawable/ic_user" />
 
         <TextView
+            android:id="@+id/userName"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:text="User Name"
-            android:id="@+id/userName"
-            android:fontFamily="sans-serif-light" />
+            android:fontFamily="sans-serif-light"
```
nielsvanvelzen, line 42: Can we rely in system font or do we specifically need this one?

app/src/main/res/layout/clock_user_bug.xml
```diff
@@ -1,39 +1,48 @@
...
+        tools:text="13:37" />
 
     <LinearLayout
-        android:orientation="horizontal"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_below="@+id/clock"
         android:layout_centerHorizontal="true"
-        android:gravity="center"
         android:divider="@drawable/blank10x10"
```
nielsvanvelzen, line 27: Remove divider en use a margin in the child element.

nielsvanvelzen (changes requested):

app/src/main/res/layout/clock_user_bug.xml
thornbill: imo this looks significantly better for the clock than the default font since it is so large. I did change the user name because the bolder font looks better there.

app/src/main/res/layout/clock_user_bug.xml
thornbill: It seems like this is possible, but seems much more complex than the current approach for little to no gain?

app/src/main/res/layout/clock_user_bug.xml
```diff
@@ -1,39 +1,48 @@
...
+
+            <ImageView
+                android:id="@+id/userImage"
+                android:layout_width="20dp"
+                android:layout_height="20dp"
+                android:layout_marginEnd="10dp"
+                android:src="@drawable/ic_user" />
+
+            <TextView
+                android:id="@+id/userName"
```
nielsvanvelzen, line 40: This might be nitpicky but I prefer `username` instead of `userName`. Although just `name` might work too

nielsvanvelzen (changes requested):

app/src/main/res/layout/clock_user_bug.xml
nielsvanvelzen: Fair enough, it shouldn't matter AFAIK

app/src/main/res/layout/clock_user_bug.xml
thornbill: It is nitpicky, but I'll allow it. :sweat_smile:

nielsvanvelzen (approved):

PR #921, opened 2021-05-26
Fix nullability issues

thornbill:
**Changes**
* Fixes nullability issues resulting in app crashes
* Marks `TvApp.getApplication()` as nullable as that seems to be a big cause of crashes

**Issues**
N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/Utils.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/Utils.java
@@ -8,7 +8,6 @@ import android.widget.PopupMenu;
 import android.widget.Toast;
 
 import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
 
 import org.jellyfin.androidtv.BuildConfig;
 import org.jellyfin.androidtv.R;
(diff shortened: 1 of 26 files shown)


app/src/main/java/org/jellyfin/androidtv/util/Utils.java
```diff
@@ -44,12 +46,12 @@ public static void showToast(Context context, int resourceId) {
...
 
-    public static int convertDpToPixel(Context ctx, int dp) {
+    public static int convertDpToPixel(@Nullable Context ctx, int dp) {
         return convertDpToPixel(ctx, (float) dp);
     }
 
-    public static int convertDpToPixel(Context ctx, float dp) {
-        float density = ctx.getResources().getDisplayMetrics().density;
+    public static int convertDpToPixel(@Nullable Context ctx, float dp) {
+        float density = ctx == null ? 1 : ctx.getResources().getDisplayMetrics().density;
```
AndreasGB, line 54: Why is the density 1 when there is no context provided?

thornbill: I thought it would be the safest default value, but I'm open to suggestions. :smile:

nielsvanvelzen: Change the context parameter to not-null. Places that don't send a context should be fixed instead

nielsvanvelzen (approved):

PR #922, opened 2021-05-27
WIP: Display next up before video ends

MrChip53:
**Changes**
Displays next up before video ends.

Should not conflict with #850 as of now (05/26/21)

**Todo**

- better time calcs
- block player ui from showing
- test, test and test
- probably some other stuff too

**Issues**


MrChip53:
Not happy with the direction I took this and how it was panning out. Going to close for now and if no one gets to it before I'm done working on some live tv stuff I'll come back to this with a fresh mind. It may also be worth waiting to implement this with the playback rewrite as the current playback is less than desirable to work with.

PR #923, opened 2021-05-27
Show login dialog when token is invalid

nielsvanvelzen:
**Changes**

- Return a boolean from SessionRepository::switchCurrentSession to indicate success
  - Same for AuthenticationRepository::setActiveSession
- Check for success in AuthenticationRepository::authenticateUser
  - If failed -> require new login or auto-login for users without password

This change fixes the issue where logins would fail when the access token was invalidated by the server.

**Issues**


thornbill (approved):
:+1: works as expected testing against the demo server!

PR #926, opened 2021-05-27
Fix direct play

nielsvanvelzen:
**Changes**

- Build a stream url when direct play url is not http/https
  - This fixes the player showing a "format not supported" toast
  - For me, in the emulator, this plays audio but video is black because the codec is not supported
 - The code is pretty much a copy from direct stream but with the play method changed to direct play

**Issues**


thornbill:
This looks good to me, but I would really like the input of someone from the server side to verify our logic matches their expectations.

crobibero:
The url builder logic looks fine to me

thornbill (approved):

PR #932, opened 2021-06-03
Use Next Up screen when playing with an external app

### LINKED ISSUE #691: Option to disable annoying questions with External Players
While using external player the UX is heavily impacted by annoying popups:
- Every time episode is finished, popup "Play next?" appears, though in settings there is option dedicated for "Queuing next episode"
- Every time episode is interrupted earlier there is a question "Mark as played?", though the jellyfin knows how much of the episode was played, and it doesn't ask questions for built-in players

I assume there is a reason behind those popups - debug probably. It would be great - to have an option to disable those

**Additional context**	
https://github.com/jellyfin/jellyfin-androidtv/blob/ccb47ab9afd6610eec98ca698203ad103eead2e4/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.java#L187
```java
if (mItemsToPlay.size() > 0) {
            //Must confirm moving to the next item or there is no way to stop playback of all the items
            new AlertDialog.Builder(this)
                    .setTitle("Next up is "+mItemsToPlay.get(mCurrentNdx).getName())
```

linetrimmer:
**Changes**

- Displays the Next Up screen instead of an alert dialog after an episode has finished
- Cancels playback if the user says to not mark the episode as watched
- Plays the next episode without waiting for user-input
  - If using the next up screen, the opportunity to stop playback is provided by the next up screen's timer
  - If the next up screen is disabled, depending on how much of the episode was watched, backing out of the player before the episode ends will either stop playback or bring up the alert dialog asking whether to mark the episode as watched. Answering no stops playback, answering yes continues to the next episode. As far as I can tell, this addresses the concern `//Must confirm moving to the next item or there is no way to stop playback of all the items`. If I'm mistaken, though, I added the behavior to autoplay when the next up screen is disabled as the fourth commit (the first three commits always use the next up screen), so rolling back a commit should solve that.

**Issues**
Fixes #691

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.java
@@ -197,8 +197,8 @@ public class ExternalPlayerActivity extends FragmentActivity {
                 mediaManager.getValue().setVideoQueueModified(true);
 
                 Intent intent = new Intent(this, NextUpActivity.class);
-                intent.putExtra("id", mItemsToPlay.get(mCurrentNdx).getId());
-                intent.putExtra("useExternalPlayer", true);
+                intent.putExtra(NextUpActivity.EXTRA_ID, mItemsToPlay.get(mCurrentNdx).getId());
+                intent.putExtra(NextUpActivity.EXTRA_USE_EXTERNAL_PLAYER, true);
                 startActivity(intent);
                 finishAfterTransition();
             } else {
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.java
@@ -18,6 +18,7 @@ import org.jellyfin.androidtv.data.compat.VideoOptions;
 import org.jellyfin.androidtv.data.service.BackgroundService;
 import org.jellyfin.androidtv.preference.UserPreferences;
 import org.jellyfin.androidtv.preference.constant.PreferredVideoPlayer;
+import org.jellyfin.androidtv.ui.playback.nextup.NextUpActivity;
 import org.jellyfin.androidtv.util.Utils;
 import org.jellyfin.androidtv.util.apiclient.ReportingHelper;
 import org.jellyfin.androidtv.util.profile.ExternalPlayerProfile;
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpActivity.kt
```diff
@@ -4,23 +4,37 @@ import android.content.Intent
...
+	private val userPreferences: UserPreferences by inject()
+	private val systemPreferences: SystemPreferences by inject()
 
 	override fun onCreate(savedInstanceState: Bundle?) {
 		super.onCreate(savedInstanceState)
+		val preferredVideoPlayer = userPreferences[UserPreferences.videoPlayer]
+		val chosenVideoPlayer = systemPreferences[SystemPreferences.chosenPlayer]
+		val usingExternalPlayer = preferredVideoPlayer == PreferredVideoPlayer.EXTERNAL
+			|| (preferredVideoPlayer == PreferredVideoPlayer.CHOOSE
+			&& chosenVideoPlayer == PreferredVideoPlayer.EXTERNAL)
```
nielsvanvelzen, line 27: Send this information via an extra/argument to the activity instead

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.java
```diff
@@ -186,28 +192,20 @@ protected void markPlayed(String itemId) {
...
-                    })
-                    .show();
+            if (userPreferences.getValue().get(UserPreferences.Companion.getNextUpEnabled())) {
+                // Set to "modified" so the queue won't be cleared
+                mediaManager.getValue().setVideoQueueModified(true);
 
+                Intent intent = new Intent(this, NextUpActivity.class);
+                intent.putExtra("id", mItemsToPlay.get(mCurrentNdx).getId());
+                startActivity(intent);
+                finish();
```
nielsvanvelzen, line 202: should use finishAfterTransition

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpActivity.kt
```diff
@@ -14,13 +15,17 @@ class NextUpActivity : FragmentActivity() {
 
 	override fun onCreate(savedInstanceState: Bundle?) {
 		super.onCreate(savedInstanceState)
+		val usingExternalPlayer = intent.getBooleanExtra("usingExternalPlayer", false)
```
nielsvanvelzen, line 18: We've moved extra/argument names to constants after this activity was created. Can you move the 2 extras (id and this one) to constants?

something like:
```
companion object {
    const val EXTRA_ID = "id"
}
```

nielsvanvelzen: Also rename to "useExternalPlayer"

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.java
```diff
@@ -186,28 +192,21 @@ protected void markPlayed(String itemId) {
...
-                    })
-                    .show();
-
+            if (userPreferences.getValue().get(UserPreferences.Companion.getNextUpEnabled())) {
+                // Set to "modified" so the queue won't be cleared
+                mediaManager.getValue().setVideoQueueModified(true);
+
+                Intent intent = new Intent(this, NextUpActivity.class);
+                intent.putExtra("id", mItemsToPlay.get(mCurrentNdx).getId());
+                intent.putExtra("useExternalPlayer", true);
```
nielsvanvelzen, line 201: `NextUpActivity.EXTRA_USE_EXTERNAL_PLAYER`

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.java
```diff
@@ -186,28 +192,21 @@ protected void markPlayed(String itemId) {
...
-                        }
-                    })
-                    .show();
-
+            if (userPreferences.getValue().get(UserPreferences.Companion.getNextUpEnabled())) {
+                // Set to "modified" so the queue won't be cleared
+                mediaManager.getValue().setVideoQueueModified(true);
+
+                Intent intent = new Intent(this, NextUpActivity.class);
+                intent.putExtra("id", mItemsToPlay.get(mCurrentNdx).getId());
```
nielsvanvelzen, line 200: `NextUpActivity.EXTRA_ID`

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

PR #938, opened 2021-06-05
Remove CONTRIBUTORS.md

nielsvanvelzen:
**Changes**

This change removes the CONTRIBUTORS.md file. It was outdated (last updated 17 months ago!) and a lot of contributors are missing from it. GitHub can show the contributors [here](https://github.com/jellyfin/jellyfin-androidtv/graphs/contributors) and you can do it from the command line with `git shortlog -sne` or a similar command. The Kotlin SDK and mobile Android app repositories don't have a contributors file either.

**Issues**


thornbill:
Nope it needs to stay for attribution purposes. Perhaps we can enforce its usage in a PR check though.

PR #940, opened 2021-06-07
Refactor all card views

nielsvanvelzen:
I'm introducing a new card view in the upcoming pull request for the login UI. This PR backports the renamed and refactored existing views.

**Changes**

- Use consistent naming for all card views
  - MyChannelCardView -> ChannelCardView
  - MyImageCardView -> LegacyImageCardView (should eventually be entirely replaced by the new one)
  - MyInfoCardView -> MediaInfoCardView
- Use view binding in all card views
- Reformatted all card view Java files

This PR does not fix issues like missing strings or generally bad code in the views.

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/card/LegacyImageCardView.java
```diff
@@ -22,13 +22,13 @@
  * Modified ImageCard with no fade on the badge
  * A card view with an {@link ImageView} as its main region.
  */
-public class MyImageCardView extends BaseCardView {
-    private ImageCardViewBinding binding = ImageCardViewBinding.inflate(LayoutInflater.from(getContext()), this);
+public class LegacyImageCardView extends BaseCardView {
```
thornbill, line 25: Should we be concerned about the missing constructor warning here?

nielsvanvelzen: For new views I'd complain about this warning, but it's not something that breaks the app. The missing constructor just causes the view to not work in the Android Studio previews. But AFAIK none of the activities/fragments using this card currently work as previews in AS.

nielsvanvelzen: A proper fix is to remove the `showInfo` parameter in the constructor and make it a class-property.

thornbill (approved):

PR #944, opened 2021-06-09
Rewrite authentication UI

### LINKED ISSUE #612: Login Redesign

nielsvanvelzen:
Started on this over a month ago and slowly made changes. I made some tweaks since my latest intensive test but it should all work now. If any issues occur or other changes are desired please consider if they can be done in a separate PR instead.

**Added**

- RecyclerView
  - Added `SpacingItemDecoration` to easily add spacing between items in a RecyclerView
- DefaultCardView
  - New card view 
  - Uses androidx.cardview for rounding the corners, no dependency on leanback.
  - Goal is to eventually make this the default card and replace other cards.
  - Supports image + text right now
  - Image should use one of the constant sizes (right now only SQUARE)
  - Uses a scaling animation on focus
- ListAdapter
  - Copied from https://github.com/nielsvanvelzen/tv-launcher/blob/e05fba217d463d0de337ff31611bd9f496ff2c1e/app/src/main/kotlin/nl/ndat/tvlauncher/ui/adapter/ListAdapter.kt
  - Should be used as a base for all adapters going forward
  - Contains functions to check change in list items to only update items that changed
    - Should drastically increase performance

**Changed**

- Dropped viewpager UI
- Removed ic_edit and tile_edit drawables (not used anymore)
- AuthenticationRepository / ServerRepository
  - Add version/login disclaimer data to Server class
  - Add new "ServerVersionNotSupported" response
  - Fix user list not using a clone in ServerRepository (could cause missing users)
- ItemLauncher
  - Added `createUserViewIntent`, basically the same as `launchUserView` except it returns the intent instead of launching it
  - `launchUserView` uses `createUserViewIntent` now
  - `createUserViewIntent` is used in the startup activity to properly launch user views when requested

**Remade**

- Redesign entire server+user selection flow
  - Now displays login disclaimer
  - No issues with viewpager anymore
  - Added a welcome message
  - Added a version check
  - Reloads server info when active
  - Slightly faster
  - Removed toolbar buttons

**Issues**

Closes  #612


nielsvanvelzen:
## Screenshots!!

image

image

image

image

nielsvanvelzen:
The linter error is caused by one of our strings not properly translated in Weblate and should be ignored for this PR (should be fixed within Weblate). PR should be ready for review now.

thornbill:
Tested on a 2nd gen Fire Stick and noticed the following issues:

* The user cards and server button on the ServerFragment do not show focus and cannot be selected. Only the "Add user" button works as expected.
* The screen margins seem to be too small. Google recommends `48dp` for horizontal margins to prevent overscan. https://developer.android.com/training/tv/start/layouts.html#overscan

thornbill:
This seems to be working brilliantly now! I have a few design ideas, but functionally it seems solid! :tada: 

The only minor issues I really noticed while testing are:

* When selecting a discovered server, the add server form is visible briefly.
* When I land on the server screen, the first user seems to have focus, but does not display larger like it does if I navigate away and back.

nielsvanvelzen:
No more opening the "AddServerAlertFragment" when connecting to discovery servers:

https://user-images.githubusercontent.com/2305178/121868788-af060200-cd01-11eb-83f1-249daae86413.mp4

- Progressbar while connecting
- Errors show a toast
- Successful connection opens the serverfragment (showing the users)

nielsvanvelzen:
Fixed a few issues:

- Fix incorrect scale in DefaultCardView when focused by default
- Fix ServerFragment keeping history when closing due too invalid server
  - Could happen when you removed the server and pressed back, now closes the app after the second press
- Fix removing user not removing the access token 
  - When the auto-signin user is "demo" and you removed "demo" but didn't logout it would still login as "demo" after restarting the app. It now signs out the user as is expected.

nielsvanvelzen:
- Fixed ellipsize for login disclaimer
- Changed cloud icon to house icon for servers
  - Also in other places that used the cloud

app/src/main/java/org/jellyfin/androidtv/auth/model/Server.kt
```diff
@@ -9,8 +11,13 @@ data class Server(
 	var id: UUID,
 	var name: String,
 	var address: String,
-	var dateLastAccessed: Date = Date(0)
+	val version: String? = null,
+	val loginDisclaimer: String? = null,
+	var dateLastAccessed: Date = Date(0),
 ) {
+	private val serverVersion = version?.let(ServerVersion::fromString)
+	val versionSupported = serverVersion != null && serverVersion >= ServerRepository.minimumServerVersion
```
thornbill, line 19: Should we consider `null` versions valid? At one point I ran into a case where the version hadn't been fetched yet so it said it was invalid, but retrying it worked as expected.

app/src/main/java/org/jellyfin/androidtv/ui/startup/StartupActivity.kt
```diff
@@ -101,47 +114,48 @@ class StartupActivity : FragmentActivity(R.layout.fragment_content_view) {
...
-				// Try opening the user view
-				val item = callApi<BaseItemDto?> { apiClient.GetItemAsync(itemId, apiClient.currentUserId, it) }
-
-				if (item != null) {
-					ItemLauncher.launchUserView(item, this, true)
-					finish()
-					return
+		// Create intent
+		val intent = when {
+			// Item is requedted
```
thornbill, line 119: "requested" :wink:

app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
```diff
@@ -19,10 +19,13 @@ import java.util.*
...
 	fun getUsers(server: UUID): List<PrivateUser>?
+	fun getUser(server: UUID, user: UUID): PrivateUser?
 	fun saveServer(id: UUID, name: String, address: String, version: String?, loginDisclaimer: String?)
 	fun authenticateUser(user: User): Flow<LoginState>
 	fun authenticateUser(user: User, server: Server): Flow<LoginState>
 	fun login(server: Server, username: String, password: String = ""): Flow<LoginState>
+	fun logout(user: User)
+	fun removeUser(user: User)
 	fun getUserImageUrl(server: Server, user: User): String?
 }
```
thornbill, line 31: Should this warning be suppressed?

thornbill (changes requested):

app/src/main/java/org/jellyfin/androidtv/auth/model/Server.kt
nielsvanvelzen: The reason it didn't work that one time for you is because you already had server info stored from a previous development version. If you add a new server right now the app fetches the version when it stores it. It shouldn't be able to be null anymore.
The nullability in the model is for compatibility with older 0.12 builds. If I removed it the app would crash for users with their own builds.

app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
nielsvanvelzen: I increased the threshold to 15

thornbill (approved):

PR #945, opened 2021-06-09
Ignore UnusedIds linter check

nielsvanvelzen:
Doesn't work properly with view binding (causing false-positives)

**Changes**
- UnusedIds check changed from warn to ignore

**Issues**


thornbill (approved):

nielsvanvelzen:
45 issues fixed!

PR #951, opened 2021-06-11
Add skip button to next up screen when shuffling

linetrimmer:
**Changes**

- Adds a button to the next up screen when shuffling to skip the queued item

I wasn't sure if there'd be any interest in skipping an item when playing in sequential order, so I didn't want to add an extra click to get to the "cancel" button in that case. The restriction can be removed if there's interest in always showing it, though.


nielsvanvelzen:
I think this is a nice feature to have eventually but I think it should be added after the playback rewrite. I don't want to introduce too much behavior related to the actual playback queues etc. until then.

The NextUpActivity is now independent of the MediaManager and can be easily re-used in the rewrite, I want to keep it that way for now.

PR #954, opened 2021-06-11
Fix StdGridFragment always refreshing on resume

### LINKED ISSUE #805: All Items view issues
Couple issues ive noticed in this section 

1. When in the all item view and you click a movie but then decide i dont want that movie and press back the all item view it goes back to start not the item you was just viewing so u have to scroll back to were u was

2. When in the all item view and scrolling across every so often the screen will go dim like it doing something refreshing the screen and at bottom were its doing a count up it would do the dim and say 1 at bottom

This is latest master on shield

nielsvanvelzen:
Now properly checks changes and updates only when necessary

**Changes**
- Fix StdGridFragment always refreshing on resume

**Issues**

Fixes #805 (first point, second one already fixed)


thornbill (approved):
:tada: works as expected!

PR #955, opened 2021-06-11
Replace ImageButton with Button.Icon style

nielsvanvelzen:
**Changes**

- Removed custom ImageButton class in favor of Android's ImageButton with our own Button.Icon style
  - Replace ImageButton in StdGridFragment
  - Replace ImageButton in AudioNowPlayingActivity
- Updated Button.Icon style:
  - Added disabled state
  - Added activated state (adds a red color, no theming support right now)
  - Increased brightness of normal state text
- Updated Button.Default style:
  - Added disabled state
  - Increased brightness of normal state text
- Remove misuses of custom ImageButton class
  - one bad type cast
  - a few constant usages although the function that receives them doesn't use the value
- Delete custom ImageButton view

**Screenshots**

Music player:

image
image

"All Items" filters:

image

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -324,7 +323,7 @@ private void play(long position, int transcodedSubtitle) {
                 if (mVideoManager.isNativeMode()) mPlaybackState = PlaybackState.PLAYING; //won't get another onprepared call
                 if (mFragment != null) {
                     mFragment.setFadingEnabled(true);
-                    mFragment.setPlayPauseActionState(ImageButton.STATE_SECONDARY);
+                    mFragment.setPlayPauseActionState(0);
```
thornbill, line 326: Does the value here not matter? It looks like the SECONDARY state was 1 previously.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -429,7 +428,7 @@ public void onClick(DialogInterface dialog, int which) {
                 playInternal(getCurrentlyPlayingItem(), position, vlcOptions, internalOptions);
                 mPlaybackState = PlaybackState.BUFFERING;
                 if (mFragment != null) {
-                    mFragment.setPlayPauseActionState(ImageButton.STATE_SECONDARY);
+                    mFragment.setPlayPauseActionState(0);
```
thornbill, line 431: Same question here.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
nielsvanvelzen: Nope, the function never uses the argument.

nielsvanvelzen: https://github.com/jellyfin/jellyfin-androidtv/blob/f668e1c66d19686c8692704ca90b2c253e1d204a/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java#L1337-L1339

thornbill (approved):

PR #957, opened 2021-06-12
Remove padding from end of toolbar content when clock is hidden

linetrimmer:
**Changes**

- Removes padding from the end of the toolbar content if the clock is hidden

Really liking the toolbar :)

**Screenshots**
<details open>
<summary>Before (no clock)</summary>

before-no-clock
</details>

<details open>
<summary>After (no clock)</summary>

after-no-clock
</details>

<details open>
<summary>Before (with clock)</summary>

before-clock
</details>

<details open>
<summary>After (with clock)</summary>

after-clock
</details>

(Before/after with the clock are the same)

### DIFF
--- a/app/src/main/res/layout/view_toolbar.xml
+++ b/app/src/main/res/layout/view_toolbar.xml
@@ -36,9 +36,9 @@
         android:gravity="bottom"
         android:orientation="horizontal"
         android:paddingStart="24dp"
-        android:paddingEnd="24dp"
         app:layout_constraintEnd_toStartOf="@id/toolbar_end"
-        app:layout_constraintStart_toEndOf="@id/toolbar_start" />
+        app:layout_constraintStart_toEndOf="@id/toolbar_start"
+        tools:ignore="RtlSymmetry" />
 
     <LinearLayout
         android:id="@+id/toolbar_end"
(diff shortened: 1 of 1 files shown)


app/src/main/res/layout/view_toolbar.xml
```diff
@@ -36,7 +36,7 @@
         android:gravity="bottom"
         android:orientation="horizontal"
         android:paddingStart="24dp"
-        android:paddingEnd="24dp"
+        android:paddingEnd="0dp"
```
thornbill, line 39: Couldn’t this be removed instead of setting it to `0`?

app/src/main/res/layout/view_toolbar.xml
```diff
@@ -60,6 +60,8 @@
             android:fontFamily="sans-serif-light"
             android:format12Hour="h:mm"
             android:format24Hour="k:mm"
+            android:paddingStart="24dp"
+            android:paddingEnd="0dp"
```
thornbill, line 64: Same question here.

thornbill (changes requested):

app/src/main/res/layout/view_toolbar.xml
linetrimmer: Definitely could, I was just trying to appease the warning it gave me about including a paddingEnd for symmetry

nielsvanvelzen: If that warning is from the Android Linter you can add `tools:ignore="warningid"` to ignore it

nielsvanvelzen (approved):

thornbill (approved):

PR #963, opened 2021-06-15
Close activity instead of crashing when playback fails

nielsvanvelzen:
**Changes**

- Add additional check to prevent an NPE I got while testing something else.

This doesn't really fix the issue but it prevents a crash when playback fails and the app tries it again.

**Issues**


thornbill:
There are so many potential NPEs in the playback code... I looked at fixing it, but it seemed like too much to take on when we plan to rewrite it anyway.

thornbill (approved):

PR #966, opened 2021-06-16
Update clock bug to match toolbar

thornbill:
**Changes**
Changes the clock bug to match the appearance of the toolbar. I think this looks nice, but maybe it's weird to only be able to select the user icon on the home screen? :man_shrugging: 

**Issues**
N/A


nielsvanvelzen (approved):

nielsvanvelzen:
> maybe it's weird to only be able to select the user icon on the home screen? 🤷‍♂️

It can be a bit confusing but it's something we have to live with for now

PR #973, opened 2021-06-17
Update playback overlay layout

thornbill:
**Changes**
* Cleanup CustomPlaybackOverlayFragment
* Update title display
  * Moves title and subtitle to the upper left corner where the logo is
  * Hides the title if a logo is available

**Movie:**
movie overlay

**Movie with Logo:**
movie logo overlay

**TV Show:**
episode overlay

**Issues**
N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
@@ -18,9 +18,7 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.view.animation.Animation;
 import android.view.animation.AnimationUtils;
-import android.widget.FrameLayout;
 import android.widget.HorizontalScrollView;
-import android.widget.ImageView;
 import android.widget.LinearLayout;
 import android.widget.RelativeLayout;
 import android.widget.TextView;
@@ -45,6 +43,8 @@ import org.jellyfin.androidtv.R;
 import org.jellyfin.androidtv.TvApp;
 import org.jellyfin.androidtv.constant.CustomMessage;
 import org.jellyfin.androidtv.data.model.DataRefreshService;
+import org.jellyfin.androidtv.databinding.OverlayTvGuideBinding;
+import org.jellyfin.androidtv.databinding.VlcPlayerInterfaceBinding;
 import org.jellyfin.androidtv.ui.AudioDelayPopup;
 import org.jellyfin.androidtv.ui.GuideChannelHeader;
 import org.jellyfin.androidtv.ui.GuidePagingButton;
@@ -77,6 +77,7 @@ import org.jellyfin.apiclient.interaction.ApiClient;
 import org.jellyfin.apiclient.interaction.EmptyResponse;
 import org.jellyfin.apiclient.interaction.Response;
 import org.jellyfin.apiclient.model.dto.BaseItemDto;
+import org.jellyfin.apiclient.model.dto.BaseItemType;
 import org.jellyfin.apiclient.model.dto.ChapterInfoDto;
 import org.jellyfin.apiclient.model.dto.UserItemDataDto;
 import org.jellyfin.apiclient.model.livetv.ChannelInfoDto;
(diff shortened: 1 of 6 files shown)


app/src/main/res/layout/vlc_player_interface.xml
```diff
@@ -91,15 +91,37 @@
             android:layout_marginTop="20sp"
             android:layout_marginEnd="40sp" />
 
-        <ImageView
-            android:id="@+id/logoImage"
-            android:layout_width="190sp"
-            android:layout_height="60sp"
+        <LinearLayout
```
nielsvanvelzen, line 94: "LinearLayout-3 can overlap @id/textClock if LinearLayout-3 grows due to localized text expansion"

This might cause issues, make sure the width is limited (by adding the clock to this linearlayout or using a constraintlayout or something)

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -1346,14 +1284,36 @@ public void setPlayPauseActionState(final int state) {
...
+            if (current.getBaseItemType() == BaseItemType.Episode) {
+                binding.itemTitle.setText(current.getSeriesName());
+                binding.itemSubtitle.setText(BaseItemUtils.getDisplayName(current, requireContext()));
+            } else {
+                binding.itemTitle.setText(current.getName());
+            }
+            // Update the logo
+            String imageUrl = ImageUtils.getLogoImageUrl(current, apiClient.getValue(), 440, false);
+            if (imageUrl != null) {
+                Glide.with(requireActivity())
```
nielsvanvelzen, line 1305: requireContext

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -1346,14 +1284,36 @@ public void setPlayPauseActionState(final int state) {
     @Override
     public void updateDisplay() {
         BaseItemDto current = mPlaybackController.getCurrentlyPlayingItem();
-        if (current != null & mActivity != null && !mActivity.isFinishing()) {
+        if (current != null && !requireActivity().isFinishing()) {
```
nielsvanvelzen, line 1287: requireactivity throws an exception if activity is null, the previous if statement just did a no-op. Might cause issues.

app/src/main/res/layout/vlc_player_interface.xml
```diff
@@ -74,32 +74,54 @@
...
         android:visibility="invisible">
 
+        <LinearLayout
+            android:id="@+id/item_info"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:orientation="vertical"
+            app:layout_constraintEnd_toStartOf="@id/textClock"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent">
```
nielsvanvelzen, line 94: This linearlayout can be removed now with constraints added to the child views (all use a constraint of "start_toendof" the previos child)

nielsvanvelzen (changes requested):

app/src/main/res/layout/vlc_player_interface.xml
```diff
@@ -74,32 +74,55 @@
...
+            android:layout_height="60dp"
+            app:layout_constraintEnd_toStartOf="@id/textClock"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent"
+            app:layout_constraintHorizontal_bias="0.0" />
+
+        <TextView
+            android:id="@+id/item_title"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
```
nielsvanvelzen, line 99: When the title of a tv series it long it will append a new line. I'm not 100% sure if we want to keep that behavior or keep it at a single line.

```suggestion
            android:layout_height="wrap_content"
            android:ellipsize="end"
            android:singleLine="true"
```

I guess the same question for the subtitle

nielsvanvelzen (approved):

app/src/main/res/layout/vlc_player_interface.xml
thornbill: It would have to be a _very_ long title... I think wrapping would be fine for those personally :man_shrugging:

nielsvanvelzen: Sounds good, I found a few things in my library with long titles but they never went over 2 lines.

PR #977, opened 2021-06-23
Fix user information not reflecting updates

thornbill:
**Changes**
Fixes user information not showing updates on the login screen

**Issues**
N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
@@ -148,8 +148,7 @@ class AuthenticationRepositoryImpl(
 				setActiveSession(user, server) -> {
 					// Update stored user
 					try {
-						val response = UserApi(userApiClient).getCurrentUser()
-						val userInfo = response.content
+						val userInfo by UserApi(userApiClient).getCurrentUser()
 						val currentUser = authenticationStore.getUser(server.id, user.id)
 
 						val updatedUser = currentUser?.copy(
@@ -163,11 +162,12 @@ class AuthenticationRepositoryImpl(
 						)
 						authenticationStore.putUser(server.id, user.id, updatedUser)
 						accountManagerHelper.putAccount(AccountManagerAccount(user.id, server.id, updatedUser.name, account.accessToken))
+
+						emit(AuthenticatedState)
 					} catch(err: ApiClientException) {
 						Timber.e(err, "Unable to get current user data")
+						emit(RequireSignInState)
 					}
-
-					emit(AuthenticatedState)
 				}
 				// Login failed
 				else -> when {
@@ -246,11 +246,13 @@ class AuthenticationRepositoryImpl(
 		val api = jellyfin.createApi(baseUrl = server.address)
 		val imageApi = ImageApi(api)
 
-		return if (user.imageTag == null) null else imageApi.getUserImageUrl(
-			userId = user.id,
-			tag = user.imageTag,
-			imageType = ImageType.PRIMARY,
-			maxHeight = ImageUtils.MAX_PRIMARY_IMAGE_HEIGHT
-		)
+		return user.imageTag?.let { imageTag ->
+			imageApi.getUserImageUrl(
+				userId = user.id,
+				tag = imageTag,
+				imageType = ImageType.PRIMARY,
+				maxHeight = ImageUtils.MAX_PRIMARY_IMAGE_HEIGHT
+			)
+		}
 	}
 }
--- a/app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
@@ -6,13 +6,13 @@ import kotlinx.coroutines.flow.emitAll
 import kotlinx.coroutines.flow.flow
 import org.jellyfin.androidtv.auth.model.*
 import org.jellyfin.androidtv.util.ImageUtils
-import org.jellyfin.androidtv.util.apiclient.callApi
-import org.jellyfin.apiclient.Jellyfin
 import org.jellyfin.apiclient.interaction.device.IDevice
-import org.jellyfin.apiclient.model.dto.ImageOptions
-import org.jellyfin.apiclient.model.entities.ImageType
-import org.jellyfin.apiclient.model.users.AuthenticationResult
-import org.jellyfin.sdk.model.serializer.toUUID
+import org.jellyfin.sdk.Jellyfin
+import org.jellyfin.sdk.api.client.exception.ApiClientException
+import org.jellyfin.sdk.api.client.extensions.authenticateUserByName;
+import org.jellyfin.sdk.api.operations.ImageApi
+import org.jellyfin.sdk.api.operations.UserApi
+import org.jellyfin.sdk.model.api.ImageType
 import timber.log.Timber
 import java.util.*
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/jellyfin/androidtv/auth/ServerRepository.kt
```diff
@@ -92,6 +92,8 @@ class ServerRepositoryImpl(
 			.forEach { user ->
 				if (users.none { it.id == user.id })
 					users.add(user)
+				else
+					users[users.indexOfFirst { it.id == user.id }] = user
```
nielsvanvelzen, line 96:
```suggestion
				val index = users.indexOfFirst { it.id == user.id }
				if (index == -1) users.add(user)
				else users[index] = user
```
This way it will not search the list twice, although one problem with this behavior is that public users will replace private users, which uses a different class and might contain additional data in the future. I'm not sure if that's something we want to do.

app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
```diff
@@ -103,7 +103,12 @@ class AuthenticationRepositoryImpl(
 			}
 
 			authenticationStore.getUser(server.id, user.id)?.let { storedUser ->
-				authenticationStore.putUser(server.id, user.id, storedUser.copy(lastUsed = Date().time))
+				authenticationStore.putUser(server.id, user.id, storedUser.copy(
```
nielsvanvelzen, line 106: The user object send to this function always comes from our own store or it's just been made in the login function. Updating the data doesn't do anything.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
thornbill: This was required in my testing because a stored user would never be updated if an auth token had been stored.

nielsvanvelzen: I think that's because it is called by the authenticateUser function which is called with the changed user data you set in the `getServerUsers` function, which shouldn't really do that.

thornbill: The big question I still have here is if we should not be updating the user in the authentication store here, then where should that be happening?

nielsvanvelzen: I think directly in in the place the changes to the user object are made. So the `getServerUsers` functions I guess.

app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
```diff
@@ -215,11 +243,14 @@ class AuthenticationRepositoryImpl(
...
 
 	override fun getUserImageUrl(server: Server, user: User): String? {
-		val apiClient = jellyfin.createApi(serverAddress = server.address, device = device)
-		return apiClient.GetUserImageUrl(user.id.toString(), ImageOptions().apply {
-			tag = user.imageTag
-			imageType = ImageType.Primary
+		val api = jellyfin.createApi(baseUrl = server.address)
+		val imageApi = ImageApi(api)
+
+		return if (user.imageTag == null) null else imageApi.getUserImageUrl(
```
nielsvanvelzen, line 249: Alternatively could use `user.imageTag?.let { imageTag -> imageApi.geturl.... }`

app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
```diff
@@ -142,7 +145,30 @@ class AuthenticationRepositoryImpl(
...
+							imageTag = userInfo.primaryImageTag
+						) ?: AuthenticationStoreUser(
+							name = userInfo.name!!,
+							requirePassword = userInfo.hasPassword,
+							imageTag = userInfo.primaryImageTag
+						)
+						authenticationStore.putUser(server.id, user.id, updatedUser)
+						accountManagerHelper.putAccount(AccountManagerAccount(user.id, server.id, updatedUser.name, account.accessToken))
+					} catch(err: ApiClientException) {
+						Timber.e(err, "Unable to get current user data")
```
nielsvanvelzen, line 167: I think we want to emit `RequireSignInState` when an apiclient exception occurs, since apparantly the authentication is bad

app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
```diff
@@ -142,7 +145,30 @@ class AuthenticationRepositoryImpl(
 			// Access token found, proceed with sign in
 			account?.accessToken != null -> when {
 				// Update session
-				setActiveSession(user, server) -> emit(AuthenticatedState)
+				setActiveSession(user, server) -> {
+					// Update stored user
+					try {
+						val response = UserApi(userApiClient).getCurrentUser()
+						val userInfo = response.content
```
nielsvanvelzen, line 152:
```suggestion
						val userInfo by UserApi(userApiClient).getCurrentUser()
```

app/src/main/java/org/jellyfin/androidtv/auth/ServerRepository.kt
```diff
@@ -102,8 +102,13 @@ class ServerRepositoryImpl(
 		getServerPublicUsers(server)
 			.sortedBy { it.name }
 			.forEach { user ->
-				if (users.none { it.id == user.id })
-					users.add(user)
+				val index = users.indexOfFirst { it.id == user.id }
+				if (index == -1) users.add(user)
+				else users[index] = (users[index] as PrivateUser).copy(
```
nielsvanvelzen, line 107: Should check if `users[index]` is a PrivateUser so this won't cause issues if the above code is changed in the future.

something like this I guess
```kt
val currentUser = users[index]
if (currentUser is PrivateUser) users[index] = currentUser.copy()
else if (currentUser is PublicUser) users[index] = user
```

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/auth/ServerRepository.kt
```diff
@@ -104,11 +104,15 @@ class ServerRepositoryImpl(
...
 				if (index == -1) users.add(user)
-				else users[index] = (users[index] as PrivateUser).copy(
-					name = user.name,
-					requirePassword = user.requirePassword,
-					imageTag = user.imageTag
-				)
+				else {
+					val currentUser = users[index]
+					if (currentUser is PublicUser) users[index] = user
+					else if (currentUser is PrivateUser) users[index] = currentUser.copy(
```
nielsvanvelzen, line 110:
```kt
users[index] = when(currentUser) {
  is PublicUser -> user
  is PrivateUser -> currentUser.copy(....)
}
```

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/auth/ServerRepository.kt
```diff
@@ -105,13 +105,14 @@ class ServerRepositoryImpl(
 				val index = users.indexOfFirst { it.id == user.id }
 				if (index == -1) users.add(user)
 				else {
```
nielsvanvelzen, line 107:
```suggestion
				else users[index] = when(val currentUser = users[index]) {
					is PublicUser -> user
					is PrivateUser -> currentUser.copy(
						name = user.name,
						requirePassword = user.requirePassword,
						imageTag = user.imageTag
					)
				}
```
🤐

nielsvanvelzen (approved):

PR #980, opened 2021-06-23
Add glide app module for configuration

thornbill:
**Changes**
Adds an app module for Glide to manage common configuration options and manage log level

**Issues**
N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/JellyfinGlideModule.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/JellyfinGlideModule.kt
@@ -10,14 +10,13 @@ import com.bumptech.glide.request.RequestOptions
 
 @GlideModule
 class JellyfinGlideModule : AppGlideModule() {
-	override fun applyOptions(context: Context, builder: GlideBuilder) {
-		builder.setDefaultRequestOptions(
+	override fun applyOptions(context: Context, builder: GlideBuilder): Unit = with(builder) {
+		setDefaultRequestOptions(
 			// Set default disk cache strategy
-			RequestOptions()
-				.diskCacheStrategy(DiskCacheStrategy.RESOURCE)
+			RequestOptions().diskCacheStrategy(DiskCacheStrategy.RESOURCE)
 		)
 
 		// Silence image load errors
-		builder.setLogLevel(Log.ERROR)
+		setLogLevel(Log.ERROR)
 	}
 }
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/jellyfin/androidtv/JellyfinGlideModule.kt
```diff
@@ -0,0 +1,23 @@
...
+	override fun applyOptions(context: Context, builder: GlideBuilder) {
+		builder.setDefaultRequestOptions(
+			// Set default disk cache strategy
+			RequestOptions()
+				.diskCacheStrategy(DiskCacheStrategy.RESOURCE)
+		)
+
+		// Silence image load errors
+		builder.setLogLevel(Log.ERROR)
+	}
```
nielsvanvelzen, line 22:
```suggestion
	override fun applyOptions(context: Context, builder: GlideBuilder) = with(builder) {
		setDefaultRequestOptions(
			// Set default disk cache strategy
			RequestOptions().diskCacheStrategy(DiskCacheStrategy.RESOURCE)
		)

		// Silence image load errors
		setLogLevel(Log.ERROR)
	}
```

nielsvanvelzen (changes requested):

linetrimmer:
I think usages of `Glide.with` need to be replaced with `GlideApp.with` to actually use the custom module.

nielsvanvelzen:
> I think usages of `Glide.with` need to be replaced with `GlideApp.with` to actually use the custom module.

No I think that's what the annotation processor does, is it a horrible way of doing it? Yes

https://bumptech.github.io/glide/doc/configuration.html#applications

linetrimmer:
>No I think that's what the annotation processor does, is it a horrible way of doing it? Yes
 https://bumptech.github.io/glide/doc/configuration.html#applications

The Flickr sample app linked in this section also uses `GlideApp.with()` instead of `Glide.with()`.

I was looking at https://bumptech.github.io/glide/doc/generatedapi.html#using-the-generated-api where it says

> Applications can use the API by starting all loads with GlideApp.with() instead of Glide.with()

My bad if it's not necessary for this, though.

app/src/main/java/org/jellyfin/androidtv/JellyfinGlideModule.kt
thornbill: I had to add a `: Unit` in there, but should be good now.

nielsvanvelzen (approved):
Can you confirm using just `Glide.xxx` instead of `JellyfinGlideMoldule.xxx` works?

thornbill:
It is definitely honoring the log configuration when just using `Glide`. It may be a bit harder to verify the cache strategy is used. :wink:

nielsvanvelzen:
If the logging works it can pick up the config so the caching should work :p

PR #995, opened 2021-07-01
Remove unnecessary createPopupMenu util

thornbill:
**Changes**
This seems unnecessary

**Issues**
N/A


nielsvanvelzen (approved):
> Use "Gravity.END" instead of "Gravity.RIGHT" to ensure correct behavior in right-to-left locales

I guess those linter issues can be fixed in this PR

nielsvanvelzen (approved):

PR #1001, opened 2021-07-02
Refactor build scripts for consistency with jellyfin-sdk-kotlin

nielsvanvelzen:
This pull request refactors our Gradle scripts to be more consistent with the SDK repository and to enable the use of multiple Gradle modules in the future. (My plan is to write the new playback code in a separate module and I want to move the current preference code and custom views to their own modules too.)

**Changes**
- Move/rename linter configs
  - detekt.yml -> detekt.yaml
  - app/lint.xml -> android-lint.xml
- Remove dependency substitution for deprecated apiclient
  - A different method using mavenLocal is available for the SDK
- Remove signing config code
  - Previously replaced with apktool in the build pipeline
- Move gradle plugin versions to buildSrc
  - Single source of truth
- Move application dependencies to Gradle version catalog
  - Single source of truth
  - Should be easier for machines to parse for dependabot (or alternative)
- Support kotlin directory for sourcesets
 -  I'm not moving our current code (yet) but new modules will use "kotlin" instead of "java" as directory name
  - Added to app for easier copy+paste build config
- Re-order config files
- Move Maven repository list to buildSrc
  - Single source of truth

This PR does not update dependency versions.

**Issues**


nielsvanvelzen:
> (My plan is to write the new playback code in a separate module and I want to move the current preference code and custom views to their own modules too.)

Not for 0.12 though 😉

app/build.gradle.kts
```diff
@@ -63,6 +52,12 @@ android {
 			resValue("string", "app_name", "@string/app_name_debug")
 		}
 	}
+
+	lintOptions {
+		lintConfig = file("$rootDir/android-android-lint.xml")
```
nielsvanvelzen, line 57:
```suggestion
		lintConfig = file("$rootDir/android-lint.xml")
```

thornbill (approved):

PR #1003, opened 2021-07-03
Fall back to item title if loading logo fails in playback overlay

linetrimmer:
**Changes**

- Falls back to the itemTitle text if the logo fails to load in the playback overlay

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
@@ -5,6 +5,7 @@ import android.content.ComponentName;
 import android.content.Context;
 import android.content.Intent;
 import android.graphics.Color;
+import android.graphics.drawable.Drawable;
 import android.media.AudioManager;
 import android.os.AsyncTask;
 import android.os.Bundle;
@@ -37,6 +38,10 @@ import androidx.leanback.widget.Row;
 import androidx.leanback.widget.RowPresenter;
 
 import com.bumptech.glide.Glide;
+import com.bumptech.glide.load.DataSource;
+import com.bumptech.glide.load.engine.GlideException;
+import com.bumptech.glide.request.RequestListener;
+import com.bumptech.glide.request.target.Target;
 
 import org.jellyfin.androidtv.R;
 import org.jellyfin.androidtv.TvApp;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -1300,13 +1305,26 @@ public void updateDisplay() {
...
+                    .listener(new RequestListener<Drawable>() {
+                        @Override
+                        public boolean onLoadFailed(@Nullable GlideException e, Object model, Target<Drawable> target, boolean isFirstResource) {
+                            binding.itemLogo.setVisibility(View.GONE);
+                            binding.itemTitle.setVisibility(View.VISIBLE);
+                            return false;
+                        }
+
+                        @Override
+                        public boolean onResourceReady(Drawable resource, Object model, Target<Drawable> target, DataSource dataSource, boolean isFirstResource) {
```
nielsvanvelzen, line 1322: Put this method above the other one. All of our callback functions are in order of success,failed.

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -1300,13 +1305,26 @@ public void updateDisplay() {
             // Update the logo
             String imageUrl = ImageUtils.getLogoImageUrl(current, apiClient.getValue(), 440, false);
             if (imageUrl != null) {
-                Glide.with(requireContext())
-                        .load(imageUrl)
-                        .centerInside()
-                        .into(binding.itemLogo);
-                binding.itemLogo.setContentDescription(current.getName());
                 binding.itemLogo.setVisibility(View.VISIBLE);
                 binding.itemTitle.setVisibility(View.GONE);
```
nielsvanvelzen, line 1309: Shouldn't these by in the `onResourceReady` function?

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
linetrimmer: Placing them in the `onResourceReady` function causes the text to be pushed down while the logo is loading. Leaving them outside the function places the subtitle text where it should be once the logo is loaded so it doesn't shift before/after the logo is loaded. The subtitle text will still shift up a little if the logo fails to load since the title text has a smaller height than `itemLogo`, but ideally the logo will load successfully if we're in this section since the imageUrl is not null.

Screenshot_1625393489

nielsvanvelzen (approved):

PR #1006, opened 2021-07-05
fix  to check for PAUSED state when running startPauseReportLoop

### LINKED ISSUE #400: Media goes back to begining while playing
While playing media, sometimes the player just skips back to the beginning of the file, and starts playback from there. This is a bug that happens consistently. Sometimes it seeks back to the last point I seeked to, sometimes it seeks back to the beginning, and it always happens after 5-ish minutes, even less sometimes.

Device: Fire TV Stick, Version: 0.11.2

In case these help for some reason:
- [Media Info](https://e-six-two.one/i/qfamvr4o.png)
- [Server Logs](https://mystb.in/raw/evagozemod)
- [ffmpeg direct stream logs](https://mystb.in/raw/isudujimuj)

dhiaayachi:
**Changes**
This change is to permit startPauseReportLoop to run effectively and avoid a web socket timeout when pausing player. 

**Issues**
Fixes #400

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -1011,7 +1011,7 @@ public class PlaybackController {
                 }
 
                 if (mPlaybackState != PlaybackState.PAUSED) {
-                    // Playback was stopped, don't report progress anymore
+                    // Playback is not paused anymore, stop reporting
                     return;
                 }
 
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -1010,7 +1010,7 @@ public class PlaybackController {
                     return;
                 }
 
-                if (mPlaybackState != PlaybackState.PLAYING) {
+                if (mPlaybackState != PlaybackState.PAUSED) {
                     // Playback was stopped, don't report progress anymore
                     return;
                 }


MrChip53:
How reliably can you reproduce issue 400?

dhiaayachi:
Hi,
I reproduce it every time I pause a video in the player for more then 5 minutes.

The behavior I observe is that the keep alive loop when paused never kick off (because of the wrong check) and I get websocket timeout log in the player.

When I play back the video it continue to play normally for the remaining of the in memory buffer before randomly jump to a timestamp earlier in the video.

thornbill:
I'm not entirely sure this will entirely fix #400, because I don't believe that check exists in 0.11.x. This is certainly an impressive find though!

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -1010,7 +1010,7 @@ public void run() {
                     return;
                 }
 
-                if (mPlaybackState != PlaybackState.PLAYING) {
+                if (mPlaybackState != PlaybackState.PAUSED) {
                     // Playback was stopped, don't report progress anymore
```
thornbill, line 1014: The comment here doesn't seem accurate.

thornbill (changes requested):

dhiaayachi:
> I'm not sure this will entirely fix #400, because I don't believe that check exists in 0.11.x. This is certainly an impressive find though!

I did not reported #400 originally so I don't now.
I was annoyed by an issue in my setup and when I searched I found #400 and I supposed that's what I'm experiencing.
After being enough annoyed I decided to debug my self and this is how I found the issue fixed in here.
That said I'm running a debug version for few days now and I'm not experiencing the problem I noticed anymore.

The problem I'm experiencing on my fire tv is the following:
- I start a video
- after few seconds I pause it
- let it on pause for a minute or 2
- continue playing 
- after few seconds the video skip to an earlier time stamp sometimes the beginning of the video

dhiaayachi:
Is this 

> I'm not sure this will entirely fix #400, because I don't believe that check exists in 0.11.x. This is certainly an impressive find though!

I think it does, is this the code released for 0.11.x?
https://github.com/jellyfin/jellyfin-androidtv/blob/03d5b88d58bb82dc6195881dc65ded407772801c/app/src/main/java/org/jellyfin/androidtv/playback/PlaybackController.java#L1008

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
dhiaayachi: Fixed

nielsvanvelzen (approved):
I'm not sure if this actually fixes #400 but the change looks like it does fix _a_ bug.

dhiaayachi:
Thank you for approving this @nielsvanvelzen 
I see this tagged for 0.12 release, do you have an ETA for it?

nielsvanvelzen:
We're aiming for a beta release in the next few weeks

thornbill (approved):

PR #1010, opened 2021-07-08
Fix auto bitrate issues

nielsvanvelzen:
**Changes**
- Use `detectBitrate` function from the SDK to calculcate the "auto bitrate"
  - The function gives better results as the old apiclient did but the results are still low because the server API has a low limit making it hard to test the bitrates
- Set default value for max bitrate to 100 (Mbit/s) instead of 0 (auto)
- Use bitrate settings for music playback

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/Utils.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/Utils.java
@@ -1,10 +1,10 @@
 package org.jellyfin.androidtv.util;
 
+import static org.koin.java.KoinJavaComponent.get;
+
 import android.content.Context;
 import android.content.res.TypedArray;
 import android.media.AudioManager;
-import android.view.View;
-import android.widget.PopupMenu;
 import android.widget.Toast;
 
 import androidx.annotation.NonNull;
@@ -19,8 +19,6 @@ import java.util.Arrays;
 import java.util.Iterator;
 
 import timber.log.Timber;
-
-import static org.koin.java.KoinJavaComponent.get;
 
 /**
  * A collection of utility methods, all static.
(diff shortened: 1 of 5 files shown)


crobibero:
If the server endpoint is the limiting factor on upload size we can just change the limit for that endpoint

nielsvanvelzen:
I think increasing the limit should allow for better measurements as right now it maxes out at 10mb: (https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Api/Controllers/MediaInfoController.cs#L314)

While large media can easily use 20mbps

app/src/main/java/org/jellyfin/androidtv/util/Utils.java
```diff
@@ -143,8 +141,11 @@ public static boolean versionGreaterThanOrEqual(String firstVersion, String seco
     public static int getMaxBitrate() {
         String maxRate = get(UserPreferences.class).get(UserPreferences.Companion.getMaxBitrate());
         Long autoRate = get(AutoBitrate.class).getBitrate();
-        float factor = Float.parseFloat(maxRate) * 10;
-        return Math.min(autoRate != null && factor == 0 ? autoRate.intValue() : ((int) factor * 100000), 100000000);
+        if (maxRate == UserPreferences.MAX_BITRATE_AUTO && autoRate != null) {
```
thornbill, line 144: Shouldn't `.equals()` be used for the `maxRate` comparison?

thornbill (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/Utils.java
nielsvanvelzen: Right this is Java

thornbill (approved):

RohanKrishna500:
Hi,i got the same issues and i want to set bitrate to 30mbps because my internet averages around that speed and my files are under than bitrate

PR #1012, opened 2021-07-09
WIP: Add a daydream that shows random movies and series

nielsvanvelzen:
A "daydream" is the original name (and the name in the code) for Android screensavers. I'm not sure if they work on Fire TV but they do on Android TV and Google TV. The feature allows users to select a screensaver which automatically starts if the device is inactive for x minutes.

**Changes**
- Add LibraryDreamService to show random movies and series
  - When started shows the Jellyfin logo for 2 seconds
  - Then shows a random movie/serie every 5 seconds
  - Currently uses the KenBurnsView for the background for some sweet animations
    - Ideally the animation would be slower but the library doesn't let me change it :(

**Issues**


nielsvanvelzen:
Preview of the current daydream using the demo server:

https://user-images.githubusercontent.com/2305178/125517556-e1493a99-29c8-4918-a5c6-fbbf3de8cde7.mp4

Each item shows for 30 seconds with a subtle animation as to not distract users when the screensaver is playing in, for example, the living room. I'm still looking into the information about the item to show, right now it just shows the title.

Issue #658, opened 2020-12-20
HEVC 4k in Firestick v1


ManuLinares:
- [X] Issue does not happen in the browser (jellyfin-web)

Mediainfo: https://pastebin.com/xqQPSrbu

Logs: https://pastebin.com/j9j2R4jh

Fire TV (  ro.build.version.fireos : [5.2.7.8]  )
Fire TV Stick (1st Gen)
Jellyfin server 10.7.7

My firestick is 1080p but jellyfin server doesn't change the resolution, it sends in original ressolution 3840x1600, unless I lower the bitrate in android app settings.

attached pics of my TV.
.
![001](https://user-images.githubusercontent.com/3720020/102703623-ff6b3e00-424f-11eb-8134-9658ac1ec71a.jpg)
![002](https://user-images.githubusercontent.com/3720020/102703625-009c6b00-4250-11eb-91ec-5339cd4b1ac8.jpg)
![003](https://user-images.githubusercontent.com/3720020/102703626-009c6b00-4250-11eb-9b2d-6fc593a5a498.jpg)
![004](https://user-images.githubusercontent.com/3720020/102703627-01350180-4250-11eb-9442-6141c34327b5.jpg)
![005](https://user-images.githubusercontent.com/3720020/102703628-01cd9800-4250-11eb-8627-710102030d30.jpg)
![006](https://user-images.githubusercontent.com/3720020/102703629-02662e80-4250-11eb-90e3-67d9593e1d94.jpg)

ManuLinares:
It's still happening in the latest beta. I have to put the bitrate on 10mbps on the settings.

Isn't there a way that the app detects my 1080p max resolution and asks the jellyfin server for a *proper* resolution?

ManuLinares:
It's still happening in the latest beta. I have to put the bitrate on 10mbps on the settings.

Isn't there a way that the app detects my 1080p max resolution and asks the jellyfin server for a proper resolution?

ManuLinares:
Isn't there a way to limit the resolution to 1080p other that lowering the max bitrate to a ludicrous low number?

thornbill:
It should already be downscaling to 1080 on that device if you are on the latest version of the app.

https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.java#L48

Can you confirm the model name of the device? It should be shown at the bottom of the settings panel in the Jellyfin app.

[...]

PR #1013, opened 2021-07-09
Playback fixes for Fire TVs

thornbill:
**Changes**
* Fixes the supported h264 level for Fire TV devices
* Adds a profile condition for devices that only support 1080p video playback (Requires a server side fix https://github.com/jellyfin/jellyfin/pull/6274 to work in 10.7.x)

Ideally both of these would be covered by properly querying the devices capabilities, but I don't want to dive into that when we want to get a beta release out soon :tm:.

**Issues**
Maybe some playback issues :man_shrugging:

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.java
@@ -2,10 +2,23 @@ package org.jellyfin.androidtv.util;
 
 import android.os.Build;
 
+import java.util.Arrays;
+
 public class DeviceUtils {
     private static final String FIRE_TV_PREFIX = "AFT";
+    // Fire TV Stick Models
     private static final String FIRE_STICK_MODEL_GEN_1 = "AFTM";
+    private static final String FIRE_STICK_MODEL_GEN_2 = "AFTT";
+    private static final String FIRE_STICK_MODEL_GEN_3 = "AFTSSS";
+    private static final String FIRE_STICK_LITE_MODEL = "AFTSS";
     private static final String FIRE_STICK_4K_MODEL = "AFTMM";
+    // Fire TV Cube Models
+    private static final String FIRE_CUBE_MODEL_GEN_1 = "AFTA";
+    private static final String FIRE_CUBE_MODEL_GEN_2 = "AFTR";
+    // Fire TV (Box) Models
+    private static final String FIRE_TV_MODEL_GEN_1 = "AFTB";
+    private static final String FIRE_TV_MODEL_GEN_2 = "AFTS";
+    private static final String FIRE_TV_MODEL_GEN_3 = "AFTN";
 
     public static boolean isFireTv() {
         return Build.MODEL.startsWith(FIRE_TV_PREFIX);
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.java
```diff
@@ -14,6 +28,22 @@ public static boolean isFireTvStickGen1() {
         return Build.MODEL.equals(FIRE_STICK_MODEL_GEN_1);
     }
 
+    public static boolean isFireTvStick4k() {
+        return Build.MODEL.equals(FIRE_STICK_4K_MODEL);
+    }
+
+    public static boolean has4kVideoSupport() {
+        return !Arrays.asList(
```
thornbill, line 36: I'm inverting this to default to future or unknown devices supporting 4k playback.

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -86,6 +92,21 @@ object ProfileHelper {
...
+		arrayOf(
+			ProfileCondition(
+				ProfileConditionType.LessThanEqual,
+				ProfileConditionValue.Width,
+				"1920"
+			),
+			ProfileCondition(
+				ProfileConditionType.LessThanEqual,
+				ProfileConditionValue.Height,
+				"1080"
```
nielsvanvelzen, line 105: Wouldn't it make more sense to use `1200` as height so 16:10 video works?

thornbill: I would have to find a video to test that. It _looks_ like 1920x1080 is the max in the Amazon docs, but it's not entirely clear imo.l

thornbill: It looks like allowing `1200` for height produces the same error about exceeding device capabilities.

nielsvanvelzen (approved):

ManuLinares:
https://github.com/jellyfin/jellyfin-androidtv/issues/658

> > It should already be downscaling to 1080 on that device if you are on the latest version of the app.
> > https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.java#L48
> > Can you confirm the model name of the device? It should be shown at the bottom of the settings panel in the Jellyfin app.
> > This does require 10.7.7 as previous versions had broken support for setting a maximum video size.
> 
> Still doesn't work.
> 
> Server: **10.7.7** App: **0.12.3** release **Firestick v1 (AFTM)**
> 
> [Server log](https://pastebin.com/huNkSvk0) [ffmpeg log](https://pastebin.com/LET0ejwY)
> 
> [Video showing everything](https://github.com/jellyfin/jellyfin-androidtv/files/7556093/VID_20211117_124856_2.m4v.zip) [Video Mirror](https://drive.google.com/file/d/1_uz9QVOspOT-c_gjDC8BZufMuXNjbbIE/view?usp=sharing)

PR #1014, opened 2021-07-11
Use SDK device for values in legacy apiclient

nielsvanvelzen:
**Changes**
- Use SDK device for values in legacy apiclient

There is a small difference in the way the name is retrieved from the system, this change makes sure both the legacy apiclient and SDK use the same values.

**Issues**


thornbill:
...so what _is_ the small difference? :smile:

nielsvanvelzen:
The AndroidDevice `fromContext` function in the old apiclient uses another way to calculate the default device name.

nielsvanvelzen:
This might actually cause issues with special characters in the old apiclient, it's better to keep current behavior to prevent this from happening.

PR #1022, opened 2021-07-16
Remove TvApp.getApplication usages in util package

nielsvanvelzen:
Everything seems to work fine so I guess we could put this in 0.12...

**Changes**
- Remove as much `TvApp.getApplication()` usages in the util package as possible
  - Some usages still left due too no context in the call stack or requires `.getCurrentActivity()`

**Issues**


nielsvanvelzen:
Merge #1023 before this one

thornbill (commented):
There seems to be several files with imports added that are unused? Perhaps those are leftover from previous changes?

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/ItemListActivity.java
```diff
@@ -44,6 +44,7 @@
 import org.jellyfin.androidtv.util.ImageUtils;
 import org.jellyfin.androidtv.util.InfoLayoutHelper;
 import org.jellyfin.androidtv.util.MathUtils;
+import org.jellyfin.androidtv.util.MediaUtils;
```
thornbill, line 47: Looks like there is one more. :wink:

thornbill (approved):

PR #1042, opened 2021-07-26
Update androidx.window to alpha09

nielsvanvelzen:
**Changes**

Noticed some crashes in Tracepot that came from the androidx.window library. Hopefully updating it to a more recent version will fix the problem.

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
@@ -19,14 +19,7 @@ import androidx.fragment.app.FragmentActivity
 import androidx.fragment.app.add
 import androidx.window.WindowManager
 import com.bumptech.glide.Glide
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.Job
-import kotlinx.coroutines.MainScope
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
```diff
@@ -40,7 +33,7 @@ import java.util.concurrent.ExecutionException
 class BackgroundService(
 	private val context: Context,
 	private val apiClient: ApiClient,
-	private val userPreferences: UserPreferences
+	private val userPreferences: UserPreferences,
```
MrChip53, line 36: You committed this comma. Is that the new style for Kotlin or just preference? Did you mean to? Nothing follows.

nielsvanvelzen: Kotlin 1.4 added trailing comma support. I'm slowly adding it to files I touch (formatter does it for me :p)

MrChip53: Ok that's what I assumed. LGTM!

MrChip53 (approved):

PR #1050, opened 2021-07-29
Update Android Gradle Plugin to 7.0.2

nielsvanvelzen:
**Changes**

- Update Android Gradle Plugin to 7.0.0
  - Changes: https://developer.android.com/studio/releases/gradle-plugin#7-0-0

**Issues**

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -6,12 +6,12 @@ plugins {
 }
 
 android {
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
The new error is actually an existing error for a wrong translation, it can be ignored for now.

thornbill (approved):

PR #1053, opened 2021-07-29
Remove GSON usages when (de)serializing activity extras

nielsvanvelzen:
Someone thought it was a great idea to send JSON blobs between activities. They use Gson for serializing/deserializing. This PR changes most of them to use the BaseItemDto from the SDK and kotlinx.serialization. I tested everything except for the Live TV activities to work.

**Changes**
- Remove GSON usages when (de)serializing for extras
  - Use SDK BaseItemDto
  - Use kotlinx.serialization

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java
@@ -32,7 +32,7 @@ public class BrowseGridFragment extends StdGridFragment {
                 ItemFields.MediaStreams
         });
         query.setParentId(mParentId.toString());
-        if (mFolder.getType() == BaseItemType.UserView.toString() || mFolder.getType() == BaseItemType.CollectionFolder.toString()) {
+        if (mFolder.getType().equalsIgnoreCase(BaseItemType.UserView.toString()) || mFolder.getType().equalsIgnoreCase(BaseItemType.CollectionFolder.toString())) {
             String type = mFolder.getCollectionType() != null ? mFolder.getCollectionType().toLowerCase() : "";
             switch (type) {
                 case "movies":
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java
```diff
@@ -32,8 +31,8 @@ protected void setupQueries(IGridLoader gridLoader) {
                 ItemFields.MediaSources,
                 ItemFields.MediaStreams
         });
-        query.setParentId(mParentId);
-        if (mFolder.getBaseItemType() == BaseItemType.UserView || mFolder.getBaseItemType() == BaseItemType.CollectionFolder) {
+        query.setParentId(mParentId.toString());
+        if (mFolder.getType() == BaseItemType.UserView.toString() || mFolder.getType() == BaseItemType.CollectionFolder.toString()) {
```
thornbill, line 35: We should be using `.equals()` here.

app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
```diff
@@ -472,7 +472,7 @@ public void onClick(View v) {
             @Override
             public void onClick(View v) {
                 Intent intent = new Intent(getActivity(), SearchActivity.class);
-                intent.putExtra("MusicOnly", "music".equals(mFolder.getCollectionType()) || mFolder.getBaseItemType() == BaseItemType.MusicAlbum || mFolder.getBaseItemType() == BaseItemType.MusicArtist);
+                intent.putExtra("MusicOnly", "music".equals(mFolder.getCollectionType()) || mFolder.getType() == BaseItemType.MusicAlbum.toString() || mFolder.getType() == BaseItemType.MusicArtist.toString());
```
thornbill, line 475: We should be using `.equals()` here.

thornbill (changes requested):

thornbill (approved):

PR #1054, opened 2021-07-30
Update VideoManager.java

jakeapp:
**Changes**
Move from setScale to AspectRatioFrameLayout for better compatibility with common androidTV devices when scaling content.  Tested and works (although maybe vertical stretch should be removed...)

**Issues**
Fixes problem where zoom, horizontal stretch and vertical stretch don't work as advertised and shift content on screen in testing on a FireTV device.


nielsvanvelzen:
> although maybe vertical stretch should be removed...

Yes this change effectively makes "ZOOM_HORIZONTAL" fill the entire screen instead of just the horizontal part. So I think it's best to remove the "ZOOM_VERTICAL" option and rename the other one.

jakeapp:
I'll close this and open a new one, tweaked a few things, open to input.  Long term maybe a preference option to set default zoom mode...

Issue #1057, opened 2021-07-31
The Playback Rewrite


nielsvanvelzen:
# :rocket: The Playback Rewrite

We've often talked about the "playback rewrite" but there was no issue tracking it's progress. I'm not promising this issue to be up-to-date all the time but I'll use it myself to track progress on the rewrite.

The goal is to rewrite all playback related code: video playback, audio playback and live tv. The new code should be easier to maintain and be prepared for a migration to Jellyfin 11. New features like SyncPlay can be added without much trouble. All code is Kotlin-first and uses the new SDK.

We've mentioned the playback rewrite to be in a 0.13 release but we've since changed our release schedule and it will not be included in 0.13 anymore. We're not giving any ETA for when it will happen.

## :game_die: Process

The initial proof-of-concept will be written on a separate branch (on my fork) until the preparations are ready. At that point it will be PR'd to the master branch in a separate Gradle module. To use it a flag (in the build file?) needs to be set. This way we can release new versions of the app while developing the new playback code on the master branch. To support a rapid development cycle PR's that are not reviewed after 5 days may be self-merged if they only touch the playback module (changes to everything outside requires a review).

Issues and pull requests related to the new playback code will be added to the "playback rewrite" project on GitHub. The "playback" label will not be used until the playback rewrite is released to the public.

## :trophy: Milestones

This list will change, some features might not be available on initial release. When possible issues/pull requests will be linked for tracking purposes. There is no particular order for this list, new features might be added before restoring older features.

### :building_construction: Preparations

Various tasks that need to be completed to work on the rewrite.

- [x] Create new Kotlin-first SDK
- [x] Add new SDK
- [ ] Support multiple Gradle modules when building

### :fireworks: Proof of Concept

- [ ] Add playlist support - everything is a playlist, even single item playback
- [ ] Add bare-bones player UI
- [x] Add ExoPlayer backend
- [ ] Load video from server
- [ ] Report playback progress to server

### :arrows_counterclockwise: Restoring features

- [ ] Support transcoding/remuxing
- [ ] Add LibVLC backend
- [ ] Add remote control API
- [ ] Add intros
- [ ] Add chapter support
- [ ] Move "next up" screen to new playback code

### :hammer_and_wrench: Integration

- [ ] Use new playback module for audio (music, podcasts, audiobooks etc.)
- [ ] Use new playback module for video (movies, series etc.)
- [ ] Use new playback module for Live TV
- [ ] Delete old playback code

[...]

PR #1064, opened 2021-08-02
Move useExternalPlayer and getPlaybackActivityClass functions to new PlaybackLauncher interface

nielsvanvelzen:
This change is to make sure development for the new playback code can happen in the master branch while still being able to release updates with the old code. It does this by adding a new interface "PlaybackLauncher" that is used when launching the playback activity and a toggle for which implementation to use.

When the initial playback module is added this will launch an activity that reads the data from the current playback code (item to play etc.) and converts it to the new playback module format before launching the new playback module activity. This is a bit hacky but good enough for development.

The implementation toggle only works on debug builds of the app and will never show up on releases.

**Changes**
- Move useExternalPlayer and getPlaybackActivityClass functions to new PlaybackLauncher interface
- Implement PlaybackLauncher interface with old code
- "Implement" PlaybackLauncher interface for new playback code 
- Add toggle to enable new playback module

**Issues**

Part of #1057

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/di/PlaybackModule.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/di/PlaybackModule.kt
@@ -20,7 +20,7 @@ val playbackModule = module {
 
 	factory {
 		val preferences = get<UserPreferences>()
-		val useRewrite = preferences[UserPreferences.playbackRewriteEnabled] && BuildConfig.DEBUG
+		val useRewrite = preferences[UserPreferences.playbackRewriteEnabled] && BuildConfig.DEVELOPMENT
 
 		// TODO Inject PlaybackLauncher for playback rewrite here
 		if (useRewrite) RewritePlaybackLauncher()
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/jellyfin/androidtv/di/PlaybackModule.kt
```diff
@@ -13,4 +17,13 @@ val playbackModule = module {
...
 		)
 	}
+
+	factory {
+		val preferences = get<UserPreferences>()
+		val useRewrite = preferences[UserPreferences.playbackRewriteEnabled] && BuildConfig.DEVELOPMENT
+
+		// TODO Inject PlaybackLauncher for playback rewrite here
+		if (useRewrite) RewritePlaybackLauncher()
+		else GarbagePlaybackLauncher(get())
```
MrChip53, line 27: :+1:

MrChip53 (approved):

PR #1084, opened 2021-08-13
Add option to always require authentication

### LINKED ISSUE #1080: Option to re-prompt for password or d-pad pin entry during startup
Would love to see options to re-prompt for password or d-pad pin entry during startup similar to Jellyfin 11.x series.

**Additional context**	
12.x series seems to have adopted a token-based logon feature that bypasses the need to authenticate each time you log in. This is no doubt super convenient for many use cases. However having a multi-user household with a great range of different age groups (3, 12, 15, and OLD) some of us rely on the requiring log on to keep adult accounts separate.

Pervious versions used server-side settings to determine which accounts require password (or pin) entry, which accounts do not require password entry, and which accounts do/do not display on the login screen. Similar options if possible would be very much appreciated.

Thanks for all your hard work!

nielsvanvelzen:
This PR adds a new option at the bottom of the authentication preference screen called "always authenticate". When enabled it changes the behavior of the app:

**Changes**

- Option enabled behavior
  - Auto login is disabled
  - System users are disabled
    - System integrations are disabled (search, leanback, daydreams etc.)
  - User is required to enter the password when selecting a user
    - Access tokens are still stored in the system
- Option disabled behavior like it always was
- Option is hidden when not currently signed in ~~although a "kid account" could still disable it after signing in~~

**Issues**

Closes #1080

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
@@ -5,16 +5,16 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.emitAll
 import kotlinx.coroutines.flow.flow
 import org.jellyfin.androidtv.auth.model.*
+import org.jellyfin.androidtv.preference.AuthenticationPreferences
 import org.jellyfin.androidtv.util.ImageUtils
 import org.jellyfin.apiclient.interaction.device.IDevice
 import org.jellyfin.sdk.Jellyfin
 import org.jellyfin.sdk.api.client.KtorClient
 import org.jellyfin.sdk.api.client.exception.ApiClientException
-import org.jellyfin.sdk.api.client.extensions.authenticateUserByName;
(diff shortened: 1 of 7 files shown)


MrChip53 (approved):

MrChip53:
TODO: In future change to use pins when support is added instead of full password so this feature isn't so horrid.

nielsvanvelzen:
> TODO: In future change to use pins when support is added instead of full password so this feature isn't so horrid.

Yup, ideally we support quick connect (waiting for 10.8 for that) and maybe the pins

MrChip53:
> > TODO: In future change to use pins when support is added instead of full password so this feature isn't so horrid.
> 
> Yup, ideally we support quick connect (waiting for 10.8 for that) and maybe the pins

Ah yes. Forgot about quick connect. Would it work with this? I was really thinking maybe just a 4-6 key length key combo, set and stored locally, would be the fastest and easiest approach.

nielsvanvelzen:
Yeah it just shows the sign-in dialog so quick connect will work

PR #1087, opened 2021-08-14
Revert FF and RW

### LINKED ISSUE #863: Seeking in Exoplayer not working on Transcoded media
Jellyfin App, Master branch.

New method of seeking in Exoplayer not working on Transcoded media. Works on direct-play media.

I think this PR #581  introduced the problem. Maybe revert?

**To Reproduce**
1- Open any media that the server transcodes to mkv on jellyfin app.
2- Try to seek by pressing >> or << on remote (one time click or continuous button down).

**Expected behavior**
I expect on 1 click on remote >> or continuous pressing down button, media to seek properly.
On app v0.11.5 , the seeking by 30 seconds increment works on direct-play and transcoded files.

**System**
Firestick (1st Gen), Fire OS 5
Jellyfin server: v10.7.5

### LINKED ISSUE #1024: Inconsistent forward and back skip/seek in latest dev builds on 2017 Nvidia Shield
**Describe the bug**

I downloaded the July 14th build to try out the latest version of Jellyfin for Android TV.  There are some great improvements (thanks!) but one thing I notice is that skipping forward 30 seconds or back 5 seconds works inconsistently.  On Jellyfin 0.11 if I hit the left button on the remote it would skip back 5 seconds, and if I hit the left button 3 times it would skip back 5 seconds each time (15 seconds total).  If I hit the right button it would skip forward 30 seconds for however many times I hit it.  This is more or less the typical behavior in most of the apps I use on the Shield.

In the latest version of the app if I hit the left button once it will skip back 10 seconds, but then it seems the search bar gets focus and the next time I hit the left button it will move backwards to the next 30 second increment.  For example if my video is at 2:35 and I hit left once it will skip back to 2:25, if I hit left again the video will pause, it will skip back to 2:00, and any further left button presses will move back in 30 second increments.  Furthermore, the first button press will move the video back by 10 seconds, but subsequent button presses will move backwards in time on the seek bar, but not actual skip the video backwards unless you press the middle button, which will then change where you are in the video.  If you wait until the seek bar disappears or you hit the back button to dismiss the seek bar you can once again hit the left button to skip backwards by 10 seconds.  The same thing is true in the forward direction (right button press) as well.  

**To Reproduce**

1. Open any video in a 0.12 nightly build of the app
2. Allow the video to play for 1.5 minutes.
3. Hit the left button one time.  Note it skips backwards 10 seconds.
4. Hit the left button again, before the seek bar disappears.  Note the seek bar moves backwards to 1:00 the video pauses, but remains at 1:20.
5. Hit the middle button to skip back to 1 minute and resume play..

**Expected behavior**

Behavior should be consistent. Each left button press should skip back by 10 seconds without having to hit the middle button or pausing the vide, each right button press should skip forward by 30.  I can understand the value in being able to move back and forth on the seek bar, but I'd argue that a different button press should trigger it.  Perhaps long-pressing left/right?

 - Android TV 8.0.2
 - Device manufacturer: Nvidia
 - Device model: Shield 2017
 - Jellyfin server version: 10.7.6

MrChip53:
**Changes**
Reverts my ill-received ff and rw changes. Pulls the FF and RW interval from web settings to make up for it though.

**Issues**
Close #1024
Close #863 

PR #1011 will need to be adjusted to accommodate these changes.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/TvApp.java
+++ b/app/src/main/java/org/jellyfin/androidtv/TvApp.java
@@ -104,7 +104,13 @@ public class TvApp extends Application {
 
     @NonNull
     public DisplayPreferences getCachedDisplayPrefs(String key) {
-        return displayPrefsCache.containsKey(key) ? displayPrefsCache.get(key) : new DisplayPreferences();
+        return getCachedDisplayPrefs(key, DISPLAY_PREFS_APP_NAME);
+    }
+
+    @NonNull
+    public DisplayPreferences getCachedDisplayPrefs(String key, String app) {
+        String cacheKey = String.format("%s.%s", app, key);
+        return displayPrefsCache.containsKey(cacheKey) ? displayPrefsCache.get(cacheKey) : new DisplayPreferences();
     }
 
     public void updateDisplayPrefs(DisplayPreferences preferences) {
@@ -112,7 +118,7 @@ public class TvApp extends Application {
     }
 
     public void updateDisplayPrefs(String app, DisplayPreferences preferences) {
-        displayPrefsCache.put(preferences.getId(), preferences);
+        displayPrefsCache.put(String.format("%s.%s", app, preferences.getId()), preferences);
         apiClient.getValue().UpdateDisplayPreferencesAsync(preferences, getCurrentUser().getId(), app, new EmptyResponse());
         Timber.d("Display prefs updated for %s isFavorite: %s", preferences.getId(), preferences.getCustomPrefs().get("FavoriteOnly"));
     }
@@ -122,9 +128,11 @@ public class TvApp extends Application {
     }
 
     public void getDisplayPrefsAsync(final String key, String app, final Response<DisplayPreferences> outerResponse) {
-        if (displayPrefsCache.containsKey(key)) {
-            Timber.d("Display prefs loaded from cache %s", key);
-            outerResponse.onResponse(displayPrefsCache.get(key));
+        String cacheKey = String.format("%s.%s", app, key);
+
+        if (displayPrefsCache.containsKey(cacheKey)) {
+            Timber.d("Display prefs loaded from cache %s", cacheKey);
+            outerResponse.onResponse(displayPrefsCache.get(cacheKey));
         } else {
             apiClient.getValue().GetDisplayPreferencesAsync(key, getCurrentUser().getId(), app, new Response<DisplayPreferences>() {
                 @Override
@@ -132,9 +140,9 @@ public class TvApp extends Application {
                     if (response.getSortBy() == null) response.setSortBy("SortName");
                     if (response.getCustomPrefs() == null)
                         response.setCustomPrefs(new HashMap<String, String>());
-                    if (app.equals(TvApp.DISPLAY_PREFS_APP_NAME))
-                        displayPrefsCache.put(key, response);
-                    Timber.d("Display prefs loaded and saved in cache %s", key);
+
+                    displayPrefsCache.put(cacheKey, response);
+                    Timber.d("Display prefs loaded and saved in cache %s", cacheKey);
                     outerResponse.onResponse(response);
                 }
 
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -43,6 +43,7 @@ import org.jellyfin.apiclient.model.dlna.SubtitleDeliveryMethod;
 import org.jellyfin.apiclient.model.dto.BaseItemDto;
 import org.jellyfin.apiclient.model.dto.BaseItemType;
 import org.jellyfin.apiclient.model.dto.MediaSourceInfo;
+import org.jellyfin.apiclient.model.entities.DisplayPreferences;
 import org.jellyfin.apiclient.model.entities.LocationType;
 import org.jellyfin.apiclient.model.entities.MediaStream;
 import org.jellyfin.apiclient.model.entities.MediaStreamType;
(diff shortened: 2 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/TvApp.java
```diff
@@ -132,8 +132,8 @@ public void onResponse(DisplayPreferences response) {
                     if (response.getSortBy() == null) response.setSortBy("SortName");
                     if (response.getCustomPrefs() == null)
                         response.setCustomPrefs(new HashMap<String, String>());
-                    if (app.equals(TvApp.DISPLAY_PREFS_APP_NAME))
```
nielsvanvelzen, line 135: Are you sure `app` always equals `TvApp.DISPLAY_PREFS_APP_NAME` here?

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/VideoPlayerAdapter.java
```diff
@@ -33,13 +34,23 @@ public void pause() {
 
     @Override
     public void rewind() {
-        playbackController.rewind();
+        DisplayPreferences cachedPrefs = TvApp.getApplication() != null ? TvApp.getApplication().getCachedDisplayPrefs("usersettings") : null;
+
+        int skipMS = cachedPrefs != null && cachedPrefs.getCustomPrefs().get("skipBackwardLength") != null ?
+                Integer.parseInt(cachedPrefs.getCustomPrefs().get("skipBackwardLength")) : 10000;
+
+        playbackController.skip(-skipMS);
```
nielsvanvelzen, line 42: This bit of code is now in 2 places; this file and the overlay fragment. Is it possible to put it in just one place without too much trouble?

app/src/main/java/org/jellyfin/androidtv/TvApp.java
MrChip53: The issue is that it doesn't. The settings we want are part of the `emby` app(jellyfin-web). That line only allowed settings to be cached if it was part of `ATV` so I removed it to cache the `emby` settings so we don't have to fetch from server everytime.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/VideoPlayerAdapter.java
MrChip53: Yes, I could bump it back out to the playback controller I'm pretty sure.

app/src/main/java/org/jellyfin/androidtv/TvApp.java
nielsvanvelzen: So this cache is removed entirely in #1011 but until that's merged we need a reasonable solution.
If we use multiple keys and want to be 100% sure it won't cause conflict we should add the app name to the cache too like you suggest, but that requires changing the getCachedDisplayPrefs function parameters for the app name.

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -83,6 +83,7 @@
 import org.jellyfin.apiclient.model.dto.BaseItemType;
 import org.jellyfin.apiclient.model.dto.ChapterInfoDto;
 import org.jellyfin.apiclient.model.dto.UserItemDataDto;
+import org.jellyfin.apiclient.model.entities.DisplayPreferences;
```
nielsvanvelzen, line 86: Unused import?

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -898,36 +912,14 @@ public void onError(Exception exception) {
         public void run() {
             if (!isPlaying()) return; // in case we completed since this was requested
 
+            // TODO move this into skip() if we are direct playing
```
nielsvanvelzen, line 915: This todo is new, should it be solved now?

nielsvanvelzen:
Just tested this PR and it does not work. These are my settings in jf-web:
image

But it always goes forward/backward by 30 seconds.

MrChip53:
> Just tested this PR and it does not work. These are my settings in jf-web:
> image
> 
> But it always goes forward/backward by 30 seconds.

First, I realized I left an unused import so don't merge anyways, but the default rewind is supposed to be 10s. I tested this before rebasing and squashing and it worked for me. IF it is really going backward by 30 seconds are you sure you are not scrolling the seek bar that is broken into 30 second segments? Just want to clarify that because 30 second rewind doesn't add up at all.

nielsvanvelzen:
I tested the wrong thing, actually seeking with the seekbar uses 30s intervals, the forward/backward buttons in the overlay work like expected.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
MrChip53: Completely forgot about that. Latest push has it implemented.

nielsvanvelzen (approved):
Changes look fine and they work for me. Waiting for @thornbill if this is the behavior he wanted.

ManuLinares:
This doesn't fixes #863

Direct play works, but when media is transcoded the seekbar doesn't update when seeking with remote.

MrChip53:
> This doesn't fixes #863
> 
> Direct play works, but when media is transcoded the seekbar doesn't update when seeking with remote.
Will look into

MrChip53:
> > This doesn't fixes #863
> > Direct play works, but when media is transcoded the seekbar doesn't update when seeking with remote.
> 
> Will look into

The functionality refered to here was lost when the playback UI got redesigned and is not a direct result of my FF and RW changes. I assume you are referring to how when you click FF or RW in 0.11.x the blue bar would move accordingly and update your `xx:xx` position before playback would actually resume. @ManuLinares

thornbill (approved):
This seems to be working much better now! Thanks for fixing this @MrChip53 and apologies again for insulting your previous code. :sweat_smile:

MrChip53:
> This seems to be working much better now! Thanks for fixing this @MrChip53 and apologies again for insulting your previous code. :sweat_smile: 

No problem lol. I will probably go ahead and look into adjusting the seekbar and video position text, although, it will have to be by manually grabbing the seek bar and adjusting the position, so another hack like the ends at time.

PR #1092, opened 2021-08-17
Add PlaybackForwardingActivity for bridging to new playback module

nielsvanvelzen:
**Changes**

This is a small change that does not modify any behavior in the normal app flow (and thus is safe for 0.12 inclusion). It adds a new activity "PlaybackForwardingActivity" that is used to bridge all playback requests to the new playback module.
Right now it extracts the to-be-played item from the current playback code and retrieves the BaseItemDto from the api using the SDK. It then quits the activity. This code is backported from my playback scratch branch.

With this the app will not crash anymore when playback is started with the new playback module enabled.

**Issues**

Part of #1057

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -167,5 +167,8 @@
             android:name=".ui.playback.nextup.NextUpActivity"
             android:noHistory="true"
             android:screenOrientation="landscape" />
+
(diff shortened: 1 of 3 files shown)


MrChip53 (approved):
Code is pretty straightforward. I don't think the comments are necessary except the TODO but I'll leave that to your discretion, otherwise LGTM.

PR #1095, opened 2021-08-20
Remove vpx (unused) as a CodecType

GodTamIt:
**Changes**
Remove the unused `vpx` video codec, which is actually ambiguous between VP8 and VP9, as part of a bigger shift to be specific between the codecs in the VPx.

**Issues**
Progress on jellyfin/jellyfin#6455


nielsvanvelzen (approved):

nielsvanvelzen:
Thanks!

crobibero:
The server doesn’t currently support sending vp8/vp9 so I don’t think this should have been merged yet

nielsvanvelzen:
This PR only removed an unused constant from the ATV source, is does not change behavior.

PR #1096, opened 2021-08-22
Fix VLC error reporting and black screens

nielsvanvelzen:
**Changes**
- Fix VLC error reporting

  No idea why but the errors where ignored, this might help fixing some of our playback issues

- Fix VLC output not using android-display

  The onNewVideoLayout function we use to scale the video view does not work on the opengl video output from LibVLC

**Issues**

Possibly a few of them..

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VlcEventHandler.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VlcEventHandler.java
@@ -9,10 +9,21 @@ public class VlcEventHandler implements org.videolan.libvlc.MediaPlayer.EventLis
     private PlaybackListener onPreparedListener;
     private PlaybackListener onProgressListener;
 
-    public void setOnCompletionListener(PlaybackListener listener) { onCompletionListener = listener; }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/VlcEventHandler.java
```diff
@@ -26,6 +37,8 @@ public void onEvent(Event event) {
             case Event.PositionChanged:
                 if (onProgressListener != null) onProgressListener.onEvent();
                 break;
+            case Event.EncounteredError:
+                if (onErrorListener != null) onErrorListener.onEvent();
```
thornbill, line 41: Should probably have a `break` here to avoid falling through unintentionally.

nielsvanvelzen: oops

thornbill:
This appears to fix the scaling issue we have had with libVLC also. :tada:

thornbill (approved):

PR #1102, opened 2021-08-25
Restore direct play support for avi with libvlc

thornbill:
**Changes**
Reverts the changes made in #604 since a proper fix for the scaling issue was added in #1096.

**Issues**
N/A


nielsvanvelzen (approved):
Reverting the revert

PR #1138, opened 2021-09-08
Remove unused tile colors and update default theme

thornbill:
**Changes**
* Removes unused tile attributes
* Updates default theme colors

The "smart screens" were the worst offenders of using completely random colors, but it can also be seen in the red background for user avatars and the various shades of blue or green used for placeholders throughout the app. This reduces the colors used to only those in [this theme](https://coolors.co/031a6b-033860-1e76ba-00a4dc-aa5cc3).

| This PR | Current |
|--|--|
| tile colors | current tile colors |

**Issues**
N/A

### DIFF
--- a/app/src/main/res/values/attrs_tiles.xml
+++ b/app/src/main/res/values/attrs_tiles.xml
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <attr name="tile_port_person_bg" format="color" />
-    <attr name="tile_port_record_bg" format="color" />
     <attr name="tile_port_series_timer_bg" format="color" />
     <attr name="tile_port_server_bg" format="color" />
     <attr name="tile_port_time_bg" format="color" />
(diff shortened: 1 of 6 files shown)


app/src/main/res/values/attrs_tiles.xml
```diff
@@ -14,11 +13,9 @@
     <attr name="tile_artists_bg" format="color" />
     <attr name="tile_audio_bg" format="color" />
     <attr name="tile_chapter_bg" format="color" />
-    <attr name="tile_edit_bg" format="color" />
     <attr name="tile_genres_bg" format="color" />
     <attr name="tile_land_folder_bg" format="color" />
     <attr name="tile_land_photo_bg" format="color" />
-    <attr name="tile_land_series_timer_bg" format="color" />
```
nielsvanvelzen, line 21: This one should be used in `tile_land_series_timer.xml` but it uses the wrong one there. Can you fix that in this PR? (probably rename the other one to `tile_land_series_timer_bg` now)

thornbill: Good catch. It looks like both port and land are needed.

nielsvanvelzen (approved):

Issue #1135, opened 2021-09-07
Sort by star rating sorts wrong (All Items / grid view)


JannikHoelling:
**Describe the bug**
When i sort movies on the grid layout by rating(star rating) they are not sorted correctly. The first movies are sorted like this: 6.4, 8.0, 8.1, 8.1, 8.5, 6.4 and it goes on randomly sorted. I can't see any system in the sorting there. The sorting is the same everytime i refresh.  

Ciritic rating actually works like expected and is sorting correctly.

**To Reproduce**

1. Go to a Tv Show or Movie library
2. Go to "All Items"(Grid layout view)
3. Click on the sort option in the top right corner (AZ)
3. Select "Rating"
4. Movies will not be correctly sorted

**Expected behavior**
Movies should be sorted by their rating. In the example above the 8.5 Movie should appear before the 6.4 one.

**Logs**
No error logs available

**Screenshots**
https://imgur.com/a/XFiSj0J

**System (please complete the following information):**
 - Android TV version: v0.12.0-beta.6
 - Device manufacturer: TCL
 - Device model: TCL U49P6046
 - Jellyfin server version: 10.7.7

nielsvanvelzen:
The sorting works as "intended" although it might look broken because the value of the field used for sorting is not shown for the movies:

- The "Rating" sort option uses the OfficialRating field
- The rating shown with a star is the CommunityRating field
- The rating shown with a tomato is the CriticRating field

So the order is not obvious for you because the sorting uses a field that is not shown.

PR #1140, opened 2021-09-09
add star rating sort to grid view

JannikHoelling:
**Changes**
Added sorting by star rating(called CommunityRating in the jellyfin-api) to the sortOption HashMap
Tested it in the android studio emulator and it worked.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-androidtv/issues/1135

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/GridFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/GridFragment.java
@@ -152,8 +152,8 @@ public class GridFragment extends Fragment {
         sortOptions.put(1, new SortOption(TvApp.getApplication().getString(R.string.lbl_date_added), "DateCreated,SortName", SortOrder.Descending));
         sortOptions.put(2, new SortOption(TvApp.getApplication().getString(R.string.lbl_premier_date), "PremiereDate,SortName", SortOrder.Descending));
         sortOptions.put(3,new SortOption(TvApp.getApplication().getString(R.string.lbl_rating), "OfficialRating,SortName", SortOrder.Ascending));
-        sortOptions.put(4,new SortOption(TvApp.getApplication().getString(R.string.lbl_stars), "CommunityRating,SortName", SortOrder.Descending));
-	sortOptions.put(5,new SortOption(TvApp.getApplication().getString(R.string.lbl_critic_rating), "CriticRating,SortName", SortOrder.Descending));
+        sortOptions.put(4,new SortOption(TvApp.getApplication().getString(R.string.lbl_community_rating), "CommunityRating,SortName", SortOrder.Descending));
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/GridFragment.java
```diff
@@ -152,8 +152,9 @@ public SortOption(String name, String value, SortOrder order) {
         sortOptions.put(1, new SortOption(TvApp.getApplication().getString(R.string.lbl_date_added), "DateCreated,SortName", SortOrder.Descending));
         sortOptions.put(2, new SortOption(TvApp.getApplication().getString(R.string.lbl_premier_date), "PremiereDate,SortName", SortOrder.Descending));
         sortOptions.put(3,new SortOption(TvApp.getApplication().getString(R.string.lbl_rating), "OfficialRating,SortName", SortOrder.Ascending));
-        sortOptions.put(4,new SortOption(TvApp.getApplication().getString(R.string.lbl_critic_rating), "CriticRating,SortName", SortOrder.Descending));
-        sortOptions.put(5,new SortOption(TvApp.getApplication().getString(R.string.lbl_last_played), "DatePlayed,SortName", SortOrder.Descending));
+        sortOptions.put(4,new SortOption(TvApp.getApplication().getString(R.string.lbl_stars), "CommunityRating,SortName", SortOrder.Descending));
```
nielsvanvelzen, line 155: I'm not sure about the used label, We might want to use "Community Rating" or something like that as label.

JannikHoelling: I can add a new string called lbl_community_rating. To make it consistent i can also change the enum RatingType to those labels. There RATING_TOMATOES with lbl_tomatoes and RATING_STARS with lbl_stars is still used. 
If that would be ok i'm gonna do that.

The jellyfin webui is currently using IMDb Rating and Critic Rating as labels. That would also be an option.

thornbill: I would recommend using the same labels as web for consistency.

nielsvanvelzen: I think we should not use the imdb name as it's trademarked, same reason that we don't use it when choosing which rating to show on cards.
Besides, the value is not always an imdb rating. different metadata provider plugins can put different values in it.

JannikHoelling: Should we also rename the option in the settings to Critic and Community Rating? Would make it equal everywhere.

nielsvanvelzen (approved):

thornbill (approved):

PR #1142, opened 2021-09-10
Show unfinished episodes in watch next row

nielsvanvelzen:
The watch next row is used by Android TV (leanback) and Google TV to display media content. Until now we've only showed "next up" items (unwatched episodes). This PR adds resumable episodes (episodes not fully watched). I specifically filtered out movies because their posters won't be displayed nicely (they would be resized to 16:9).

I've also marked the "my media" row (shows all libraries) as a default row, this means that users upgrading from 0.11 -> 0.12 will automatically see this row on their home screen.

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
@@ -105,7 +105,7 @@ class LeanbackChannelWorker(
 	 * update or create the channel. The [name] parameter is used to store the id and should be
 	 * unique.
 	 */
-	private fun getChannelUri(name: String, settings: Channel): Uri {
+	private fun getChannelUri(name: String, settings: Channel, default: Boolean = false): Uri {
 		val store = context.getSharedPreferences("leanback_channels", Context.MODE_PRIVATE)
 
 		val uri = if (store.contains(name)) {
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
Funny enough the code to show the episode's progress was already written but never activated :p

mueslimak3r:
> I specifically filtered out movies because their posters won't be displayed nicely (they would be resized to 16:9).

Is there a way to make movies look 'good enough' so they can be displayed too? Perhaps by sizing down their image, preserving aspect ratio, and having black bars on the sides. It may be worthwhile trade off for many users for the benefit of having all media available, and not just TV.

Another option could be to use the 'cover' photo instead of the poster. Or to have a settings that allows users to choose to use a resized poster, cover photo, or to hide movies. This could be formatted similar to the "home" user settings on the server, where you can select what media appears.

nielsvanvelzen:
> > I specifically filtered out movies because their posters won't be displayed nicely (they would be resized to 16:9).
> 
> Is there a way to make movies look 'good enough' so they can be displayed too? Perhaps by sizing down their image, preserving aspect ratio, and having black bars on the sides. It may be worthwhile trade off for many users for the benefit of having all media available, and not just TV

Yes this is possible and will be added eventually, but it's out of the scope for this pull request,

thornbill (approved):
Could you just use the thumb image for movies if one exists?

nielsvanvelzen:
> Could you just use the thumb image for movies if one exists?

Probably but I'd rather add it properly in a separate PR.

PR #1144, opened 2021-09-11
Fix NPE when autologin is disabled

MrChip53:
**Changes**
Move the restore session call to onCreate in StartupActivity to fix the race condition.

**Issues**


MrChip53:
@nielsvanvelzen I'm honestly not sure if moving away from the generic lifecycle event and into the specific activity has any ill effects so if you would like a different implementation

MrChip53:
Unused import I left in again. Damn auto import on copy/paste lol

nielsvanvelzen:
This PR effectively reverts an important part of #1082 so in it's current state it's a no-go.

MrChip53:
#1145 instead

PR #1145, opened 2021-09-11
Fix crash when autologin is off

nielsvanvelzen:
Please test this thoroughly before approve/merge 

**Changes**
- Use `destroyCurrentSession` instead of `setCurrentSession` with null value
- Change `destroyCurrentSession` to set api etc. to null in a **synchronous** way instead of calling `setCurrentSession`

**Issues**
Supersedes #1144


MrChip53 (approved):
Good fix let's moon this baby! :rocket: :rofl:

thornbill (approved):
I could no longer cause a crash with these changes! :tada:

PR #1155, opened 2021-09-23
Add multiple versions support

danieladov:
Adds a button to choose which video versions you want to watch when more than one is available


tomaswl1:
how does this work does the different file versions have to be in the same folder as its not working for me?

nielsvanvelzen:
You need to have multiple versions of a file, how that works can be found in the documentation (https://jellyfin.org/docs/general/server/media/movies.html#multiple-versions-of-a-movie). It will then show a button to select the version in the overview screen for that item.

tomaswl1:
Thank you very much that worked appreciate your time

PR #1172, opened 2021-10-03
Replace UUIDSerializer with SDK provided version

nielsvanvelzen:
The SDK implementation is the same, I copied it to this repo when we didn't use the SDK yet.

**Changes**

- Replace UUIDSerializer with SDK provided version

**Issues**


nielsvanvelzen:
See https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/master/jellyfin-model/src/jvmMain/kotlin/org/jellyfin/sdk/model/serializer/UUIDSerializer.kt

Biggest difference is that the SDK uses a class instead of object for KMPP purposes

thornbill (approved):

PR #1173, opened 2021-10-03
Remove unused ApiClient extensions

nielsvanvelzen:
**Changes**

- Remove unused ApiClient extensions
- Remove unused LoadingState class (left over from earlier login stuff)

**Issues**


nielsvanvelzen:
Apparently I removed them in another PR too, so after a rebase only the LoadingState removal is still there.

PR #1178, opened 2021-10-04
Explicitly specify android:exported

ElegyD:
**Changes**
Removes the following warning in Android Studio:
> When using intent filters, please specify android:exported as well

It is true per default for activities and receivers that do have intent filters.


nielsvanvelzen (approved):
Thank you! We went from 1823 linter issues to 1820.

PR #1180, opened 2021-10-05
Fix several lint issues

ElegyD:
**Changes**
Use property access syntax
Unified static import ordering (`import static` before `import`)
Added missing/Removed unnecessary `@NonNull` `@Nullable` annotations
Removed unnecessary semicolons
Removed unused imports

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/GridFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/GridFragment.java
@@ -5,10 +5,10 @@ import android.view.Gravity;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.FrameLayout;
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
+import androidx.annotation.NonNull;
 import androidx.fragment.app.Fragment;
 import androidx.leanback.widget.BaseGridView;
 import androidx.leanback.widget.ObjectAdapter;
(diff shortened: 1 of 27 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/GridFragment.java
```diff
@@ -294,7 +294,7 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
     }
 
     @Override
-    public void onViewCreated(View view, Bundle savedInstanceState) {
+    public void onViewCreated(@NonNull View view, Bundle savedInstanceState) {
```
nielsvanvelzen, line 297: Why did you add NonNull for the first parameter but did nothing for the second parameter?

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -110,7 +110,7 @@
     private Display.Mode[] mDisplayModes;
     private boolean refreshRateSwitchingEnabled;
 
-    public PlaybackController(List<BaseItemDto> items, IPlaybackOverlayFragment fragment) {
+    public PlaybackController(List<BaseItemDto> items, @Nullable IPlaybackOverlayFragment fragment) {
```
nielsvanvelzen, line 113: Why did you add Nullable for the second parameter but did nothing for the first parameter?

ElegyD: I reverted this one, because all calls to it don't pass null and shouldn't probably.

PR #1213, opened 2021-10-19
Add preferences UI for home sections

nielsvanvelzen:
**Changes**

Add a new preference category "Home" allowing the user to change home sections - no need to do it in jf-web anymore!

image image image

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/constant/HomeSectionType.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/constant/HomeSectionType.kt
@@ -1,6 +1,8 @@
 package org.jellyfin.androidtv.constant
 
+import org.jellyfin.androidtv.R
 import org.jellyfin.androidtv.preference.PreferenceEnum
+import org.jellyfin.androidtv.ui.preference.dsl.EnumDisplayOptions
 
 /**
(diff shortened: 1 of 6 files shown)


MrChip53 (dismissed):
Homesection 1 says not set. It is supposed to be my media.

Screenshot from 2021-10-20 22-12-14

I did some investigating and concluded its because small library titles has hidden = true. Are these settings then synced to the server after they are changed or are they stored locally? If they are synced to the server it would probably be best to not hide this option even though in ATV we treat it the same as library buttons it seems.

dkanada:
Mentioned this in chat but I forget if there was a conversation about it, I would suggest an unordered list rather than manually including each section as a preference. People might want to set one or two sections instead for example.

thornbill (approved):
:+1: lgtm... could we force the home screen to reload when these preferences are changed?

nielsvanvelzen:
Unfortunately not possible without significant changes, ideally we want to reload the home section in the `onResume()` function so it updates every time the home screen is opened (closing prefs, ending playback etc.). But since the code that "refreshes" the rows just throws the current ones away before loading the sections/items it would cause the currently selected item to reset and screen to flash.

PR #1217, opened 2021-10-21
Use default transition in PhotoPlayer 

nielsvanvelzen:
**Changes**

- Use default transition in PhotoPlayer 
  - Removes the copied code that changes a single line
  - Also removes a bunch of stuff from MathUtils
- Some more unused code cleaning

**Issues**

> _"You do nothing but remove code here."_
> \- @MrChip53 2021


Maxr1998 (approved):
Seems good, though the transition change might need an additional approval by ATV folks.

PR #1222, opened 2021-10-22
Merge ByLetterActivity and ByGenreActivity into a single activity

nielsvanvelzen:
**Changes**
- Merged 3 activities into 1
- Removed one of the fragments because it wasn;t used
- merged the base of the fragment because only 1 usage
- remove garbage code never called or used from other code
- more cleaning stuff
- help this code is a mess

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdBrowseFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdBrowseFragment.java
@@ -3,9 +3,7 @@ package org.jellyfin.androidtv.ui.browsing;
 import static org.koin.java.KoinJavaComponent.inject;
 
 import android.os.Bundle;
-import android.os.Handler;
 import android.view.Gravity;
-import android.view.KeyEvent;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.FrameLayout;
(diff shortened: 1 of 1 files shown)


mueslimak3r (commented):
It builds and runs fine 👍
The amount of consolidation is impressive

Potential changes:
* It would be great to have a comment at the top of BrowseFolderFragment that explains its purpose as a parent to the genre and letter fragments. It's not clear from the name or the contents of the file (and it looks like it's only used as a parent for those).

Questions:
* I assume the key and message listeners, and refreshRows are removed from stdBrowseFragment because they aren't used, is that correct?

* If refreshRows isn't used, is that because the rows never need to be refreshed, or because it's handled elsewhere?

nielsvanvelzen:
> It would be great to have a comment at the top of BrowseFolderFragment that explains its purpose as a parent to the genre and letter fragments. It's not clear from the name or the contents of the file (and it looks like it's only used as a parent for those).

This PR is just an in-between state. The goal is to move everything "browsing" related to a single activity with proper viewmodels. This includes: the home screen, detail screen, library views, search screen and possibly some more I forgot about.

> I assume the key and message listeners, and refreshRows are removed from stdBrowseFragment because they aren't used, is that correct?

Yeah if I removed them and the code compiles it means they weren't used

> If refreshRows isn't used, is that because the rows never need to be refreshed, or because it's handled elsewhere?

Most UI isn't reactive, the parts that are reactive are mostly cheating (like the home screen that removes all data and rebuilds all the rows from scratch).

mueslimak3r (approved):
looks good!

PR #1226, opened 2021-10-23
Only show the nextup button for a series after a valid response from nextUpQuery

mueslimak3r:
**Changes**
These changes will hide the 'play next up' button by default for series and then query the server with the same request that pressing the button would. If the query finds episodes then the button will be shown and given focus.

**Issues**
Often series will end up without 'next up' episodes. Currently the android-tv client will always show the 'play next up' button for series even when invalid. Clicking on the button results in the toast "Unable to find next up episode".


mueslimak3r:
Couldn't find a good solution

PR #1243, opened 2021-11-07
fix to default preferred video player not actually using exoplayer

mueslimak3r:
**Changes**
updates migrations so that exoplayer is the new default preferred video player

**Issues**
Even though #1212 changed the companion object value, the new setting didn't apply for new or updated installs.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -185,11 +185,12 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 	init {
 		// Migrations
 		// v0.10.x to v0.11.x: Old migrations
+		// this is also used for debug versions
 		migration(toVersion = 2) {
 			// Migrate to video player enum
 			// Note: This is the only time we need to check if the value is not set yet because the version numbers were reset
 			if (!it.contains("video_player"))
-				putEnum("video_player", if (it.getBoolean("pref_video_use_external", false)) PreferredVideoPlayer.EXTERNAL else PreferredVideoPlayer.AUTO)
+				putEnum("video_player", if (it.getBoolean("pref_video_use_external", false)) PreferredVideoPlayer.EXTERNAL else PreferredVideoPlayer.EXOPLAYER)
 		}
 
 		// v0.11.x to v0.12.x: Migrates from the old way of storing preferences to the current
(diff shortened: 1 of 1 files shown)


mueslimak3r:
I'm not sure if I applied this to the correct migrations, and if having it twice is okay. For the debug build I need to have it in the `toVersion = 2` migration.

Also, should the default `liveTvVideoPlayer` be exoplayer too?

nielsvanvelzen:
## Preference defaults

A default value for a preference is used when there is no known value. When a user changes a preference (including to the same value as the default) the user's preference is stored and the default will be ignored. The only way to get back to a default is by using the `reset` function (which we don't currently use or most preferences).

Changing a default value thus only applies to new users and users that haven't changed the particular preference.

## How do migrations work

Migration use incremental identifiers for their definition. A single migration works somewhat like a transaction. When the user opens the app the migration definitions are read and compared to the currently stored version of the store. For example: a new user will go trough all migrations, a user upgrading from 0.11->0.12 should go to migration 3, 4 and 5.

Migrations are normally somewhat bound to version releases. So migration 2 migrates 0.10 preferences to 0.11 and migration 3 from 0.11 to 0.12. Two exceptions here are migrations 4 and 5.

We don't normally use migrations to change existing values because the defaults can be used for that.

## Goal of #1212

With the "Change default video player to ExoPlayer instead of auto detect" we wanted to change the default video player to ExoPlayer. The PR did exactly that. It did not, however, change the value from users that manually changed their video player option. Some might have changes it to Exo and then back to auto - those people are still using the auto option. But this is intended, as we don't want people that use LibVLC only to suddenly go to ExoPlayer only. So the PR did exactly what we intended.

## Changing the default video player for Live TV

I've been told that LibVLC works more reliably for Live TV, apparently ExoPlayer doesn't work at all in some cases. I can't verify this myself as I (still) don't have Live TV setup on my Jellyfin server.

## This PR

So with the above in mind, if we really wanted to put everyone on ExoPlayer in a .z release (like 0.12.4) we'd have to write a new migration (as all the current ones are already applied for users running 0.12).

Modifying existing migrations is normally done only in dev builds for migrations not in a release. As developers should be able to fix their installation manually.

---

I hope this clears up the confusion a bit, thanks for thinking along with us.

mueslimak3r:
Thank you for the thorough explanation, the use-case for using and not using migrations makes more sense now.

My reason for looking for a way to force the preferences change was because I noticed that the preferred player setting was auto after clearing the play store release's cache, data, uninstalling, and then reinstalling. All the other settings went to the defaults.

PR #1244, opened 2021-11-07
Add MigrationContext for improved preference migration behavior

nielsvanvelzen:
To fix an issue where new installs would execute all migrations I made some changes to the code.

**Changes**
- Add new MigrationContext
  - Implementation unaware, meaning we can add migrations for display preferences at some point
  - Works in 2 steps:
    1. Define migrations
    2. Run migrations
  - Skips migrating when current store version is -1 (new install)
  - Skips migrations already executed (previous behavior)
  - All migration run in their own transaction (previous behavior)
- Remove .migration from SharedPreferencesStore and add .runMigrations instead
- Merge UserPreferences migrations meant for 0.11->0.12 into a single block. (keeps the toVersion for compatibility with current installs)

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/SharedPreferenceStore.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/SharedPreferenceStore.kt
@@ -1,6 +1,7 @@
 package org.jellyfin.androidtv.preference
 
 import android.content.SharedPreferences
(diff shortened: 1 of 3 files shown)


mueslimak3r (approved):
Tested:
* fresh-installing
* installing a different debug build and changing some user preferences, and then installing this build

Defaults are used on fresh installs, as they should be
My preferences weren't modified when I installed this build over another debug build
looks good!

PR #1248, opened 2021-11-10
Misc playback adjustments (d-pad click can play/pause, stop drift during pause, libvlc time fix)

mueslimak3r:
1) during transcoded playback libvlc `getTime()` and other libvlc methods don't know its playback ticks or duration. libvlc `getTime()` reports ms since playback started, not actual position. The patch in this PR adds that value to the position in the media when playback began.

2) clicking the d-pad center button when the exo glue overlay is hidden no longer just calls `pause()`. It now calls `playPause()`. This allows for predictable behavior, and avoids any potential issues from excessive calls to `pause()`.

3) when playback is paused there is a slight drift between the position the player reports and playbackController's mCurrentPosition. My patch for that is to update mCurrentPosition during `PauseReportLoop`.


mueslimak3r:
Will clean up, squash, and reopen a new one when ready for review

PR #1251, opened 2021-11-17
Fixed libvlc reporting incorrect currentPosition when transcoding

mueslimak3r:
**Changes**
libvlc `getTime()` reports ms from stream start during transcode. Now, that value alone is not used to report position.

In videoManager a new property `mMetaVLCStreamStartPosition` stores the timestamp of the media where playback began.
This value is added to the return from `getTime` to get an accurate timestamp from the player.

The values being added is done in `getCurrentPosition()`

The value of `mMetaVLCStreamStartPosition` is set before `videoManager.start()` is called, and before seeking an mkv transcode.

**Issues**
during transcoded playback, libvlc `getTime()` is used by `videoManager` to get the current player position in `getCurrentPosition()`, which is used to display timestamps in the overlay and report progress to the server. `getTime()` only reports ms since stream start in this situation, and that breaks things like the seekbar, timestamps, and sending erroneous reports to the server.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -305,7 +305,7 @@ public class PlaybackController {
     }
 
     private void play(long position, int transcodedSubtitle) {
-        Timber.d("Play called with pos: %d and sub index: %d", position, transcodedSubtitle);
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (approved):

nielsvanvelzen:
There is a merge conflict in PlaybackController.java. Can you rebase the PR @mueslimak3r?

PR #1252, opened 2021-11-17
prevent double pausing, allow d-pad center to unpause with overlay hidden

mueslimak3r:
**Changes**
1) `CustomPlaybackOverlayFragment` will now `playPause()` when it catches a
`DPAD_CENTER` keyEvent instead of just calling `pause()`. If allowing playing in this context isn't desired, 2) will prevent double pausing and this change could be excluded.

2) `pause()` in `playbackController` now checks if already paused and returns if it is to prevent double pausing.

**Issues**
When the seekbar is interacted with `pause()` is always called. 1) stops this from potentially causing issues

When playback is paused and the overlay is hidden `CustomPlaybackOverlayFragment` called `pause()` on d-pad center keyEvent, even if already paused.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -793,7 +793,7 @@ public class PlaybackController {
         // if playback is paused and the seekbar is scrubbed, it will call pause even if already paused
         if (mPlaybackState == PlaybackState.PAUSED) {
             Timber.d("already paused, ignoring");
-            return ;
+            return;
         }
         mPlaybackState = PlaybackState.PAUSED;
         mVideoManager.pause();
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -790,6 +790,11 @@ public void onError(Exception ex) {
     }
 
     public void pause() {
+        // if playback is paused and the seekbar is scrubbed, it will call pause even if already paused
+        if (mPlaybackState == PlaybackState.PAUSED) {
+            Timber.d("already paused, ignoring");
+            return ;
```
nielsvanvelzen, line 796:
```suggestion
            return;
```

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

PR #1255, opened 2021-11-19
Choose player only once

mueslimak3r:
**Changes**
Finalize the player choice and adhere to that choice.

When I player is choosen ensure that it's used since the server gives a response based on that choice, and the player is fed parameters and player options specific to that choice. Feeding parameters meant for, for instance libVLC to exoplayer, may cause issues

**Issues**
Player parameters are generated for libVLC or Exo in play and playInternal, and then startItem may override that choice but still use the parameters generated in prior steps


mueslimak3r:
This is a fairly major restructure so I'm just looking for feedback

PR #1258, opened 2021-11-19
Fix now-playing row not updating when queue changes

mueslimak3r:
**Changes**
when queue is changed, remove the nowplaying row and re-add it

**Issues**
now-playing row would only generate once and never update unless audio playback stopped

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentNowPlayingRow.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentNowPlayingRow.kt
@@ -34,8 +34,12 @@ class HomeFragmentNowPlayingRow(
 
 	fun update(rowsAdapter: ArrayObjectAdapter) {
 		if (mediaManager.isPlayingAudio) {
-			remove(rowsAdapter)
-			add(rowsAdapter)
+			if (row != null) {
+				rowsAdapter.remove(row)
+				row = ListRow(HeaderItem(context.getString(R.string.lbl_now_playing)), mediaManager.managedAudioQueue)
+				rowsAdapter.add(0, row)
+			}
+			else add(rowsAdapter)
 		}
 		else remove(rowsAdapter)
 	}
--- a/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentNowPlayingRow.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentNowPlayingRow.kt
@@ -33,7 +33,10 @@ class HomeFragmentNowPlayingRow(
 	}
 
 	fun update(rowsAdapter: ArrayObjectAdapter) {
-		if (mediaManager.isPlayingAudio) add(rowsAdapter)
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentNowPlayingRow.kt
```diff
@@ -33,7 +33,10 @@ class HomeFragmentNowPlayingRow(
 	}
 
 	fun update(rowsAdapter: ArrayObjectAdapter) {
-		if (mediaManager.isPlayingAudio) add(rowsAdapter)
+		if (mediaManager.isPlayingAudio) {
+			remove(rowsAdapter)
+			add(rowsAdapter)
```
nielsvanvelzen, line 38: I think this should use the update function in the adapter

https://developer.android.com/reference/androidx/leanback/widget/ArrayObjectAdapter#replace

mueslimak3r: done. I'm not super familiar with the nuances of garbage collection in java. I hope just reassigning `row` to a new `ListRow` doesn't cause a leak.

37: `if (row != null) {`
38: `row = ListRow(HeaderItem(context.getString(R.string.lbl_now_playing)), mediaManager.managedAudioQueue)`
39: `rowsAdapter.replace(0, row)`

mueslimak3r:
I would also like to prevent pausing from destroying the row/stopping playback. Pause should == pausing. Then there could be a stop button added to the nowplaying activity.

nielsvanvelzen:
The `NowPlayingBug` uses `mediaManager.hasAudioQueueItems` instead of `mediaManager.isPlayingAudio`. That should probably work to still show the row when paused.

mueslimak3r:
> The `NowPlayingBug` uses `mediaManager.hasAudioQueueItems` instead of `mediaManager.isPlayingAudio`. That should probably work to still show the row when paused.

Done. It seems to work as expected. My only concern is that the queue remains after pausing, playing a movie, and stopping the movie. However, that's not an issue with the `nowPlayingRow` but a (bug/**feature**) of the `mediaManager`

app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
```diff
@@ -193,7 +193,12 @@ private void reportProgress() {
 
     private void onComplete() {
         ReportingHelper.reportStopped(mCurrentAudioItem, mCurrentAudioStreamInfo, mCurrentAudioPosition);
-        nextAudioItem();
+        if (hasNextAudioItem()) {
+            nextAudioItem();
+        }
+        else if (hasAudioQueueItems()){
```
nielsvanvelzen, line 199:
```suggestion
        } else if (hasAudioQueueItems()) {
```

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

Issue #1260, opened 2021-11-19
After update 1.12.3, the playback has drastically worsened.


JVT038:
Basically as the title says.
I'm using android tv and the player used to work normally.

But after updating, it can't resume where I left off. It says "Can't search" (at least, that's the best translation I can think of right now).

Anyway, when I start playing something, the app will suddenly freeze after around 30min of watching. No idea why. I have to reboot my TV to fix this.

And sometimes, when it doesn't freeze, it just randomly jumps back to the beginning of the video for no reason. 

And lastly, the subtitles used to load in very smoothly. No freezing of the movie I'm watching. The movie would just play normally and subtitles would appear as soon as it was done loading. It was great.
Now, after the update, the movie freezes for a couple of seconds while loading the subtitles, which is kind of annoying. Though it isn't as annoying as any of the other issues I mentioned.

I hope the devs will fix this soon, because it's really annoying to use the android tv app right now. 

Extra info: I use a Philip's Smart TV with Android 8.0.0 and an Android kernel with version 3.10.79. My Jellyfin server is version 10.7.7

JVT038:
So in the log files, I see this a lot:
`[2021-11-19 21:18:34.051 +00:00] [WRN] [69] Jellyfin.Server.Middleware.ResponseTimeMiddleware: Slow HTTP Response from "https://jellyfin.mydomain.com/Sessions/Playing/Progress" to "myipaddress" in 0:00:00.9589688 with Status Code 204`
What's 204? Perhaps this has something to do with it?

nielsvanvelzen:
Are you using ExoPlayer, LibVLC or Auto as preferred video player?

JVT038:
> Are you using ExoPlayer, LibVLC or Auto as preferred video player?

I'm using Auto

mueslimak3r:
> > Are you using ExoPlayer, LibVLC or Auto as preferred video player?
> 
> I'm using Auto

Can you test using exoplayer and libVLC selected as preferred and see if one or both have these issues?

Another way to tell which one it's using when this issue happens is to look for the resize/zoom icon. It's only in exoplayer:

![20211119_230901.jpg](https://user-images.githubusercontent.com/28824542/142717979-e07f2b40-d59b-4ba3-a6e1-73d39ef0d716.jpg)

JVT038:
The bugs appear in both Exoplayer and libvlc

layanto:
[...]

PR #1261, opened 2021-11-20
reworked playbackController position tracking and reporting system

mueslimak3r:
**Changes**
1) use tmp position value for seeking instead of modifying mCurrentPosition right away
2) new method for updating mCurrentPosition. This is called not only from progressListener, but from the audio and subtitle switching methods to make sure everything is synced
3) queries to playbackController for currentPosition will return seeked position if relevant
4) seekbar will hopefully be **extra** stable now. no more jumping from 0:00 to resume position
5) adjusted directplay seeking to prevent seekbar pos irregularities
6) startReportLoop calls stopReportLoop before starting. I saw a log where it looked like there were two instances running. This should prevent that.

**Issues**
playbackController would report wrong position during seeking and when initializing playback
seekbar starts at 0:00 then jumps to resume position during transcode
seekbar occasionally jumped around still during seeking
Hopeful fixes for #1260 and #1239

**Other notes**
I left the debug printouts in this one so people can test it and send logs
Based on the couple issues opened since 0.12.3, this is an attempt to reassess my approach to "fixing" transcoding seeking.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -305,32 +305,27 @@ public class PlaybackController {
     }
 
     // central place to update mCurrentPosition
-    // use getRealTimeProgress for liveTV
-    // use mSeekedPosition while seeking
-    // when videoManager is actually playing reset mSeekedPosition and stop using it
     private void refreshCurrentPosition() {
         long newPos = -1;
 
         if (updateProgress == true) {
             if (isLiveTv && mCurrentProgramStartTime > 0) {
-                Timber.d("PlaybackController - using live-tv time: %s", newPos); // debug - will remove
                 newPos = getRealTimeProgress();
+                // live tv
             }
             else if (mVideoManager != null) {
                 if (!isPlaying() && mSeekedPosition != -1) {
                     newPos = mSeekedPosition;
-                    Timber.d("PlaybackController - using SeekedPos: %s", newPos); // debug - will remove
+                    // use seekedPosition until playback starts
                 }
                 else if (isPlaying()) {
                     newPos = mVideoManager.getCurrentPosition();
                     mSeekedPosition = -1;
-                    Timber.d("PlaybackController - real time: %s", newPos); // debug - will remove
+                    // playback is happening - get current position and reset seekedPosition
                 }
             }
         }
-        else {
-            Timber.d("PlaybackController - keeping original time: %s", mCurrentPosition); // debug - will remove
-        }
+        // use original value if new one isn't available
         mCurrentPosition = newPos != -1 ? newPos : mCurrentPosition;
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -71,7 +71,7 @@
     private Lazy<MediaManager> mediaManager = inject(MediaManager.class);
 
     List<BaseItemDto> mItems;
-    VideoManager mVideoManager;
+    VideoManager mVideoManager = null;
```
nielsvanvelzen, line 74: Isn't it null already by default? 🤔

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -381,6 +414,9 @@ public void onClick(DialogInterface dialog, int which) {
                 isLiveTv = item.getBaseItemType() == BaseItemType.TvChannel;
                 startSpinner();
 
+                // undo setting mSeekedPosition for liveTV
+                if (isLiveTv) { mSeekedPosition = -1;  }
```
nielsvanvelzen, line 418:
```suggestion
                if (isLiveTv) mSeekedPosition = -1;
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -1048,10 +1109,11 @@ public void run() {
...
                 } else {
                     // do the seek
-                    if (mVideoManager.seekTo(position) < 0)
+                    if (mVideoManager.seekTo(position) < 0) {
                         Utils.showToast(TvApp.getApplication(), TvApp.getApplication().getString(R.string.seek_error));
-
-                    mPlaybackState = PlaybackState.PLAYING;
+                        pause();
+                    }
+                    else { mPlaybackState = PlaybackState.PLAYING; }
```
nielsvanvelzen, line 1116:
```suggestion
                    } else {
                        mPlaybackState = PlaybackState.PLAYING;
                    }
```

nielsvanvelzen (changes requested):
Works fine in emulator 👍

mueslimak3r:
fixed the formatting, and all new code I write will use the preferred formatting. I also took out the `= null`. As a C person I do that by habit but I see that java (mostly) does it for me for objects.

I also see a new issue opened: #1266 that I think may be caused by my previous "fix" for seeking. I suspect this because that change only sets `mCurrentPosition`'s value upon seek, and it is likely changed again in the `onProgressListener` before the seeking finished. In the context of the seek that occurs upon playback first starting, this could cause instability.
I haven't been able to reproduce those issues on my end, but the changes in this PR force the use of the current position to be consistent within `playbackController` and when its requested through `getCurrentPosition()`

nielsvanvelzen (approved):

PR #1263, opened 2021-11-22
audioNowPlaying button state, queue state, and screensaver fixes

mueslimak3r:
**Changes**
* Use the player's onProgressListener to update the playPause button state after an event is caught. A slight negative side-effect of this change is that navigating into the activity when playback is paused shows the playPause button in the "playing" state briefly and then correctly changes. I think this is a fair trade-off to have the playPause button display correctly at all times.

* Fixed queue row not scrolling to selected song `onResume`

* Gave the `mPlayPauseButton` button the same `mainAreaFocusListener` as the other buttons in that row for consistency

* tweaked the screensaver activation timer so all user interactions would delay its opening

* Button presses while screen saver is active now close the screen saver when appropriate instead of activating buttons/behaving as if the screen saver wasn't active

* modified `KeyProcessor` so its method `createItemMenu` returns the created menu

* the popup menu opened by selecting an item from the queue will be dismissed by the screensaver activating or the queue changing

* playPause button is now focused when the screensaver is closed

**Issues**
* When playback first starts, `AudioNowPlayingActivity` catches `onResume` and `onQueueStatusChanged` before playback actually starts or has the ability to query `mediaManager`. Because of that, `isPlayingAudio()` is always false is that context. This is why the delay is needed when updating the `playPause` button state by checking `isPlaying()` in the `onResume` and `onQueueStatusChanged` methods

* When activity is resumed the queue would be positioned at the beginning

* the screensaver didn't prevent interactions with the buttons/items underneath it

* the popup menu from clicking on a queue item stayed open with the screen saver active

**To do for another time**
* fix focus on queue row not scrolling to the bottom of the view
* -done- fix nowPlaying "screen saver" not preventing interaction with the nowPlayingActivity UI
* fix performing "remove from queue" on currently playing song causing multiple subsequent songs to display the "playing" triangle icon

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingActivity.java
@@ -354,6 +354,8 @@ public class AudioNowPlayingActivity extends BaseActivity {
                 if (currentItem != mBaseItem) {
                     // new item started
                     loadItem();
+                    // immediately move the queue row to the current song
+                    // disable queue refresh in onProgress since the queue position is already set
                     if (mAudioQueuePresenter != null && mediaManager.getValue().hasAudioQueueItems()) {
                         mAudioQueuePresenter.setPosition(mediaManager.getValue().getCurrentAudioQueuePosition());
                         shouldRefreshQueue = false;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingActivity.java
```diff
@@ -330,7 +335,13 @@ public void onProgress(long pos) {
         public void onQueueStatusChanged(boolean hasQueue) {
             if (hasQueue) {
                 loadItem();
-                updateButtons(mediaManager.getValue().isPlayingAudio());
+                //Make sure our initial button state reflects playback properly accounting for late loading of the audio stream
+                mLoopHandler.postDelayed(new Runnable() {
```
nielsvanvelzen, line 339: I'm not a fan of delaying code, this currently happens a lot in the app and is really a sign of bad code. It introduces all kinds of race conditions and bad UX (buttons showing up later).

I'd prefer a solution that doesn't need a handler for this.

mueslimak3r:
removed all the delayed code and moved the logic to the `audioEventListener`'s `onProgress`, which only runs the code if a flag and/or incremental update condition is met.

I also fixed the screensaver allowing buttons to be interacted with and the screensaver should now switch tracks properly

mueslimak3r:
migrated the `mediaManager` changes to #1273. I will open a new PR with just the `audioNowPlaying` changes

PR #1274, opened 2021-11-28
allow setting the nextup timer duration to infinite

mueslimak3r:
**Changes**
* in the preferences display class 0s is displayed as the infinity character
* in `nextUpButtons` if the value of the timer is 0s it won't start the timer

**Issues**
https://features.jellyfin.org/posts/1383/next-up-episode-queue-without-autoplay

**Notes**
100 is the actual value used to check if the timer is 0s. I'm not sure where that comes from but 0s = 100, 1s = 1100, etc

If this is an acceptable implementation then it might be good to add a migration for users who have their timeout set to 0s, and change it to 1s

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
@@ -9,20 +9,20 @@ import android.widget.FrameLayout
 import android.widget.ProgressBar
 import androidx.appcompat.widget.AppCompatButton
 import org.jellyfin.androidtv.R
-import org.jellyfin.androidtv.preference.UserPreferences
-import org.koin.core.component.KoinComponent
-import org.koin.core.component.get
 
 class NextUpButtons(
 	context: Context,
 	attrs: AttributeSet? = null,
 	defStyleAttr: Int = 0,
 	defStyle: Int = 0
-) : FrameLayout(context, attrs, defStyleAttr, defStyle), KoinComponent {
+) : FrameLayout(context, attrs, defStyleAttr, defStyle) {
 	constructor(context: Context, attrs: AttributeSet) : this(context, attrs, 0, 0)
 
 	private var countdownTimer: CountDownTimer? = null
 	private val view = View.inflate(context, R.layout.fragment_next_up_buttons, null)
+
+	var countdownTimerEnabled: Boolean = true
+	var duration: Long = 6
 
 	init {
 		addView(view)
@@ -35,15 +35,13 @@ class NextUpButtons(
 		}
 	}
 
-	fun startTimer(): Boolean {
-		val duration = get<UserPreferences>()[UserPreferences.nextUpTimeout].toLong()
+	fun startTimer() {
 
 		// Cancel current timer if one is already set
 		countdownTimer?.cancel()
 
-		// value of 0 disables timer
-		if (duration == 0L) {
-			return false
+		if (!countdownTimerEnabled) {
+			return
 		}
 
 		// Create timer
@@ -60,7 +58,6 @@ class NextUpButtons(
 				view.findViewById<AppCompatButton>(R.id.fragment_next_up_buttons_play_next).performClick()
 			}
 		}.start()
-		return true
 	}
 
 	fun stopTimer() {
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
@@ -41,8 +41,8 @@ class NextUpButtons(
 		// Cancel current timer if one is already set
 		countdownTimer?.cancel()
 
-		// for some reason a timer duration of 0s == 100
-		if (duration == 100L) {
+		// value of 0 disables timer
+		if (duration == 0L) {
 			return false
 		}
 
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
@@ -21,8 +21,8 @@ class NextUpButtons(
 	private var countdownTimer: CountDownTimer? = null
 	private val view = View.inflate(context, R.layout.fragment_next_up_buttons, null)
 
-	var countdownTimerEnabled: Boolean = true
-	var duration: Long = 6
+	var countdownTimerEnabled: Boolean = false
+	var duration: Long = 0
 
 	init {
 		addView(view)
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
@@ -40,9 +40,7 @@ class NextUpButtons(
 		// Cancel current timer if one is already set
 		countdownTimer?.cancel()
 
-		if (!countdownTimerEnabled) {
-			return
-		}
+		if (!countdownTimerEnabled) return
 
 		// Create timer
 		countdownTimer = object : CountDownTimer(duration, 1) {
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
@@ -35,11 +35,16 @@ class NextUpButtons(
 		}
 	}
 
-	fun startTimer() {
+	fun startTimer(): Boolean {
 		val duration = get<UserPreferences>()[UserPreferences.nextUpTimeout].toLong()
 
 		// Cancel current timer if one is already set
 		countdownTimer?.cancel()
+
+		// for some reason a timer duration of 0s == 100
+		if (duration == 100L) {
+			return false
+		}
 
 		// Create timer
 		countdownTimer = object : CountDownTimer(duration, 1) {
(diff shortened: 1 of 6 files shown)


nielsvanvelzen:
You actually found a bug in the DurationSeekbarPreference. The `min` value doesn't work. The seekbar should actually go from 3 until 30.
I'm not sure why it defaults to 100 as min though.

mueslimak3r:
> You actually found a bug in the DurationSeekbarPreference. The `min` value doesn't work. The seekbar should actually go from 3 until 30. I'm not sure why it defaults to 100 as min though.

From what I can tell, there isn't a bug in DurationSeekbarPreference.

Instead, It looks like the underlying `SeekBarPreference` doesn't work with `min` values over 100. From what I can tell in the docs, the current version of the androidx.preferences library should support this (on API >= 26), but on my Google TV it will still default to 100 if I increase `min`, and setting values <= 100 work fine. I'm not sure if there's something I'm missing as far as something being out of date.

Assuming `min` is limited to 100, a possible solution is to rework the slider to use seconds (3 - 30), and including this "infinite timeout" 3 - 31 with 31 displaying the infinity character. The issue with that is it would require a migration to divide users' current preference by 1000

If that all sounds like too much hassle than it's worth, a range of 1000 - 30000 with 100 being unlimited isn't the worst option.

mueslimak3r:
updated the code to use a valid range of 0 - 30,000 with 0 disabling the timeout and showing the infinity character

app/src/main/java/org/jellyfin/androidtv/ui/preference/category/playback.kt
```diff
@@ -53,11 +53,11 @@ fun OptionsScreen.playbackCategory(
 	seekbar {
 		setTitle(R.string.pref_next_up_timeout_title)
 		setContent(R.string.pref_next_up_timeout_summary)
-		min = 3000
+		min = 0
 		max = 30000
 		increment = 1000
 		valueFormatter = object : DurationSeekBarPreference.ValueFormatter() {
-			override fun display(value: Int) = "${value / 1000}s"
+			override fun display(value: Int) = if (value == 0) "∞" else "${value / 1000}s"
```
nielsvanvelzen, line 60: I'd prefer a when statement here for readability

nielsvanvelzen (changes requested):
Instead of using preferences in a view we should set the preferences from the NextUpFragment.

At line 33-36 (the apply{} call) we could pass two values: timerEnabled and timerDuration. That way we don't need to use preferences inside of a view.

Additionally, setting `timerStarted` to true in the onStart function should always happen, even if the timer is disabled, the boolean is to prevent a timer from starting after the activity was paused.

mueslimak3r:
> Instead of using preferences in a view we should set the preferences from the NextUpFragment.
> 
> At line 33-36 (the apply{} call) we could pass two values: timerEnabled and timerDuration. That way we don't need to use preferences inside of a view.
> 
> Additionally, setting `timerStarted` to true in the onStart function should always happen, even if the timer is disabled, the boolean is to prevent a timer from starting after the activity was paused.

I'm definitely a Kotlin noob but I think I did this right.
* moved the preference getting and the duration setting to `NextUpFragment`
* removed `startTimer` returning a Boolean and use the original logic for `timerStarted`
* in `playback.kt` use `when {}`
* in `playback.kt` change `increment` to 3s so the original intended minimum is still used (inf/disabled, 3s, 6s, etc)
* in `UserPreferences` changed default timeout from 7s to 6s to match the changed increment

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpFragment.kt
```diff
@@ -31,6 +34,11 @@ class NextUpFragment : Fragment() {
 		}
 
 		binding.fragmentNextUpButtons.apply {
+			duration = get<UserPreferences>()[UserPreferences.nextUpTimeout].toLong()
```
mueslimak3r, line 37: I'm unsure about how I'm setting duration here. In `NextUpButtons` I don't set `duration` or `countdownTimerEnabled` as nullable.
The intention is for duration to be the default value defined in `NextUpButtons` if for some reason getting the user preference fails/returns null.

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
```diff
@@ -9,21 +9,21 @@ import android.widget.FrameLayout
...
 	defStyle: Int = 0
-) : FrameLayout(context, attrs, defStyleAttr, defStyle), KoinComponent {
+) : FrameLayout(context, attrs, defStyleAttr, defStyle) {
 	constructor(context: Context, attrs: AttributeSet) : this(context, attrs, 0, 0)
 
 	private var countdownTimer: CountDownTimer? = null
 	private val view = View.inflate(context, R.layout.fragment_next_up_buttons, null)
 
+	var countdownTimerEnabled: Boolean = true
+	var duration: Long = 6
```
nielsvanvelzen, line 25:
```suggestion
	var countdownTimerEnabled: Boolean = false
	var duration: Long = 0
```

So we can keep the defaults in just one place (UserPreferences)

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
```diff
@@ -36,11 +36,14 @@ class NextUpButtons(
...
 
 	fun startTimer() {
-		val duration = get<UserPreferences>()[UserPreferences.nextUpTimeout].toLong()
 
 		// Cancel current timer if one is already set
 		countdownTimer?.cancel()
 
+		if (!countdownTimerEnabled) {
+			return
+		}
```
nielsvanvelzen, line 45:
```suggestion
		if (!countdownTimerEnabled) return
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpFragment.kt
```diff
@@ -7,10 +7,13 @@ import android.view.ViewGroup
...
 import org.jellyfin.androidtv.data.service.BackgroundService
 import org.jellyfin.androidtv.databinding.FragmentNextUpBinding
+import org.jellyfin.androidtv.preference.UserPreferences
 import org.koin.android.ext.android.inject
 import org.koin.androidx.viewmodel.ext.android.sharedViewModel
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.get
 
-class NextUpFragment : Fragment() {
+class NextUpFragment : Fragment(), KoinComponent {
```
nielsvanvelzen, line 16:
```suggestion
class NextUpFragment : Fragment(){
```

AFAIK koin is already available in fragments

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpFragment.kt
nielsvanvelzen, line 37: inject UserPreferences as class property like backgrroundservice etc.

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpButtons.kt
nielsvanvelzen, line 25:
```suggestion
	var duration: Int = 6
```

Also use an int as the preference is an int too. Integers are big enough for this usecase. Alternatively a Kotlin Duration instance could be used but that's probably overkill.

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpFragment.kt
nielsvanvelzen: If this part of the code fails the complete fragment wouldn't work (setters for callbacks are below this line)
current approach seems fine to me, set the duration first and then checking if it should disable

app/src/main/java/org/jellyfin/androidtv/ui/preference/category/playback.kt
```diff
@@ -53,11 +53,14 @@ fun OptionsScreen.playbackCategory(
 	seekbar {
 		setTitle(R.string.pref_next_up_timeout_title)
 		setContent(R.string.pref_next_up_timeout_summary)
-		min = 3000
+		min = 0
 		max = 30000
-		increment = 1000
+		increment = 3000
```
nielsvanvelzen, line 58: Increment should stay at 1 second, not 3

app/src/main/java/org/jellyfin/androidtv/ui/preference/category/playback.kt
```diff
@@ -53,11 +53,14 @@ fun OptionsScreen.playbackCategory(
...
 		valueFormatter = object : DurationSeekBarPreference.ValueFormatter() {
-			override fun display(value: Int) = "${value / 1000}s"
+			override fun display(value: Int) = when {
+				value == 0 -> "∞" // disables the timer and displays infinity character
```
nielsvanvelzen, line 61: It might be better to create a constant for the 0/infinite value somewhere and use it both here and in the nextupfragment.

Additionally, the string should be in strings.xml so use getString here for the infinite character.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/nextup/NextUpFragment.kt
```diff
@@ -31,6 +34,11 @@ class NextUpFragment : Fragment() {
 		}
 
 		binding.fragmentNextUpButtons.apply {
+			duration = userPreferences[UserPreferences.nextUpTimeout]
+			countdownTimerEnabled = when {
+				duration == NEXTUP_TIMER_DISABLED -> false
+				else -> true
+			}
```
nielsvanvelzen, line 41:
```suggestion
			countdownTimerEnabled = duration != NEXTUP_TIMER_DISABLED
```

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

PR #1284, opened 2021-12-14
fixed add artist to queue, added shuffle artist

mueslimak3r:
**Changes**
* added "add to queue" and "shuffle all" buttons to `fullDetailsActivity` for music artists
* updated `playbackHelper`'s artist query so all song by an artist can be played, arranged by album, and songs keep their order within the album

**Issues**
* `FullDetailsActivity.addItemToQueue()` would pass `mBaseItem` to `mediaManager` even when `mBaseItem` was a "parent" item like an artist or an album
(I added logic for albums just in case even though they arent handled here. There is logic elsewhere in `FullDetailsActivity` for `MusicAlbum` so I added it just to be safe)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
@@ -830,9 +830,17 @@ public class FullDetailsActivity extends BaseActivity implements IRecordingIndic
     }
 
     private void addItemToQueue() {
-        if (mBaseItem.getBaseItemType() == BaseItemType.Audio) {
-            mediaManager.getValue().addToAudioQueue(Arrays.asList(mBaseItem));
-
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -830,9 +830,17 @@ private String getEndTime() {
     }
 
     private void addItemToQueue() {
-        if (mBaseItem.getBaseItemType() == BaseItemType.Audio) {
-            mediaManager.getValue().addToAudioQueue(Arrays.asList(mBaseItem));
-
+        if (mBaseItem.getBaseItemType() == BaseItemType.Audio || mBaseItem.getBaseItemType() == BaseItemType.MusicAlbum || mBaseItem.getBaseItemType() == BaseItemType.MusicArtist) {
+            if (mBaseItem.getBaseItemType() == BaseItemType.MusicAlbum || mBaseItem.getBaseItemType() == BaseItemType.MusicArtist) {
+                PlaybackHelper.getItemsToPlay(mBaseItem, false, false, new Response<List<BaseItemDto>>() {
```
nielsvanvelzen, line 835: This is technically using the legacy java apiclient 👀

nielsvanvelzen (approved):

PR #1290, opened 2021-12-15
prevent unwanted interactions when music player screensaver active

mueslimak3r:
**Changes**
* made `KeyProcessor.createItemMenu()` public so it can be called directly, and it now returns the created menu so the caller can dismiss it
* if the created item menu is open during various events it's closed (screensaver starting, queue replaced)
* added checks for all interactive objects to account for the screensaver being active
* updated the animation for the screensaver to make the fade in and out look the same
* play/pause button gains focus when the screensaver is stopped for user convenience

**Issues**
* the screensaver being active didn't prevent interactions with objects like the buttons or the queue items
* the dropdown menu opened by clicking on a queue item would stay open if the screensaver starts and after a queue replace (which could result in the clicked item no longer being in the queue)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingActivity.java
@@ -12,6 +12,7 @@ import android.view.KeyEvent;
 import android.view.View;
 import android.widget.ImageButton;
 import android.widget.ImageView;
+import android.widget.PopupMenu;
 import android.widget.ProgressBar;
 import android.widget.RelativeLayout;
 import android.widget.ScrollView;
@@ -101,6 +102,8 @@ public class AudioNowPlayingActivity extends BaseActivity {
     private Lazy<BackgroundService> backgroundService = inject(BackgroundService.class);
     private Lazy<MediaManager> mediaManager = inject(MediaManager.class);
 
+    private PopupMenu mPopupMenu;
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
@@ -129,12 +132,18 @@ public class AudioNowPlayingActivity extends BaseActivity {
 
         mPlayPauseButton = findViewById(R.id.playPauseBtn);
         mPlayPauseButton.setContentDescription(getString(R.string.lbl_pause));
+        mPlayPauseButton.setOnFocusChangeListener(mainAreaFocusListener);
         mPrevButton = findViewById(R.id.prevBtn);
         mPrevButton.setContentDescription(getString(R.string.lbl_prev_item));
         mPrevButton.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View v) {
-                mediaManager.getValue().prevAudioItem();
+                if (ssActive) {
+                    stopScreenSaver();
+                } else {
+                    mediaManager.getValue().prevAudioItem();
+                }
+                lastUserInteraction = System.currentTimeMillis();
             }
         });
         mPrevButton.setOnFocusChangeListener(mainAreaFocusListener);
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingActivity.java
```diff
@@ -101,6 +102,8 @@
     private Lazy<BackgroundService> backgroundService = inject(BackgroundService.class);
     private Lazy<MediaManager> mediaManager = inject(MediaManager.class);
 
+    private PopupMenu mPopupMenu;
```
nielsvanvelzen, line 105:
```suggestion
    private PopupMenu popupMenu;
```

We don't use those c-style prefixes for variables

app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingActivity.java
```diff
@@ -460,7 +506,12 @@ public void onItemClicked(final Presenter.ViewHolder itemViewHolder, Object item
                                   RowPresenter.ViewHolder rowViewHolder, Row row) {
 
             if (!(item instanceof BaseRowItem)) return;
-            KeyProcessor.HandleKey(KeyEvent.KEYCODE_MENU, (BaseRowItem) item, mActivity);
+            lastUserInteraction = System.currentTimeMillis();
+            if (ssActive) {
+                stopScreenSaver();
+            } else {
+                mPopupMenu =  KeyProcessor.createItemMenu((BaseRowItem) item, ((BaseRowItem) item).getBaseItem().getUserData(), mActivity);
```
nielsvanvelzen, line 513:
```suggestion
                mPopupMenu = KeyProcessor.createItemMenu((BaseRowItem) item, ((BaseRowItem) item).getBaseItem().getUserData(), mActivity);
```

app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
```diff
@@ -221,7 +221,8 @@ public static boolean HandleKey(int key, BaseRowItem rowItem, Activity activity)
         return false;
     }
 
-    private static void createItemMenu(BaseRowItem rowItem, UserItemDataDto userData, Activity activity) {
+    // return the created PopupMenu so that the caller can dismiss it if needed
```
nielsvanvelzen, line 224: This comment doesn't seem that necessary to me. Besides, it doesn't use Javadoc/Kdoc so it's not even useful in an IDE.

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

nielsvanvelzen:
There is a bug report about the screensaver in the AudioNowPlayingActivity on Reddit:

https://www.reddit.com/r/jellyfin/comments/t3k3to/android_tv_beta_artist_background_bug_wmult/

I'm not 100% sure, but I think this PR might cause it. Have not been able to reproduce the issue though.

mueslimak3r:
> There is a bug report about the screensaver in the AudioNowPlayingActivity on Reddit:
> 
> https://www.reddit.com/r/jellyfin/comments/t3k3to/android_tv_beta_artist_background_bug_wmult/
> 
> I'm not 100% sure, but I think this PR might cause it. Have not been able to reproduce the issue though.
> 
> Edit: Could also be #703. I don't think we should call `setBackground` inside of rotateBackdrops because the background service already cycles between multiple backdrops.

Not much immediately stands out as potentially causing that behavior in this PR, or even in the most recent version of `nowPlayingActivity`.

The one thing that does stand out in the video as potentially related to the `nowPlayingActivity` is how right after the background switched back to the first image the time stamp jumps back from 3:23 -> 3:18.

This is done with `setCurrentTime(pos)` in the progress listener, and it's also done when starting and stopping the screensaver using `setCurrentTime(mediaManager.getValue().getCurrentAudioPosition())`
Perhaps that could give a hint on what's going on.

I'll see if I can reproduce this by adding some background photos for an artist.

wouldn't rule out #703 too

PR #1291, opened 2021-12-15
bring firetv controls in line with the new player controls

mueslimak3r:
**Changes**
* during playback with the overlay hidden, d-pad left/right should reveal the overlay, typically used to begin scrobbling the seekbar. I removed a check that would disable this behavior on fire tv devices
* `videoPlayerAdapter` would always return false on firetv when `canSeek()` is called (this is used by various objects like the seekbar and control glue to allow/disallow seeking & seekbar interaction)

**Issues**
https://www.reddit.com/r/jellyfin/comments/rfs766/comment/hogi4ih/?utm_source=share&utm_medium=web2x&context=3

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
@@ -569,7 +569,7 @@ public class CustomPlaybackOverlayFragment extends Fragment implements IPlayback
                     }
 
                     if (!mIsVisible) {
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (approved):
Not sure if anyone with Fire TV wanted to test this, changes look fine though.

mueslimak3r:
The redditor that posted about the bug said they're gonna try it out tomorrow and report back

mueslimak3r:
The redditor tested it and it works for them:
"After testing with my FireTV 4k I can confirm it now functions exactly as you described, pausing the video, scrolling through timestamps and, once selected, resuming the video from the selected position."

PR #1299, opened 2021-12-19
Target API 31, update all dependencies

nielsvanvelzen:
**Changes**
- Update all dependencies
- Always use version references in version catalog
- Target API 31
- Fix some code afer dep updates
  - ExoPlayer: update error listener signature
  - androidx.window: update to new method

**Issues**


nielsvanvelzen:
ExoPlayer goes from 2.14.0 to 2.16.1 which might fix (and introduce!) some issues:
https://github.com/google/ExoPlayer/blob/release-v2/RELEASENOTES.md

mueslimak3r:
It might be good to replace simpleexoplayer and its prepare logic, which are both depricated by this update. idk if it would be best to do that in this PR or a separate one. I have logic ready to replace the depricated prepare

nielsvanvelzen:
This PR updates the dependencies and the code changes necessary to run again, which for ExoPlayer is the error listener. I have no intention to spend significant time on the old playback code as I'm working on the rewrite. But feel free to create a separate PR to replace SimpleExoPlayer.

mueslimak3r (approved):
I'll take a look at the simpleexoplayer stuff and see if it's fine as-is or would benefit from replacing. Otherwise there's no need 👍

PR #1189, opened 2021-10-09
Add video playback speed controls

bakatrouble:
**Changes**
Add video playback speed controls

**Issues**
Fixes #1037

PR #1302, opened 2021-12-20
Add Playback speed controls

DavidFair:
**Changes**
Adds Playback speed controls:
Use the same icon / general UX style as Jellyfin-Android's playback button
A playback speed will apply to the current and future videos for the current application launch. I.e. setting speed to 2.00x then playing the next episode will also play at 2.00x

This PR Contains the following:
- Add some unit tests for functionality I've touched where it's possible to automatically test
- Add a `VideoSpeedController` to keep logic / state nice and encapsulated with the associated `PlaybackSpeedAction`. This should hopefully make it easy to incorporate into #1057 

See each commit for more detailed descriptions.

This is my first foray into Kotlin, and I haven't done Android dev in 4-5 years. So I'm happy to take any/all criticisms, feedback and change ideas.

**Issues**
Fixes: #1189 

**Testing**
- Try playing a video with speed set to various levels
- Try seeking around in the video, the speed should be preserved
- Play a different video. It should continue at the same speed as the last video
- Ensure unit tests run / pass


app/src/test/java/android/util/Log.java
```diff
@@ -0,0 +1,35 @@
+package android.util;
+
+/**
+ * Workaround slf4j trying to invoke Android.Log.isLoggable
+ * by providing a shim which simply uses println on everything.
+ */
+public class Log {
```
mueslimak3r, line 7: The rest of the project uses Timber for logging. Why use System.out instead?

DavidFair: So there's a bit of a hidden chain going on. Tl;dr anything that uses logs (such as Mockk or tested code) will invariantly require Android classes that aren't available in JVM only tests.

The longer explanation:
`Timber` and `SLF4J` are logging frameworks, which indirectly pipe to a logging API (usually the Android system). As `Mockk` will emit debug logging to allow you to track invocations / usage of the mocks this also pulls in the logging frameworks. Unit testing code which logs could potentially do the same thing too (but as of this PR nothing we test touches logs).
As this is a JVM unit test (as opposed to running on the Android Emulator) all the `Android.*` imports are unavailable and will produce the following error:
```
Caused by: java.lang.RuntimeException: Method isLoggable in android.util.Log not mocked. See http://g.co/androidstudio/not-mocked for details.
	at android.util.Log.isLoggable(Log.java)
	at uk.uuid.slf4j.android.LogAdapter.getNativeLogLevel(LogAdapter.java:113)
	at uk.uuid.slf4j.android.LogAdapter.<init>(LogAdapter.java:65)
	at uk.uuid.slf4j.android.LoggerFactory.<clinit>(LoggerFactory.java:46)
	... 50 more
```

So ultimately we're left with 3 options:
- Run the unit test with the Android emulator, just to put the logging API we will ignore in
- Before each test, mock out `Android.utils.Log` to make the framework happy (but this has a chicken and egg problem too so we'd need something like Powermockito)
- Have our own implementation of `Android.utils.Log` that only exists for JVM test targets that simply forwards to `stdout` as per this (SO post)[https://stackoverflow.com/a/46793567]

If these were integration tests I'd opt for the former. But this would still prevent us writing unit tests which (IMO) are a nice form of test for people to write and something is better than nothing in my mind.

I hope that explains the rationale, and I'm happy to hear your thoughts on this?

nielsvanvelzen:
Thanks for your pull request(s). Before I review it I'd like to ask you to split all the unit testing changes to a separate PR as it's not simply adding a few tests but adding dependencies etc. with it. We  try to focus pull requests to one particular change to make it easier to review, and with this PR I feel like it might take some time to get merged purely because of a secondary change (testing changes).

app/src/test/java/android/util/Log.java
DavidFair: I'll mark this as resolved, as the offending commit is now primarily in #1312

DavidFair:
Blocked by:
#1312 
#1313

DavidFair:
> Thanks for your pull request(s). Before I review it I'd like to ask you to split all the unit testing changes to a separate PR as it's not simply adding a few tests but adding dependencies etc. with it.

Done, I've rejigged the history of both branches to drop any non-essential commits into their own PRs.

> with this PR I feel like it might take some time to get merged purely because of a secondary change

No worries, I'm also aware of the amount of maintainers time required especially with the time commitments from the festive time of the year coming up too. In this case I just got on a bit of a roll and ended up with a much larger PR that originally envisioned....

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -192,6 +192,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		var defaultSubtitlesSize = Preference.int("subtitles_size", 28)
 	}
 
+	override operator fun <T : Preference<V>, V : Enum<V>> get(preference: T): V {
+		// Mock-able point for unit tests
+		return super.get(preference)
+	}
```
nielsvanvelzen, line 198: Preference implementations should not override their parent functions, I don't even understand this one as it doesn't even do anything.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -86,6 +87,7 @@
     private VideoOptions mCurrentOptions;
     private int mDefaultSubIndex = -1;
     private int mDefaultAudioIndex = -1;
+    private double mRequestedPlaybackSpeed = -1.0;
```
nielsvanvelzen, line 90: Why not 1.0 by default?

app/src/main/res/values/strings.xml
```diff
@@ -201,6 +201,7 @@
     <string name="lbl_stretch">Stretch</string>
     <string name="lbl_fit">Normal</string>
     <string name="lbl_audio_track">Select audio track</string>
+    <string name="lbl_playback_speed">Playback Speed</string>
```
nielsvanvelzen, line 204:
```suggestion
    <string name="lbl_playback_speed">Playback speed</string>
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/action/PlaybackSpeedAction.kt
```diff
@@ -0,0 +1,62 @@
...
+class PlaybackSpeedAction(
+	context: Context,
+	customPlaybackTransportControlGlue: CustomPlaybackTransportControlGlue,
+	playbackController: PlaybackController
+) : CustomAction(context, customPlaybackTransportControlGlue) {
+	private val speedController = VideoSpeedController(playbackController)
+	private val speeds = VideoSpeedController.Companion.SpeedSteps.values()
+
+	init {
+		initializeWithIcon(R.drawable.exo_ic_speed)
```
nielsvanvelzen, line 23: Don't use drawables from libraries, add a new ic_speed to our own drawables instead.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/action/PlaybackSpeedAction.kt
```diff
@@ -0,0 +1,62 @@
...
+		context: Context, view: View
+	) {
+		val speedMenu = populateMenu(context, view, speedController)
+
+		speedMenu.setOnDismissListener { leanbackOverlayFragment.setFading(true) }
+
+		speedMenu.setOnMenuItemClickListener { menuItem ->
+			speedController.setNewSpeed(speeds[menuItem.itemId])
+			speedMenu.dismiss()
+			return@setOnMenuItemClickListener true
```
nielsvanvelzen, line 38:
```suggestion
			
			true
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/action/PlaybackSpeedAction.kt
```diff
@@ -0,0 +1,62 @@
...
+			// Since this is purely numeric data, coerce to en_us to keep the linter happy
+			menu.add(0, i, i, String.format(Locale.US, "%.2fx", selected.speed))
+		}
+
+		menu.setGroupCheckable(0, true, true)
+		menu.getItem(speeds.indexOf(speedController.getCurrentSpeed())).isChecked = true
+		return speedMenu
+	}
+
+
```
nielsvanvelzen, line 61:
```suggestion
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/action/PlaybackSpeedAction.kt
```diff
@@ -0,0 +1,62 @@
...
+		val menu = speedMenu.menu
+		speeds.forEachIndexed { i, selected ->
+			// Since this is purely numeric data, coerce to en_us to keep the linter happy
+			menu.add(0, i, i, String.format(Locale.US, "%.2fx", selected.speed))
+		}
+
+		menu.setGroupCheckable(0, true, true)
+		menu.getItem(speeds.indexOf(speedController.getCurrentSpeed())).isChecked = true
+		return speedMenu
+	}
```
nielsvanvelzen, line 59:
```suggestion
	) = PopupMenu(context, view, Gravity.END).apply {
		speeds.forEachIndexed { i, selected ->
			// Since this is purely numeric data, coerce to en_us to keep the linter happy
			menu.add(0, i, i, String.format(Locale.US, "%.2fx", selected.speed))
		}

		menu.setGroupCheckable(0, true, true)
		menu.getItem(speeds.indexOf(speedController.getCurrentSpeed())).isChecked = true
	}
```

nielsvanvelzen (changes requested):

app/src/main/res/values-en-rGB/strings.xml
```diff
@@ -294,6 +294,7 @@
     <string name="lbl_Reset">Reset</string>
     <string name="pref_button_remapping_description">Press any button on the TV remote to assign it</string>
     <string name="pref_subtitle_track_button">Subtitle track button</string>
+    <string name="lbl_playback_speed">Playback Speed</string>
```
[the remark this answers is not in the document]
DavidFair: Ah, what's the correct way of adding a placeholder for a new string?

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
DavidFair: When I was writing this my thought process was to fail very obviously if for some reason we get the playback controller going without the playback action configuring a speed.
This should help us notice any bugs where the playback action somehow isn't created before the controller, which would also break carrying a users speed forwards

nielsvanvelzen: Sounds like a good idea

app/src/main/res/values-en-rGB/strings.xml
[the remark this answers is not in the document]
DavidFair: I've dropped this out of the commit too so it doesn't affect Weblate
All the other changes will go in a new commit to make it easier to review

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
DavidFair: So I'm a bit stumped how to resolve this
Removing this line causes the following exception:

```
java.lang.NullPointerException
	at org.jellyfin.androidtv.preference.SharedPreferenceStore.get(SharedPreferenceStore.kt:65)
	at org.jellyfin.androidtv.ui.playback.PlaybackControllerTest$prepDiMocks$1.invoke(PlaybackControllerTest.kt:38)
```

The cause of this is:
- We mock out `UserPreferences` to avoid `init` trying to run migrations on the shared preferences
- This implicitly mocks the parent abstract class `SharedPreferenceStore` so the internal val `sharedPreferences` is null (as the class constructor does not fire)
- As the return types are generic Mockk invokes the method with the passed types to determine the underlying types
- This naturally causes the real impl to run which throws a NPE

What I've tried:
- Mocking `SharedPreferences` and create a concrete `UserPreferences`, this triggers the migrations to run from init causing more problems
- Dropping type hints, the compiler cannot deduce the mocked types though
- Moving the type hints from `any` to `get<T, V>`, same problem
- Searching for the majority of the day if there's a way to workaround this from Mockk. But it appears there isn't.

So off the top of my head the solutions I've come up with (none of which I really like):
- Mark `UserPreferences` as open, create a shim with a setter to set the protected val (which needs changing to a var)
- Create a shim object which is passed to Koin, this will hold a real or injected `UserPreferencesStore` and forward to it
- Explicitly create an override with the types in, so Mockk does not have to deduce the types used (this current approach)

So I'm open to ideas or suggestions on this?

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
mueslimak3r: I'm more on the side of breaking gracefully and creating useful debug output, and not impacting the user experience

DavidFair: So coming from a C/C++ background is there some way of detecting if were in Debug mode:

`mRequestedPlaybackSpeed = DEBUG ? -1.0 : 1.0`
That way it "just works" for users but should be painfully obvious for devs?

mueslimak3r: What I was thinking is just set it to 1.0 so you don't "debug by letting stuff break". Then you could have a timber.d() debug log somewhere that would catch it at the beginning of the playback session lifecycle. You could probably put it in play() or the onPrepared listener

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -409,6 +416,14 @@ public void setAudioTrack(int ndx, List<MediaStream> allStreams) {
         }
     }
 
+    public void setPlaybackSpeed(@NonNull Double speed) {
+        if (nativeMode) {
```
mueslimak3r, line 420: If you want playbackSpeed to have a default value of `-1` and you think there could be errors caused by that, you could protect against that and log the error here

```suggestion
        if (speed < 0.25f) {
            Timber.d("invalid playback speed");
            return;
        }
        if (nativeMode) {
```

mueslimak3r (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
DavidFair: I think this is a good safety check we were missing
I've bumped it to warning in this case because if we ever get a value like `-1.0` it means the model controlling playback speed is completely broken too, so users likely won't be able to change their speed either / something is very wrong

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
DavidFair: After a couple of days of thinking, I realised we should just inject the interface. This matches traditional DI and resolves all the issues quite nicely without large sweeping changes

app/src/main/java/org/jellyfin/androidtv/preference/IUserPreferences.kt
```diff
@@ -0,0 +1,4 @@
+package org.jellyfin.androidtv.preference
+
+// Stub interface to allow Koin to resolve between different preference stores
+interface IUserPreferences : PreferenceStore
```
nielsvanvelzen, line 4: I don't see a point in an empty interface either, the way preference stores are meant to work is to have a class extending one of the preference store implementations (right now that is sharedprefs or displayprefs). No interfaces needed or desired.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -13,7 +13,7 @@ import org.jellyfin.androidtv.util.DeviceUtils
  *
  * @param context Context to get the SharedPreferences from
  */
-class UserPreferences(context: Context) : SharedPreferenceStore(
+class UserPreferences(context: Context) : IUserPreferences, SharedPreferenceStore(
```
nielsvanvelzen, line 16: Please don't format this file as the formatting was perfectly fine.

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoSpeedController.kt
```diff
@@ -0,0 +1,41 @@
+package org.jellyfin.androidtv.ui.playback
+
+class VideoSpeedController(playbackController: PlaybackController) {
+	companion object {
+		enum class SpeedSteps(val speed: Double) {
```
nielsvanvelzen, line 5: Move the speedsteps enum out of the companion object to make it easier accessible (VideoSpeedController.SpeedSteps instead of VideoSpeedController.Companion.SpeedSteps)

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -13,7 +13,7 @@ import org.jellyfin.androidtv.util.DeviceUtils
  *
  * @param context Context to get the SharedPreferences from
  */
-class UserPreferences(context: Context) : SharedPreferenceStore(
+class UserPreferences(context: Context) : PreferenceStore, SharedPreferenceStore(
```
nielsvanvelzen, line 16: SharedPreferenceStore already implements PreferenceStore

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -127,12 +127,14 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		/**
 		 * Shortcut used for changing the audio track
 		 */
-		var shortcutAudioTrack = Preference.int("shortcut_audio_track", KeyEvent.KEYCODE_MEDIA_AUDIO_TRACK)
+		var shortcutAudioTrack =
```
nielsvanvelzen, line 130: > Please don't format this file as the formatting was perfectly fine.

app/src/main/java/org/jellyfin/androidtv/di/PreferenceModule.kt
```diff
@@ -10,6 +11,9 @@ val preferenceModule = module {
 	single { LiveTvPreferences(get(userApiClient)) }
 	single { UserSettingPreferences(get(userApiClient)) }
 	single { AuthenticationPreferences(androidApplication()) }
-	single { UserPreferences(androidApplication()) }
+	single { UserPreferences(androidApplication()) } binds arrayOf(
+		UserPreferences::class,
+		PreferenceStore::class
+	)
```
nielsvanvelzen, line 17: No need to bind to the UserPreferences class as the app should never want to retrieve a store by this interface as it doesn't contain any preferences

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -66,7 +68,7 @@
 
     private Lazy<ApiClient> apiClient = inject(ApiClient.class);
     private Lazy<PlaybackManager> playbackManager = inject(PlaybackManager.class);
-    private Lazy<UserPreferences> userPreferences = inject(UserPreferences.class);
+    private Lazy<PreferenceStore> userPreferences = inject(PreferenceStore.class);
```
nielsvanvelzen, line 71: We use the UserPreferences store so use that one, not an interface.

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoSpeedController.kt
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	companion object {
+		private var previousSpeedSelection = SpeedSteps.SPEED_1_00
+		fun resetPreviousSpeedToDefault() {
+			previousSpeedSelection = SpeedSteps.SPEED_1_00
+		}
+	}
+
+	private val parentController = playbackController
```
nielsvanvelzen, line 23: Just make the constructor parameter a private val instead of this

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoSpeedController.kt
```diff
@@ -0,0 +1,41 @@
...
+	init {
+		// Carry forward the user's recent speed selection onto the next video(s)
+		setNewSpeed(previousSpeedSelection)
+	}
+
+	fun getCurrentSpeed(): SpeedSteps {
+		// Currently getCurrentSpeed uses previousSpeedSelection (from the companion)
+		// but this is an implementation detail I'd rather not leak in-case we ever need
+		// to separate out the two details. So implement a custom named getter...
+		return previousSpeedSelection
```
nielsvanvelzen, line 34: Rename "previousSpeedSelection" to "speed" or whatever and make it public. Implement a setter that implements the behavior of setNewSpeed.

That way no get/set functions are needed.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
DavidFair: I've turned off auto-formatting as it's causing more hassle than good :+1:

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
DavidFair: But this circles back to breaking mocks:
By specifying a concrete type, our mock also has to pull in the parent class when generics are involved. The abstract base class then tries to interact with a real preferences store (which is null as we're in a JVM unit test) and fails.

Injecting an interface allows us to still use all the same members, but removes the need for the mocking lib to inherit from the parent class

nielsvanvelzen: But that's a problem to solve in the tests, not in the preferences library.

DavidFair: I'm going to push for this. 

I understand I'm simply a contributor and I'm not trying to do anything hostile / force my approach onto this project in any way, hence I'm happy to make as many changes as required or wanted. But in this case using interfaces instead of implementations is a widely accepted pattern used in software development to make things more testable:

General:
https://stackoverflow.com/questions/2697783/what-does-program-to-interfaces-not-implementations-mean
https://stackoverflow.com/questions/3194278/should-you-always-code-to-interfaces-in-java
https://betterprogramming.pub/code-against-interfaces-not-implementations-37b30e7ab992

More relevant to this case with relation to mocks:
https://softwareengineering.stackexchange.com/questions/159813/do-i-need-to-use-an-interface-when-only-one-class-will-ever-implement-it

And the main server repo uses this pattern to generate / inject mocks too (I simply searched `interface` and grabbed the latest PR at random) :
https://github.com/jellyfin/jellyfin/pull/7039/files

----

Long-term I'd like to help write more tests for this project to help with some of the stability problems people encounter on the diverse (and sometimes painful) Android TV ecosystem.

Whilst Android development isn't my speciality, my hope is adding better / new tests helps maintainers at least know that things they fix don't regress, giving you more time to focus on the things I'm not qualified to help with i.e. Playback rewrite.

But this either requires significant architectural changes with a factory / injection focus for testability or small shims and simple interfaces on existing code, which I would argue is better than nothing.

nielsvanvelzen: I get your point, I too believe we could use more tests in the app. Unfortunately the app is still 80% legacy code (pretty much all Java is) and a lot of the newly written Kotlin code must be kept compatible with the old code. That's why some weird architectural choices were made. This is also why I personally do not create tests right now.

I'm not against interfaces. But in this particular situation it just does not work. The preferences library is not made to work like this and would need more significant changes to be properly testable. The type that a preference is read from **must** be the same as where the preference definition comes from, this is for various safety reasons. The Kotlin compiler doesn't check for this right now due to the way I implemented the system. If I'd rewritten the preferences code today I would use a different approach and architecture for it.

For now, I'd propose dropping the tests from this PR and moving them to a new one. That way the speed controls can be merged as-is and we can further discuss changes needed for these specific tests. It would also make it easier to review, just like when we moved the addition of Mockk etc. to separate PR's.

DavidFair: > The Kotlin compiler doesn't check for this right now due to the way I implemented the system.

Sorry, I completely misread your previous comment as we won't take this change, rather than this is a technical limitation around how types are implemented.

> ...I'd propose dropping the tests from this PR and moving them to a new one...
I completely agree about getting this in rather than stalling out. I would argue we can simply drop the PlaybackController tests all-together until the rewrite is done. Since trying to figure out how to make everything play nice, then simply ripping it out later on seems like a waste.

Also I have to say, thank you for your patience with me so far whilst I try to get the hang of this

nielsvanvelzen (approved):

mueslimak3r:
The speed controls are available for live tv playback. Setting the speed to > 1 breaks playback by constantly running out of buffer.

Can you patch this?
Maybe just have a flag to constrain the speed adjustment to values <= 1, and that would be used for live tv. Also, the speed would also need to be reset at playback start for live tv since otherwise it persists across sessions.

Dnkhatri:
@DavidFair  When playing faster speed pass through audio should be disabled as otherwise we get the high pitch audio,

DavidFair:
> The speed controls are available for live tv playback. Setting the speed to > 1 breaks playback by constantly running out of buffer.
Maybe just have a flag to constrain the speed adjustment to values <= 1, and that would be used for live tv. Also, the speed would also need to be reset at playback start for live tv since otherwise it persists across sessions.

I'll put a patch in tonight to check if we're playing back live tv and simply reset the speed from the speed controller.
I'm sure there is a way to have faster playback up until we hit the end of the buffer, then reset back to 1. But I can't really test it here so I'm not in a position to have a crack at it.

>  When playing faster speed pass through audio should be disabled as otherwise we get the high pitch audio
I'm not sure disabling pass through audio will fix the pitch, as we need some sort of audio scale filter. If it does I'm happy to apply that as a short term fix.

If it doesn't we'll have to leave it as a limitation, where users can choose to play faster with the current caveat that the pitch is wrong. I'll have a look what's available too, but AFAIK Chromium, Firefox and MPV all get around this by implementing a scale filter. So it depends on if ExoPlayer and / or libVLC have the appropriate audio filters available

DavidFair:
@Dnkhatri I can't reproduce this with Exoplayer locally, shifting 1x to 2x maintains the same sang note on my content
Could you check you're using Exoplayer locally? If you're still seeing this please feel free to open an issue and I'll be happy to look at this.

Dnkhatri:
@DavidFair 
is this commit not part of the new beta as that does not have playback speed control option

mueslimak3r:
@Dnkhatri It should be there. The only place it is disabled is with live tv

Issue #1055, opened 2021-07-31
subtitle font size and drop shadow


baburam0007:
**Checks**
- [x] Issue does not happen in the browser (jellyfin-web)

**Media Info of the file**

**Logs**

**FFmpeg Logs**

**System (please complete the following information):**
 - Android TV : Android 9 and above
 - Device manufacturer: Sony
 
 - Jellyfin server version: 10.7.5

**Additional context**
There is no option to disable the drop shadow in the subtitle section and select the font size. The font size is too big.

thornbill:
Can you please provide the version of the app you are using? Thanks!

baburam0007:
Currently, I'm testing on 0.12.0 beta 2. Thanks!

thornbill:
Can you provide a screenshot/image of what the subtitles look like?

baburam0007:
![Jellyfin dropdown shadow and font size](https://user-images.githubusercontent.com/70329150/127888918-419f36aa-c207-4cf8-8785-1f867a9ad9ce.jpg)
In the android tv app, there is no option to change to dropdown and font size

ManuLinares:
There is not, but you can change them in these places and compile it (for now I assume)

_app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java_

line 1478: change **R.color.black_opaque** to **R.color.transparent**

_app/src/main/res/layout/vlc_player_interface.xml_

line 55: change **android:textSize="28sp"** to whatever you like, I use **android:textSize="20sp"** (and I also change android:layout_marginBottom="20dp")

zilexa:
Can the default size please be changed?

Subtitle font size is really HUGE. 
It's one of the main reasons I'm still using Kodi with the add-on on AndroidTV. Default font size in Kodi is of normal size.

DemirG:
> Can the default size please be changed?
> 
> Subtitle font size is really HUGE.
> It's one of the main reasons I'm still using Kodi with the add-on on AndroidTV. Default font size in Kodi is of normal size.

They have been made smaller in the first beta. Which is way better then it was before.✌️

Imo they are still a bit to big. As an European I use subtitles for everything so it a bit of a nuisance. If it's possible an option with say 10 different sizes to pick from would be great.👌

As TVs a getting bigger its looks a bit comically with large subtitles on a 77"+ TV 😁

magounet:
[...]

Issue #437, opened 2020-04-16
Very big subtitles that never changes even if you change the size settings


tayhr:
No matter what size I set in the server, I think that the subtitle size is always 'larger' by default.

Jellyfin Version: 10.5.4:stable
Operating System: Windows 10
Architecture: X64
Installed Plugins: Open subtitles, MusicBrainz, AudioDB
Base URL: none
Storage: Local
Client: Mi Box, Android Version 6

My configs:
Preferred subtitle language: Any Language
Subtitle mode: Default
Burn subtitles: Auto
Text size: Normal
Font: Default
Drop shadow: Drop shadow

Subtitle type: external .SRT

nielsvanvelzen:
The app does not use the users preferences for subtitles. Marking this as enhancement.

tayhr:
Is there any possibility to decrease the subtitle size? It is very large and sometimes occupies a very large space on the screen

thornbill:
Yes. Some improvements are already in the queue for the next major release! https://github.com/jellyfin/jellyfin-androidtv/pull/396

Cloneg:
Android tv still has big subtitles and no option to change. 1080p video. Default video player. Chromecast with GoogleTV.

thornbill:
Yes we are aware. The next release (0.12) has not shipped yet.

DemirG:
> Yes we are aware. The next release (0.12) has not shipped yet.

Hi, what is the estimated date of release for 0.12?

DemirG:
> Yes we are aware. The next release (0.12) has not shipped yet.

I see that the subtitles are changed to a smaller font in the new beta. Thank you very much for the work! Looks better now.

2pandaemonium2:
I just switched from Plex to Jellyfinn, and still cannot figure out why it's such a big deal to just add a feature to change the font on the device you're watching it. I mean there are so many features but subtitles are a pig part of watching movies. The GUI looks great and even the metadata algorithm works great. If this subtitle feature could be added it can really be a competitor to PLEX. Please consider that guys. thx!

nielsvanvelzen:
> to change the font 

This issue isn't even about changing fonts

2pandaemonium2:
> > to change the font
> 
> This issue isn't even about changing fonts

i know, i meant font/subtitle size

2pandaemonium2:
> > to change the font
> 
> This issue isn't even about changing fonts

[...]

PR #1303, opened 2021-12-21
Add subtitles font size and background configurations

siankatabg:
**Changes**

Added seekbar for choosing subtitles font size (18-38 range, keeping 28 by default)
Added text outlining for better legibility and an option to disable the subtitles black background.

[Few screenshots and debug apk for who wants to test](https://drive.google.com/drive/folders/1Lk0FQIEC1klFdcbZ5DarzyCXobGYHjcc?usp=sharing)

**Issues**

Fixes: #1055
Fixes: #437

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -180,6 +180,16 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Enable series thumbnails in home screen rows
 		 */
 		var seriesThumbnailsEnabled = Preference.boolean("pref_enable_series_thumbnails", true)
+			
+		/**
+		 * Enable subtitles background
+		 */
+		var subtitlesBackgroundEnabled = Preference.boolean("subtitles_background_enabled", true)
+
+		/**
+		 * Set default subtitles font size
+		 */
+		var defaultSubtitlesSize = Preference.int("subtitles_size", 28)
 	}
 
 	init {
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
@@ -1476,7 +1476,9 @@ public class CustomPlaybackOverlayFragment extends Fragment implements IPlayback
 
             final SpannableString span = new SpannableString(TextUtilsKt.toHtmlSpanned(htmlText));
             span.setSpan(new ForegroundColorSpan(Color.WHITE), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
-            span.setSpan(new PaddedLineBackgroundSpan(ContextCompat.getColor(requireContext(), subtitlesBackgroundEnabled ? R.color.black_opaque : R.color.transparent), SUBTITLE_PADDING), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
+            if (subtitlesBackgroundEnabled) {
+                span.setSpan(new PaddedLineBackgroundSpan(ContextCompat.getColor(requireContext(), R.color.black_opaque), SUBTITLE_PADDING), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
+            }
             span.setSpan(new OutlineSpan(), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
 
             binding.subtitlesText.setText(span);
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
@@ -47,6 +47,7 @@ import com.bumptech.glide.request.target.Target;
 import org.jellyfin.androidtv.R;
 import org.jellyfin.androidtv.TvApp;
 import org.jellyfin.androidtv.constant.CustomMessage;
+import org.jellyfin.androidtv.preference.UserPreferences;
 import org.jellyfin.androidtv.data.model.DataRefreshService;
 import org.jellyfin.androidtv.databinding.OverlayTvGuideBinding;
 import org.jellyfin.androidtv.databinding.VlcPlayerInterfaceBinding;
@@ -68,6 +69,7 @@ import org.jellyfin.androidtv.ui.playback.overlay.LeanbackOverlayFragment;
 import org.jellyfin.androidtv.ui.presentation.CardPresenter;
 import org.jellyfin.androidtv.ui.presentation.ChannelCardPresenter;
 import org.jellyfin.androidtv.ui.presentation.PositionableListRowPresenter;
+import org.jellyfin.androidtv.ui.shared.OutlineSpan;
 import org.jellyfin.androidtv.ui.shared.PaddedLineBackgroundSpan;
 import org.jellyfin.androidtv.util.DeviceUtils;
 import org.jellyfin.androidtv.util.ImageUtils;
@@ -148,7 +150,9 @@ public class CustomPlaybackOverlayFragment extends Fragment implements IPlayback
     private static final long SUBTITLE_RENDER_INTERVAL_MS = 50;
     private SubtitleTrackInfo subtitleTrackInfo;
     private int currentSubtitleIndex = 0;
+    private int subtitlesSize = KoinJavaComponent.<UserPreferences>get(UserPreferences.class).get(UserPreferences.Companion.getDefaultSubtitlesSize());
     private long lastSubtitlePositionMs = 0;
+    private boolean subtitlesBackgroundEnabled = KoinJavaComponent.<UserPreferences>get(UserPreferences.class).get(UserPreferences.Companion.getSubtitlesBackgroundEnabled());
 
     private final Lazy<ApiClient> apiClient = inject(ApiClient.class);
     private final Lazy<MediaManager> mediaManager = inject(MediaManager.class);
@@ -257,6 +261,9 @@ public class CustomPlaybackOverlayFragment extends Fragment implements IPlayback
         // This configuration is required for the PaddedLineBackgroundSpan to work
         binding.subtitlesText.setShadowLayer(SUBTITLE_PADDING, 0, 0, Color.TRANSPARENT);
         binding.subtitlesText.setPadding(SUBTITLE_PADDING, 0, SUBTITLE_PADDING, 0);
+
+        // Subtitles font size configuration
+        binding.subtitlesText.setTextSize(subtitlesSize);
 
         //pre-load animations
         fadeOut = AnimationUtils.loadAnimation(requireContext(), R.anim.abc_fade_out);
@@ -1469,7 +1476,8 @@ public class CustomPlaybackOverlayFragment extends Fragment implements IPlayback
 
             final SpannableString span = new SpannableString(TextUtilsKt.toHtmlSpanned(htmlText));
             span.setSpan(new ForegroundColorSpan(Color.WHITE), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
-            span.setSpan(new PaddedLineBackgroundSpan(ContextCompat.getColor(requireContext(), R.color.black_opaque), SUBTITLE_PADDING), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
+            span.setSpan(new PaddedLineBackgroundSpan(ContextCompat.getColor(requireContext(), subtitlesBackgroundEnabled ? R.color.black_opaque : R.color.transparent), SUBTITLE_PADDING), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
+            span.setSpan(new OutlineSpan(), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
 
             binding.subtitlesText.setText(span);
             binding.subtitlesText.setVisibility(View.VISIBLE);
--- a/app/src/main/java/org/jellyfin/androidtv/ui/preference/category/playback.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/preference/category/playback.kt
@@ -159,9 +159,6 @@ fun OptionsScreen.playbackCategory(
 		setTitle(R.string.pref_subtitles_size)
 		min = 18
 		max = 38
-		valueFormatter = object : DurationSeekBarPreference.ValueFormatter() {
-			override fun display(value: Int) = "${value}"
-		}
 		bind(userPreferences, UserPreferences.defaultSubtitlesSize)
 	}
 }
--- a/app/src/main/java/org/jellyfin/androidtv/ui/preference/category/playback.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/preference/category/playback.kt
@@ -148,4 +148,20 @@ fun OptionsScreen.playbackCategory(
 		}
 		depends { userPreferences[UserPreferences.videoPlayer] == PreferredVideoPlayer.EXTERNAL }
 	}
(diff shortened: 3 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -464,4 +464,12 @@
     <string name="home_section_i">Home section %1$d</string>
     <string name="searchable_hint">Search Jellyfin</string>
     <string name="searchable_settings_description">Media</string>
+    <string name="pref_subtitles_background_title">Subtitles background</string>
+    <string name="pref_subtitles_background_summary">Show a black background behind the subtitles</string>
+    <string name="subs_size_xlarge">XLarge</string>
```
mueslimak3r, line 469: I think it would look nicer to not user abbreviated titles for the sizes. eg: "Extra Large" instead of "XLarge".
Also there's a typo: "Lanrge"

mueslimak3r (changes requested):

app/src/main/res/values/strings.xml
```diff
@@ -464,4 +464,12 @@
...
+    <string name="subs_size_large">Lanrge</string>
```
mueslimak3r, line 470:
```suggestion
    <string name="subs_size_large">Large</string>
```

mueslimak3r (changes requested):

thornbill:
A shadow is not the same as an outline for text and imho is an insufficient solution to maintain subtitle legibility.

siankatabg:
All recommendations have been implemented

siankatabg:
Sorry. I forgot to include a link to the source of the class that I've used for outlining the text. :)

siankatabg:
I've removed most of the abbreviations, only "subs" is left, so let me know if you want that fully spelled too

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -180,6 +180,16 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
...
+			
+		/**
+		 * Enable subtitles background
+		 */
+		var subtitlesBackgroundEnabled = Preference.boolean("pref_enable_subtitles_background", true)
+
+		/**
+		 * Set default subtitles font size
+		 */
+		var defaultSubtitlesSize = Preference.enum("pref_subtitles_size", SubtitlesSize.SUBTITLES_SIZE_NORMAL)
```
nielsvanvelzen, line 192:
```suggestion
		var defaultSubtitlesSize = Preference.enum("subtitle_size", SubtitlesSize.SUBTITLES_SIZE_NORMAL)
```

I'd prefer storing an integer/double with the used size.

app/src/main/java/org/jellyfin/androidtv/ui/shared/OutlineSpan.kt
```diff
@@ -0,0 +1,65 @@
...
+import androidx.annotation.ColorInt
+import androidx.annotation.Dimension
+
+/**
+ * Source: https://github.com/santaevpavel/OutlineSpan
+ *
+ * A class that draws the outlines of a text when given a stroke color and stroke width.
+ */
+
+class OutlineSpan(
```
nielsvanvelzen, line 15: This class seems simple enough to write manually instead of bringing in a file from another (differently licensed) project.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/shared/OutlineSpan.kt
siankatabg: The guy maintains a library, I just took the class and I actually modified it a bit. I thought I've to mention from where I took most of it and unfortunately I don't know what you want me to do about it.

mueslimak3r: Technically the license that code uses (Apache 2) conflicts with the license of this project (GPL 2). Additionally, the terms of the Apache 2 license require more than the simple "source: url" credit you use here.

You could ask the author(s) of the class/file to license it differently for use here, with one compatible with GPL 2.

But it may be simpler to just implement your own code or find something different that is licensed appropriately.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -180,6 +180,16 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
...
+			
+		/**
+		 * Enable subtitles background
+		 */
+		var subtitlesBackgroundEnabled = Preference.boolean("subtitles_background_enabled", true)
+
+		/**
+		 * Set default subtitles font size
+		 */
+		var defaultSubtitlesSize = Preference.string("subtitles_size", "28")
```
nielsvanvelzen, line 192: Use an integer or double.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -180,6 +180,16 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Enable series thumbnails in home screen rows
 		 */
 		var seriesThumbnailsEnabled = Preference.boolean("pref_enable_series_thumbnails", true)
+			
+		/**
+		 * Enable subtitles background
+		 */
+		var subtitlesBackgroundEnabled = Preference.boolean("subtitles_background_enabled", true)
```
nielsvanvelzen, line 187: > suggestion: we could also store an RGBA value so the color and transparency can be changed. If the color is fully transparent the background wouldn't be visible.

nielsvanvelzen (changes requested):

siankatabg:
Ok guys, my last try to do this the way you want... I wrote the outlining class from scratch and I made it as simple as possible. I also used seekbar for the subtitles size as I didn't found a way of doing it as a list and I don't see why I should use ARGB instead of a simple boolean to enable/disable the background.

Anyway, coding random fixes and mods for the things I'm using is just a hobby to me(I'm self taught) and I just wanted to help with something that many people request from a long time (and looked simple enought to do), but it seems that my skills are not enough to make you guys happy, so fell free to help me out, or just close the request.

Happy holydays to all!

nielsvanvelzen:
> I also used seekbar for the subtitles size as I didn't found a way of doing it as a list with int preference 

I don't think we support lists with int right now so this solution is fine for now

> and I don't see why I should use ARGB instead of a simple boolean to enable/disable the background.

It was a suggestion, it's not required or anything

> Anyway, coding random fixes and mods for the things I'm using is just a hobby to me(I'm self taught) and I just wanted to help with something that many people request from a long time (and looked simple enough to do), but it seems that my skills are not enough to make you guys happy, so fell free to help me out, or just close the request.

We definitely appreciate the pull request, replying "out of my knowledge" or "not interested" to a suggestion is valid.

> Happy holidays to all!

Same to you! I'll try to review this PR again when I find some time, I think it's mostly fine now :)

mueslimak3r (approved):

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -1469,7 +1476,8 @@ private void renderSubtitles(@Nullable final String text) {
 
             final SpannableString span = new SpannableString(TextUtilsKt.toHtmlSpanned(htmlText));
             span.setSpan(new ForegroundColorSpan(Color.WHITE), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
-            span.setSpan(new PaddedLineBackgroundSpan(ContextCompat.getColor(requireContext(), R.color.black_opaque), SUBTITLE_PADDING), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
+            span.setSpan(new PaddedLineBackgroundSpan(ContextCompat.getColor(requireContext(), subtitlesBackgroundEnabled ? R.color.black_opaque : R.color.transparent), SUBTITLE_PADDING), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
```
nielsvanvelzen, line 1479:
```suggestion
            if (subtitlesBackgroundEnabled) {
                span.setSpan(new PaddedLineBackgroundSpan(ContextCompat.getColor(requireContext(), R.color.black_opaque), SUBTITLE_PADDING), 0, span.length(), Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
            }
```

I believe doing it like this would make it slightly more readable

app/src/main/java/org/jellyfin/androidtv/ui/preference/category/playback.kt
```diff
@@ -148,4 +148,20 @@ fun OptionsScreen.playbackCategory(
...
+		bind(userPreferences, UserPreferences.subtitlesBackgroundEnabled)
+	}
+
+	seekbar {
+		setTitle(R.string.pref_subtitles_size)
+		min = 18
+		max = 38
+		valueFormatter = object : DurationSeekBarPreference.ValueFormatter() {
+			override fun display(value: Int) = "${value}"
+		}
```
nielsvanvelzen, line 164:
```suggestion
```

The default formatter already does a .toString, no need to override

app/src/main/java/org/jellyfin/androidtv/ui/shared/OutlineSpan.kt
```diff
@@ -0,0 +1,54 @@
...
+import android.graphics.Canvas
+import android.graphics.Color
+import android.graphics.Paint
+import android.text.style.ReplacementSpan
+
+/**
+ * A class that draws the outlines of a text
+ */
+class OutlineSpan(
+): ReplacementSpan() {
```
nielsvanvelzen, line 12:
```suggestion
class OutlineSpan : ReplacementSpan() {
```

app/src/main/java/org/jellyfin/androidtv/ui/shared/OutlineSpan.kt
```diff
@@ -0,0 +1,54 @@
...
+
```
nielsvanvelzen, line 13:
```suggestion
```

app/src/main/java/org/jellyfin/androidtv/ui/shared/OutlineSpan.kt
```diff
@@ -0,0 +1,54 @@
...
+		if (fm != null) {
+			fm.ascent = paint.fontMetricsInt.ascent
+			fm.bottom = paint.fontMetricsInt.bottom
+			fm.descent = paint.fontMetricsInt.descent
+			fm.leading = paint.fontMetricsInt.leading
+			fm.top = paint.fontMetricsInt.top
+		}
+
+		return paint.measureText(text, start, end).toInt()
+	}
```
nielsvanvelzen, line 30: Not 100% sure about this but I think this function does nothing right now, some values in `fm` get changes but `fm` is never used after that.

app/src/main/java/org/jellyfin/androidtv/ui/shared/OutlineSpan.kt
```diff
@@ -0,0 +1,54 @@
...
+		y: Int,
+		bottom: Int,
+		paint: Paint
+	) {
+		val initialColor = paint.color
+
+		paint.color = Color.BLACK
+		paint.style = Paint.Style.STROKE
+		paint.strokeWidth = 4f
+		canvas.drawText(text, start, end, x, y.toFloat(), paint)
```
nielsvanvelzen, line 48: By adding something like

```kotlin
val strokePaint = Paint().apply {
	color = Color.BLACK
	style = Paint.Style.STROKE
	strokeWidth = 4f
}
```

To a companion object, you can create a re-usable paint for stroking. It prevents allocations during a draw (which is expensive). This function could then look something like this:

```kotlin
canvas.drawText(text, start, end, x, y.toFloat(), strokePaint)
canvas.drawText(text, start, end, x, y.toFloat(), paint)
```

app/src/main/res/values/strings.xml
```diff
@@ -464,4 +464,7 @@
     <string name="home_section_i">Home section %1$d</string>
     <string name="searchable_hint">Search Jellyfin</string>
     <string name="searchable_settings_description">Media</string>
+    <string name="pref_subtitles_background_title">Subtitles background</string>
```
nielsvanvelzen, line 467:
```suggestion
    <string name="pref_subtitles_background_title">Enable subtitle background</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -464,4 +464,7 @@
...
+    <string name="pref_subtitles_background_summary">Show a black background behind the subtitles</string>
+    <string name="pref_subtitles_size">Subtitles font size</string>
```
nielsvanvelzen, line 469:
```suggestion
    <string name="pref_subtitles_size">Subtitle size</string>
```

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/shared/OutlineSpan.kt
siankatabg: Without that when two line subtitles are shown they the first line is half cut and displaying over the other line.

ManuLinares:
Could a subtitle ¿height? (bottom distance) be added as a slider also?

siankatabg:
> Could a subtitle ¿height? (bottom distance) be added as a slider also?
> 
> P.S.: The outline looks amazing, thanks for all of your work.

I also think the subtitles bottom padding(the empty space underneath them) is too big, but I don't think we need a slider for it, maybe finding a better value would do the job (now on most of the movies when 2 lines of subtitles are shown, the first one is mostly over the video, the other one under it on the black lane).

I was thinking of doing another PR about it as I've tested with different values and find the value of 20 to be the best (imo) on most of the media I've and even with 2 lines of subtitles shown (both lines are under the video in the black lane and also looking good when the video is full screen with no black lane)

ManuLinares:
> > Could a subtitle ¿height? (bottom distance) be added as a slider also?
> > P.S.: The outline looks amazing, thanks for all of your work.
> 
> I also think the subtitles bottom padding(the empty space underneath them) is too big, but I don't think we need a slider for it, maybe finding a better value would do the job (now on most of the movies when 2 lines of subtitles are shown, the first one is mostly over the video, the other one under it on the black lane).
> 
> I was thinking of doing another PR about it(so we can comment it/more ideas/opinions and so on) as I've tested with different values and find the value of 20(currently 48) to be the best (imo) on most of the media I've and even when 2 lines of subtitles are shown they both are under the video in the black lane and also looks much better when the video is full screen (with no black lanes on top/bottom).
> 
> But we could discuss it here also, as it's a really small (codewise) change!?

I'm using a value of 20 bottom padding in my build. It works best and I tried with several aspect ratios.

siankatabg:
> > > Could a subtitle ¿height? (bottom distance) be added as a slider also?
> > > P.S.: The outline looks amazing, thanks for all of your work.
> > 
> > 
> > I also think the subtitles bottom padding(the empty space underneath them) is too big, but I don't think we need a slider for it, maybe finding a better value would do the job (now on most of the movies when 2 lines of subtitles are shown, the first one is mostly over the video, the other one under it on the black lane).
> > I was thinking of doing another PR about it(so we can comment it/more ideas/opinions and so on) as I've tested with different values and find the value of 20(currently 48) to be the best (imo) on most of the media I've and even when 2 lines of subtitles are shown they both are under the video in the black lane and also looks much better when the video is full screen (with no black lanes on top/bottom).
> > But we could discuss it here also, as it's a really small (codewise) change!?
> 
> I'm using a value of 20 bottom padding in my build. It works best and I tried with several aspect ratios.

I'm also experimenting with slight to even zero padding, but the first line to be on the same place all the time (ie when a second line should be shown it is underneath the first one which is on a "constant" place(when not more than 2 lines are shown))

I know that's not a standard solution and most probably won't be merged here, but I guess my custom builds would be using it.

Anyway, I probably should really do a new PR and we should move that convo there as that's a thing out of this PR topic right now :)

nielsvanvelzen (approved):

mueslimak3r:
Hey @siankatabg it looks like this PR broke subtitle line wrapping. This is especially noticeable on large sizes.
Can you submit a patch for this?

Issue #1241, opened 2021-11-06
Improve error message when network is unavailable


krutburken:
### Describe the feature you'd like

When attempting login though the app and network is unavailable, the app will complain about the password being wrong instead of a network error. 

My parents seem to run into this a lot since they aren't very technical and keep messing with their router, and the error message being shown is very misleading (thus they call me instead of just fixing it themselves)

sourpatched:
There was a planned bug fix for this in 0.14.0. The app no longer crashes and should give an error message?

PR #1309, opened 2021-12-22
Network State Handling and Fix Login Error With No Network

DavidFair:
**Changes**

- Add a Util `NetworkStatus` to register for [NetworkCallbacks](https://developer.android.com/reference/kotlin/android/net/ConnectivityManager.NetworkCallback)
- Make this observable / stateful (for toast and login flow)
- Add a toast tied to the application lifetime, that will notify users when the network disconnects / reconnects. This should allow users to actively see they're going to have problems. Especially, as the connection status on Android TV requires users to actively go to their settings, instead of glancing at a status bar.
- Add a check to `login` that the network is connected, and bail out if we're not to emit the correct error.

Fixes: #1241 

**To Test**
(I'd recommend using a virtual device for this, these are the tests I've run through locally too)
- Disconnect / Reconnect on the network whilst remaining in the app.
- If you are using a virtual device, ADB can be used to avoid going out / in the app: 
```
adb devices  # Check the device appears first
adb shell svc wifi disable
adb shell svc wifi enable
```
- Alternatively, yank the power cable out of the wireless AP / remove ethernet cable on a real device
- Check the toast appears for connecting / disconnecting
- Check the login screen gives a sensible prompt for starting the app without a connection
- Check you can sign in after re-enabling the connection / with a good connection from the start.


DavidFair:
Blocked by:
#1312

nielsvanvelzen:
We could also check the specific ApiClientException we receive here

https://github.com/jellyfin/jellyfin-androidtv/blob/960212b832a2c2cef61c9367dd0e93783a4008a0/app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt#L199-L203

And emit the ServerUnavilableState

nielsvanvelzen:
I think the [TimeoutException](https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/master/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/exception/TimeoutException.kt) from the SDK can be used. The doc even says "This can happen when the host is unreachable or the network is offline.".

DavidFair:
> I think the [TimeoutException](https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/master/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/exception/TimeoutException.kt) from the SDK can be used. The doc even says "This can happen when the host is unreachable or the network is offline.".

Can we have both checks?

The network state check allows us to definitively say to users, "you need to connect to a network" as we know exactly what the problem is before attempting a connection
Whilst the timeout exception allows us to say, "you might need to connect to a network, or there might be a connectivity problem to the server".

Also as an aside, are there plans for localisation in the SDK library, or do we need to check the exception message (or better still an enum on the SDK side) then select the correct string for the user error?

nielsvanvelzen:
> Can we have both checks?
> 
> The network state check allows us to definitively say to users, "you need to connect to a network" as we know exactly what the problem is before attempting a connection Whilst the timeout exception allows us to say, "you might need to connect to a network, or there might be a connectivity problem to the server".

I don't really see the benefit, android tv already shows a big "There is no internet connection" warning on the home screen. Besides, the connectivity check in Android tries an actual internet connection last time I checked. The app works without internet via LAN, this change would effectively disable that.

> Also as an aside, are there plans for localisation in the SDK library, or do we need to check the exception message (or better still an enum on the SDK side) then select the correct string for the user error?

No, the SDK does not contain any strings that need translation (Exception messages are meant for developers only). We always use enums/sealed classes to communicate state.

DavidFair:
I've dropped the check before running the auth flow, so this shows the toast notification when the network state changes (connect to disconnected) now simply

nielsvanvelzen:
Sorry, but I still don't really see a point in this change. It adds a lot of code that needs to be maintained just to show a toast when the network state changes. The Android launcher already shows this though, why would you open an app expecting it to work when the launcher literally shows a big "not connected" banner?

DavidFair:
> Sorry, but I still don't really see a point in this change. It adds a lot of code that needs to be maintained just to show a toast when the network state changes. 

I agree on this being a lot of code for a simple toast, originally it also checked before attempting to login.
I cover my rationale below for this change ...

Maybe we can extend this with some sort of persistent status bar instead?

The nice thing is this only triggers on change, so if you start off in an obscure config that appears offline but isn't the bar won't appear. If you toggle between this will correctly flag up.

> The Android launcher already shows this though, why would you open an app expecting it to work when the launcher literally shows a big "not connected" banner?

Not all TVs unfortunately show this on their main screen. IIRC my Bravia requires me to go to settings. But unfortunately I can't confirm this as the TV is currently in a box waiting for me to move.

However even if the status is plainly in the launcher, this a UX thing for your average user more than anything. If you use something maintained byarge UX teams, like YouTube on Android mobile, you can clearly see the connections in the status bar. Switch to aeroplane mode and the app will still show no connection at the bottom in plain text.

The problem is a technical person knows where to look for their connectivity. As a technical people developing apps it's blindly obvious that this is an OS concern and something we shouldn't worry about.

But the average user simply opens "Netflix" or "YouTube" or "Jellyfin" for example.
For them it either works or it doesn't. The problem is if it stops working because their cheap router has flaked out, or someone has been "fiddling with the cables at the back of the TV" breaks their playback they will blame that app.

Netflix preempts this by producing an error effectively saying, "hey your connection dropped during playback" and users don't get angry and the company thinking, "Netflix has broken again".

We just silently stop working, or kick the user back to a login screen which is currently the only place we give feedback that it's not apps fault.
Instead I find the average user simply gives up at that point because it requires thought and they wanted to watch TV. 
The thought process is, "your Jellyfin thing broke again it's so unreliable" then they switch to a other another app that at least tells them in plain English they're offline. That way they can either fix it themselves or message someone with more technical know-how to say, restart your WiFi router.

DavidFair:
@nielsvanvelzen I was thinking something like this (minus the control bar and a touch larger)? I'm genuinely happy to ditch or fundamentally change this PR though based on feedback

Screenshot_20220125-172530

This would allow a user to continue watching their video until the buffer runs out whilst prompting them that something is wrong without having to exit the app

mueslimak3r:
On my Chromecast w Google TV theres a system popup that says I lost internet.

I assume this was added in a new-ish version of Android.

Currently in the app when internet is dropped I get timeout exceptions and it kicks me back to the login screen.

Perhaps a new activity or pop-up in this situation could make what's going on a little clearer to the user. It could say something like:
"Connection to the server is unavailable"
And have a button to dismiss, and a button to go select a different server.

Just throwing ideas out here, but if an in-app connection issues handler also handled jf API connection issues, then it could show the pop-up for both server related issues and device internet connection issues.

nielsvanvelzen (commented):
I now agree this could be useful. Let's start out with the simple toast from this PR and possibly extend it in a future PR (like showing the network state in the login-loading-screen etc.)


Your current implementation uses your own observer implementation, it should instead use LiveData or Kotlin Flow. The data should be kept in a repository from which it can be accessed in the application.


Look at UserViewsRepository for a sample implementation for our usage of the repository pattern.

DavidFair:
@nielsvanvelzen I've switched to a StateFlow which has massively simplified this (thanks).

Unfortunately I'm not sure how to join up the view-model with Jellyfin's current architecture. My goal was to setup the observer in a "global activity" so we don't need to add code to each activity to observe the current value. But I'm not sure where something like this would go?
```
viewModel.networkState.asLiveData().observe(this) { state ->
    when(state) {
        NetworkState.CONNECTED -> showToast(this, "... connected ..")
        NetworkState.DISCONNECTED -> showToast(this, "...disconnected...")
        NetworkState.UNKNOWN -> showToast(this, "...unknown...")
    }
}
```

nielsvanvelzen:
@DavidFair sorry for the late reply; I normally don't look at PR's in draft state so I ignore their notifications. Just looked here and it looks like this one is ready for review again. Can you confirm that?

DavidFair:
No need to apologise, I ended up figuring out the fragment part ...etc. after a couple of days reading up on Android components. Plus I was happy to see 0.13 get polished up and shipped.

So the only thing stopping me marking this ready is the lifecycle. We've got a fragment which holds the `StateFlow`, but I don't really want to attach the fragment to every activity. Is there a way we can assign an application lifecycle or do I need to convert from a fragment to something else?

I'll mark this as ready to review, since everything else should be reviewable, it's just not fully integrated into the app yet

nielsvanvelzen:
One long term goal I have is to have only one activity in the app and implementing all "pages" as separate fragments. Unfortunately that's not an easy task.

I guess adding a new activity lifecycle callback might work for the current "show toast" approach.
See:
- https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/di/ActivityLifecycleCallbacksModule.kt

app/src/test/kotlin/org/jellyfin/androidtv/data/repository/NetworkStatusRepositoryTest.kt
```diff
@@ -0,0 +1,27 @@
+package org.jellyfin.androidtv.data.repository
+
+import io.kotest.core.spec.style.ShouldSpec
+import io.kotest.matchers.shouldBe
+import io.mockk.mockk
+
+class NetworkStatusImplTest : ShouldSpec({
```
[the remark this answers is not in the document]
nielsvanvelzen: file is also in the wrong dir

nielsvanvelzen, line 7: Preferably use "FunSpec", although if there's a good reason to use "ShouldSpec" that _should_ be fine.

app/src/main/res/values/strings.xml
```diff
@@ -486,4 +486,6 @@
     <string name="pref_direct_stream_live_off">Transcoding is used when necessary</string>
     <string name="pref_direct_stream_live_on">Transcoding is disabled</string>
     <string name="home_section_resume_book">Continue reading</string>
+    <string name="toast_disconnected_network">Network Unavailable</string>
+    <string name="toast_connected_network">Network Reconnected</string>
```
nielsvanvelzen, line 490:
```suggestion
    <string name="toast_disconnected_network">Network unavailable</string>
    <string name="toast_connected_network">Network reconnected</string>
```

We might want to use more descriptive messages

app/src/main/java/org/jellyfin/androidtv/ui/global/NetworkStatusViewModel.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import androidx.lifecycle.ViewModel
+import org.jellyfin.androidtv.data.repository.NetworkState
+import org.jellyfin.androidtv.data.repository.NetworkStatusRepository
+
+class NetworkStatusViewModel (
+	private val repo: NetworkStatusRepository
+) : ViewModel() {
+	val getNetworkState = repo.state
+	fun getCurrentNetworkState() : NetworkState = repo.state.value
```
nielsvanvelzen, line 11: We probably want to use LiveData here so we can dynamically show the connection state in the toolbar/other places at some point.

nielsvanvelzen, line 11:
```suggestion
	val getNetworkState = repo.state
	fun getCurrentNetworkState(): NetworkState = repo.state.value
```

app/src/main/java/org/jellyfin/androidtv/data/repository/NetworkStatusRepository.kt
```diff
@@ -0,0 +1,86 @@
...
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.update
+import timber.log.Timber
+
+
+enum class NetworkState {
+	CONNECTED,
+	DISCONNECTED,
+	UNKNOWN
+}
```
nielsvanvelzen, line 21: Probably makes sense to put this in a separate file (constant package maybe?)

app/src/main/java/org/jellyfin/androidtv/data/repository/NetworkStatusRepository.kt
```diff
@@ -0,0 +1,86 @@
...
+	connectivityManagerDi: ConnectivityManager? = null,
+	private val networkRequestBuilderDi: NetworkRequest.Builder? = null
+) : NetworkStatusRepository {
+
+	private val _state = MutableStateFlow(NetworkState.UNKNOWN)
+	override val state = _state.asStateFlow()
+
+	// Some manual DI, since using Koin here is overkill
+	private val connectivityManager = connectivityManagerDi
+		?: appContext.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
```
nielsvanvelzen, line 44: Add a `factory<ConnectivityManager> { applicationContexT().getSystemService(Context.CONNECTIVITY_SERVICE) }` to Koin so we can just use the service via DI.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/global/NetworkStatusViewModel.kt
DavidFair: So I went for StateFlow as the official `AndroidDevelopers` blog are nudging people towards `StateFlows`: https://medium.com/androiddevelopers/migrating-from-livedata-to-kotlins-flow-379292f419fb

I'm tempted to drop `getCurrentNetworkState` (which is unused) and we can simply have two methods which returns a `StateFlow` or `LiveData` depending on requirements? 
The former seems to be preferred as the Fragment can write lifecycle aware handling (using `repeatOnLifestyle`) by the look of it.

nielsvanvelzen: I was not aware that LiveData is not recommended anymore. I already preferred using flows so that's a win-win :p.

I think dropping the function is fine, just expose the flow from the viewmodel so we can use it from the activity/fragment/other lifecycle. No need for livedata.

DavidFair:
@nielsvanvelzen Still not having much luck on the lifecycle from. I can either:
- convert the fragment to a `AbstractActivityLifecycleCallbacks` class, but then there is no `lifecycleScope` available to attach our coroutine to
- keep the fragment and create a new class. But there is no way (AFAIK) to dynamically inject a fragment in `onActivityCreated`

In this case searching the internet is leading to blanks so I'm throwing it out there for other ideas to look at whilst keeping the `StateFlow`

nielsvanvelzen:
> convert the fragment to a `AbstractActivityLifecycleCallbacks` class, but then there is no `lifecycleScope` available to attach our coroutine to

Check if the activity is an instance of `ComponentActivity` (which it always is) and then it will have the lifecycle stuff. Like so:

```kotlin
if (activity is ComponentActivity) activity.lifecycleScope.launch { /* stuff */ }
```

DavidFair:
So I've still not had much luck with Fragments and/or callbacks, but looking around I stumbled onto this: 
https://stackoverflow.com/questions/45728334/register-and-unregister-the-broadcastreceiver-in-application-class

The correct way to do this looking at the APIs available is to register a `BroadcastReceiver` at application level. 
This will then notify a component whenever the network status changes. I can still use a lot of the `ConnectivityManager` code but the viewmodel and fragment code will need to be scrapped

nielsvanvelzen:
Closing as inactive.

PR #1311, opened 2021-12-23
Emit JUnit output whilst running Gradle through CLI

DavidFair:
**Changes**
Tweaks the gradle script to forward on test failures in a way that the
stack trace is preserved.

Previously, the stack pointed to a line after the test suite, e.g. line
101 on an 80 line file.

**To Test**
- Go to a test, e.g. `TimeUtilsTest` and change the expected output
- Run `gradlew test`, check the output points to the line changed

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -23,3 +23,20 @@ subprojects {
 		}
 	}
 }
+
+tasks.withType<Test> {
+	// Ensure Junit emits the full stack trace when a unit test fails through gradle
+	useJUnit()
+
+	testLogging {
+		events(
+			org.gradle.api.tasks.testing.logging.TestLogEvent.FAILED,
+			org.gradle.api.tasks.testing.logging.TestLogEvent.STANDARD_ERROR,
+			org.gradle.api.tasks.testing.logging.TestLogEvent.SKIPPED
+		)
+		exceptionFormat = org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
+		showExceptions = true
+		showCauses = true
+		showStackTraces = true
+	}
+}


nielsvanvelzen:
Without this change the logs look somewhat like this:

```
> Task :app:testDebugUnitTest FAILED

org.jellyfin.androidtv.util.TimeUtilsTest > secondsToMillis FAILED
    java.lang.AssertionError at TimeUtilsTest.java:14
```

This seems clear enough to me, I can open TimeUtilsTest on line 14 and use the "go to declaration" command in Android Studio to open the faulty function.

With this change the log looks like this:

```
> Task :app:testDebugUnitTest FAILED

org.jellyfin.androidtv.util.TimeUtilsTest > secondsToMillis FAILED
    java.lang.AssertionError: expected:<14250> but was:<1250>
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at org.junit.Assert.assertEquals(Assert.java:633)
        at org.jellyfin.androidtv.util.TimeUtilsTest.secondsToMillis(TimeUtilsTest.java:14)
```

Which is the same information as I had before but now it includes the assert functions in the stacktrace which I don't really need

DavidFair:
Sorry my test example is poor, it's exceptions that come from the code.
In `TimeUtils:secondsToMillis` add an obvious NPE:
```
        Integer foo = null;
        Integer bar = foo + 2;
        return Math.round(seconds * MILLIS_PER_SEC);
```

Without the change we get:
```
> Task :app:testDebugUnitTest FAILED

org.jellyfin.androidtv.util.TimeUtilsTest > secondsToMillis FAILED
    java.lang.NullPointerException at TimeUtilsTest.java:11

5 tests completed, 1 failed
```

With this change we get:

```
org.jellyfin.androidtv.util.TimeUtilsTest > secondsToMillis FAILED
    java.lang.NullPointerException
        at org.jellyfin.androidtv.util.TimeUtils.secondsToMillis(TimeUtils.java:28)
        at org.jellyfin.androidtv.util.TimeUtilsTest.secondsToMillis(TimeUtilsTest.java:11)

5 tests completed, 1 failed
```

Though I agree the extra noise on test failures aren't great

app/build.gradle.kts
```diff
@@ -148,3 +151,15 @@ dependencies {
 	testImplementation(libs.koin.test)
 	testImplementation(libs.koin.test.junit4)
 }
+
+tasks.withType<Test> {
+    useJUnit()
+
+    testLogging {
+        events(FAILED, STANDARD_ERROR, SKIPPED)
+        exceptionFormat = FULL
```
nielsvanvelzen, line 160: We don't use imports in our Gradle files, use the complete package instead like so:

```
exceptionFormat = org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
```

nielsvanvelzen (changes requested):
Since we have multiple Gradle modules this change should be in the root build.gradle.kts instead of the app one.

build.gradle.kts
```diff
@@ -23,3 +23,18 @@ subprojects {
...
 }
+
+tasks.withType<Test> {
+	// Ensure Junit emits the full stack trace when a unit test fails through gradle
+    useJUnit()
+
+    testLogging {
+        events(org.gradle.api.tasks.testing.logging.TestLogEvent.FAILED, 
+			   org.gradle.api.tasks.testing.logging.TestLogEvent.STANDARD_ERROR,
+			   org.gradle.api.tasks.testing.logging.TestLogEvent.SKIPPED)
```
nielsvanvelzen, line 34:
```suggestion
		events(
			org.gradle.api.tasks.testing.logging.TestLogEvent.FAILED, 
			org.gradle.api.tasks.testing.logging.TestLogEvent.STANDARD_ERROR,
			org.gradle.api.tasks.testing.logging.TestLogEvent.SKIPPED
		)
```

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

PR #1312, opened 2021-12-23
Add Mockk and one test to exercise the new dep

DavidFair:
**Changes**
    Add Mockk to test framework deps
    
    Add Mockk as this is aware of Kotlin inheritence unlike Mockito.
    This primarily shows up where classes are not marked open. We can
    either:
    - Mark all unit tested classes open (least viable option)
    - Add boilerplate to Mockito each time we test a Kotlin class
    - Add a mocking framework that is Kotlin aware.
    
    In this case I've opted for the latter, since it also avoids us having
    to look up the boilerplate for static mocks...etc.

I've also added a single set of simple tests to exercise Mockk / demo it's working. This blurs single change per PR, but without it it's not really possible to prove Mockk is working.

For more details see each commit message

**To Test**
- Run DeviceUtils test


nielsvanvelzen:
We already have Mockito as a mocking library for Java. It's only used by `CustomSeekProviderTest` at this moment. Would it be possible to migrate these tests to Mockk to avoid having multiple mocking libraries?

app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.java
```diff
@@ -25,39 +26,45 @@
     // Nvidia Shield TV Model
     private static final String SHIELD_TV_MODEL = "SHIELD Android TV";
 
+    static String getBuildModel(){
```
nielsvanvelzen, line 29:
```suggestion
    static String getBuildModel() {
```

app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.java
```diff
@@ -5,6 +5,7 @@
 import java.util.Arrays;
 
 public class DeviceUtils {
+
```
nielsvanvelzen, line 8:
```suggestion
```

app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.java
```diff
@@ -25,39 +26,45 @@
...
             FIRE_TV_MODEL_GEN_1,
             FIRE_TV_MODEL_GEN_2
-        ).contains(Build.MODEL);
+        ).contains(getBuildModel());
     }
 
     public static boolean is60() {
         return Build.VERSION.SDK_INT >= Build.VERSION_CODES.M;
     }
+
```
nielsvanvelzen, line 69:
```suggestion
```

app/build.gradle.kts
```diff
@@ -144,5 +144,6 @@ dependencies {
 
 	// Testing
 	testImplementation(libs.junit)
-	testImplementation(libs.mockito)
+	testImplementation(libs.mockk)
+
```
nielsvanvelzen, line 148:
```suggestion
```

app/src/test/java/org/jellyfin/androidtv/util/DeviceUtilsTest.kt
```diff
@@ -0,0 +1,91 @@
...
+import io.mockk.*
+import org.junit.After
+import org.junit.Test
+import org.junit.Assert.*
+
+class DeviceUtilsTest {
+	@After
+	fun tearDown(){
+		unmockkAll()
+	}
```
nielsvanvelzen, line 13:
```suggestion
	fun tearDown() = unmockkAll()
```

app/src/test/java/org/jellyfin/androidtv/util/DeviceUtilsTest.kt
```diff
@@ -0,0 +1,91 @@
...
+
+	@Test
+	fun testBuildModelIsNull(){
```
nielsvanvelzen, line 16:
```suggestion
	fun testBuildModelIsNull() {
```
same for other functions

nielsvanvelzen (changes requested):

app/src/test/java/org/jellyfin/androidtv/util/DeviceUtilsTest.kt
DavidFair: In this case, the built in formatter will revert it back to the original form each time
Can we leave this one as-is to save me going back and changing it each time I touch this file?

nielsvanvelzen: That doesn't happen to me with the default settings of Android Studio. Are you sure you removed the curly braces?

DavidFair: Ah showing my inexperience with Kotlin, didn't realise that was valid syntax :+1:

app/src/test/java/android/util/Log.java
```diff
@@ -0,0 +1,35 @@
+package android.util;
+
+/**
+ * Workaround slf4j trying to invoke Android.Log.isLoggable
+ * by providing a shim which simply uses println on everything.
+ */
+public class Log {
```
nielsvanvelzen, line 7: If it's slf4j causing this issue we should probably drop slf4j-android from the test dependencies and use another logger. No need to add this class then.

app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.java
```diff
@@ -25,36 +25,41 @@
     // Nvidia Shield TV Model
     private static final String SHIELD_TV_MODEL = "SHIELD Android TV";
 
+    static String getBuildModel() {
```
nielsvanvelzen, line 28: function isn't marked as nullable, might also be better to use a default value (like "Unknown" or something like that) so we don't need to check nullability everywhere,

nielsvanvelzen (changes requested):

app/src/test/java/android/util/Log.java
DavidFair: We could use the [impl](https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-slf4j-impl) for tests only?

For our Gradle files, how can we structure it so we only include `slf4j-android` in our app and not tests?

DavidFair:
Not sure why this is being flagged up since it last changed in October: https://github.com/jellyfin/jellyfin-androidtv/pull/1312/files#annotation_2492482636

Possibly related to: https://android-review.googlesource.com/c/platform/sdk/+/46741/1

app/src/test/java/android/util/Log.java
nielsvanvelzen: > We could use the impl for tests only?

I'm not sure why you're linking to a log4j module, we don't use log4j.

> For our Gradle files, how can we structure it so we only include slf4j-android in our app and not tests?

I looked a bit into this and it doesn't seem that easy



I think a good solution here is to replace the slf4j-android dependency with something else, I'm looking into that and will open a pull request for that. So we can keep this class for now.

nielsvanvelzen (approved):

nielsvanvelzen:
> Not sure why `HomePreferencesScreen` is being flagged up since it last changed in October: https://github.com/jellyfin/jellyfin-androidtv/pull/1312/files#annotation_2492482636
> 
> Possibly related to: https://android-review.googlesource.com/c/platform/sdk/+/46741/1

In this case it is triggered by a bad translation, it's safe to ignore in the context of this pull request.

PR #1313, opened 2021-12-23
Add Koin unit testing deps implementation

DavidFair:
**Changes**
This allows Koin to do DI during unit tests and allows us to establish
Junit rules which involve Koin

This is essential for testing some classes as the playback code, where
DI is used heavily.

Prerequisite of #1302

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -145,4 +145,6 @@ dependencies {
 	// Testing
 	testImplementation(libs.junit)
 	testImplementation(libs.mockito)
+	testImplementation(libs.koin.test)
+	testImplementation(libs.koin.test.junit4)
 }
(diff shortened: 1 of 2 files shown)


app/build.gradle.kts
```diff
@@ -145,4 +145,7 @@ dependencies {
 	// Testing
 	testImplementation(libs.junit)
 	testImplementation(libs.mockito)
+	testImplementation(libs.koin.test)
+	testImplementation(libs.koin.test.junit4)
+
```
nielsvanvelzen, line 150:
```suggestion
	testImplementation(libs.koin.test.junit4)
```

nielsvanvelzen (changes requested):
This change would've been fine to be included in the PR actually using it. But we can merge it as-is.

nielsvanvelzen (approved):

PR #1324, opened 2021-12-31
reset player error count if playback stabilizes

mueslimak3r:
**Changes**
* record the timestamp of the most recent error when it's caught
* in `OnProgressListener` check if more than 30 seconds has passed since the last player error, and reset the retry count if so

**Issues**
* users who experience repeated player errors that are recoverable would still have their stream stopped by the "too many errors, giving up" condition, despite the errors being recoverable.

**Expected behavior**
* Unrecoverable errors where the retries repeatedly fail in a short time will still end the session
* Repeated recoverable errors will not end the session until they either 1) happen too close to each other (within 30s) 2) the player does not recover within the 3 attempts allowed

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -225,6 +225,13 @@ public class PlaybackController {
 
     public void playerErrorEncountered() {
         if (mVideoManager.isNativeMode()) exoErrorEncountered = true; else vlcErrorEncountered = true;
+
+        // reset the retry count if it's been more than 30s since previous error
+        if (playbackRetries > 0 && System.currentTimeMillis() - lastPlaybackError > 30000) {
+            Timber.d("playback stabilized - retry count reset to 0 from %s", playbackRetries);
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -1282,6 +1284,13 @@ public void onEvent() {
                             // crossed fire off an async routine to update the program info
                             updateTvProgramInfo();
                         }
+
+                        // if no additional player errors occur within 30 seconds, reset the count
+                        if (playbackRetries > 0 && System.currentTimeMillis() - lastPlaybackError > 30000) {
+                            Timber.d("resetting playback errror count from %s to 0 because playback has become stable", playbackRetries);
+                            playbackRetries = 0;
+                        }
```
nielsvanvelzen, line 1292: Wouldn't it make more sense to do this check in the top of `playerErrorEncountered()` instead?

mueslimak3r: absolutely. That makes way more sense

nielsvanvelzen (approved):

PR #1326, opened 2022-01-02
Run formatter over playback controller

DavidFair:
**Changes**

Runs the Android studio formatter over the playback controller. This
should make it easier to use automated tooling and clear up some diffs (#1302)
which have touched this file too.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -176,14 +176,31 @@ public class PlaybackController {
         }
     }
 
-    public StreamInfo getCurrentStreamInfo() { return mCurrentStreamInfo; }
-    public boolean canSeek() {return !isLiveTv;}
-    public boolean isLiveTv() { return isLiveTv; }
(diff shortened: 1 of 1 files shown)


mueslimak3r (approved):
Thanks for tidying this up 👍

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -176,14 +176,31 @@ public MediaSourceInfo getCurrentMediaSource() {
...
+    public boolean isLiveTv() {
+        return isLiveTv;
+    }
+
+    public int getSubtitleStreamIndex() {
+        return (mCurrentOptions != null && mCurrentOptions.getSubtitleStreamIndex() != null) ? mCurrentOptions.getSubtitleStreamIndex() : -1;
+    }
+
+    public @Nullable
+    Integer getAudioStreamIndex() {
```
nielsvanvelzen, line 196: I don't think this should be multiline?

```suggestion
    public @Nullable Integer getAudioStreamIndex() {
```

DavidFair: The formatter will split it back into multi line
Would the following be better?
```
    @Nullable
    public Integer getAudioStreamIndex() {
```

nielsvanvelzen: Yeah that would be better

nielsvanvelzen (approved):

PR #1337, opened 2022-01-08
Overhaul Preference typing / handling

DavidFair:
**Changes**
This work came out of my trying to get my head around how Java/Kotlin generics interact. It started off as a small prototype but I saw benefits to the codebase so carried on.

Fundamentally this change boils down to:
- Use SealedClass for Preference Types instead of generics on the library side
```
    Use a list of sealed classes for preference types. This gives a number
    of advantages:
    
    - A strict list of accepted types is known at compile-time
    - The compiler can warn about missing clauses in when branches
    - We can elide throwing, as we know the types creatable
    - Interface is a lot cleaner, as we don't have nested generics / enum
    cases
    - Don't have to deal with the quirks of Java type erasure for all but Enum types
```
- Extract common functionality into a new abstract class `BasicPreferenceStore`:

```
    Unifies the common bits, this will make it significantly easier to write
    future stores.
    It also means the most risky parts (de-serializing enums and dynamic
    dispatch) are kept in one place. Making debugging / fixes significantly
    easier.
```

**Testing**
- Open preferences in the library selection window
- Change some around, close and open the app. Make sure they are saved.
- Do the same for preferences in something like a TV library (as Display Settings are their own model)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/auth/LegacyAccountMigration.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/auth/LegacyAccountMigration.kt
@@ -10,6 +10,7 @@ import kotlinx.serialization.modules.contextual
 import org.jellyfin.androidtv.auth.model.AccountManagerAccount
 import org.jellyfin.androidtv.auth.model.AuthenticationStoreServer
 import org.jellyfin.androidtv.auth.model.AuthenticationStoreUser
+import org.jellyfin.androidtv.preference.PreferenceVal
 import org.jellyfin.androidtv.preference.SystemPreferences
 import org.jellyfin.sdk.model.serializer.UUIDSerializer
 import org.jellyfin.sdk.model.serializer.toUUIDOrNull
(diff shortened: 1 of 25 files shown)


nielsvanvelzen:
I'm not sure what your goal is with this PR but I'm stopping it right now: this is definitely not a change I'd accept.

The preference stores are meant to work with property delegation to make it extremely easy on the user side, with all complicated code hidden in the implementation. This PR does the opposite by adding the `PreferenceVal` class

DavidFair:
> The preference stores are meant to work with property delegation to make it extremely easy on the user side, with all complicated code hidden in the implementation. This PR does the opposite by adding the `PreferenceVal` class

If I changed the `set` method so users of the property didn't haven't to see / know that it was backed by a `PreferenceVal` would that be better? I would then revert changes in all unrelated code.

> I'm not sure what your goal is with this PR but I'm stopping it right now: this is definitely not a change I'd accept.

To create a union type originally. This evolved into called `PreferenceT` to replace the generic property handling. On the implementation side this makes it trivial for the compiler to see / warn about not handling handling all cases (as they form the set of sealed classes). 

This should make it easier to write and maintain bespoke models like `DisplayPreferencesStore`:
Currently in SharedPreferences if we added support for a new type, say Array<T>, it would be completely up to the reviewer and implementor to go and check each location now handles it. If they miss a location this would result in a runtime error, since our current system with generics has no bounds on the type T we pass in.

With this change the compiler can see we've added a new type, `ArrayT` from our sealed class, and inform us if there are any `when` clauses that haven't updated yet. This is because `T` is now bound by `PreferenceVal` rather than  `Any`

If this is a hard NACK and this pattern won't be accepted period I'm happy to close it. Originally I opened it as a draft as a RFC idea since I'd already done the work to get a feel for Kotlins reflection / typing system.

PR #1340, opened 2022-01-09
Implement server connection feedback

nielsvanvelzen:
This is one of my items I wanted to implement for 0.13. This feature was introduced in the v1.1 version of the SDK. It communicates the issues it found while connecting to a server. See https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/master/docs/migration/v1.1.md#error-reporting-in-recommendedserverdiscovery. I've also made a feature request to the awesome developers of the SDK requesting a new issue type for SSL issues so we don't need the current workaround for it (https://github.com/jellyfin/jellyfin-sdk-kotlin/issues/351).

The UI doesn't look amazing but it works fine, if anyone with UI knowledge wants to improve it; go ahead!

**Changes**
- Communicate "issues" from ServerRepository to UI
- Add helper functions to stringify server issues
- Show issues in login UI

image

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/RecommendedServerIssueExtensions.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/RecommendedServerIssueExtensions.kt
@@ -1,0 +1,59 @@
+package org.jellyfin.androidtv.util
+
+import android.content.Context
+import org.jellyfin.androidtv.R
+import org.jellyfin.androidtv.auth.ServerRepository
+import org.jellyfin.sdk.api.client.exception.ApiClientException
+import org.jellyfin.sdk.discovery.RecommendedServerIssue
+import javax.net.ssl.SSLHandshakeException
+
+private fun ApiClientException.getFriendlyMessage(context: Context): String? = when (cause) {
+	is SSLHandshakeException -> context.getString(R.string.server_issue_ssl_handshake)
+	else -> null
+}
+
+/**
+ * Create a localized string with the message to show an end-user of the app.
+ */
+fun RecommendedServerIssue.getFriendlyMessage(context: Context): String = when (this) {
+	is RecommendedServerIssue.MissingSystemInfo -> throwable?.let { throwable ->
+		// Search for exceptions we want to display but the SDK does not have
+		// an RecommendedServerIssue member for
+		if (throwable is ApiClientException) throwable.getFriendlyMessage(context)
+		else null
+	} ?: context.getString(R.string.server_issue_unable_to_connect)
+
+	is RecommendedServerIssue.InvalidProductName -> context.getString(R.string.server_issue_invalid_product)
+
+	RecommendedServerIssue.MissingVersion -> context.getString(R.string.server_issue_missing_version)
+
+	is RecommendedServerIssue.UnsupportedServerVersion -> context.getString(
+		R.string.server_issue_unsupported_version,
+		version.toString(),
+	)
+
+	is RecommendedServerIssue.OutdatedServerVersion -> context.getString(
+		R.string.server_issue_outdated_version,
+		version.toString(),
+		ServerRepository.minimumServerVersion.toString()
+	)
+
+	is RecommendedServerIssue.SlowResponse -> context.getString(R.string.server_issue_timeout)
+}
+
+/**
+ * Get the summary for a collection of issues. This is the most significant issue in the collection
+ * passed trough [RecommendedServerIssue.getFriendlyMessage].
+ */
+@Suppress("MagicNumber")
+fun Collection<RecommendedServerIssue>.getSummary(context: Context): String? = maxByOrNull {
+	// Assign a "score" to each issue type so we can return the most important one
+	when (it) {
+		is RecommendedServerIssue.MissingSystemInfo -> 5
+		is RecommendedServerIssue.UnsupportedServerVersion -> 4
+		is RecommendedServerIssue.OutdatedServerVersion -> 3
+		is RecommendedServerIssue.SlowResponse -> 2
+		RecommendedServerIssue.MissingVersion -> 1
+		is RecommendedServerIssue.InvalidProductName -> 0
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/jellyfin/androidtv/util/RecommendedServerIssueExtensions.kt
```diff
@@ -0,0 +1,59 @@
...
+@Suppress("MagicNumber")
+fun Collection<RecommendedServerIssue>.getSummary(context: Context): String? = maxByOrNull {
+	// Assign a "score" to each issue type so we can return the most important one
+	when (it) {
+		is RecommendedServerIssue.MissingSystemInfo -> 5
+		is RecommendedServerIssue.InvalidProductName -> 0
+		RecommendedServerIssue.MissingVersion -> 1
+		is RecommendedServerIssue.UnsupportedServerVersion -> 4
+		is RecommendedServerIssue.OutdatedServerVersion -> 3
+		is RecommendedServerIssue.SlowResponse -> 2
```
DavidFair, line 57: Nit: Could we also order these in-code in the same order as their weighting?

app/src/main/java/org/jellyfin/androidtv/util/RecommendedServerIssueExtensions.kt
```diff
@@ -0,0 +1,59 @@
...
+import android.content.Context
+import org.jellyfin.androidtv.R
+import org.jellyfin.androidtv.auth.ServerRepository
+import org.jellyfin.sdk.api.client.exception.ApiClientException
+import org.jellyfin.sdk.discovery.RecommendedServerIssue
+import javax.net.ssl.SSLHandshakeException
+
+private fun ApiClientException.getFriendlyMessage(context: Context): String? = when (cause) {
+	is SSLHandshakeException -> context.getString(R.string.server_issue_ssl_handshake)
+	else -> null
```
DavidFair, line 12: Could we print something like, "Unknown connection issue" here?

DavidFair (commented):
Thought I'd have a quick skim as there is some overlap with the Networking PR
Just some minor things I noticed skimming rather than a complete review

app/src/main/java/org/jellyfin/androidtv/util/RecommendedServerIssueExtensions.kt
nielsvanvelzen: No because we want to ignore other exception kinds here, I only used the when statement in case we want to match other exceptions in the future but that's unlikely.

app/src/main/java/org/jellyfin/androidtv/util/RecommendedServerIssueExtensions.kt
nielsvanvelzen: Definitely, I planned to do that after I decided the scores but forgot.

app/src/main/java/org/jellyfin/androidtv/util/RecommendedServerIssueExtensions.kt
```diff
@@ -0,0 +1,59 @@
...
+		version.toString(),
+	)
+
+	is RecommendedServerIssue.OutdatedServerVersion -> context.getString(
+		R.string.server_issue_outdated_version,
+		version.toString(),
+		ServerRepository.minimumServerVersion.toString()
+	)
+
+	is RecommendedServerIssue.SlowResponse -> context.getString(R.string.server_issue_timeout)
```
mueslimak3r, line 41: It's unfortunate that the name for this property is SlowResponse and not Timeout. I see that's defined by the SDK though.

Is it safe to assume that SlowResponse always means timeout, or could SlowResponse be caught in other situations?

I see "slow http response" in my server logs all the time. I assume that's unrelated to anything the "discovery" part of the SDK you're using here would test for.

nielsvanvelzen: SlowResponse is literally "it took to long for a proper reply so we gave up trying this server". It's sort of a timeout but it doesn't use the timeout value from JellyfinOptions. For discovery the value is 3 seconds [^1] because the response is not difficult to calculate and if this response is 3+ seconds, others will definitely be more making a client unusable.

A timeout in the ApiClient is reported as MissingSystemInfo with the [TimeoutException](https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/master/jellyfin-api/src/commonMain/kotlin/org/jellyfin/sdk/api/client/exception/TimeoutException.kt) as cause.

[^1]: https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/master/jellyfin-core/src/commonMain/kotlin/org/jellyfin/sdk/discovery/RecommendedServerDiscovery.kt#L29

nielsvanvelzen: The "slow http response" in server logs is something calculated by the server and unrelated to the SDK/app(s).

mueslimak3r: When I put in the IP address of a computer that is on my LAN but isn't running a jellyfin server I get this:
`D/RecommendedServerDiscovery: Could not connect to http://192.168.3.200:8096
    org.jellyfin.sdk.api.client.exception.TimeoutException: HTTP request timed out`

And the feedback is all "Unable to connect". Is that expected behavior?

Otherwise everything looks good!
I tested with the demo server, my server, and the badssl example

nielsvanvelzen: >  Is that expected behavior (unable to connect instead of timed out)?

Yes because it couldn't connect because of a timeout

mueslimak3r (approved):
looks good!
I tested with the demo server, my server, and the badssl example

PR #1341, opened 2022-01-09
Refactor duplicated preferences code

DavidFair:
On reflection I agree that the `PreferenceVal` implementation in #1337 was to complex in the implementation details. So I salvaged the maintenance part of the work from that PR.

**Changes**
- Remove duplicated code for handling Preference types (this means to introduce new pref stores we simply need to write code for interacting with the backing store)
- Add an interface, so we can code to an interface rather than an implementation in the future (making testing much much easier)
- Unit tests


app/src/main/java/org/jellyfin/androidtv/preference/BasicPreferenceStore.kt
```diff
@@ -0,0 +1,70 @@
+package org.jellyfin.androidtv.preference
+
+abstract class BasicPreferenceStore : IPreferenceStore {
```
nielsvanvelzen, line 3: Merge BasicPreferenceStore and IPreferenceStore (and rename it back to just PreferenceStore, we don't prefix interfaces with `I`)

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/preference/BasicPreferenceStore.kt
DavidFair: I've renamed `IPreferenceStore` to `PreferenceStore`. Thanks for the heads up

Merging the two is a bit more difficult in an interface we can't annotate methods as `protected` or `private`:
My goal was to keep `getInt` `getBool`...etc. methods protected so you users of this interface are forced to use the `get()` operator directly.
Additionally, we can't mark the enum serialization code private at all so the interface becomes polluted with this too.

Our options are:
- Leave this as a public interface and abstract class
- Merge them together, then convert PreferenceStore from an interface to an abstract class
- Move all the methods into the interface but have the API polluted by "public" private methods (hence I'm not keen)
- Something else? Happy for any ideas here

nielsvanvelzen: - [x] Merge them together, then convert PreferenceStore from an interface to an abstract class

This one seems like the proper approach.

app/src/main/java/org/jellyfin/androidtv/preference/PreferenceStore.kt
```diff
@@ -1,27 +1,82 @@
...
+	// it in the abstract common functionality (where it is used)
+	protected abstract fun getInt(keyName: String, defaultValue: Int): Int
+	protected abstract fun getLong(keyName: String, defaultValue: Long): Long
+	protected abstract fun getBool(keyName: String, defaultValue: Boolean): Boolean
+	protected abstract fun getString(keyName: String, defaultValue: String): String
+
+	protected abstract fun setInt(keyName: String, value: Int)
+	protected abstract fun setLong(keyName: String, value: Long)
+	protected abstract fun setBool(keyName: String, value: Boolean)
+	protected abstract fun setString(keyName: String, value: String)
```
nielsvanvelzen, line 59: `keyName` => `key`

app/src/main/java/org/jellyfin/androidtv/preference/PreferenceStore.kt
```diff
@@ -1,27 +1,82 @@
...
+			is String -> setString(preference.key, value)
+			is Enum<*> -> setEnum(preference, value)
+			else -> throw IllegalArgumentException("${preference.type.simpleName} type is not supported")
+		}
 
 	// store.getDefaultValue(Preference.x)
-	fun <T : Preference<V>, V : Any> getDefaultValue(preference: T): V {
+	fun <T : Any> getDefaultValue(preference: Preference<T>): T {
 		return preference.defaultValue
 	}
```
nielsvanvelzen, line 39:
```suggestion
	fun <T : Any> getDefaultValue(preference: Preference<T>): T = preference.defaultValue
```

app/src/main/java/org/jellyfin/androidtv/preference/PreferenceStore.kt
```diff
@@ -1,27 +1,82 @@
...
+			return preference.defaultValue
+		}
+
+		val loadedVal = preference.type.java.enumConstants?.find {
+			(it is PreferenceEnum && it.serializedName == stringValue) || it.name == stringValue
+		} ?: preference.defaultValue
+		return loadedVal
+	}
+
+	private fun <V : Enum<V>> setEnum(preference: Preference<*>, value: Enum<V>) =
```
nielsvanvelzen, line 75: This should be in the implementation, although our two current impelmentations (shared prefs and display prefs) stringify it doesn't mean other stores will too.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/preference/PreferenceStore.kt
```diff
@@ -1,27 +1,82 @@
...
+	protected abstract fun getBool(keyName: String, defaultValue: Boolean): Boolean
+	protected abstract fun getString(keyName: String, defaultValue: String): String
+
+	protected abstract fun setInt(keyName: String, value: Int)
+	protected abstract fun setLong(keyName: String, value: Long)
+	protected abstract fun setBool(keyName: String, value: Boolean)
+	protected abstract fun setString(keyName: String, value: String)
+
+	// Private Enum handling, all Enum types are serialized to / from String types
+	private fun <T : Enum<T>> getEnum(preference: Preference<T>): T {
```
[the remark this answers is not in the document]
DavidFair: :+1: 
Could we move it to a Util class like `PrefEnumSerializer` (or something better named)?

I agree about not having everyone try to serialize enums to string by default, but also would like to avoid duplicating complex code like this into different stores as we have to then maintain n copies of this code.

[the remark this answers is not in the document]
nielsvanvelzen: I don't think that's necessary. It's only 3 lines (4 when including the part to get the string value) of code that work, and I don't expect they need to be maintained ever.

app/src/main/java/org/jellyfin/androidtv/preference/DisplayPreferencesStore.kt
```diff
@@ -71,47 +71,59 @@ abstract class DisplayPreferencesStore(
...
+		val stringValue = getString(preference.key, "")
+
+		if (stringValue.isBlank()) {
+			return preference.defaultValue
+		}
+
+		val loadedVal = preference.type.java.enumConstants?.find {
+			(it is PreferenceEnum && it.serializedName == stringValue) || it.name == stringValue
+		} ?: preference.defaultValue
+		return loadedVal
```
nielsvanvelzen, line 116: Not entirely sure why you changed the implementation, but using null as default is faster as it doesn't allocate an empty string, it also should not use the getString() function but read from the cached data directly.

(same for the implementation in SharedPrefs)
```suggestion
		val stringValue = cachedPreferences[preference.key]

		return if (stringValue.isNullOrBlank()) preference.defaultValue
		else preference.type.java.enumConstants?.find {
			(it is PreferenceEnum && it.serializedName == stringValue) || it.name == stringValue
		} ?: preference.defaultValue
```

DavidFair: In `SharedPreferenceStore` I've opted for the following:
```
		val stringValue = getString(preference.key, "")
		return if (stringValue.isBlank()) preference.defaultValue
		// <unmodified>
```

This is so all callers of `getString` don't have to handle the null case which is an implementation detail for `getEnum`. For the shared store I'm tempted to leave it, it's a small performance pessimism but avoids us having to put null checks everywhere when we (de)serialise strings elsewhere

app/src/test/java/org/jellyfin/androidtv/preference/PreferenceStoreTest.kt
```diff
@@ -0,0 +1,96 @@
...
+import io.kotest.core.spec.style.FunSpec
+import io.kotest.core.test.TestCase
+import io.kotest.matchers.shouldBe
+
+/**
+ * Tests that the generic set/get methods dispatch to the correct
+ * internal getT/setT methods
+ */
+class PreferenceStoreTest : FunSpec() {
+	class TestStub : PreferenceStore() {
```
nielsvanvelzen, line 12: Move out of PreferenceStoreTest (but in the same file) and mark private.

app/src/test/java/org/jellyfin/androidtv/preference/PreferenceStoreTest.kt
```diff
@@ -0,0 +1,96 @@
...
+		}
+	}
+
+	var instance = TestStub()
+	override fun beforeTest(f: suspend (TestCase) -> Unit) {
+		super.beforeTest(f)
+		instance = TestStub()
+	}
+
+	enum class TestEnum { NOT_SET, SET }
```
nielsvanvelzen, line 66: Move out of PreferenceStoreTest (but in the same file) and mark private.

app/src/test/java/org/jellyfin/androidtv/preference/PreferenceStoreTest.kt
```diff
@@ -0,0 +1,96 @@
...
+		override fun <T : Any> delete(preference: Preference<T>) {
+			throw NotImplementedError("Not required for tests")
+		}
+	}
+
+	var instance = TestStub()
+	override fun beforeTest(f: suspend (TestCase) -> Unit) {
+		super.beforeTest(f)
+		instance = TestStub()
+	}
```
nielsvanvelzen, line 64: Wouldn't it be easier to just do `val instance = TestStub()` in the tests itself?

app/src/main/java/org/jellyfin/androidtv/preference/SharedPreferenceStore.kt
```diff
@@ -27,58 +27,53 @@ abstract class SharedPreferenceStore(
...
-			else -> value.toString()
-		})
-	}
+	override fun <V : Enum<V>> setEnum(preference: Preference<*>, value: Enum<V>) =
+		setString(
+			preference.key, when (value) {
+				is PreferenceEnum -> value.serializedName
+				else -> value.toString()
+			}
+		)
```
nielsvanvelzen, line 73: I would personally format this like so:

```suggestion
	override fun <V : Enum<V>> setEnum(preference: Preference<*>, value: Enum<V>) =
		setString(preference.key, when (value) {
			is PreferenceEnum -> value.serializedName
			else -> value.toString()
		})
```

nielsvanvelzen (changes requested):
Looks good, some comments on the tests though.

app/src/test/java/org/jellyfin/androidtv/preference/PreferenceStoreTest.kt
DavidFair: Huh didn't know you could mark a class private in Kotlin. That's good to know :+1:

app/src/main/java/org/jellyfin/androidtv/preference/SharedPreferenceStore.kt
DavidFair: :+1: I'm not sure why Android Studio is so keen on adding whitespace everywhere, but good catch :+1:

app/src/test/java/org/jellyfin/androidtv/preference/PreferenceStoreTest.kt
```diff
@@ -0,0 +1,90 @@
...
+}
+
+private enum class TestEnum { NOT_SET, SET }
+
+/**
+ * Tests that the generic set/get methods dispatch to the correct
+ * internal getT/setT methods
+ */
+class PreferenceStoreTest : FunSpec() {
```
nielsvanvelzen, line 61:
```suggestion
class PreferenceStoreTest : FunSpec({
```

When doing it like this you don't need the `init {}` block. This is how the other tests work too.

nielsvanvelzen (changes requested):

app/src/test/java/org/jellyfin/androidtv/preference/PreferenceStoreTest.kt
DavidFair: It was like this because of the nested class (I didn't even know private classes were a thing). I'll revert it to match the prevailing style :+1:

nielsvanvelzen (approved):

PR #1342, opened 2022-01-12
use HLS for exoplayer with transcoded streams

mueslimak3r:
**Changes**
* `ExoPlayerProfile` creates its own `transcodingProfiles` (I'd prefer it have it add to the default ones instead of replacing)
* `ExoPlayerProfile`'s `transcodingProfiles` use `hls` with an (f)mp4 container, h264 codec and hevc if the device supports it

* when playing with exoplayer (native mode), seeking and playback startup behavior uses the direct play/stream logic. Otherwise (with libVLC), the old logic is still used
* replaced `progressiveMediaSource` with `defaultMediaSource`, built with a `mediaItem`
* exoplayer uses `addMediaSource` instead of adding the media in `prepare()`


app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -325,9 +326,13 @@ public void setVideoPath(@Nullable String path) {
...
 
-                // reset position to the new videos default position
-                // exoplayer currently sees transcode streams as live windows, a fix is needed.
-                mExoPlayer.prepare(new ProgressiveMediaSource.Factory(dataSourceFactory).createMediaSource(Uri.parse(path)), true, false);
+                MediaItem.Builder mMediaItemBuilder = new MediaItem.Builder();
+                mMediaItemBuilder.setUri(Uri.parse(path));
+
+                // reset position when changing video so the player doesn't use the position from the prior video stream
+                mExoPlayer.setMediaSource(new DefaultMediaSourceFactory(dataSourceFactory).createMediaSource(mMediaItemBuilder.build()), true);
+                //mExoPlayer.setMediaSource(new ProgressiveMediaSource.Factory(dataSourceFactory).createMediaSource(mMediaItemBuilder.build()), true);
```
nielsvanvelzen, line 334:
```suggestion
```

app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
```diff
@@ -22,6 +22,30 @@ class ExoPlayerProfile(
...
+		transcodingProfiles = arrayOf(
+			// MP4 video profile
+			TranscodingProfile().apply {
+				type = DlnaProfileType.Video
+				context = EncodingContext.Streaming
+				container = ContainerTypes.MP4
+				videoCodec = buildList {
+					if (deviceHevcCodecProfile.ContainsCodec(CodecTypes.HEVC, ContainerTypes.MP4)) add(CodecTypes.HEVC)
+					add(CodecTypes.H264)
+				}.toTypedArray().joinToString(",")
```
nielsvanvelzen, line 34:
```suggestion
				}.joinToString(",")
```

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

PR #1345, opened 2022-01-13
patched subtitle outline PR breaking line wrapping

mueslimak3r:
**Notes**
* I'm opening this pending a response from the author of the original PR in case they'd prefer to author a patch instead.

**Changes**
* remove `OutlineSpan` which extends `ReplacementSpan`. `ReplacementSpan` doesn't support line wrapping, which explains the breakage.
* Instead of `OutlineSpan`, combine two `TextView`s. One is the original white text body. The other draws the outline using `Paint.Style.Stroke` in black.
* Added conditions to only draw the outline when the subtitle background is disabled
* Removed the `ForegroundColorSpan` since the text is already white, and defining text color in the span breaks the outline `TextView` being defined as black in XML

**Issues**
#1303 broke subtitle line wrapping


siankatabg:
Hey,

I saw your message last night, but didn't had a chance to check what's going on.
I actually had issues with my server and didn't used the app the last few days (power outages where my server is), so thanks for finding and providing a fixing for the bug. 

I just read about the ReplacementSpan and found out the line wrapping doesn't work with it and I'm sorry I didn't read (and tested) more before using it... 

Anyway, I actually did a different solution to the problem and I'm just testing it, so I'll upload the patch a bit later and I'll let you guys decide which one to use (as you guys are the pro's and my coding skills are limited as you already saw). I also don't have issues with the authorship at all, I just want the app to get better 👍

PR #1349, opened 2022-01-15
Preference UI refactor

nielsvanvelzen:
**Changes**

- Use default width for preferences overlay
- Use separate screens for a few main categories
- Categorize preferences
- Reorder preferences/categories/screens
- Rename strings to make more sense
- Add missing descriptions
- Rename "settings" to "preferences"
- Rename "user" to "account"
- Use Normal Capitalization Instead Of Title Casing
- Add icons to preference links and about items
- Remove Utils.getVersionString
- Fix bad string formatting in server issue strings
- Merge lazyOptionsScreen and optionsScreen into a single (non-lazy) getter
- Use slide right animation for preferences theme

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/preference/category/about.kt
```diff
@@ -1,28 +1,31 @@
...
 import org.jellyfin.androidtv.ui.preference.screen.LicensesScreen
-import org.jellyfin.androidtv.util.Utils
 
 fun OptionsScreen.aboutCategory() = category {
 	setTitle(R.string.pref_about_title)
 
 	link {
-		setTitle(R.string.lbl_version)
-		content = Utils.getVersionString(context)
+		title = "Jellyfin app version"
```
nielsvanvelzen, line 14: ^ This is explicitly not translated to make it easier to find when we ask for the app version during troubleshooting/bug reports.

mueslimak3r (approved):
looks good!!
Finding settings is much more intuitive now, and less scrolling is cool too.

siankatabg:
It really looks/works much better!

PR #1350, opened 2022-01-15
fixed some playback leaks

mueslimak3r:
**Changes**
* fixed leaks caused by the `runnable` report loop in `videomanager`

* clear references to `videomanager` and `overlay fragment` in `playbackcontroller` when playback ends

* disabled screensaver

* handle home button interrupting playback by ending the playback activity

**Validating fixes**
_Validated results with a heap dump_
* start playback and then end it by pressing back
* start playback and then end it by pressing "home" on the controller
* start playback and skip to the next or previous item

**Remaining leaks I found**
* stopping playback from the jellyfin remote
* the live tv player is full of leaks
* a leak in a `textClock` leftover from the playback session when nextup is shown

**Reproducing remaining leaks I found**
* start playback and skip to the end so nextup is shown
* start playback and stop it from another client using the jellyfin remote
* play live tv and then stop playback

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
@@ -259,7 +259,6 @@ public class VideoManager implements IVLCVout.OnNewVideoLayoutListener {
                 Timber.e("Attempt to play before surface ready");
                 return;
             }
-
             if (!mVlcPlayer.isPlaying()) {
                 mVlcPlayer.play();
             }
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -252,7 +252,6 @@ public void start() {
                 return;
             }
             mExoPlayer.setPlayWhenReady(true);
-            mExoPlayerView.setKeepScreenOn(true);
```
nielsvanvelzen, line 255: There's still calls for "mSurfaceView.setKeepScreenOn(false);" which we probably want to remove too.

mueslimak3r: whoops

nielsvanvelzen (approved):

PR #1351, opened 2022-01-15
Add opus as supported audio codec for ExoPlayer

nielsvanvelzen:
Tested on Nvidia Shield Pro

**Changes**

- Add opus to ExoPlayer direct play profile when down mixing is disabled

**Issues**


thornbill:
It looks like some older Fire TV devices do not support opus 😕

nielsvanvelzen:
According to the Android documentation opus decoding should work on Android 5>=. I don't really understand why it doesn't work on older Fire TV devices because we only target Android 5 and newer.

https://developer.android.com/guide/topics/media/media-formats#audio-formats

mueslimak3r (approved):
`exoplayerprofile` is currently only used with devices >= android 6

nielsvanvelzen:
This change still seems fine to me, if there's any issues reported after merging we can revert it.

PR #1352, opened 2022-01-16
Add libraries link to customization preferences

nielsvanvelzen:
This is a shortcut to open the display preferences for user views

**Changes**

- Add "rebuild()" function to trigger a preferences screen rebuilding
  - Useful when data is asynchronously loaded
- Add "UserViewsRepository"
  - It's a layer between UI and API to get user view (library) specific things
- Add "LibrariesPreferencesScreen" to open display preference screens
- Move user view related logic to UserViewsRepository
  - The list of (un)supported collection types
  - The collection types that allow view selection

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/repository/UserViewsRepository.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/repository/UserViewsRepository.kt
@@ -1,0 +1,24 @@
+package org.jellyfin.androidtv.data.repository
+
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.liveData
+import org.jellyfin.androidtv.ui.itemhandling.ItemRowAdapter
+import org.jellyfin.sdk.api.client.ApiClient
+import org.jellyfin.sdk.api.client.extensions.userViewsApi
+import org.jellyfin.sdk.model.api.BaseItemDto
+
+interface UserViewsRepository {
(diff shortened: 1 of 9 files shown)


mueslimak3r:
I've tested this feature and it works great 👍 I haven't reviewed the code though since I really don't know kotlin yet.

mueslimak3r (approved):
I read up on it and I think I understand enough to review the code.

It looks like collections and their corresponding views which can be:
* static (only one way of organizing/presenting the items)
* sortable, and able to be shown in different arrangements with their own view, like by genre, letter, artist, etc.

for the latter type of collection, "view selection" is used for "smart screen" and those views have the settings menu you've linked in this PR. Those views are also what are integrated into the leanback home rows.

the `UserViewsRepository` uses `LiveData` to facilitate the settings menu, and potentially other features in the future, to easily access views for collections which qualify

It looks like this line is what registers a collection to the repository:
https://github.com/jellyfin/jellyfin-androidtv/blob/b9b920a608c6afe53abd472411f026693cca7768/app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java#L469

mueslimak3r:
I got this error in the logs just now and it also printed the exception in a toast

image

nielsvanvelzen:
Hurray for nullability in Java code, creating a PR in a bit.

PR #1354, opened 2022-01-17
added text outline to the video title in the playback overlay

mueslimak3r:
**Changes**
* use the new StrokeTextView to make the video title and subtitle more visible over light media


nielsvanvelzen (commented):
I'm not a fan of this one, it looks rather ugly for the title/subtitle of a video. The overlay already dims the video, the title should be readable enough with that I think.

siankatabg:
The overlay sometimes fades before the sub/title which makes it at times unreadable and not looking right, so maybe a lower width? Like 2.0 or something which won't be that visible, but still useful when the background is clear white. 

Otherwise I agree that 5.0 width looks strange and I don't like it too :)

PR #1362, opened 2022-01-18
Bypass hls seeking behavior for devices not using exoplayerprofile

### LINKED ISSUE #1353: latest changes broke seeking in exoplayer
### Describe the bug

Latest changes broke seeking in exoplayer.

After a seek, player starts from zero.

sad 😢

### Application version

latest dev build

### Device information

fire stick v1

### Android version

FireOS 5

### Jellyfin server version

10.7.7

mueslimak3r:
**Changes**
* added conditions for the new HLS behavior so it can be ignored in cases where exoplayer is used for transcodes but hls isn't used. This applies when the baseprofile instead of the exoplayerprofile is used.

**Notes**
It's likely that devices which aren't eligible to use exoplayerprofile could still use hls.

In that case I guess a new exoplayer profile could be created which adds basic exoplayer-specific features.

They could be renamed to exoplayer-base and exoplayer-enhanced, or something.

**Issues**
Fixes #1353

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -1343,7 +1343,11 @@ public class PlaybackController {
                     boolean continueUpdate = true;
                     if (!spinnerOff) {
                         if (mStartPosition > 0) {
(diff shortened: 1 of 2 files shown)


ManuLinares:
> **Notes** It's likely that devices which aren't eligible to use exoplayerprofile could still use hls.
> 
> In that case I guess a new exoplayer profile could be created which adds basic exoplayer-specific features.
> 
> They could be renamed to exoplayer-base and exoplayer-enhanced, or something.

This closes #1353, If you need more testing, I'm happy to help, just tag my username.

nielsvanvelzen (approved):

PR #1364, opened 2022-01-18
Cap playback speed to 1 watching live tv

DavidFair:
**Changes**
    Cap speed at 1.0 when live TV is on
    
    We could allow various things, like a speed less than 1 or greater than
    until we exhaust the buffer.
    
    Unfortunately I don't have the equipment to test more complex
    interactions, so lets play it safe and force it always to 1.0
    Additionally we hide the playback speed button in live TV too

**Issues**
Mentioned in #1302 

**To test**
Could someone check the button is hidden watching live TV, as I don't have this available locally

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -753,7 +753,9 @@ public class PlaybackController {
 
         // get subtitle info
         mSubtitleStreams = response.GetSubtitleProfiles(false, apiClient.getValue().getApiUrl(), apiClient.getValue().getAccessToken());
-        if (mVideoManager != null) mVideoManager.setPlaybackSpeed(mRequestedPlaybackSpeed);
+
+        // set playback speed to user selection, or 1 if we're watching live-tv
+        if (mVideoManager != null) mVideoManager.setPlaybackSpeed(isLiveTv() ? 1.0 : mRequestedPlaybackSpeed);
 
         if (mFragment != null) mFragment.updateDisplay();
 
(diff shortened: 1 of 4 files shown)


mueslimak3r (changes requested):
This PR misses the most likely situation that could catch up users: the playback speed is set to > 1 in the last session, and then they play live tv.

If the playback speed controls are removed for users watching live tv and they have no way of lowering it, the speed needs to be reset to 1 when live tv playback is started.

There are some publicly accessible m3u8/xml tv sources like PBS and C-SPAN in the US that you could use for testing. That's what I do and it works great.

Also yes, the button is hidden. I just tested.

DavidFair:
> This PR misses the most likely situation that could catch up users: the playback speed is set to > 1 in the last session, and then they play live tv.

:man_facepalming: I've added a check in the speed controller for this case and unit test case, then stopped creating that action (and controller) in the second commit.

My thought on how to tackle this it to revert the second commit (i.e. always create the playback speed action), but hide the icon. That way the speed controller is always created and `init` will correctly detect / fix-up the live TV case. Any thoughts on this approach / another preferred approach?

mueslimak3r:
That could work.
Another option would be to make the change in PlaybackController and just check if live tv when starting:

`if (mVideoManager != null) mVideoManager.setPlaybackSpeed(isLiveTv() ? 1.0f : mRequestedPlaybackSpeed);`

https://github.com/jellyfin/jellyfin-androidtv/blob/c85751ed3139b33a0905b523d2dc209a91071fac/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java#L756

Its debatable whether that's delegating the responsibility of the SpeedController too much to PlaybackController, or whether your solution is doing the opposite.

DavidFair:
Lets do this in `PlaybackController` so we don't instantiate the `SpeedController` with live TV.

In the future if we make this smarter (supporting speeds < 1, or >1 until we run out of buffer) we can re-enable the speed controller and remove this check out of `PlaybackController`.

That way it's obvious that `SpeedController` isn't a factor if we see playback issues with live TV at the moment

mueslimak3r (approved):
Nice!
It seems to work as intended:
1) play a movie or tv episode and set the speed to 2
2) then play live tv, speed is 1 while live tv is playing
3) go back to a movie or tv episode and the speed is 2 again

PR #1373, opened 2022-01-22
FullDetailsActivity - dynamically fill button row and hide extra

mueslimak3r:
**Notes**
* the row = mDetailsOverviewRow
* the main drawback to this approach is that it assumes all the less important buttons, which should be hidden first, are visible when they are sorted. If they're not, they will be shown first and the more important buttons would be hidden.
Currently though, it looks like all the buttons (or almost all) are visible when `showMoreButtonIfNeeded()` is called.

**Changes**
* given a set of buttons to consider showing or hiding, loop through them and hide buttons until there either is extra space available or the buttons are already shown. the hidden buttons become part of the "more options" menu.

* the more options menu only shows items that are hidden in the row

* added the shuffle button and play trailers button to the more options menu. The trailer button is last in the list to be hidden so it's very unlikely that it ever will be. Future button could be added to the "more options" menu by default, in addition to the row, so that the row can always offload them into the menu if needed.

* renamed a couple buttons to standardize the naming.

* moved the onClick logic for the play trailers button to its own method so it can be called from the more options menu

**Issues**
* buttons were overflowing past the 5 button limit of the row
* there was a static set of buttons that would always go to the "more options" menu if the total number of buttons exceeded five, even if some could fit in the row

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
@@ -813,6 +813,21 @@ public class FullDetailsActivity extends BaseActivity implements IRecordingIndic
 //        }
     }
 
+    private void playTrailers() {
+        apiClient.getValue().GetLocalTrailersAsync(TvApp.getApplication().getCurrentUser().getId(), mBaseItem.getId(), new Response<BaseItemDto[]>() {
+            @Override
+            public void onResponse(BaseItemDto[] response) {
+                play(response, 0, false);
+            }
+
+            @Override
+            public void onError(Exception exception) {
+                Timber.e(exception, "Error retrieving trailers for playback");
+                Utils.showToast(mActivity, R.string.msg_video_playback_error);
+            }
+        });
+    }
+
     private String getRunTime() {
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (approved):

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -1327,29 +1332,48 @@ public void onClick(View v) {
...
+                if (favButton != null && !favButton.isVisible()) {
                     if (mBaseItem.getUserData().getIsFavorite()) {
                         more.getMenu().getItem(0).setVisible(false);
                         more.getMenu().getItem(1).setVisible(true);
                     } else {
                         more.getMenu().getItem(0).setVisible(true);
                         more.getMenu().getItem(1).setVisible(false);
                     }
-                } else {
+                } else if (favButton != null) {
```
mueslimak3r, line 1344: this shouldn't be here

nielsvanvelzen: good way to bump your PR count :p

PR #1376, opened 2022-01-22
Remove unused stuff from MyDetailsOverviewRow

nielsvanvelzen:
Found this one while working on some other stuff. The studio image is loaded but never used. Removing it for a small performance boost.

**Changes**

- Remove unused stuff from MyDetailsOverviewRow
- Remove studio image loading
- Remove getPrimaryImageUrl for StudioDto in ImageUtils

**Issues**


mueslimak3r (approved):
Nice find. I think the images for items in ItemRowAdapter might be loaded more than once too

PR #1377, opened 2022-01-22
Initial ImageUtils SDK migration

nielsvanvelzen:
Most of the image url building in the app is done with the ImageUtils class. I've migrated all functions I could test (everything not live tv) to use the SDK. This generates slightly different URL's.

**Changes**

- Add ImageHelper to create image urls using the SDK
  - Migrate all non-livetv image url builder functions in ImageUtils to new ImageHelper
- Remove apiclient parameter from ImageUtils functions

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/NowPlayingBug.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/NowPlayingBug.java
@@ -21,9 +21,7 @@ import org.jellyfin.androidtv.ui.playback.MediaManager;
 import org.jellyfin.androidtv.ui.playback.PlaybackController;
 import org.jellyfin.androidtv.util.ImageUtils;
 import org.jellyfin.androidtv.util.TimeUtils;
-import org.jellyfin.apiclient.interaction.ApiClient;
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/NowPlayingBug.java
```diff
@@ -125,7 +123,7 @@ private void setInfo(BaseItemDto item) {
         if (item == null) return;
 
         Glide.with(context)
-                .load(ImageUtils.getPrimaryImageUrl(item, KoinJavaComponent.<ApiClient>get(ApiClient.class)))
+                .load(ImageUtils.getPrimaryImageUrl(item))
```
mueslimak3r, line 126: this image didn't load before this PR, and it still doesn't

nielsvanvelzen: This one works for me, I think it requires a primary image for the track (not the album) to show up.

mueslimak3r (approved):
looks like a pretty straightforward java -> kotlin translation + the sdk stuff 👍

PR #1385, opened 2022-01-25
fix crash from empty or null browsecommand sent from android client

mueslimak3r:
**Changes**
* catch the `IllegalArgumentException` GetItemAsync throws when `itemId` is null or empty in this situation

**Issues**
* `command.getItemId()` could return a null or empty string when remote controlling the app from the android client and quickly pressing the back button (on the phone) until the app is closed

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/eventhandling/TvApiEventListener.java
+++ b/app/src/main/java/org/jellyfin/androidtv/data/eventhandling/TvApiEventListener.java
@@ -140,13 +140,17 @@ public class TvApiEventListener extends ApiEventListener {
                 Timber.i("Command ignored due to no activity or playback in progress");
                 return;
             }
-            client.GetItemAsync(command.getItemId(), TvApp.getApplication().getCurrentUser().getId(), new Response<BaseItemDto>() {
-                @Override
-                public void onResponse(BaseItemDto response) {
-                    //Create a rowItem and pass to our handler
-                    ItemLauncher.launch(new BaseRowItem(0, response), null, -1, TvApp.getApplication().getCurrentActivity(), true);
-                }
-            });
+            try {
+                client.GetItemAsync(command.getItemId(), TvApp.getApplication().getCurrentUser().getId(), new Response<BaseItemDto>() {
+                    @Override
+                    public void onResponse(BaseItemDto response) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/data/eventhandling/TvApiEventListener.java
```diff
@@ -140,13 +140,17 @@ public void onBrowseCommand(ApiClient client, BrowseRequest command) {
...
             }
-            client.GetItemAsync(command.getItemId(), TvApp.getApplication().getCurrentUser().getId(), new Response<BaseItemDto>() {
-                @Override
-                public void onResponse(BaseItemDto response) {
-                    //Create a rowItem and pass to our handler
-                    ItemLauncher.launch(new BaseRowItem(0, response), null, -1, TvApp.getApplication().getCurrentActivity(), true);
-                }
-            });
+            try {
+                client.GetItemAsync(command.getItemId(), TvApp.getApplication().getCurrentUser().getId(), new Response<BaseItemDto>() {
```
nielsvanvelzen, line 144: Why not check if `getItemId()` is null/empty and return instead?

mueslimak3r: yeah that's much simpler

app/src/main/java/org/jellyfin/androidtv/data/eventhandling/TvApiEventListener.java
```diff
@@ -133,24 +133,21 @@ else if(playbackController != null && mediaManager.hasVideoQueueItems() && playb
     @Override
     public void onBrowseCommand(ApiClient client, BrowseRequest command) {
         Timber.d("Browse command received");
+        if (command.getItemId() == null || command.getItemId().equals("")) return;
```
nielsvanvelzen, line 136: In Java code use one of these:

https://github.com/jellyfin/jellyfin-androidtv/blob/0de3437e4e8d06e849f667ba683506a6423acac5/app/src/main/java/org/jellyfin/androidtv/util/Utils.java#L80-L86

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

PR #1387, opened 2022-01-25
shuffle the audio queue locally

mueslimak3r:
**Changes**

* allow `mediaManager` to manage shuffling a new audio queue. This means clicking "shuffle (all)" will create a queue with the shuffle button enabled, and toggling it off will give the user an un-shuffled queue.

* gave the relevant `playNow()` methods a `shuffle` (boolean) parameter

* bypassed `getItemsToPlay()` handling shuffling audio items. This could be done more cleanly by having it internally ignore the `shuffle` parameter for audio

* fixed the query sorting of MusicArtist so songs are sorted by album and are in order within the albums

**Issues**
* starting audio playback by using the "shuffle (all)" buttons didn't allow for truly un-shuffling
* clicking "play all" on a music artist would sometimes (but not always) queue the songs unordered

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/eventhandling/TvApiEventListener.java
+++ b/app/src/main/java/org/jellyfin/androidtv/data/eventhandling/TvApiEventListener.java
@@ -193,7 +193,7 @@ public class TvApiEventListener extends ApiEventListener {
                                     TvApp.getApplication().getCurrentActivity().startActivity(intent);
                                     break;
                                 case "Audio":
-                                    mediaManager.playNow(Arrays.asList(response.getItems()), startIndex);
+                                    mediaManager.playNow(Arrays.asList(response.getItems()), startIndex, false);
                                     break;
                             }
                         }
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
@@ -1599,9 +1599,7 @@ public class FullDetailsActivity extends BaseActivity implements IRecordingIndic
     }
 
     protected void play(final BaseItemDto item, final int pos, final boolean shuffle) {
-        boolean shuffleQuery = item.getBaseItemType() == BaseItemType.MusicAlbum || item.getBaseItemType() == BaseItemType.MusicArtist || item.getBaseItemType() == BaseItemType.Audio ? false : shuffle;
-
-        PlaybackHelper.getItemsToPlay(item, pos == 0 && item.getBaseItemType() == BaseItemType.Movie, shuffleQuery, new Response<List<BaseItemDto>>() {
+        PlaybackHelper.getItemsToPlay(item, pos == 0 && item.getBaseItemType() == BaseItemType.Movie, shuffle, new Response<List<BaseItemDto>>() {
             @Override
             public void onResponse(List<BaseItemDto> response) {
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/jellyfin/androidtv/data/eventhandling/TvApiEventListener.java
```diff
@@ -193,7 +193,7 @@ public void onResponse(ItemsResult response) {
                                     TvApp.getApplication().getCurrentActivity().startActivity(intent);
                                     break;
                                 case "Audio":
-                                    mediaManager.playNow(Arrays.asList(response.getItems()), startIndex);
+                                    mediaManager.playNow(Arrays.asList(response.getItems()), startIndex, false);
```
DavidFair, line 196: For these values could we use a named bool so future devs don't have to dive into the docs?

```kotlin
shuffle = false
mediaManager.playNow(Arrays.asList(response.getItems()), startIndex, shuffle);
```

I find it prevents code from gradually turning into this:

```kotlin
mediaManager.foo(response, true, false);
```

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
```diff
@@ -1599,15 +1599,16 @@ public void onResponse(UserItemDataDto response) {
     }
 
     protected void play(final BaseItemDto item, final int pos, final boolean shuffle) {
+        boolean shuffleQuery = item.getBaseItemType() == BaseItemType.MusicAlbum || item.getBaseItemType() == BaseItemType.MusicArtist || item.getBaseItemType() == BaseItemType.Audio ? false : shuffle;
```
DavidFair, line 1602: Can this be moved into it's own private method? That way the logic can be broken down and documented without causing `play` to grow

app/src/main/java/org/jellyfin/androidtv/data/eventhandling/TvApiEventListener.java
mueslimak3r: I'm definitely open to this.
I do think this accommodation is only useful in IDEs without syntax highlighting / introspection features.
In android studio, parameters like this have their name shown beside them.

DavidFair: Ah feel free to disregard this 👍 
I tend to use both vim and the IDE so this is why I spotted it intislly.
But I'm also not against either jumping to the source or using the proper tool compared to adding noise to the source code. Plus multiple booleans is indicative of a hidden dataclass we can pull out

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
mueslimak3r: I went ahead and removed all the uses of `shuffleQuery`.
Instead I just removed the use of `shuffle` for creating the queries for `MusicAlbum` & `MusicArtist` in `getItemsToPlay`. If shuffling audio queues locally is accepted, leaving `shuffle` in as an option for the queries, and working around it externally, is messy.

app/src/main/java/org/jellyfin/androidtv/data/eventhandling/TvApiEventListener.java
mueslimak3r: I'm a fan of vim too
Have you tried plugins like this:
https://github.com/uiiaoo/java-syntax.vim

I just tested with neovim and it does add nice colors to the code but it still doesn't provide the parameter names for methods.

nielsvanvelzen (approved):

PR #1394, opened 2022-01-28
Upgrade Android Gradle Plugin to 7.1.0

nielsvanvelzen:
Nothing special actually

**Changes**
- Upgrade Android Gradle Plugin to 7.1.0
- https://developer.android.com/studio/releases/gradle-plugin#settings-gradle
- Suppress RestrictedApi issues for Leanback library in LeanbackChannelWorker
  - Google marked pretty much all of the tvprovider/leanback functions as restricted but theres no other way to use them lol

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
@@ -130,6 +130,7 @@ class LeanbackChannelWorker(
 	/**
 	 * Updates the "my media" row with current media libraries.
 	 */
+	@Suppress("RestrictedApi")
 	private suspend fun updateMyMedia() {
 		// Get channel
 		val channelUri = getChannelUri("my_media", Channel.Builder()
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -130,6 +130,7 @@ class LeanbackChannelWorker(
 	/**
 	 * Updates the "my media" row with current media libraries.
 	 */
+	@Suppress("RestrictedApi")
```
mueslimak3r, line 133: Could one of the annotations in this doc be used instead to conditionally enable/disable the lean back channel stuff?

The doc says when it's eventually (and maybe already?) blocked in new versions of android, it could cause breakage.

https://developer.android.com/about/versions/10/non-sdk-q#code-annotations

nielsvanvelzen: The used API's are all from androidx.tvprovider. They shouldn't be restriced, just Google not maintaining their libraries for Android TV like always.
The linked annotations are meant for functions that use the platform API under the hood, tvprovider does that but detects if the platform functions are available and does nothing as fallback (so no crashes).

The reason I added the suppress annotations in this PR is because this version of the Gradle plugin added new (or improved existing) linter issues to find usages of restricted API's. If usage of these functions would break something it should've happened when we changes the Android SDK target in #1299.

mueslimak3r (approved):

PR #1369, opened 2022-01-20
Include Jellyfin Exoplayer FFmpeg dep

DavidFair:
**Changes**
Port the includes from the Android repo as-is into this repo to include
the dependency

Convert our initialisation to prefer the built in extensions. This
should avoid flaky device logic and matches the Android app choice too.

**Issues**
Partially Completes: #1308 

**To Test**
- Grab a video with DTS audio (free trailers with DTS can be downloaded from: https://thedigitaltheater.com/dts-trailers/)
- Ensure the playback is set to Exoplayer 
- Ensure playback is direct (not downsampling)
- Play media, ensure audio now plays

This should hopefully fix some of the issues people (including myself) are seeing with certain audio formats

PR #1401, opened 2022-01-31
simplified exoplayer code

mueslimak3r:
**Changes**
* put all the factory-related code in one place
* Replaced deprecated `DefaultDataSourceFactory` with `DefaultHttpDataSource.Factory`
* Replaced deprecated `Player.EventListener` with `Player.Listener`
* cut out unnecessary steps in setting the media

**Notes**
The `MediaSourceFactory` is only used to add the user agent (via `DefaultHttpDataSource`). If the user agent isn't actually necessary, that can be removed too. Then it would be all defaults, and would be a simple:
1) make the player builder, set the renderer, and build
2) set the media item and prepare

This cleanup was done while trying to figure out why the last few seconds of hls streams repeat. I tested adding a `DefaultExtractorsFactory` and setting:
`extractorsFactory.setFragmentedMp4ExtractorFlags(FragmentedMp4Extractor.FLAG_WORKAROUND_IGNORE_EDIT_LISTS);`
This is what the exoplayer docs recommended for parts of the video repeating. It didn't work. Neither did the corresponding flag for the `Mp4Extractor` 

I also tested using an `HlsMediaSource` and setting:
`exoPlayerBuilder.setMediaSourceFactory(new HlsMediaSource.Factory(baseDataSourceFactory).setAllowChunklessPreparation(true));`
This sped up video loading. However, there wasn't a simple way do this only when transcoding, so I ditched it.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
@@ -148,7 +148,7 @@ public class VideoManager implements IVLCVout.OnNewVideoLayoutListener {
                 new DefaultRenderersFactory(context) {
                     @Override
                     protected void buildTextRenderers(Context context, TextOutput output,
-                                                      Looper outputLooper, int extensionRendeexrerMode,
+                                                      Looper outputLooper, int extensionRendererMode,
                                                       ArrayList<Renderer> out) {
                         // Do not add text renderers since we handle subtitles
                     }
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
@@ -93,10 +93,7 @@ public class VideoManager implements IVLCVout.OnNewVideoLayoutListener {
         mSubtitlesSurface.setZOrderMediaOverlay(true);
         mSubtitlesSurface.getHolder().setFormat(PixelFormat.TRANSLUCENT);
 
-        ExoPlayer.Builder exoPlayerBuilder = new ExoPlayer.Builder(activity);
-        exoPlayerBuilder.setRenderersFactory(createExoplayerRenderersFactory(activity));
-        mExoPlayer = exoPlayerBuilder.build();
-
+        mExoPlayer = configureExoplayerBuilder(activity).build();
         mExoPlayerView = view.findViewById(R.id.exoPlayerView);
         mExoPlayerView.setPlayer(mExoPlayer);
         mExoPlayer.addListener(new Player.Listener() {
@@ -142,9 +139,11 @@ public class VideoManager implements IVLCVout.OnNewVideoLayoutListener {
      * ExoPlayer to silently fallback to software renderers.
      *
      * @param context The associated context
-     * @return A configured factory for Exoplayer
+     * @return A configured builder for Exoplayer
      */
-    private DefaultRenderersFactory createExoplayerRenderersFactory(Context context) {
+    private ExoPlayer.Builder configureExoplayerBuilder(Context context) {
+        ExoPlayer.Builder exoPlayerBuilder = new ExoPlayer.Builder(context);
+
         DefaultRenderersFactory defaultRendererFactory =
                 new DefaultRenderersFactory(context) {
                     @Override
@@ -156,7 +155,9 @@ public class VideoManager implements IVLCVout.OnNewVideoLayoutListener {
                 };
         defaultRendererFactory.setEnableDecoderFallback(true);
         defaultRendererFactory.setExtensionRendererMode(DefaultRenderersFactory.EXTENSION_RENDERER_MODE_ON);
-        return defaultRendererFactory;
+        exoPlayerBuilder.setRenderersFactory(defaultRendererFactory);
+
+        return exoPlayerBuilder;
     }
 
     public boolean isInitialized() {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -98,13 +98,18 @@ public VideoManager(PlaybackOverlayActivity activity, View view) {
         mSubtitlesSurface.setZOrderMediaOverlay(true);
         mSubtitlesSurface.getHolder().setFormat(PixelFormat.TRANSLUCENT);
 
-        mExoPlayer = new ExoPlayer.Builder(TvApp.getApplication(), new DefaultRenderersFactory(TvApp.getApplication()) {
+        ExoPlayer.Builder exoPlayerBuilder = new ExoPlayer.Builder(TvApp.getApplication());
```
DavidFair, line 101: (Pinching the comments from nielsvanvelzen's review on the same part of code)
Both `TvApp.getApplication()` calls should be replaced with `activity`

DavidFair (commented):
Just an FYI we're going to have some merge conflicts in VideoManager depending on who lands first as we've both changed the factory invocation:

https://github.com/jellyfin/jellyfin-androidtv/pull/1369/files#diff-45fafbc82b0431551a7258bf23dc1e1d47de746431e359cd1b4812fa37e20c42R102

mueslimak3r:
> Just an FYI we're going to have some merge conflicts in VideoManager depending on who lands first as we've both changed the factory invocation:

Yeah. Fortunately all I did is separate constructing the exoplayer builder and setting the factories, so it's less cluttered.
Instead of adding each factory as a parameter to the builder you just make the builder, and then do `exoPlayerBuilder.setSomeFactory(...)`

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -98,17 +98,22 @@ public VideoManager(PlaybackOverlayActivity activity, View view) {
...
+        ExoPlayer.Builder exoPlayerBuilder = new ExoPlayer.Builder(activity);
+        exoPlayerBuilder.setRenderersFactory(new DefaultRenderersFactory(activity) {
             @Override
             protected void buildTextRenderers(Context context, TextOutput output, Looper outputLooper, int extensionRendererMode, ArrayList<Renderer> out) {
                 // Do not add text renderers since we handle subtitles
             }
-        }).build();
+        });
+
+        DataSource.Factory baseDataSourceFactory = new DefaultHttpDataSource.Factory().setUserAgent("ATV/ExoPlayer");
```
nielsvanvelzen, line 109: If we're adding a user agent here we should probably add the app version too.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
mueslimak3r: I only included the data source factory stuff because the removed `DefaultDataSourceFactory` used a user agent. If it's unneeded I'd rather just remove the whole thing.

Otherwise I can look into adding the app version to the user agent

nielsvanvelzen: I don't think we really need to set the user agent. The SDK doesn't set one either (although the http library might set one).

nielsvanvelzen:
Can you rebase this PR? Looks like we both changed some stuff in the contractVideo function which could cause issues after merge.

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -544,7 +535,7 @@ private void createPlayer(int buffer, boolean isInterlaced) {
             options.add("-v");
             options.add("--vout=android-opaque,android-display");
 
-            mLibVLC = new LibVLC(TvApp.getApplication(), options);
+            mLibVLC = new LibVLC(mActivity, options);
```
mueslimak3r, line 538: was the `TvApp.getApplication()` just missed in a previous PR and removing it is okay, or was it left in for a reason?
Same for the toast on line 556

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -147,7 +144,7 @@ public void onTimelineChanged(Timeline timeline, int reason) {
      * @param context The associated context
      * @return A configured factory for Exoplayer
      */
-    private DefaultRenderersFactory configureExoplayerFactory(Context context) {
+    private DefaultRenderersFactory createExoplayerRenderersFactory(Context context) {
```
mueslimak3r, line 147: @DavidFair does this rename look okay? imo it makes sense when it's called with `exoPlayerBuilder.setRenderersFactory(createExoplayerRenderersFactory(activity));`

mueslimak3r: Nevermind, I'm gonna change the name back and have it return the exoplayer builder.

Then any future factory stuff (like setting a track selector) can go inside that function

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
nielsvanvelzen: Probably forgotten

nielsvanvelzen (approved):

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
DavidFair: Yeah forgotten, I tend to be good with big picture stuff but miss little details like this :+1: 
Thanks for fixing that

PR #1416, opened 2022-02-02
Remove "i" prefix from interfaces

nielsvanvelzen:
Consistency stuff. A bunch of interfaces used the prefix and a bunch don't. Now they all don't.

**Changes**
- Remove "I" prefix from interfaces (so IGridLoader becomes GridLoader)
- Remove GridLoader interface (unnecessary)

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/GuideChannelHeader.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/GuideChannelHeader.java
@@ -14,7 +14,7 @@ import androidx.core.content.ContextCompat;
 import com.bumptech.glide.Glide;
 
 import org.jellyfin.androidtv.R;
-import org.jellyfin.androidtv.ui.livetv.ILiveTvGuide;
+import org.jellyfin.androidtv.ui.livetv.LiveTvGuide;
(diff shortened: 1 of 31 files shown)


DavidFair:
Could we create some sort of wiki page / .md file / something else to note down various project coding standards, as I've been caught out by this too?

The points I've picked up so far which apply to all PRs for example are:
- `TvApp.getApplication` should replaced with activities where possible
- Use fragments instead of new activities
- Interfaces do not use an `I` prefix
...etc.

It would be nice to have a single de-facto page for new and existing contributors to work off

thornbill:
Swiftfin uses a CONTRIBUTING.md file for those types of notes. It might be a good idea to do the same here... in a different PR of course. 😄

mueslimak3r (approved):

nielsvanvelzen:
I did make a list of reviewing rules in my personal project board: https://github.com/jellyfin/jellyfin-androidtv/projects/15. Eventually this could be moved to a new contributing document, I never bothered with such a document because we didn't see a lot of contributors in this repository, and most of them wouldn't even read such a document (unfortunately).

PR #1432, opened 2022-02-09
exoplayer track selection for audio

mueslimak3r:
**Changes**
* track selection for exoplayer - matches MediaStream index to exoplayer tracks. I left in the ability to select subtitles, though it isn't used. If needed in the future, video track selection should work too.
* removed the buildTextRenderers override. Instead, subtitle track selection is disabled by default.
* updated selectAudioIndex to allow changing exoplayer audio track without transcoding (for embedded tracks)
* Removed the code that forced transcoding due to not having exo track selection

**Issues**
* playing an audio track other than the first required transcoding for exoplayer

**Notes**
* To take advantage of these changes the audio track need to be embedded, and playback needs to be direct
* Disable "downmix to stereo" when testing to prevent audio transcoding - also I suspect this setting might not be needed at all for exoplayer, and it needlessly causing transcoding


app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -405,6 +423,135 @@ public boolean setSubtitleTrack(int index, @Nullable List<MediaStream> allStream
...
+                    // exoplayer wont have the external tracks so don't count them
+                    tmpNdx++;
+                }
+            }
+        }
+
+        if (mediaStreamFilteredIndex < 0)
+            return false;
+
+        // loop through exoplayer's tracks, counting up when the track type matches
```
mueslimak3r, line 496:
```suggestion
        // loop through exoplayer's tracks, counting up for each track matching the chosen type
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -405,6 +423,135 @@ public boolean setSubtitleTrack(int index, @Nullable List<MediaStream> allStream
         return false;
     }
 
+    public int getExoPlayerTrack(MediaStreamType streamType, @Nullable List<MediaStream> allStreams) {
```
nielsvanvelzen, line 426: Please add nullable notations for new (non-Kotlin) functions like the linter warning asks.

nielsvanvelzen (approved):

PR #1435, opened 2022-02-11
show buffered position in seekbar for exoplayer

mueslimak3r:
**Changes**
* get the buffered position from exoplayer and use it to update the overlay's seekbar
  this is disabled for libVLC since afaik it can't report buffered position.
* move getDuration from the VideoAdapter to PlaybackController
* removed check for finishedInitialSeek when setting mFragment.setCurrentTime. I think this was left in by mistake, and leaving it in causes the seekbar to show 0 until playback starts.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -1458,6 +1458,27 @@ public class PlaybackController {
         });
     }
 
+    public long getDuration() {
+        if (hasInitializedVideoManager()) {
+            return mVideoManager.getDuration();
+        } else if (getCurrentlyPlayingItem() != null && getCurrentlyPlayingItem().getRunTimeTicks() != null) {
+            return getCurrentlyPlayingItem().getRunTimeTicks() / 10000;
+        }
+        return -1;
+    }
+
+    public long getBufferedPosition() {
+        long bufferedPosition = -1;
+
+        if (hasInitializedVideoManager() && mVideoManager.getBufferedPosition() > -1) {
+            bufferedPosition = mVideoManager.getBufferedPosition();
+        } else {
+            bufferedPosition = getDuration();
+        }
+
+        return bufferedPosition;
+    }
+
     public long getCurrentPosition() {
         // don't report the real position if seeking
         return !isPlaying() && mSeekPosition != -1 ? mSeekPosition : mCurrentPosition;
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -221,6 +221,22 @@ public long getDuration() {
         }
     }
 
+    public long getBufferedPosition() {
+        if (!isInitialized())
+            return -1;
+
+        // to do: add libVLC support
```
nielsvanvelzen, line 228: todo comments should be ` // TODO ` or `// FIXME`.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
mueslimak3r: Good to know for the future. I turned this comment from a TODO into a "why this is exoplayer only" instead

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -1458,6 +1458,27 @@ public void onEvent() {
         });
     }
 
+    public long getDuration() {
```
nielsvanvelzen, line 1461: There is a few usages of mVideoManager.getDuration in this same file (seek functions). Should those be changed to use this one?

mueslimak3r: I replaced mVideoManager.getDuration where appropriate, and changed getDuration and getBufferedPosition to fallback to 0 instead of -1. I also moved the logic for calculating a safe skip/seek position into a utility so it can be used by `skip()` and `seek()`

nielsvanvelzen (approved):

nielsvanvelzen:
@mueslimak3r I found an issue with this PR after I merged it and started testing other PR's (I often rebase PR's I test on top of master).

When playing an intro before a movie the duration of the video is not updated. So in my case the intro is 4 seconds and the movie is 10 minutes. After the intro ended the movie starts playing, the progress increments as expected, but the duration of the movie is 4 seconds.

Before PR:

image

After PR:

image

The "ends at" time is correct though.

mueslimak3r:
> @mueslimak3r I found an issue with this PR after I merged it and started testing other PR's (I often rebase PR's I test on top of master).
> 
> When playing an intro before a movie the duration of the video is not updated. So in my case the intro is 4 seconds and the movie is 10 minutes. After the intro ended the movie starts playing, the progress increments as expected, but the duration of the movie is 4 seconds.
> 
> The "ends at" time is correct though.

Is that using jellyfin-plugin-intros - https://github.com/dkanada/jellyfin-plugin-intros?

nielsvanvelzen:
Yes

mueslimak3r:
fixed

PR #1436, opened 2022-02-11
Refactor unit tests to Kotlin with Kotest and Mockk

nielsvanvelzen:
Refactor all current unit tests to a consistent way using Kotest and Mockk. I did not change the behavior of the tests, although the names or implementation may be changed. The CustomSeekProviderTests file is the easiest to compare before<>after.

**Changes**
- All unit tests are now Kotlin (converted TimeUtilsTests)
- All unit tests are now in the "kotlin" source directory
  - No top-level package directory, package-info.kt as helper for IntelliJ based editors
- Disabled wildcard imports for unit tests (complete project in upcoming future PR)
- Remove JUnit, kotlin.test and Koin test dependencies
- Added Kotest for test definitions and assertion
  - Always use FunSpec
  - Use infix assertion functions where possible
- Change `mockk()` calls to use DSL for the `every` definitions
- Always use `[classname]Tests` as name (was `[classname]Test` before)
- Use descriptive names for tests that include spaces to make it easy to read

**Issues**


nielsvanvelzen:
@Maxr1998 I'd like to get testing consistent between jellyfin-androidtv, jellyfin-sdk-kotlin and jellyfin-android. If you have some spare time, could you look at this PR and let me know what you think? That way we don't need to update this one later when we start adding tests the android app or refactor those in the SDK.

Maxr1998:
Sorry for the late reply. Looks good to me. Have been using Kotest for most of my other projects too (albeit with a different test style, but FunSpec is ok for me).

We only should ensure that
```kt
class TimeUtilsTests : FunSpec({
    …
})
```
won't get reformatted to
```kt
class TimeUtilsTests : FunSpec(
    {
        …
    }
)
```
since that's rather ugly :sweat_smile:

nielsvanvelzen:
> We only should ensure that
> 
> ```kotlin
> class TimeUtilsTests : FunSpec({
>     …
> })
> ```
> 
> won't get reformatted to
> 
> ```kotlin
> class TimeUtilsTests : FunSpec(
>     {
>         …
>     }
> )
> ```

Definitely agree with this, looks like the build-in formatter in Android Studio / IntelliJ agrees as well so no problems there.

Maxr1998 (approved):

PR #1440, opened 2022-02-15
only use HLS for live tv - otherwise use progressive mkv

### LINKED ISSUE #1410: 0.13beta1 track transcoding bug
### Describe the bug

1.Select exoPlayer to play
2.Audio selection pass through
Then, On devices that support AC3 and DTS, track transcoding is AAC and silent.
If mixed stereo is selected, the track is transcoded to AAC and works is OK.
The old version will be transcoded to AC3 according to the option, and it works well.
We hope to add the option of forced AC3 / DTS transcoding to make more devices work.This is very important.Android devices are difficult to set up multi-channel, and the transcoding of the old version of AC3 / DTS is a miracle.
Thanks.

### Logs

_No response_

### Application version

0.13 beta1

_No response_

### Device information

android TV

### Android version

Android 7,Android 9

### Jellyfin server version

10.8.0 alpha5

mueslimak3r:
**Changes**
* revert to using progressive mkv for everything but live tv which will still use HLS
* added a condition to `getBufferedPosition()` that excludes live-tv. The logic for getting the current position when playing live tv is kinda broken, especially if the player is paused and then resumed.

**Issues**
* using mkv again should resolve #1410 though I'm not sure about live-tv contexts
* after seeking using HLS, the video would be slightly early but the timestamp wouldn't reflect it.
  This caused:
  * subtitles to be out of sync
  * the last few seconds of a video looping

  According to ffmpeg docs, the flag `-noaccurate_seek`, which jellyfin uses, supplies a stream starting at the key-frame prior to the requested timestamp. This is incompatible with exoplayer's frame-accurate seeking.
The ffmpeg docs (https://trac.ffmpeg.org/wiki/Seeking):
  > -ss is now also "frame-accurate" even when used as an input option. Previous behavior (seeking only to the nearest 
     preceding keyframe, even if not precisely accurate) can be restored with the -noaccurate_seek option

  To resolve this and use HLS in contexts that allow seeking, the version of exoplayer we use will need to support `SeekParameters` for HLS. As far as I can tell, the flag we'd use is ``SeekParameters.PREVIOUS_SYNC``
  The relevant issue is here:
  https://github.com/google/ExoPlayer/issues/2882
  And here is the merged PR that _should_ allow this once the next exoplayer release comes out:
  https://github.com/google/ExoPlayer/pull/9536

**Notes**
* using HLS has made the player **far** more stable. If there's a way to synchronize the subtitles despite the player being unaware of the desync then that would be much better than abandoning HLS.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -1448,7 +1448,8 @@ public class PlaybackController {
     public long getBufferedPosition() {
         long bufferedPosition = -1;
 
-        if (hasInitializedVideoManager())
+        // FIXME handle live tv buffered position and position while paused
+        if (!isLiveTv() && hasInitializedVideoManager())
             bufferedPosition = mVideoManager.getBufferedPosition();
 
         if (bufferedPosition < 0)
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
@@ -45,29 +45,50 @@ class ExoPlayerProfile(
 	init {
 		name = "AndroidTV-ExoPlayer"
 
-		transcodingProfiles = arrayOf(
-			// MP4 video profile
-			TranscodingProfile().apply {
-				type = DlnaProfileType.Video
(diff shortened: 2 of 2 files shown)


mueslimak3r:
I'm leaving this as a draft pending this server issue since it seems to only happen with mkv, and while testing this PR I encountered it a bunch:
https://github.com/jellyfin/jellyfin/issues/6957

I've been digging around for a client-side cause of this but haven't found anything

PR #1463, opened 2022-02-22
Update process lifecycle to avoid startup crash

nielsvanvelzen:
An update in androidx.startup and androidx.lifecycle fixed an issue where the app _could_ crash during startup. I haven't encountered the issue but it was a good excuse to update the code to not use the deprecated annotation based lifecycle event listeners.

**Changes**
- Update androidx.startup to 1.1.1
- Update androidx.lifecycle to 2.4.1
- Update process lifecycle code to not use annotations

**Issues**


app/src/main/java/org/jellyfin/androidtv/JellyfinApplication.kt
```diff
@@ -1,10 +1,7 @@
 package org.jellyfin.androidtv
 
 import android.content.Context
-import androidx.lifecycle.Lifecycle
-import androidx.lifecycle.LifecycleObserver
-import androidx.lifecycle.OnLifecycleEvent
-import androidx.lifecycle.ProcessLifecycleOwner
+import androidx.lifecycle.*
```
crobibero, line 4: Think this will conflict with the other PR?

nielsvanvelzen: Yup!

mueslimak3r (approved):

PR #1466, opened 2022-02-23
use ts for exoplayer transcodes

### LINKED ISSUE #1410: 0.13beta1 track transcoding bug

mueslimak3r:
**Changes**
* remove the code that prevents seeking when the container is TS
* replaced MP4 with TS (still uses HLS) in `ExoplayerProfile`
* modify the TS search bytes to match what jellyfin-android does:
  https://github.com/jellyfin/jellyfin-android/blob/6d42621e16bc38d805cf2861a6a10f65ca2d626c/app/src/main/java/org/jellyfin/mobile/app/AppModule.kt#L77
* update `addAc3Streaming` so it can enable AC3 for more than one container (mkv, ts)

**Issues**
* fixes #1410

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
@@ -139,11 +139,6 @@ object ProfileHelper {
 			)
 		}
 
-	private fun findTranscodingProfile(
-		deviceProfile: DeviceProfile,
-		container: String
-	) = deviceProfile.transcodingProfiles.find { it.container == container }
-
 	internal fun subtitleProfile(
 		format: String,
 		method: SubtitleDeliveryMethod
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
@@ -143,12 +143,16 @@ object ProfileHelper {
 	fun addAc3Streaming(profile: DeviceProfile, primary: Boolean) {
 		if (Utils.downMixAudio()) return
 
-		val mkvProfile = findTranscodingProfile(profile, ContainerTypes.MKV) ?: return
+		val containers = arrayOf(ContainerTypes.MKV, ContainerTypes.TS)
+		containers.forEach { c ->
+			Timber.d("trying to match transcoding profile for %s", c)
+			val transcodingProfile = findTranscodingProfile(profile, c) ?: return@forEach
 
-		Timber.i("*** Adding AC3 as supported transcoded audio")
-		mkvProfile.audioCodec = when (primary) {
-			true -> "${CodecTypes.AC3},${mkvProfile.audioCodec}"
-			false -> "${mkvProfile.audioCodec},${CodecTypes.AC3}"
+			Timber.i("*** Adding AC3 as supported transcoded audio for %s", c)
+			transcodingProfile.audioCodec = when (primary) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -143,12 +143,16 @@ object ProfileHelper {
...
+		val containers = arrayOf(ContainerTypes.MKV, ContainerTypes.TS)
+		containers.forEach { c ->
+			Timber.d("trying to match transcoding profile for %s", c)
+			val transcodingProfile = findTranscodingProfile(profile, c) ?: return@forEach
+
+			Timber.i("*** Adding AC3 as supported transcoded audio for %s", c)
+			transcodingProfile.audioCodec = when (primary) {
+				true -> "${CodecTypes.AC3},${transcodingProfile.audioCodec}"
+				false -> "${transcodingProfile.audioCodec},${CodecTypes.AC3}"
+			}
```
nielsvanvelzen, line 155: Looking at the usages of `addAc3Streaming` the primary argument is always true so this function can be simplified a bit.

nielsvanvelzen: In fact, it's only used for exoplayer profiles so we might just pass a boolean to the exoplayerprofile constructor whether to enable or disable ac3 and remove this helper function.

https://github.com/jellyfin/jellyfin-androidtv/blob/8424e581583743d7d0a09902716defbfde875bc7/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java#L498-L502

mueslimak3r: `UserPreferences.Companion.getAc3Enabled()` looks like it just resolves to !DeviceUtils.isFireTvStickGen1()
https://cs.github.com/jellyfin/jellyfin-androidtv?q=pref_bitstream_ac3

So anything using exoplayer profile will qualify, and ac3 might as well just be added to the profile by default?

nielsvanvelzen: The preference can still be set in the app. It's disabled by default on fire tv.

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -139,23 +139,6 @@ object ProfileHelper {
...
-
-			Timber.i("*** Adding AC3 as supported transcoded audio for %s", c)
-			transcodingProfile.audioCodec = when (primary) {
-				true -> "${CodecTypes.AC3},${transcodingProfile.audioCodec}"
-				false -> "${transcodingProfile.audioCodec},${CodecTypes.AC3}"
-			}
-		}
-	}
-
 	private fun findTranscodingProfile(
```
nielsvanvelzen, line 142: This one is unused now and can be removed

nielsvanvelzen (approved):

PR #1470, opened 2022-02-25
Move to view binding in most views

nielsvanvelzen:
Rewrite layout inflation to use view binding in custom views / fragments / activities where possible. Some places I did not convert are live tv related and/or hacks.

findViewById usages went from 169 to 85

**Changes**
- Move to view binding
- Remove a bunch of unused code from changed files

**Issues**


mueslimak3r (approved):
nice, the ui is noticeably snappier

Issue #1482, opened 2022-03-02
refresh rate switching also change resolution


Arial-Z:
### Describe the bug

Using the refresh rate switch on 4K display when playing 720p/1080p media, the refresh rate is switched to match the video as intended, but the display resolution also switch to 1080p which will block the needed upscale from the nvidia shield.

Tested 4K media working as intended (4K resolution and 23.976Hz)

Tested with both exoplayer and libVLC.

### Application version

0.13.0-beta.4

### Device information

Nvidia Shield Pro 2019

### Android version

Android 11

### Jellyfin server version

10.7.7

mueslimak3r:
Can you provide some details about the display you're using?

mueslimak3r:
can you try side loading this build from #1484  (the apk is in the build artifacts after it finishes)
https://github.com/jellyfin/jellyfin-androidtv/actions/runs/1926392404

Arial-Z:
Display is a LG55C1
Will test the APK today

Arial-Z:
Build fixed the issue 23,976 FPS 1080p files play at 23.976Hz 4K

Don't know if i should create a new issue but i tested  some other files 24FPS and 25 FPS
25 FPS play at 50Hz whick is fine 1080p in beta4 and 4K in dev build
Vut 24 FPS play at 60Hz 1080p in beta4 and 4K in dev build

mueslimak3r:
> 25 FPS play at 50Hz whick is fine 1080p in beta4 and 4K in dev build

Is that good (I'm not too familiar with what's ideal for variable refresh rate displays but 50Hz being 2x 25Hz is even frame pacing

>  24 FPS play at 60Hz 1080p in beta4 and 4K in dev build

Is the problem it being 60Hz or it being 4K? If the problem is it being 60Hz that's due to the logic that's supposed to handle 2:3 pull down (https://en.wikipedia.org/wiki/Three-two_pull_down#2:3) but idk if that's working as expected.

> Don't know if i should create a new issue

nope, I think this is all relevant to this issue.

Arial-Z:
25 FPS at 50Hz is good yes it's a round multiple 1:2, for 24 FPS i would have expected the same, display at 24Hz or 48Hz because if it's playing at 60Hz with a pull down what the point of using the option to sync the display refresh rate with the video, 60Hz pull down is the normal behavior.

mueslimak3r:
[...]

PR #1484, opened 2022-03-03
prefer changing refresh rate without changing resolution

mueslimak3r:
**Changes**
* modes that have the same resolution as the default mode will rank higher than other modes. This will hopefully result in the resolution changing in addition to the refresh rate only if there are no other options.

**Issues**
#1482 

**Notes**
* My display doesn't support the refresh rates needed for this feature. Someone else will need to test it.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -317,11 +317,13 @@ public class PlaybackController {
 
             Timber.d("qualifying display mode: %sx%s @%sfps", mode.getPhysicalWidth(), mode.getPhysicalHeight(), mode.getRefreshRate());
 
-            int weight;
-            if (mode.getPhysicalWidth() == defaultMode.getPhysicalWidth() && mode.getPhysicalHeight() == defaultMode.getPhysicalHeight())
-                weight = (rate * 10000) + 9999;
-            else
-                weight = (rate * 10000) + 9999 - mode.getPhysicalWidth() - videoStream.getWidth();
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (approved):

KyleSanderson:
@nielsvanvelzen can you revert this? This damages every single user not on a shield as they can't use their TV upscaler anymore.

mueslimak3r:
> This damages every single user not on a shield as they can't use their TV upscaler anymore.

Which shield model do you have?
It's interesting that it broke the upscaler for you but fixed the upscaler for the user who opened #1482

KyleSanderson:
> > This damages every single user not on a shield as they can't use their TV upscaler anymore.
> 
> Which device are you using? It's interesting that it broke the upscaler for you but fixed the upscaler for the user who opened #1482

>Using the refresh rate switch on 4K display when playing 720p/1080p media, the refresh rate is switched to match the video as intended, but the display resolution also switch to 1080p which will block the needed upscale from the nvidia shield.

I do not want to use my device upscaler. My TV is the correct upscaler, as it is for everyone outside of Shield users. The native resolution going to the TV is correct, and should be a "disable" for Shield users as an opt-in only as it's absolutely not desirable behaviour.

Dnkhatri:
I think Kodi implementation of whitelist for resolution and refresh rate matching would probably the best way to go.  

https://github.com/xbmc/xbmc/pull/13274
https://github.com/xbmc/xbmc/pull/16216

nielsvanvelzen:
@KyleSanderson can you open a new issue for your problem? It's easier to track then in the comments of a merged PR. I guess the "easy" solution for now is to have 2 toggles; one for refresh rate switching and one for resolution switching.

PR #1469, opened 2022-02-24
use track id for exoplayer track selection instead of guessing

mueslimak3r:
**Changes**
* rely on matching exoplayer track IDs (start at 1) to MediaStream IDs (start at 0) by subtracting 1
   This definitely needs plenty of testing to verify the exoplayer track IDs can always be matched this way.

**Issues**
fixes #1460

PR #1465, opened 2022-02-23
subtitle fixes for PR 1455

mueslimak3r:
**Changes**
* set `mDefaultSubIndex` when a subtitle track is selected/cleared
* don't search for subtitle events if the current position is before the start of the first event
* clear subtitles  after seeking

**Issues**
* if subtitles were enabled and then disabled, they would be re-enabled after seeking
* a subtitle event might not clear after seeking, and would stay on screen until a new event is ready (new from #1455)
* if playback starts at the beginning/before the start of the first subtitle event, each call to `updateSubtitles()` would search through the whole list of events without doing anything (new from #1455)

PR #1487, opened 2022-03-03
fixes for crashes from subtitle rendering and audio track selection

### LINKED ISSUE #1488: Crash when Live TV playback starts
### Describe the bug

Recently started to crash when I try to view any channel in Live TV. Picture and sound is present just for an instant then the app crashes. I got beta version from playstore and guess it occured now with the latest beta release.

Didn't change any settings what I can remember, updated my jellyfin server aswell to see if that would fix it, without success.

1. Go to Live TV
2. Play any channel

### Logs

_No response_

### FFmpeg logs

_No response_

### Media info of the file

_No response_

### Application version

0.13.0-beta.4

### Device information

Nvidia Shield TV 2019

### Android version

Android 11

### Jellyfin server version

10.8.0

mueslimak3r:
**Changes**
* catch subtitle track events size being < 1
* catch `NumberFormatException` if an exoplayer track ID isn't/can't be parsed to an integer, and return -1 so another method is used to get the current track, and to switch tracks (transcoding).

**Issues**
* potentially fixes #1488 
* https://github.com/jellyfin/jellyfin-androidtv/pull/1469#issuecomment-1058027334
* https://github.com/jellyfin/jellyfin-androidtv/pull/1465#issuecomment-1058029794

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
@@ -464,7 +464,13 @@ public class VideoManager implements IVLCVout.OnNewVideoLayoutListener {
                 if (trackType == chosenTrackType) {
                     if (groupInfo.isTrackSelected(i)) {
                         if (trackFormat.id != null) {
-                            int id = Integer.parseInt(trackFormat.id) - 1;
+                            int id;
+                            try {
+                                id = Integer.parseInt(trackFormat.id) - 1;
+                            } catch (NumberFormatException e) {
+                                Timber.d("failed to parse track ID [%s]", trackFormat.id);
+                                return -1;
+                            }
                             if (id >= 0 && id < allStreams.size())
                                 return id;
                         }
@@ -535,7 +541,14 @@ public class VideoManager implements IVLCVout.OnNewVideoLayoutListener {
 
                 if (trackType != chosenTrackType || trackFormat.id == null)
                     continue;
-                if (Integer.parseInt(trackFormat.id) != exoTrackID) {
+
+                int id;
+                try {
+                    id = Integer.parseInt(trackFormat.id);
+                    if (id != exoTrackID)
+                        continue;
+                } catch (NumberFormatException e) {
+                    Timber.d("failed to parse track ID [%s]", trackFormat.id);
                     continue;
                 }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -464,7 +464,13 @@ public int getExoPlayerTrack(@Nullable MediaStreamType streamType, @Nullable Lis
...
                     if (groupInfo.isTrackSelected(i)) {
                         if (trackFormat.id != null) {
-                            int id = Integer.parseInt(trackFormat.id) - 1;
+                            int id;
+                            try {
+                                id = Integer.parseInt(trackFormat.id) - 1;
+                            } catch (NumberFormatException e) {
+                                Timber.d("failed to parse track ID [%s]", trackFormat.id);
+                                return -1;
+                            }
```
nielsvanvelzen, line 473: I don't think we should ever assume the id is an integer at all. There must be a reason the ExoPlayer developers made it a string.

mueslimak3r: It looks like mpeg-ts is the main culprit as far as using something other than an integer for the ID.
I've come across two variations so far:
* Menu ID/ID (this is what's seen in all of the crashes eg: "17103/221")
* codec:title (eg: aac:rus a1)

Here's a good example:
https://github.com/google/ExoPlayer/issues/8824#issuecomment-823184070

I don't know of any ways to match a MediaStream track to an Exoplayer track other than using ordering or matching a property like the ID. In this case, perhaps ordering could be used second option though it would open up the possibility of guessing wrong.

As it is, as long as the exception is caught, if the track matching fails it will fall-back to switching tracks by transcoding like it did before track selection was added (and the exo track selection code isn't used while transcoding).

nielsvanvelzen: Would it be possible to return the index in the track group instead of the id?

mueslimak3r: > Would it be possible to return the index in the track group instead of the id?

* From my testing, there is 1 track per `TrackGroup`, and 1 `TrackGroup` per `TrackGroupInfo`.
* The docs do say that `TrackGroup` should have its own ID
https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/source/TrackGroup.html
  but trying to use it in the project says it doesn't exist. 

> Or maybe even use a string in our own code.

Some tracks do have labels, but not all do. And there's no guarantee that there wouldn't be multiple tracks with the label "english" for instance.

nielsvanvelzen (approved):

PR #1489, opened 2022-03-04
Open context menu on long press

nielsvanvelzen:
_technically a feature but let's sneak it in for 0.13..._

This is a bit of a hack, but it will do for now.

**Changes**
- Emit a "menu" key press when long pressing an LegacyImageCardView
  - This causes the context menu to show for most cards
  - Enables a feature previously unusable on remotes without a dedicated menu buttons (like the Nvidia Shield)

**Issues**


nielsvanvelzen:
image

for reference, this is the menu that now works on long press

thornbill (approved):
Untested, but looks safe enough :rocket:

PR #1490, opened 2022-03-04
Remove like & dislike

nielsvanvelzen:
Remove the like/dislike functionality from the app. AFAIK we're the only client that used it and the like/dislike state was never shown in any UI.

**Changes**
- Remove "like", "dislike", "unlike" and "undislike" (lol) from the item menu
- Remove API calls to set like/dislike from the KeyProccessor
- Remove now unused strings related to likes/dislikes
  - Also updated one string to not mention likes
- Remove some other unused variables from KeyProcessor

**Issues**


thornbill (approved):
:fire:

PR #1501, opened 2022-03-07
require at least server version 10.8.0 to use HLS

mueslimak3r:
**Changes**
* added a method to return whether the current server meets a given minimum version requirement
  _it should probably be moved somewhere more suitable_
* exoplayer profile only adds its transcoding profile if HLS is enabled

* fixed transcoding profile still using `ContainsCodec(CodecTypes.HEVC, ContainerTypes.MP4)` instead of `ContainsCodec(CodecTypes.HEVC, ContainerTypes.TS)`
* removed unused variable in playback controller

**Issues**
* HLS transcoding is partially broken prior to 10.8.0 PR:
  https://github.com/jellyfin/jellyfin/pull/7338
  prior to that PR, HLS transcode streams may repeat the last few seconds of a video or not seek accurately
  
  **Notes**
Please suggest a better place to store the `getServerMeetsMinimumVersion()` method

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
@@ -45,6 +45,9 @@ import com.bumptech.glide.request.target.Target;
 
 import org.jellyfin.androidtv.R;
 import org.jellyfin.androidtv.TvApp;
+import org.jellyfin.androidtv.auth.AuthenticationStore;
+import org.jellyfin.androidtv.auth.SessionRepository;
+import org.jellyfin.androidtv.auth.model.AuthenticationStoreServer;
 import org.jellyfin.androidtv.constant.CustomMessage;
 import org.jellyfin.androidtv.data.model.DataRefreshService;
 import org.jellyfin.androidtv.databinding.OverlayTvGuideBinding;
@@ -85,12 +88,14 @@ import org.jellyfin.apiclient.model.livetv.ChannelInfoDto;
 import org.jellyfin.apiclient.model.livetv.SeriesTimerInfoDto;
 import org.jellyfin.apiclient.model.mediainfo.SubtitleTrackEvent;
 import org.jellyfin.apiclient.model.mediainfo.SubtitleTrackInfo;
+import org.jellyfin.sdk.model.ServerVersion;
 import org.koin.java.KoinJavaComponent;
 
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.List;
+import java.util.UUID;
 
 import kotlin.Lazy;
 import timber.log.Timber;
@@ -150,6 +155,8 @@ public class CustomPlaybackOverlayFragment extends Fragment implements LiveTvGui
 
     private final Lazy<ApiClient> apiClient = inject(ApiClient.class);
     private final Lazy<MediaManager> mediaManager = inject(MediaManager.class);
+    private final Lazy<SessionRepository> sessionRepository = inject(SessionRepository.class);
+    private final Lazy<AuthenticationStore> authenticationStore = inject(AuthenticationStore.class);
 
     @Override
     public void onCreate(Bundle savedInstanceState) {
@@ -1486,5 +1493,30 @@ public class CustomPlaybackOverlayFragment extends Fragment implements LiveTvGui
             binding.subtitlesText.setText(span);
             binding.subtitlesText.setVisibility(View.VISIBLE);
         });
+    }
+
+    public boolean getServerMeetsMinimumVersion(@NonNull String minServerVersion) {
+        if (sessionRepository.getValue().getCurrentSession().getValue() == null) {
+            Timber.d("failed to get current session");
+            return false;
+        }
+
+        UUID serverId = sessionRepository.getValue().getCurrentSession().getValue().getServerId();
+        AuthenticationStoreServer server = authenticationStore.getValue().getServer(serverId);
+        if (server == null || server.getVersion() == null) {
+            Timber.d("server is null or version is null");
+            return false;
+        }
+
+        ServerVersion version = ServerVersion.Companion.fromString(server.getVersion());
+        ServerVersion minVersion = ServerVersion.Companion.fromString(minServerVersion);
+        if (version == null || minVersion == null) {
+            Timber.d("failed to create ServerVersions from strings: %s %s", version, minVersion);
+            return false;
+        }
+
+        int comparison = version.compareTo(minVersion);
+        Timber.d("current server version [%s] is %s [%s]", version, comparison >= 0 ? ">=" : "<", minVersion);
+        return comparison >= 0;
     }
 }
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -114,8 +114,6 @@ public class PlaybackController {
     private int playbackRetries = 0;
     private long lastPlaybackError = 0;
 
-    private boolean updateProgress = true;
-
     private Display.Mode[] mDisplayModes;
     private boolean refreshRateSwitchingEnabled;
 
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -1487,4 +1494,29 @@ private void renderSubtitles(@Nullable final String text) {
             binding.subtitlesText.setVisibility(View.VISIBLE);
         });
     }
+
+    public boolean getServerMeetsMinimumVersion(@NonNull String minServerVersion) {
```
nielsvanvelzen, line 1498:
```suggestion
    public boolean getServerMeetsMinimumVersion(@NonNull ServerVersion minServerVersion) {
```
Use an instance of server version instead of a string, the constructor is easy to use. Should look something like `new ServerVersion(10, 8, 0);` from Java.

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -1487,4 +1494,29 @@ private void renderSubtitles(@Nullable final String text) {
...
+        if (sessionRepository.getValue().getCurrentSession().getValue() == null) {
```
nielsvanvelzen, line 1499: Assign the session to a variable, the value might have changed after this check. (although unlikely)

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -1487,4 +1494,29 @@ private void renderSubtitles(@Nullable final String text) {
...
+        ServerVersion version = ServerVersion.Companion.fromString(server.getVersion());
+        ServerVersion minVersion = ServerVersion.Companion.fromString(minServerVersion);
+        if (version == null || minVersion == null) {
+            Timber.d("failed to create ServerVersions from strings: %s %s", version, minVersion);
+            return false;
+        }
+
+        int comparison = version.compareTo(minVersion);
+        Timber.d("current server version [%s] is %s [%s]", version, comparison >= 0 ? ">=" : "<", minVersion);
+        return comparison >= 0;
```
nielsvanvelzen, line 1520: In AuthenticationRepository, replace the current getServers function with this:

```kotlin
	private fun AuthenticationStoreServer.asServer(id: UUID) = Server(
		id = id,
		name = name,
		address = address,
		version = version,
		loginDisclaimer = loginDisclaimer,
		dateLastAccessed = Date(lastUsed),
	)

	override fun getServers() = authenticationStore.getServers()
		.map { (id, entry) -> entry.asServer(id) }
		.sortedWith(serverComparator)

	override fun getServer(id: UUID) = authenticationStore.getServer(id)?.asServer(id)
```

(also add `fun getServer(id: UUID): Server?` to the interface, after getServers)

Then, in the server data class, change serverVersion to public or add a function to compare with a ServerVersion instance.

Then get the server via the AuthenticationRepository, or maybe even add a getServer() to the ServerRepository (although that feels a bit redundant..). And use that for the comparison here

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
nielsvanvelzen:
```patch
diff --git a/app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt b/app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
index 61cfef7b3..cdfb9556b 100644
--- a/app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/auth/AuthenticationRepository.kt
@@ -31,6 +31,7 @@ import java.util.UUID
 
 interface AuthenticationRepository {
 	fun getServers(): List<Server>
+	fun getServer(id: UUID): Server?
 	fun getUsers(server: UUID): List<PrivateUser>?
 	fun getUser(server: UUID, user: UUID): PrivateUser?
 	fun saveServer(id: UUID, name: String, address: String, version: String?, loginDisclaimer: String?)
@@ -54,16 +55,20 @@ class AuthenticationRepositoryImpl(
 	private val serverComparator = compareByDescending<Server> { it.dateLastAccessed }.thenBy { it.name }
 	private val userComparator = compareByDescending<PrivateUser> { it.lastUsed }.thenBy { it.name }
 
-	override fun getServers() = authenticationStore.getServers().map { (id, info) ->
-		Server(
-			id = id,
-			name = info.name,
-			address = info.address,
-			version = info.version,
-			loginDisclaimer = info.loginDisclaimer,
-			dateLastAccessed = Date(info.lastUsed),
-		)
-	}.sortedWith(serverComparator)
+	private fun AuthenticationStoreServer.asServer(id: UUID) = Server(
+		id = id,
+		name = name,
+		address = address,
+		version = version,
+		loginDisclaimer = loginDisclaimer,
+		dateLastAccessed = Date(lastUsed),
+	)
+
+	override fun getServers() = authenticationStore.getServers()
+		.map { (id, entry) -> entry.asServer(id) }
+		.sortedWith(serverComparator)
+
+	override fun getServer(id: UUID) = authenticationStore.getServer(id)?.asServer(id)
 
 	private fun mapUser(
 		serverId: UUID,
```

mueslimak3r:
I added the changes suggested above.

I do still think it would be nice to move `getServerMeetsMinimumVersion` (now renamed) somewhere more appropriate

app/src/main/java/org/jellyfin/androidtv/auth/model/Server.kt
```diff
@@ -16,9 +16,16 @@ data class Server(
 	val loginDisclaimer: String? = null,
 	var dateLastAccessed: Date = Date(0),
 ) {
-	private val serverVersion = version?.let(ServerVersion::fromString)
+	val serverVersion = version?.let(ServerVersion::fromString)
```
nielsvanvelzen, line 19: Variable is not used outside of this class so it can be kept private

app/src/main/java/org/jellyfin/androidtv/auth/model/Server.kt
```diff
@@ -16,9 +16,16 @@ data class Server(
...
 	val versionSupported = serverVersion != null && serverVersion >= ServerRepository.minimumServerVersion
 
+
+	fun isVersionEqualOrGreater(otherVersion: ServerVersion?): Boolean {
```
nielsvanvelzen, line 23: I think we can make this an operator function [^1]. That way the Kotlin usages look a bit nicer.

[^1]: https://kotlinlang.org/docs/operator-overloading.html#comparison-operators

app/src/main/java/org/jellyfin/androidtv/auth/model/Server.kt
```diff
@@ -16,9 +16,16 @@ data class Server(
...
+		val serverVersion: ServerVersion? = version?.let { ServerVersion.fromString(it) }
```
nielsvanvelzen, line 24: Re-use serverVersion from the class, no need to parse it again

nielsvanvelzen: (so just remove this line)

app/src/main/java/org/jellyfin/androidtv/auth/model/Server.kt
nielsvanvelzen, line 23: Should be not-null

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -1487,4 +1494,25 @@ private void renderSubtitles(@Nullable final String text) {
...
+        Server server = authenticationRepository.getValue().getServer(session.getServerId());
+        if (server == null) {
+            Timber.d("couldn't compare server versions - server is null");
+            return false;
+        }
+
+        boolean result = server.isVersionEqualOrGreater(minServerVersion);
+        Timber.d("current server version [%s] is %s [%s.%s.%s]",
+                server.getVersion(), result ? ">=" : "<",
+                minServerVersion.getMajor(), minServerVersion.getMinor(), minServerVersion.getPatch());
```
nielsvanvelzen, line 1514: You should be able to just do `minServerVersion.toString()` to get the version as "10.8.0"

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/auth/model/Server.kt
mueslimak3r: like this?

Wouldn't a comparison overload be for comparing two objects of the same type?

`operator fun compareTo(other: ServerVersion?): Int  = serverVersion?.let { other?.compareTo(it) } ?: -1`

nielsvanvelzen (approved):

nielsvanvelzen:
I'm not the greatest fan of having getServerVersionEqualOrGreater in the CustomPlaybackOverlayFragment either. The next release (0.14) will require the server to be 10.8 or newer though, so we can remove it after 0.13 is released.

app/src/main/java/org/jellyfin/androidtv/auth/model/Server.kt
nielsvanvelzen: The compareTo function is not restricted to the same type, same for other operators actually.

PR #1507, opened 2022-03-09
Add ripple effect to DefaultCardView and fix the one in LegacyImageCardView

nielsvanvelzen:
**Changes**
- Fix ripple effect in LegacyImageCardView
  - (used pretty much everywhere in the app)
  - Previously, the effect would go over the text underneath the image. This looked off.
  - Now it only shows on top of the image part
- Add ripple effect to DefaultCardView
  - (used for users in the login screen)
  - Show ripple on top of the image part
- Remove some unused stuff from CustomListRowPresenter

**Issues**


nielsvanvelzen:
It's a bit annoying to capture the ripple as it's an animation and it looks quite bad in the emulator. But this is how it looked, you can see the white rectangle extending beneath the image:

image

mueslimak3r (approved):

PR #1508, opened 2022-03-09
crash fix from item being null when playback controller play() is called

mueslimak3r:
**Changes**
* guard against `item` (BaseItemDto) being null or `mItems` being null when play() is called

**Issues**
* tracepot issue `688e6a080001267b`

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -231,7 +231,7 @@ public class PlaybackController {
     }
 
     public boolean hasNextItem() {
-        return mCurrentIndex < mItems.size() - 1;
+        return mItems != null && mCurrentIndex < mItems.size() - 1;
     }
 
     public BaseItemDto getNextItem() {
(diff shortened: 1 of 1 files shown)


thornbill (approved):

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -421,8 +421,12 @@ private void play(long position, @Nullable Integer transcodedSubtitle) {
                 BaseItemDto item = getCurrentlyPlayingItem();
 
                 // make sure item isn't missing
-                if (item.getLocationType() == LocationType.Virtual && mFragment != null) {
-                    if (hasNextItem()) {
+                if (item == null || item.getLocationType() == LocationType.Virtual) {
+                    if (mFragment == null) {
+                        Timber.d("item is null!");
+                        if (TvApp.getApplication() != null)
+                            Utils.showToast(TvApp.getApplication(), TvApp.getApplication().getString(R.string.msg_cannot_play));
```
nielsvanvelzen, line 428: Don't use TvApp.getApplication....

mueslimak3r: If mFragment is null is there another way to get context?

If not I could just remove the toast

nielsvanvelzen: If mFragment is null that means that the playbackcontroller is already dead and shouldn't be called anyway.

mueslimak3r: I made a new PR with TvApp.getApplication removed

PR #1516, opened 2022-03-12
Add fastlane structure to repository

nielsvanvelzen:
Just like our mobile app, adding the fastlane structure to the repository helps us getting releases out on F-Droid and allow us to automatically publish to the Google Play store in the future.

The descriptions were taken from our current Play store listing and slightly modified. Some parts are copied from the mobile app. I've remade the icon/featureGraphic & tvBanner images. They now use the same assets generated from the vector XML's in this repo. The screenshots are freshly made using 0.13. Although we might want to add some more.

The goal with this PR is to have all stuff used for publishing in the repository. This will help us to get consistent listings on the different stores (Amazon Appstore, Google Play Store, F-Droid) with the ability for translators to localize the descriptions and us (the developers) to update the screenshots when necessary.

**Changes**
- Add fastlane structure

**Issues**


fastlane/metadata/android/en-US/title.txt
```diff
@@ -0,0 +1 @@
+Jellyfin for Android TV
```
nielsvanvelzen, line 1: We might want to rename the app to "Jellyfin for TV" to have a unified name between Fire TV and Android TV.

DavidFair: I feel like I'm [bike shedding](https://en.wikipedia.org/wiki/Law_of_triviality) a bit, but I think `Jellyfin TV` is a bit easier to say out loud there's a potential renaming to make it more generic

thornbill: Jellyfin TV sounds like a streaming service and Jellyfin for TV sounds rather broad since it doesn't cover **all** TVs... so I have no idea what the right answer is here. 🤷‍♂️

PR #1522, opened 2022-03-13
Add ripple effect to all buttons

nielsvanvelzen:
Follow up for #1507. This PR adds a ripple effect to all buttons[^1] in the app.

[^1]: except for live tv if there are any custom buttons

**Changes**
- Add ripple effect to:
  - NowPlayingBug
  - ServerButtonView
  - ItemRow (track list in audio album view)
  - TextUnderButton (the buttons on details page)
  - Button.Default (buttons used in the login screen)
  - Button.Icon (buttons used in various toolbars)
  - AlphaPicker buttons
- Always use `?attr/buttonRounding` for roundness of buttons
  - Fixes inconsistency between various button types
  - Set `?attr/buttonRounding` to `3dp` for the default theme
  - The Muted Purple theme still uses `7dp` roundness
- Always send theme to ResourcesCompat.getDrawable calls
- Reformat some changed XML files
- Add `?attr/buttonDefaultHighlightBackground` to change the button background color on highlight
  - Muted Purple now uses a purple highlight background for all buttons (was only TextUnderButton before)

**Issues**


mueslimak3r (changes requested):
This changes all the buttons to light grey with an even lighter grey highlight. I assume this isn't intentional since it isn't mentioned in the changes.

The Muted Purple theme appears to display colors as expected. All the other themes no longer display their colors (and are grey on grey as shown below)

nielsvanvelzen:
> This changes all the buttons to light grey with an even lighter grey highlight. I assume this isn't intentional since it isn't mentioned in the changes.
> 
> The Muted Purple theme appears to display colors as expected. All the other themes no longer display their colors (and are grey on grey as shown below)

This was partially intended, the buttons now use the styling from Button.Default which has slightly different colors. Some of the buttons didn't use the proper text color though. I've fixed that now.

mueslimak3r:
It looks like only the emerald green theme changed the card background color to match the theme.

I don't think it (or any themes) should do this since it makes the progress bar almost invisible when the progress bar also uses the theme's highlight color.

nielsvanvelzen:
The card color was not changed in this PR.

mueslimak3r (approved):

PR #1537, opened 2022-03-20
Remove srt as external from ExoPlayer subtitle profiles

### LINKED ISSUE #1534: Embedded subtitles fail to load
### Describe the bug

I have noticed that all of my embedded subtitles fail to load. When I select them, first I get "Loading subtitles" message and after a while the message disappears and a popup appears: "Failed to load subtitles". External subtitles are working.

I believe this is an error on jellyfin androidtv part, since other types of clients can display the very same subtitles OK. Also, it doesn't matter whether the media is transcoded or direct played, the issue occurs in both cases.

### Logs

_No response_

### Application version

0.13.0

Google Play

### Device information

Sony Bravia KD-55XF9005

### Android version

9

### Jellyfin server version

10.7.7

nielsvanvelzen:
We had 2 separate entries for srt: external and embed. I guess the server just used the first entry (external). It doesn't really do much because we don't support embedded subs in ExoPlayer anyway.

**Changes**
- Remove srt as external from ExoPlayer subtitle profiles
- Remove srt as external from libVLC subtitle profiles

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
@@ -217,7 +217,6 @@ class ExoPlayerProfile(
 		}.toTypedArray()
 
 		subtitleProfiles = arrayOf(
-			subtitleProfile("srt", SubtitleDeliveryMethod.External),
 			subtitleProfile("srt", SubtitleDeliveryMethod.Embed),
 			subtitleProfile("subrip", SubtitleDeliveryMethod.Embed),
 			subtitleProfile("ass", SubtitleDeliveryMethod.Encode),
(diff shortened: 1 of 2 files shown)


mueslimak3r:
this breaks direct playing external srt subtitles and forces them to be burned in. Tested with external .srt and a video that doesn't have any embedded subtitles.

I've personally not had issues playing embedded subtitles. But my guess is that the "fall through to external handling" logic here may be responsible for issues.

https://github.com/jellyfin/jellyfin-androidtv/blob/7f77814e7584a5dc51cd1ff93bb9eca19cb42c7d/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java#L953

nielsvanvelzen:
So that means that the server actually allows for multiple delivery methods for a subtitle format.. Didn't know that.

PR #1546, opened 2022-03-21
Make cancel default on delete/clear

DavidFair:
**Changes**
Select cancel by default on confirmation dialog
    
Select cancel so a user doesn't accidentally delete media by hitting
delete twice. Since we ask it's worth adding this extra step, if we need
to remove the extra step it's arguable we can completely remove alert
dialog all-together

**Testing**
Ensure that the user account can delete media
- Go to a random video
- Attempt to delete
- Ensure cancel is selected by default


mueslimak3r (approved):

nielsvanvelzen (approved):

nielsvanvelzen:
Some while ago we discussed the delete function and I believe we decided the app shouldn't have a delete button for items. So we'll probably remove it at some point.

PR #1547, opened 2022-03-21
Add handling for updating end time based on playback speed

### LINKED ISSUE #1509: Ends at time is not recalculated when playing the video at faster or slower speed
### Describe the bug

Ends at time is not recalculated when playing the video at faster or slower speed. 

### Logs

_No response_

### Application version

0.13 Beta 5

Sideloaded APK

### Device information

Nvidia Shield  2015

### Android version

Android 11

### Jellyfin server version

10.8 latest nightly

DavidFair:
**Changes**
- Add handling for updating end time based on playback speed. This is a bit hacky due to the current architecture:
```
    Exoplayer is built on the assumption that we
    have bi-directional communication to our UI layer. Instead the
    CustomPlaybackTransportControlGlue is uni-directional polling the video
    manager.
    
    We unfortunately have to use hacky poll tricks to work-around the fact
    we can't wire up on the "onPlaybackParametersChanged" call to notify the
    glue class to update its value. However, in the short-term a 5s "delay" between setting 
    and updating is more favourable than setting up reverse listening support.
```

**Issues**
Fixes #1509 

**Testing**
- Play a video, check the finish time
- Change the speed to 0.25 and/or 2x
- Check the finish time updates accordingly (after 5s)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -88,7 +88,7 @@ public class PlaybackController {
     private int mDefaultSubIndex = -1;
     private int mDefaultAudioIndex = -1;
     private boolean burningSubs = false;
-    private double mRequestedPlaybackSpeed = -1.0;
+    private float mRequestedPlaybackSpeed = -1.0f;
 
     private PlayMethod mPlaybackMethod = PlayMethod.Transcode;
 
@@ -163,7 +163,13 @@ public class PlaybackController {
         mPlaybackMethod = value;
     }
 
-    public void setPlaybackSpeed(@NonNull Double speed) {
+    public float getPlaybackSpeed() {
+        // Actually poll the video manager, since exoplayer can revert back
+        // to 1x if it can't go faster, so we should check directly
+        return mVideoManager.getPlaybackSpeed();
+    }
+
+    public void setPlaybackSpeed(@NonNull float speed) {
         mRequestedPlaybackSpeed = speed;
         if (hasInitializedVideoManager()) {
             mVideoManager.setPlaybackSpeed(speed);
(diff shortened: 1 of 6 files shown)


mueslimak3r:
We had issues (and reverted) last time the switch from EventListener -> Listener was made. It looks like your approach is a little different (skips handling `playWhenReady`/ the event for it) so hopefully it does the trick!

DavidFair:
I was struggling to get my head around the new `playWhenReady` and how it interacts when paused...etc. when I saw the docs for `onIsPlaying`. I'm guessing we're not the only project who had problems switching to the more complex handler for playback start and paused. 

I'm hoping since we're early in 0.14 these changes can bake and any problems flush out too.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -163,6 +163,12 @@ public void setPlaybackMethod(@NonNull PlayMethod value) {
         mPlaybackMethod = value;
     }
 
+    public double getPlaybackSpeed() {
```
nielsvanvelzen, line 166: The setter should use `double` or the getter should use `Double`.

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -642,6 +649,16 @@ public int setVLCAudioTrack(int ndx) {
         return ndx;
     }
 
+    public double getPlaybackSpeed(){
```
nielsvanvelzen, line 652: The getter should use `Double` or the setter should use `double`.

nielsvanvelzen:
> I'm hoping since we're early in 0.14 these changes can bake and any problems flush out too.

If possible we'd like to include this change to 0.13.1, so keeping the listener in a separate PR would help with that. For 0.14 development you're right, we have quite some time to fix issues that may be caused by the switch.

nielsvanvelzen (approved):

PR #1556, opened 2022-03-23
only parse exoplayer TracksInfo when needed - instead of once every 0.5-2 seconds

mueslimak3r:
**Changes**
* store the current exoplayer audio track index/id and reuse it instead of parsing `TracksInfo` every time `getExoPlayerTrack()` is called

**Issues**
* every time `reportProgress()` runs it gets the current audio index (calls `getExoPlayerTrack()`), which currently parses `TracksInfo`. I'm not sure how expensive this is but some users have reported worse performance in 0.13 specifically with exoplayer. This may be contributing to that.


mueslimak3r:
I'm not sure exactly how to read this but it does look like it was expensive prior to this change

before:
Screen Shot 2022-03-22 at 11 23 09 PM

after:
Screen Shot 2022-03-22 at 11 29 30 PM

nielsvanvelzen (approved):

PR #1568, opened 2022-03-26
fixed playFrom() crash, support queueing entire row (eg: last played)

mueslimak3r:
**Changes**
* fixed crash when calling playFrom() where the player was unintentionally released
* clicking on an audio item will now:
> * open NowPlayingActivity if it is the currently playing item
> * use `playFrom()` to skip within the queue (and not affect things like shufffle, etc) if the item belongs to the queue
> * otherwise, start a new session by adding the items in the row to the queue and playing from the selected item's index. This means we can easily play all "last played" songs

* `playNow()` will loop through its input list and create a new list of only audio items. This protects against accidentally adding video items if an entire row is added to the queue. I don't know of any rows that have mixed A/V items but better safe than sorry, and the methods for adding to the video queue already do this.

* changed deprecated `mExoPlayer.stop(boolean)` -> `mExoPlayer.stop()`

**Issues**
* the audio player was being released unintentionally
* playFrom() in the context menu was missing a return statement, which caused some weird behavior

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemLauncher.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemLauncher.java
@@ -140,14 +140,16 @@ public class ItemLauncher {
                         Timber.d("got pos %s", pos);
                         if (rowItem.getBaseItem() == null)
                             return;
+                        MediaManager mediaManager = KoinJavaComponent.<MediaManager>get(MediaManager.class);
+
                         // if the song currently playing is selected (and is the exact item - this only happens in the nowPlayingRow), open AudioNowPlayingActivity
-                        if (KoinJavaComponent.<MediaManager>get(MediaManager.class).hasAudioQueueItems() && rowItem.getBaseItem() == KoinJavaComponent.<MediaManager>get(MediaManager.class).getCurrentAudioItem()) {
+                        if (mediaManager.hasAudioQueueItems() && rowItem.getBaseItem() == mediaManager.getCurrentAudioItem()) {
                             // otherwise, open AudioNowPlayingActivity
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemLauncher.java
```diff
@@ -139,17 +138,28 @@ public static void launch(final BaseRowItem rowItem, ItemRowAdapter adapter, int
...
-                            KoinJavaComponent.<MediaManager>get(MediaManager.class).playFrom(rowItem.getIndex());
-                        } else {
+                        if (rowItem.getBaseItem() == null)
+                            return;
+                        // if the song currently playing is selected (and is the exact item - this only happens in the nowPlayingRow), open AudioNowPlayingActivity
+                        if (KoinJavaComponent.<MediaManager>get(MediaManager.class).hasAudioQueueItems() && rowItem.getBaseItem() == KoinJavaComponent.<MediaManager>get(MediaManager.class).getCurrentAudioItem()) {
                             // otherwise, open AudioNowPlayingActivity
                             Intent nowPlaying = new Intent(activity, AudioNowPlayingActivity.class);
                             activity.startActivity(nowPlaying);
+                        } else if (KoinJavaComponent.<MediaManager>get(MediaManager.class).hasAudioQueueItems() && rowItem instanceof AudioQueueItem && pos < KoinJavaComponent.<MediaManager>get(MediaManager.class).getCurrentAudioQueueSize()) {
```
nielsvanvelzen, line 148: Probably more efficient (and readable) to assign the mediamanager to a variable here instead of writing `KoinJavaComponent.<MediaManager>get(MediaManager.class)` multiple times.

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

PR #1574, opened 2022-03-27
Refactor play logic in PlaybackController

DavidFair:
**Changes**
- Refactors the video option building from `PlaybackController:play` into private methods. This first step was done as-is so no modifications were made to the actual methods
- Removes the three call sites to `Utils.getMaxBitrate` to a single location

This is a precursor to #1569 and should allow us to:
- Make sense of the diff in the next PR
- Allow us to easily refactor away deprecated `VideoOptions` when the replacement
- Makes the code more readable

**Testing**
- Try to play a video with libVLC and ExoPlayer

PR #1575, opened 2022-03-27
Refactor existing callbacks into interface

DavidFair:
Builds on #1574 

**Changes**
  Many call-backs already exist between the VideoManager and
  PlaybackController. But they are created on an ad-hoc basis in various
  methods.
  
  By extracting these out to a method it means we can:
  - Remove the boilerplate of @Override methods in PlaybackController
  - Easily refactor these into a seperate class long-term
  - Add new handlers
  - Pass parameters around in the future
  - Not have everything dispatch to a simple onEvent.
  
  This also makes extending the interface easier where we have exoplayer
  only notifications (such as onPlaybackSpeedChanged). Since the VLC code
  can simply call the notifier without going through a PlaybackListener

--

I'm not super happy with `PlaybackControllerNotifiable` so I'm open to any/all suggestions on this front, maybe something along the lines of `VideoManagerSignals` (coming from a Qt background)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -143,7 +143,7 @@ public class PlaybackController implements PlaybackControllerNotifiable {
         return mFragment;
     }
 
-    public void init(VideoManager mgr, CustomPlaybackOverlayFragment fragment) {
+    public void init(@NonNull VideoManager mgr, @NonNull CustomPlaybackOverlayFragment fragment) {
         mVideoManager = mgr;
         mVideoManager.subscribe(this);
         mFragment = fragment;
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -57,7 +57,7 @@ import java.util.List;
 import kotlin.Lazy;
 import timber.log.Timber;
 
-public class PlaybackController {
+public class PlaybackController implements PlaybackControllerNotifiable {
     // Frequency to report playback progress
     private final static long PROGRESS_REPORTING_INTERVAL = TimeUtils.secondsToMillis(3);
     // Frequency to report paused state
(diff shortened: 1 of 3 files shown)


DavidFair:
I've taken the copy from `master` and extract it into the method as-is, hopefully that unblocks this PR

mueslimak3r:
> I've taken the copy from `master` and extract it into the method as-is, hopefully that unblocks this PR

I'm not sure what you mean, are you talking about the latest commit (dfa9f8b)? Blocked just means "this needs another PR first", which in this case is #1574. Both because you explicitly say so and because this PR shares changes with it. The label just helps us keep track of it so they get reviewed and merged in the intended order. So this will get unblocked when the other PR is merged

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -161,6 +166,11 @@ public void onTracksInfoChanged(TracksInfo tracksInfo) {
         });
     }
 
+    public void subscribe(PlaybackControllerNotifiable notifier){
```
mueslimak3r, line 169: Unknown nullability; explicitly declare as @Nullable or @NonNull to improve Kotlin interoperability

```suggestion
    public void subscribe(@NonNull PlaybackControllerNotifiable notifier) {
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -146,9 +146,9 @@ public CustomPlaybackOverlayFragment getFragment() {
 
     public void init(VideoManager mgr, CustomPlaybackOverlayFragment fragment) {
```
mueslimak3r, line 147: You could go ahead an annotate these too. I think these would both be @NonNull

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -58,7 +58,7 @@
 import kotlin.Lazy;
 import timber.log.Timber;
 
-public class PlaybackController {
+public class PlaybackController implements PlaybackControllerNotifiable {
```
mueslimak3r, line 61: > I'm not super happy with PlaybackControllerNotifiable

You could use the same nomenclature as SessionRepository. Interface: `SessionRepository`, Class: `SessionRepositoryImpl`

mueslimak3r (commented):
I'll try to do a functional test of this later today

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
DavidFair: In this case the `Impl` class provides the implementation whereas our interface is an observer, so what do we call the equivalent of `SessionRepository` in your example?

`PlaybackControllerRepository` doesn't really fit as its not a MVVM component :thinking:

mueslimak3r (approved):
I tested movies/series playback, live tv playback, and checked for leaks. Looks good!

PR #1597, opened 2022-04-02
Remove hotdog theme

nielsvanvelzen:
April fools is over so the theme can go now. With the recent UI changes it wasn't really working properly in 0.14 anyway.

**Changes**

- Remove hotdog theme

**Issues**


mueslimak3r (approved):
If a new hot dog theme is added in the future I think it should be more than just red. I was hoping for some yellow and was disappointed 😂

nielsvanvelzen:
It did use yellow at first but that wasn't visible at all in 0.14 unfortunately.

PR #1605, opened 2022-04-03
Require context in Utils.downMixAudio()

nielsvanvelzen:
**Changes**
- Require context in Utils.downMixAudio()
  - I tried to use a proper context where possible but some places use TvApp.getApplication right now because ItemLauncher/MediaManager have no context.

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
@@ -198,7 +198,7 @@ public class MediaManager {
             return false;
         }
 
-        return createPlayer(600);
+        return createPlayer(TvApp.getApplication(), 600);
     }
 
     private boolean isPaused() {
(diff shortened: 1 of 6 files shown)


mueslimak3r (approved):
Looks good.

When testing with downmix to stereo enabled I do see some weird behavior that isn't related to this PR (I checked stable and it's there too). With it enabled the video is transcoded due to "audio not supported" as expected, however ffmpeg runs with `-ac 6` (audio channels = 6) as if the setting is just removing codec support but not limiting the number of channels. I tested this with exoplayer.

nielsvanvelzen:
> When testing with downmix to stereo enabled I do see some weird behavior that isn't related to this PR (I checked stable and it's there too). With it enabled the video is transcoded due to "audio not supported" as expected, however ffmpeg runs with `-ac 6` (audio channels = 6) as if the setting is just removing codec support but not limiting the number of channels. I tested this with exoplayer.

Probably caused by this line, I'm not that familiar with the profile system though;

https://github.com/jellyfin/jellyfin-androidtv/blob/d7ee86ebb7c051fe84c4d3c9dfa9694d7e7b6517/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt#L197

PR #1609, opened 2022-04-04
Use context in MediaManager

nielsvanvelzen:
**Changes**
- Send context to MediaManager constructor
- Always use context in MediaManager (replaced TvApp.getApplication)
- Change some function signatures to use Context instead of Activity

This is basically the same behavior as before, as the context injected now is the application context.

**Issues**


nielsvanvelzen:
Replaced with #1610

PR #1617, opened 2022-04-11
Bump to Exoplayer 1.27.1

DavidFair:
This Exoplayer release has been out for 4 weeks (at the time of this commit) so it
reasonably well baked.

Some highlights that might help us from their release notes:

-----------
( https://github.com/google/ExoPlayer/releases/tag/r2.17.0 )

- Update video and audio track selection logic to only choose formats for adaptive selections that have the same level of decoder and hardware support
- Update video track selection logic to prefer more efficient codecs if multiple codecs are supported by primary, hardware-accelerated decoders
- Prefer audio content preferences (for example, the "default" audio track or a track matching the system locale language) over technical track selection constraints (for example, preferred MIME type, or maximum channel count).
- Fix decoder fallback logic for Dolby Vision to use a compatible H264/H265 decoder if needed.
- Fix decoder fallback logic for Dolby Atmos (E-AC3-JOC) to use a compatible E-AC3 decoder if needed.
- (HLS) Use chunkless preparation by default to improve start up time.


nielsvanvelzen:
[jellyfin-exoplayer-ffmpegextension](https://github.com/jellyfin/jellyfin-exoplayer-ffmpeg-extension) needs to be bumped too, it needs to be in sync with the exoplayer version.

nielsvanvelzen (approved):

PR #1620, opened 2022-04-12
Refactor preference code

nielsvanvelzen:
Some changes to make it easier to work with and maintain the preference code. Blocking function calls removed from the module and replaced with a utility function in the app (only needed for Java compat right now).

**Changes**

- Allow migrations in all store types
- Remove blocking functions from the preference module
- Move Preference.xx functions to global functions

**Issues**

Depends on ##1619

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/CoroutineUtils.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/CoroutineUtils.kt
@@ -1,0 +1,8 @@
+@file:JvmName("CoroutineUtils")
+package org.jellyfin.androidtv.util
+
+import kotlinx.coroutines.CoroutineScope
+
+fun <T: Any> runBlocking(block: suspend CoroutineScope.() -> T) = kotlinx.coroutines.runBlocking {
+	block()
+}
--- a/preference/src/main/kotlin/store/PreferenceStore.kt
+++ b/preference/src/main/kotlin/store/PreferenceStore.kt
@@ -1,6 +1,7 @@
 package org.jellyfin.preference.store
 
 import org.jellyfin.preference.Preference
+import org.jellyfin.preference.migration.MigrationContext
 
 /**
  * Abstract class defining the required functions for a preference store.
@@ -8,7 +9,8 @@ import org.jellyfin.preference.Preference
  * This implements shared functionality (such as Enum handling), whilst allowing for
  * different backing stores.
  */
-abstract class PreferenceStore {
+@Suppress("TooManyFunctions")
+abstract class PreferenceStore<ME, MV> {
 	// val value = store[Preference.x]
 	@Suppress("UNCHECKED_CAST")
 	operator fun <T : Any> get(preference: Preference<T>): T =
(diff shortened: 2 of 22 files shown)


app/src/main/java/org/jellyfin/androidtv/util/CoroutineUtils.kt
```diff
@@ -0,0 +1,8 @@
+@file:JvmName("CoroutineUtils")
+package org.jellyfin.androidtv.util
+
+import kotlinx.coroutines.CoroutineScope
+
+fun <T> runBlocking(block: suspend CoroutineScope.() -> T) = kotlinx.coroutines.runBlocking {
```
mueslimak3r, line 6: the linter says this needs to be annotated as null or non-null

preference/src/main/kotlin/store/PreferenceStore.kt
```diff
@@ -1,14 +1,16 @@
...
 
 /**
  * Abstract class defining the required functions for a preference store.
  *
  * This implements shared functionality (such as Enum handling), whilst allowing for
  * different backing stores.
  */
-abstract class PreferenceStore {
+@Suppress("TooManyFunctions")
+abstract class PreferenceStore<ME, MV> {
```
mueslimak3r, line 13: When <ME, MV> is used in this PR, it represents `Migration<E, V>`, right?

```

data class Migration<E, V>(
		val toVersion: Int,
		val body: E.(V) -> Unit
	)

```

I'm not sure what this is doing, specifically the `E.(V) -> Unit` part. What would I google to read up on it?
I'm already reading about data classes

mueslimak3r (approved):

preference/src/main/kotlin/store/PreferenceStore.kt
nielsvanvelzen: > When <ME, MV> is used in this PR, it represents Migration<E, V>, right?

Yes, the "M" prefix is used to indicate it's for the migration part.

The "ME" means the editor type to use and the "MV" means the current data to send to the editor. You can look at the SharedPreferencesStore to see how it's used:

`PreferenceStore<SharedPreferences.Editor, SharedPreferences>()`

Which then causes the "runMigrations" function to use SharedPreferences.Editor as this scope and SharedPreferences as "it". (See usage in UserPreferences).

> I'm not sure what this is doing, specifically the E.(V) -> Unit part.

`() -> Unit` would mean: a function with no parameters that returns nothing (Unit)
`(V) -> Unit` would mean that is contains one parameter, of type V
`E.(V) -> Unit` would mean that the function is in the (this) scope of type E

app/src/main/java/org/jellyfin/androidtv/util/CoroutineUtils.kt
nielsvanvelzen: Fixed, had to use `<T: Any>` so it can never be null.

preference/src/main/kotlin/store/PreferenceStore.kt
mueslimak3r: That makes sense now, thanks for the explanation 🙂

PR #1622, opened 2022-04-14
Refactor authentication code

### LINKED ISSUE #1621: removing the client from the list of 'devices' authenticated with the server causes the app to always crash on launch
### Describe the bug

**Steps to reproduce**
2) install the latest build off master (commit: 0d12a149f84fc31148d39a04f665421850b389c8)
1) sign into a jellyfin server and keep the app open
2) in the web client find the android tv client in the dashboard under 'devices' and delete it
3) back out of the android tv app and then click on it again
4) the app will crash

**Context for the logs**
1) the app is force stopped
2) start logging
3) open the app and let it crash back to the launcher
4) stop logging

### Logs

```shell
client logs:
https://pastebin.com/dAJACb15
```

### Application version

dev build off master

Sideloaded APK

### Device information

Chromecast w/ Google TV

### Android version

10

### Jellyfin server version

10.8 beta 1

nielsvanvelzen:
This PR refactors the authentication code to make it more maintainable and easier to work with. It also drops the "system user" feature, which now makes it easier to use the SDK.

**Changes**
- Refactor authentication
- Rethink all repositories in the auth package
- Fix various issues (like #1621)
- Prepare code for Quick Connect implementation
- Update coroutine usages to be slightly more efficient

The AndroidModule is partially borrowed from #1309.

**Issues**

Fix #1621 

- There is one new notable issue: when using autologin the progress screen isn't shown properly. You'll see the server/user selection screen for a short duration. A fix for this will come later.


mueslimak3r (commented):
This does fix the infinite crashing after de-authing the client but it still crashes once before recovering.

context for the logs:
1) opened app then de-authed client
2) backed out of the app
3) opened the app and it crashes back to the launcher
4) opened the app again and this time it goes to the login page

logs:
https://pastebin.com/mH9TmS0Z

nielsvanvelzen:
When cold starting the app it will always open the StartupActivity, which will then deal with authentication stuff before opening the next activity (most of the time this is the MainActivity).

After that, going to the launcher and back into the app won't trigger a cold start, but instead resumes the previous activity, so we're not refreshing the session at this point. The API calls that happen in the activity end up crashing the app.

To fix the we'd need to properly handle errors for API calls, which will slowly happen while moving stuff from the old apiclient to the new SDK, with a proper repository pattern. I have some ideas on how to patch it right now, I could work on those and include it in this PR or create a separate PR (might be easier since this one is pretty big already?).

mueslimak3r (approved):

PR #1681, opened 2022-05-02
Redesign AddServerFragment

nielsvanvelzen:
Use the same styling as the new login fragment for the add server fragment.

**Changes**
- Use new input style for server address input
- Use left-aligned layout
- Add ViewModel for server add thingy
- Remove KeyboardFocusChangeListener - no longer used
- Fix some formatting in the login fragment
- Fix some bad `+id/` in the login xml

**Before**
image

**After**
image

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/startup/fragment/ServerAddFragment.kt
```diff
@@ -0,0 +1,110 @@
...
+					}
+					is UnableToConnectState -> {
+						// Enable form
+						binding.address.isEnabled = true
+						binding.confirm.isEnabled = true
+						// Update state text
+						binding.error.text = getString(
+							R.string.server_connection_failed_candidates,
+							state.addressCandidates
+								.map { "${it.key} ${it.value.getSummary(requireContext())}" }
```
mueslimak3r, line 85: I think this is the place where this is set. A hyphen or some other separator might be nice to have between the address and the reason for failure.
eg:
> Unable to connect to server, tried the following addresses:
https://demo.jellyfin.org/stable Unable to connect

becomes

> Unable to connect to server, tried the following addresses:
https://demo.jellyfin.org/stable - Unable to connect

```suggestion
								.map { "${it.key} - ${it.value.getSummary(requireContext())}" }
```

nielsvanvelzen:
image

mueslimak3r (approved):

PR #1741, opened 2022-05-28
Propagate errors from ItemRowAdapter to RetrieveFinishedListener

nielsvanvelzen:
Previously the app would occasionally show volley clienterror toasts that provide 0 feedback to the end-user about the issue. Now the error is propagated to the RetrieveFinishedListener (in StdGridGragment and SearchRunnable) which both casually ignore the error.

**Changes**
- Remove stupid volley clienterror toast
- Send exceptions to RetrieveFinishedListener where they can potentially be better handled

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
@@ -505,7 +505,7 @@ public class ItemRowAdapter extends ArrayObjectAdapter {
     }
 
     private void retrieveNext() {
-        if(fullyLoaded || isCurrentlyRetrieving()) {
+        if (fullyLoaded || isCurrentlyRetrieving()) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
```diff
@@ -446,7 +446,8 @@ public void setPosition(int pos) {
         }
     }
 
-    public @Nullable String getStartLetter() {
+    public @Nullable
```
mueslimak3r, line 449: Do you want this style (`public @Nullable` on its own line) used for all new code when a return type has an annotation?

mueslimak3r (approved):

app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
nielsvanvelzen: The build-in formatter did that, it wasn't intended oops

PR #1760, opened 2022-06-07
Home notifications

nielsvanvelzen:
This one was on my list for a while now. Right now it implements a single message but some others I'm considering are:

- Outdated server ("A newer version of Jellyfin is available: Jellyfin 10.8.2")
- Incompatible server ("The next version of the app requires Jellyfin 10.8.0 or newer, you are still using 10.7.0.")
- Offline (#1309)

**Changes**
- Show "notifications" on the home screen
  - Managed by NotificationRepository
- Add notification about the UI mode of the system (detect TV devices, show incompatibility warning)
- Show notification in a home section
  - Autohides when empty
  - Click on a notification to hide it (until you restart the app)

**Screenshots**
image

**Issues**


thornbill:
I think this is a great idea! Should we maybe make it the full width of the screen?

nielsvanvelzen:
There can be multiple notifications, which would then show next to each other. Making it full width would look weird on a TV device. Open for suggestions though.

thornbill:
Maybe use a bulleted list in a single full-width box for multiple? I don't know really... just trying to think of some conventions I have seen used before. That wouldn't really work if they are individually dismissable though.

I'm sure it would be fine as is too. :sweat_smile:

PR #1767, opened 2022-06-09
Fix EAC3 transcoding in libvlc on 10.8.0-beta3

### LINKED ISSUE #1723: Investigate transcoding EAC-3 in 10.8.0
### Describe the bug

There have been a couple reports on Reddit that EAC-3 is being transcoded (or not playing at all?) when it should direct play when using a 0.13.x release of the app and a 10.8.0 alpha/beta server release.

**References:**
* https://www.reddit.com/r/jellyfin/comments/sbiovm/android_tv_app_and_shield_transcoding_eac3_audio/
* https://www.reddit.com/r/jellyfin/comments/urj1fa/regression_all_video_files_with_eac3_fail_to/

### Logs

_No response_

### FFmpeg logs

_No response_

### Media info of the file

media info

### Application version

N/A, 0.13.5

### Device information

Nvidia Shield Pro, Sony Bravia TV

### Android version

N/A, Android 9

### Jellyfin server version

N/A, 10.8.0 beta 3

Fallenbagel:
Fixes EAC3 transcoding in libvlc of 10.8.0-beta3

**Changes**
Previously, it would transcode when EAC3 files are played (especially audio channels ≥ 6), and if direct play is forced, they do not play at all with the error `Cannot play this video format`. Therefore, I added the codec EAC3 as an audio profile to the libVLC profile. This has fixed the issue where when EAC3 files are played using libvlc on Jellyfin, it resorts to transcoding on Jellyfin Server v10.8.0-beta3.

**Issues**
Fixes #1723

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/LibVlcProfile.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/LibVlcProfile.kt
@@ -59,6 +59,7 @@ class LibVlcProfile(
 					Codec.Audio.MP3,
 					Codec.Audio.MP2,
 					Codec.Audio.AC3,
+					Codec.Audio.EAC3,
 					Codec.Audio.WMA,
 					Codec.Audio.WMAV2,
 					Codec.Audio.DCA,


mueslimak3r (approved):

bentrop:
> Fixes EAC3 transcoding in libvlc of 10.8.0-beta3

Without this change both LibVLC **and ExoPlayer** refuse to direct play EAC3 audio.

I do not understand why adding this codec to [LibVlcProfile.kt](https://github.com/jellyfin/jellyfin-androidtv/pull/1767/files/596dfedba25c94387a90fabc83a4756f88a43f34#diff-5c388e6836a5fd404b57c319059faf568de520e5572da1c40ce8c9dc9cffc8d0) _also_ fixes ExoPlayer?

It shouldn't, right? But it does.

While this fixes playback, I feel like there is a deeper issue here that remains unresolved.

Also: why is this profile only an issue with 10.8.0 but both LibVLC and ExoPlayer direct play EAC3 just fine from 10.7.7 even without this change?

PR #1772, opened 2022-06-12
Add external trailer support

nielsvanvelzen:
From my observations this doesn't completely work in the emulator because the YouTube app crashes for some reason. It worked fine on my shield. The YouTube app also "traps" the user by going back to it's own home screen instead of our app when pressing back, so the UX is pretty terrible. Not much to do about that unfortunately. I expect third-party YouTube apps to behave better but haven't tested those.

The "trailer" row is not shown for external trailers, it's only the button that shows up for external ones. The code tries to play any of the remote trailers before falling back to local trailers. If nothing works it will show a toast that playback failed.

**Changes**
- Add ability to open external trailers

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
@@ -4,6 +4,7 @@ import static org.koin.java.KoinJavaComponent.inject;
 
 import android.app.Activity;
 import android.app.AlertDialog;
+import android.content.ActivityNotFoundException;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.graphics.Point;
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
**This PR needs to be tested on a Fire TV device before merging.**

nielsvanvelzen:
I want to make this a bit smarter and hide the trailer button if we can't open an external app (devices without YouTube installed for example). Marking it as draft for now.

nielsvanvelzen:
Some additional changes pushed:
- Check if the external player can be played before showing the trailers button
  - Now this button only shows when we know we can play a video
  - Ignore Android TV stub package (which doesn't actually open the video but shows an "no app found" toast)

morkyy:
Why not use exoplayer to stream the youtube trailer from within the Jellyfin app? That would make this feature more seamless and resolve the issue where the user is locked in the youtube app.

Just a suggestion for a feature improvement in the future. Unsure how complicated that would be.

nielsvanvelzen:
We can't play a YouTube video in-app. That's why we need to attempt to open the YouTube app or webbrowser.

This is also stated on the ExoPlayer website: https://exoplayer.dev/troubleshooting.html#can-i-play-youtube-videos-directly-with-exoplayer. The YouTube Android Player API is proprietary so we won't be adding that to the app either. If you want a good experience playing trailers you should have them locally.

chr-zz:
The trailer button appears on the current beta releases.
on Amazon Stick devices, viewing external trailers works
but not returning. youtube must be closed, jellyfin restarted, and go back to the film.

tested on 0.14.beta3  Amazon Stick 2Gen

nielsvanvelzen:
From the draft of the 0.14 blog post:

> Unfortunately opening external apps does not work great on some television platforms and you might see an error on your device, telling you there is no app to open the trailer. If this happens you should check if the YouTube/Vimeo/other app is installed.
> For a better experience we recommend using local trailers.

We can't do much about this behavior, it's just the crappy way the TV version of YouTube works.

PR #1775, opened 2022-06-12
Add showing additional parts

dspencer001:
**Changes**
Adds showing Additional parts above Cast / Crew for Movies. Includes what translations I could get from jellyfin-web. I have very limited Android app experience, so this is mostly copied from how the specials and suggestions are retrieved and displayed. Tested with Lord of the Rings extended editions on an Nvidia Shield TV Pro.


app/src/main/java/org/jellyfin/androidtv/data/querying/AdditionalPartsQuery.java
```diff
@@ -0,0 +1,13 @@
+package org.jellyfin.androidtv.data.querying;
```
nielsvanvelzen, line 1: All new files should be written in Kotlin

app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
```diff
@@ -1316,6 +1329,41 @@ public void onError(Exception exception) {
 
     }
 
+    private void retrieve(final AdditionalPartsQuery query) {
+        final ItemRowAdapter adapter = this;
+        apiClient.getValue().GetAdditionalParts(query.getItemId(), KoinJavaComponent.<UserRepository>get(UserRepository.class).getCurrentUser().getValue().getId().toString(), new Response<ItemsResult>() {
```
nielsvanvelzen, line 1334: All new API related code should use the SDK instead of the apiclient

app/src/main/res/values-ar/strings.xml
```diff
@@ -96,6 +96,7 @@
     <string name="lbl_specials">مقاطع خاصة</string>
     <string name="lbl_chapters">الأجزاء</string>
     <string name="lbl_cast_crew">الممثلون</string>
+    <string name="lbl_additional_parts">أدوار إضافية</string>
```
nielsvanvelzen, line 99: Only english strings (values/strings.xml) should be added in git. Everything else should be done in Weblate (translate.jellyfin.org).

(Make sure this doesn't stay in the git history to prevent breakage)

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/data/querying/AdditionalPartsQuery.java
dspencer001: Done.

app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
dspencer001: I'm running into some difficulties with this, I'm hoping due to my inexperience with Jellyfin development.

I think I have retrieving the additional parts with the SDK working, but the BaseRowItem class expects the old BaseItemDto class, and I only see a way to convert from the apiclient BaseItemDto class to SDK BaseItemDto. Is there a way to go the other way, or if there's not, should I implement one? Or is there something entirely different that I should be doing?

nielsvanvelzen: Yeah looks like using the SDK here requires a lot of changes, so we can using the apiclient here for now is fine.

app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
```diff
@@ -1316,6 +1329,41 @@ public void onError(Exception exception) {
...
+                    // no results - don't show us
+                    removeRow();
+                }
+
+                notifyRetrieveFinished();
+            }
+
+            @Override
+            public void onError(Exception exception) {
+                Timber.e(exception, "Error retrieving similar series items");
```
nielsvanvelzen, line 1360: Error message should be updated

app/src/main/java/org/jellyfin/androidtv/data/querying/AdditionalPartsQuery.kt
```diff
@@ -0,0 +1,7 @@
+package org.jellyfin.androidtv.data.querying;
+
+class AdditionalPartsQuery(private val ItemId: String) {
+        fun getItemId(): String {
+            return ItemId
+        }
+}
```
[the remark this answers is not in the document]
dspencer001: That's cool, done.

app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
dspencer001: Fixed.

dspencer001:
Do you want me to add the `@NonNull` annotations to the ItemRowAdapter?

PR #1783, opened 2022-06-17
Fix ItemRowAdapter.retrieveNext() Query StartIndex

Andy2244:
**Changes**
Make sure we reset the StartIndex, since the Query object is reused, otherwise all following queries will break.

**Issues**
Issue steps:
* start a library/grid view with a lib > 50 entries
* change focus until retrieveNext() is triggered via loadMoreItemsIfNeeded()
* now enable any filter like "watched" or "favorites"
* the "new" query will have the StartIndex, from the last retrieveNext() call and either display nothing or some single items or a mix of other stuff depending on the set "sortby" filter

PS: @nielsvanvelzen just a quick fix that drove me crazy, since i though my refactor stuff was the cause, yet turned out its a master branch bug.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
@@ -509,6 +509,7 @@ public class ItemRowAdapter extends ArrayObjectAdapter {
             return;
         }
 
+        Integer savedIdx = null;
         switch (queryType) {
             case Persons:
                 if (mPersonsQuery == null) {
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
Wouldn't it make more sense to reset the start index when the filters change instead?

Andy2244:
> Wouldn't it make more sense to reset the start index when the filters change instead?

Still mainly working on the Grid related classes, so have no clue if other stuff may want to set a persistent start index. Just noticed that retrieveNext is changing stuff willy-nilly.

My first instinct was to add a copy constructor to ItemQuery, so queries that only need temporal changes can use a copy. Yet seemed overkill so i opted for a quick fix.

Andy2244:
> set a persistent start index

mhh just noticed, so actually we would want to save/restore the index state tobe 100% consistent?
The issue i see is setFilters() has nothing todo with setStartIndex(), so we cant really safeguard those.

nielsvanvelzen (approved):
I think this behavior is fine for now. We can always tweak it some more. Thanks!

PR #1812, opened 2022-07-03
rework, optimize GridFragment

### LINKED ISSUE #1774: Layout/UI poster wall issues in 4k, lots of wasted space?
### Describe the bug

On my Hisense 4k TV the UI layout while in a 4k resolution looks off and has lots of wasted space?
_Thats from a Zidoo Z9X box running Android 9, not AndroidTV._

* the home item rows are unnecessary big for 4k resolution and cant be changed in size via settings
* all poster wall layouts have massive empty spaces at the sides/bottom
* the vertical layouts have strangely cut off rows
* the poster size for "small" in 4k is a little too small and medium too large for my taste, would have liked something in between

Would like to see a layout option that mimics something most Kodi skins do by default and use all available screen-space for poster rows/columns display.

Layouts are "small" and "default" and horizontal/vertical, _Images are blurred for privacy reasons_

jf_home
jf_horizontal_medium
jf_horizontal_small
jf_vertical_default
jf_vertical_small

### Logs

_No response_

### Application version

0.13.6, master

Sideload

### Device information

Zidoo Z9X

### Android version

Android 9

### Jellyfin server version

10.8

Andy2244:
**Changes**
The grid now auto adapts to the actual screen size/dimensions, while the grid itself is density independent. So the size settings will always produce a consistent row/columns view result. The spacing's/padding, chunk size now also auto adapts to whats actually visible on-screen. We also now use the resource scaling % directly to determine padding/spacing.
All card sizes, padding, is adapted within a 2px delta.
The "default" naming, was also confusing and inconsistent. The user should directly see what the actual values are and not guess what "default" means.
My beta testers also reported issues that the layout was too big for large screens, so after feedback we decided on 2 new size options to address this.
There was also the feature request to directly being able to use "play" from the grid view, so i added this feature.

- add auto image size logic
- add ability to play directly from grid
- add/change Ui size names
- remove superfluous "Default" naming
- fix grid layout fragment to following screen layout/size
- fix card refresh not working after resume
- fix/optimize chunk size logic
- fix card padding/spacing
- remove vertical grid "hacks"
- remove unnecessary grid reload's
- reduce status-line update timer 400->250ms to feel more responsive

**Issues**
fixes #1774


nielsvanvelzen:
Right now this is a mix of fixes and features. It will be a lot easier to review when split into multiple pull requests. Can you do that?

nielsvanvelzen:
To be more specific, can you split the "play from the grid view" feature to a separate PR. Everything else can stay in this one.

Andy2244:
> To be more specific, can you split the "play from the grid view" feature to a separate PR. Everything else can stay in this one.

ok will try, i know its not a ideal PR, but i had to-do quite some iterations, before i ended-up with this version, being new to the code and all.
On the upside, myself and a couple of beta testers are using the changes for two weeks now and i have no outstanding issues, only feature requests.

Andy2244:
btw i wish i could add clean kotlin classes, but i never used kotlin and some of the syntax looks like haskel "magic" to me. Yet i noticed its quite clean and has some cool language stuff and superior threading/co-routines support.

Maybe if i spend more time on the project, its worth learning some more kotlin.

Andy2244:
ok only has the grid changes/fixes now

app/src/main/java/org/jellyfin/androidtv/util/Utils.java
```diff
@@ -21,6 +21,9 @@
  * A collection of utility methods, all static.
  */
 public class Utils {
+
+    static public final long RUNTIME_TICKS_TO_MS = 10000;
```
nielsvanvelzen, line 25: Looks like this is unused

nielsvanvelzen (changes requested):
Opening the albums grid view for a music libraries with medium poster images in a vertical direction doesn't behave properly.

Same happens for artists, I think because they both use a square size instead of poster size when set to "poster".

Andy2244:
> Opening the albums grid view for a music libraries with medium poster images in a vertical direction doesn't behave properly.
> 
> Same happens for artists, I think because they both use a square size instead of poster size when set to "poster".

Ah ok, i remember there was special music poster handling, should be easy to fix.

Andy2244:
ok music views are fixed, yet check the horizontal vs vertical card sizes/aspect. I understand that the horizontal special handling was to allow for slightly none 1.0 aspect posters or combined front+back covers. 

Yet i find the varying grid item offsets that is the result of this kinda weird and visually unpleasing. So i personally rather scale/cutoff those none 1.0 cases, instead of adapting the card to it? We could still allow the special handling for large 1 row grids?

Just a idea.

nielsvanvelzen:
Yeah I just noticed when viewing music artists that I have a single entry with a 1182x548 image, it shows as a square now in the vertical view but looks really bad and displaces all other items in the horizontal view. I think forcing all items to a single size would be best (even for single row).

nielsvanvelzen:
For translations we only modify the `values/strings.xml` file. Translations can then be added on our Weblate instance at https://translate.jellyfin.org.
We don't allow translations in pull requests because those caused a lot of conflicts in the past, even causing Weblate to stop pushing changes to the repository.

The translations are made by actual people, not by machine. We'd rather have the app show an English string instead of a badly machine translated string.

app/src/main/res/values/strings.xml
```diff
@@ -365,12 +365,15 @@
     <string name="enable_smart_view">Enable smart screen</string>
     <string name="enable_smart_view_description">Smart screen shows relevant items instead of a grid</string>
     <string name="image_type_default">Default</string>
```
nielsvanvelzen, line 367: Remove this string from the source file and keep the translations (will be auto-removed after merge)

app/src/main/res/values/strings.xml
```diff
@@ -365,12 +365,15 @@
...
+    <string name="image_type_poster">Poster</string>
     <string name="image_type_thumbnail">Thumbnail</string>
     <string name="image_type_banner">Banner</string>
     <string name="image_size_auto">Default</string>
```
nielsvanvelzen, line 371: Remove this string from the source file and keep the translations (will be auto-removed after merge)

app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
```diff
@@ -70,119 +76,324 @@ public class StdGridFragment extends GridFragment {
...
+        setupEventListeners();
+    }
 
-        setGridSizes();
+    protected boolean isDirty() {
+        return mDirty;
+    }
 
-        mJumplistPopup = new JumplistPopup();
+    protected BrowseRowDef getRowDef() {
```
nielsvanvelzen, line 140: Looks unused, can probably be removed

app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
```diff
@@ -70,119 +76,324 @@ public class StdGridFragment extends GridFragment {
...
+        mCardHeight = height;
+    }
+
+    protected int getCardHeight() {
+        return mCardHeight;
     }
 
-    private void setGridSizes() {
-        Presenter gridPresenter = getGridPresenter();
+    public void printViewStats(View view) {
```
nielsvanvelzen, line 162: Looks unused, can probably be removed

app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
```diff
@@ -289,50 +505,16 @@ protected void buildAdapter(BrowseRowDef rowDef) {
...
-                case MED:
-                    return mImageType.equals(ImageType.BANNER) ? MED_BANNER : MED_CARD;
-                case LARGE:
-                    return mImageType.equals(ImageType.BANNER) ? LARGE_BANNER : LARGE_CARD;
-                default:
-                    return mImageType.equals(ImageType.BANNER) ? SMALL_BANNER : SMALL_CARD;
-            }
-        }
+//        printViewStats(getGridView());
+//        printViewStats(mGridDock);
```
nielsvanvelzen, line 517: can probably be removed

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
Andy2244: its a helper to debug the view, would like to keep this so i don't have to always add it again just to check whats happening with a view.

app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
Andy2244: done

app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
Andy2244: done

app/src/main/res/values/strings.xml
Andy2244: ok

nielsvanvelzen (approved):

nielsvanvelzen:
Works perfect now, thanks!

PR #1818, opened 2022-07-06
add default menu hotkey actions to grid

Andy2244:
**Changes**
- add default play action to grid items
- add default menu action to grid items

somewhat depends on #1812 so setupEventListeners() order is fully working on resume() + changes

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
@@ -14,6 +14,9 @@ import android.widget.LinearLayout;
 import android.widget.PopupMenu;
 import android.widget.PopupWindow;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.leanback.widget.BaseGridView;
 import androidx.leanback.widget.OnItemViewClickedListener;
 import androidx.leanback.widget.OnItemViewSelectedListener;
 import androidx.leanback.widget.Presenter;
@@ -50,6 +53,8 @@ import org.jellyfin.androidtv.ui.shared.MessageListener;
 import org.jellyfin.androidtv.util.CoroutineUtils;
 import org.jellyfin.androidtv.util.KeyProcessor;
 import org.jellyfin.androidtv.util.Utils;
+import org.jellyfin.androidtv.util.apiclient.BaseItemUtils;
+import org.jellyfin.androidtv.util.apiclient.PlaybackHelper;
 import org.jellyfin.apiclient.interaction.EmptyResponse;
 import org.jellyfin.apiclient.model.dto.BaseItemType;
 import org.jellyfin.apiclient.model.entities.CollectionType;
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
```diff
@@ -514,8 +519,35 @@ public void dismiss() {
...
+                return false;
+            }
+            if ((event.getKeyCode() == KeyEvent.KEYCODE_MEDIA_PLAY || event.getKeyCode() == KeyEvent.KEYCODE_MEDIA_PLAY_PAUSE) && BaseItemUtils.canPlay(mCurrentItem.getBaseItem())) {
+                mediaManager.getValue().setCurrentMediaAdapter(mGridAdapter);
+                mediaManager.getValue().setCurrentMediaPosition(mCurrentItem.getIndex());
+                mediaManager.getValue().setCurrentMediaTitle(mFolder.getName());
+                //default play action
+                mHandler.postDelayed(() -> PlaybackHelper.playOrPlayNextUp(mCurrentItem.getBaseItem(), requireActivity()), 10);
+                return true;
+            }
```
nielsvanvelzen, line 536: The keyprocessor already deals with KEYCODE_MEDIA_PLAY, why did you make a separate implementation here?

Andy2244: I was not sure if changing from `createPlayMenu`  to a actual play action was just a oversight in the org. code or had some use in other views?

My goal was to get a actual play action, not just a menu if u hit the play-button, since there is a dedicated menu action otherwise.

I can rework this in the keyprocessor for series/seasons/episode/movie/video cases?

Andy2244: ah also the org. code for BaseActivity did set those parameters, so i just mimic-ed this logic.
```
                mediaManager.getValue().setCurrentMediaAdapter(mGridAdapter);
                mediaManager.getValue().setCurrentMediaPosition(mCurrentItem.getIndex());
                mediaManager.getValue().setCurrentMediaTitle(mFolder.getName());
```

nielsvanvelzen: > My goal was to get a actual play action, not just a menu if u hit the play-button, since there is a dedicated menu action otherwise.

It only shows a menu when there's multiple options, I always use the play button on my remote to immediately start items from the play next home section. So I don't see the need for duplicate code here.

Andy2244: Not sure what you mean?
Ofc it plays from the play next home section, since it only actually shows possible play-next items?

So whats the conclusion now? Do i move the new play-code to the respective basetypes inside the keyhandler?

Andy2244: Ok just noticed that the "play" case for any collection basetype makes no sense, it always just brings-up up a menu, even for music.

That's not how the webclient or any music-player works, if i hit "play" on a album/playlist i expect that it actually starts. 
So we should change all those collection cases to a actual play, so this is consistent with the webclient and normal user expectations.

Andy2244:
ok just closing for now and reworking the play logic, there is a lot more that needs changes.

PR #1820, opened 2022-07-06
Rewrite crash reporting

nielsvanvelzen:
This PR replaces our current crash reporting with a new implementation that sends the logs to the last used Jellyfin server. The reports use markdown with a YAML header indicating the format. This can be used in the future if we create a log viewer in jellyfin-web.

The crash reporting functionality caches the Jellyfin server url and access token when restoring a session (login/app open). When the app crashes these values are used to manually connect to the Jellyfin server instead of using the SDK. This is so we can catch SDK issues too.

**Changes**

- Fix ACRA process running Application.onCreate
- Move crash reporting to new TelemetryService
- Use ACRA Toast instead of dialog (always auto-submits now)
  - This is to avoid an issue where the dialog is inresponsive
- Upload crash report to Jellyfin server instead of Tracepot
- Remove ACRA limitter

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/JellyfinApplication.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/JellyfinApplication.kt
@@ -14,16 +14,12 @@ import androidx.work.await
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
-import org.acra.config.dialog
-import org.acra.config.httpSender
-import org.acra.config.limiter
-import org.acra.data.StringFormat
-import org.acra.ktx.initAcra
-import org.acra.sender.HttpSender
+import org.acra.ACRA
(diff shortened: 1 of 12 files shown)


nielsvanvelzen:
Right now 2 preferences are removed:

- pref_acra_alwaysaccept - This is now the (only) default option, the dialog is removed
- pref_acra_syslog - System logs are now always included, I'm not sure if we want to keep this option

Additionally, because the options moved to a different preference store, the value is reset and reporting is now enabled for everyone when updating.

nielsvanvelzen:
Additional changes pushed:

- Add back option to enable/disable logs (previously called system logs), enabled by default
  - The description previously stated that this could include logs from other applications, this is not true so the naming has changed to not mention "system logs" but just "logs" instead.
- Rename the strings related to crash reports, they don't use "acra" in the name anymore and this forces all translations to be rewritten.

PR #1822, opened 2022-07-07
adjust TV detection logic

Andy2244:
**Changes**
- adjust TV detection logic

**Issues**
All TV media boxes running normal Android, instead of AndroidTV will fail this check. 
All none AndroidTV certified devices run Android, which is the largest segment for TV media devices. 
The main reason is the harder AndroidTV certification process and licensing cost, so this will not change in the future.

So we should never rely on anything specific AndroidTv related.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/repository/NotificationsRepository.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/repository/NotificationsRepository.kt
@@ -39,6 +39,8 @@ class NotificationsRepositoryImpl(
 		val isTouch = context.packageManager.hasSystemFeature("android.hardware.touchscreen")
 		val hasHdmiCec = context.packageManager.hasSystemFeature("android.hardware.hdmi.cec")
 
-		if (invalidUiMode && isTouch && !hasHdmiCec) addNotification(context.getString(R.string.app_notification_uimode_invalid))
+		if (invalidUiMode && isTouch && !hasHdmiCec) {
+			addNotification(context.getString(R.string.app_notification_uimode_invalid))
+		}
 	}
 }
--- a/app/src/main/java/org/jellyfin/androidtv/data/repository/NotificationsRepository.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/repository/NotificationsRepository.kt
@@ -36,7 +36,9 @@ class NotificationsRepositoryImpl(
 	private fun addUiModeNotification() {
 		val supportedUiModes = setOf(Configuration.UI_MODE_TYPE_TELEVISION, Configuration.UI_MODE_TYPE_UNDEFINED)
 		val invalidUiMode = !supportedUiModes.contains(uiModeManager.currentModeType)
+		val isTouch = context.packageManager.hasSystemFeature("android.hardware.touchscreen")
+		val hasHdmiCec = context.packageManager.hasSystemFeature("android.hardware.hdmi.cec")
 
-		if (invalidUiMode) addNotification(context.getString(R.string.app_notification_uimode_invalid))
+		if (invalidUiMode && isTouch && !hasHdmiCec) addNotification(context.getString(R.string.app_notification_uimode_invalid))
 	}
 }


nielsvanvelzen:
The Android TV app is not meant for normal Android devices, so the notification does exactly what it should do. The mobile app is for non-TV devices.

Andy2244:
> The Android TV app is not meant for normal Android devices, so the notification does exactly what it should do. The mobile app is for non-TV devices.

This ignores the reality that the majority of TV media-boxes run Android and not AndroidTV. 
No user of those boxes wants to use the Android version, since they still use a TV and have no touch inputs.

Most of the devices in those list all run Android and are yet only designed for TV use:
https://androidtvbox.eu/category/tv-box/
https://www.androidtv-guide.com/streaming-gaming/

So "Jellyfin Android TV" runs fine on all those devices, since there is not any AndroidTV specific code and this should remain so.
This client should really just be a TV optimized variant and not a variant that runs only on AndroidTV specifically.

Andy2244:
Here is a sort-able list with OS column:
https://tvboxstop.com/tv-box-rank-chart-2020/

nielsvanvelzen:
Ok so we'd need to change the detection logic because those manufacturers didn't change the device mode flag. I don't think adding the "normal" type to the list is the right fix as that hides the notification on phones/tablets, where it should show.

Andy2244:
> Ok so we'd need to change the detection logic because those manufacturers didn't change the device mode flag. I don't think adding the "normal" type to the list is the right fix as that hides the notification on phones/tablets, where it should show.

I will see what i can find, possible solutions are:

- do a 2k, 4k with 16:9+ check
- check if we have a hdmi TV device

any other ideas?

Andy2244:
> I think we can check for touch screen support.

Depends on how this is detected?
You are aware that some TV's have "air mouse" enabled remotes, so they emulate a pointing device?
You can also order a BT "Air mouse" for those media-boxes, which is done to somewhat make none TV app's usable.

So if we can detect real touch-input and distinguish it from those Air-mouse devices, this could work along with a basic aspect-ratio check?

nielsvanvelzen:
A cursor is not a touch screen. I think using `!packageManager.hasSystemFeature("android.hardware.touchscreen")` should therefor be reliable enough. Together with the existing check it should support most TV boxes that don't run Android TV.

Andy2244:
ah cool, will test this asap on my device. 

_Btw i sold my ShieldTv/FireTv, so keep in mind that i only test via one of those Android 9 devices._

Andy2244:
@nielsvanvelzen touch detection does not work, my device reports a touchpad and even a accelerometer... I think thats all just to emulate things.

Yet i found the "android.hardware.hdmi.cec" feature, which "should" only exist on those media boxes?
Can you verify it works for your cases as well?

nielsvanvelzen:
Tested on a bunch of devices & emulator:

- ATV emulator: doesn't show ✔️
- Nvidia Shield Pro: doesn't show ✔️
- Android (mobile) emulator: shows ✔️
- Oneplus 6: shows ✔️
- Galaxy Tab S 8.4 (with unofficial lineageos rom): shows ✔️

So I think it works as expected

app/src/main/java/org/jellyfin/androidtv/data/repository/NotificationsRepository.kt
```diff
@@ -36,7 +36,9 @@ class NotificationsRepositoryImpl(
 	private fun addUiModeNotification() {
 		val supportedUiModes = setOf(Configuration.UI_MODE_TYPE_TELEVISION, Configuration.UI_MODE_TYPE_UNDEFINED)
 		val invalidUiMode = !supportedUiModes.contains(uiModeManager.currentModeType)
+		val isTouch = context.packageManager.hasSystemFeature("android.hardware.touchscreen")
+		val hasHdmiCec = context.packageManager.hasSystemFeature("android.hardware.hdmi.cec")
 
-		if (invalidUiMode) addNotification(context.getString(R.string.app_notification_uimode_invalid))
+		if (invalidUiMode && isTouch && !hasHdmiCec) addNotification(context.getString(R.string.app_notification_uimode_invalid))
```
nielsvanvelzen, line 42:
```suggestion
		if (invalidUiMode && isTouch && !hasHdmiCec) {
			addNotification(context.getString(R.string.app_notification_uimode_invalid))
		}
```

nielsvanvelzen (changes requested):

Andy2244:
> So I think it works as expected

Ah nice, btw any chance to get the new dolby vision stuff working for BaseItemDto or in some other way?
We also miss DV/Atmos text/icons for the various views.

nielsvanvelzen:
> btw any chance to get the new dolby vision stuff working for BaseItemDto or in some other way?
We also miss DV/Atmos text/icons for the various views.

Rewrite all playback code to use the SDK, that's the only way.

nielsvanvelzen (approved):

PR #1825, opened 2022-07-08
fix, don't send invalid sub/audio index

Andy2244:
**Changes**
- fix server broken "-rangetype" behavior
- fix don't send -1 subIndex
- fix DlnaProfileType

**Issues**
In transcode cases the client will always trigger a video-stream transcode, since the rangetype is always missing.
Hotfix until jellyfin/jellyfin#8078 is merged and we get a new server release.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/GetPlaybackInfoResponse.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/GetPlaybackInfoResponse.java
@@ -90,8 +90,11 @@ public class GetPlaybackInfoResponse extends Response<PlaybackInfoResponse> {
 
         streamInfo = new StreamInfo();
         streamInfo.setItemId(options.getItemId());
-        streamInfo.setMediaType(DlnaProfileType.Audio);
-
+        if (isVideo) {
+            streamInfo.setMediaType(DlnaProfileType.Video);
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
The current master branch is targetting 10.8>= but is not in any release/beta yet. I don't think a stable release will be made before 10.8.2 is ready so this fix shouldn't be necessary then.

Andy2244:
> I don't think a stable release will be made before 10.8.2 is ready so this fix shouldn't be necessary then.

If you think so ok, so also no 0.13.x hotfix for this with the streamcopy code enabled?
Its super annoying, since this server bug wont allow any audio/sub/container only transcodes without also triggering a video transcode.

We should keep the "-1" checks, since sending any value to the server will also prevent a stream copy.

nielsvanvelzen:
> We should keep the "-1" checks, since sending any value to the server will also prevent a stream copy.

Isn't that fixed in your server PR?

Andy2244:
> > We should keep the "-1" checks, since sending any value to the server will also prevent a stream copy.
> 
> Isn't that fixed in your server PR?

Yes, but only for this specific check, in general the server does not check against invalid values, just against null values. So we need to try not sending junk values.

nielsvanvelzen (approved):

PR #1828, opened 2022-07-09
add default menu, play actions to grid

Andy2244:
Update on #1818 

**Changes**
add default menu, play actions to grid
fix default play actions for collections
fix album play action

**Issues**
Add default key handling/menu's to the grid views.
Now a play hotkey action does trigger a actual "play" in all cases, where this is possible.
Exception is "Audio" via AudioQueueItem, which needs additional fixes, as it does not support the menu key as example.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
@@ -526,6 +526,9 @@ public class StdGridFragment extends GridFragment {
             mActivity.registerKeyListener(new KeyListener() {
                 @Override
                 public boolean onKeyUp(int key, KeyEvent event) {
+                    if (mGridDock == null || !mGridDock.hasFocus()) {
+                        return false;
+                    }
                     if (key == KeyEvent.KEYCODE_MEDIA_PLAY || key == KeyEvent.KEYCODE_MEDIA_PLAY_PAUSE) {
                         mediaManager.getValue().setCurrentMediaAdapter(mGridAdapter);
                         mediaManager.getValue().setCurrentMediaPosition(mCurrentItem.getIndex());
--- a/app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
@@ -95,15 +95,14 @@ public class KeyProcessor {
                                 return true;
                             case Series:
                             case Season:
-                            case BoxSet:
-                                createPlayMenu(rowItem.getBaseItem(), true, false, activity);
+                                // play next, resume or play
+                                PlaybackHelper.playOrPlayNextUp(item, activity);
                                 return true;
+                            case BoxSet:
                             case MusicAlbum:
                             case MusicArtist:
-                                createPlayMenu(rowItem.getBaseItem(), true, true, activity);
-                                return true;
                             case Playlist:
-                                createPlayMenu(rowItem.getBaseItem(), true, "Audio".equals(rowItem.getBaseItem().getMediaType()), activity);
+                                PlaybackHelper.retrieveAndPlay(item.getId(), false, activity);
                                 return true;
                             case Photo:
                                 // open photo player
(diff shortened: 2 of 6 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
```diff
@@ -514,8 +519,32 @@ public void dismiss() {
...
+                return false;
+            }
+            if (event.getKeyCode() == KeyEvent.KEYCODE_MEDIA_PLAY || event.getKeyCode() == KeyEvent.KEYCODE_MEDIA_PLAY_PAUSE) {
+                mediaManager.getValue().setCurrentMediaAdapter(mGridAdapter);
+                mediaManager.getValue().setCurrentMediaPosition(mCurrentItem.getIndex());
+                mediaManager.getValue().setCurrentMediaTitle(mFolder.getName());
+            }
+            return KeyProcessor.HandleKey(event.getKeyCode(), mCurrentItem, requireActivity());
+        }
+    };
```
nielsvanvelzen, line 536: This is basically duplicated code from setupEventListeners()

app/src/main/java/org/jellyfin/androidtv/ui/browsing/StdGridFragment.java
```diff
@@ -514,8 +519,32 @@ public void dismiss() {
...
 
+    @Override
+    public void onMessageReceived(CustomMessage message) {
+        switch (message) {
+            case RefreshCurrentItem:
+                refreshCurrentItem();
+                break;
+        }
+    }
```
nielsvanvelzen, line 545: This is basically duplicated code from setupEventListeners()

nielsvanvelzen (changes requested):
Looks like the StdGridFragment already supports the play key but the GenericGridActivity doesn't extend BaseActivity causing it to not work. I guess just changing that would be enough.

Andy2244:
> I guess just changing that would be enough.

@nielsvanvelzen I thought "BaseActivity" is deprecated, also this will send the key-events for the whole activity, instead of just whan the grid is active? I guess we can add extra filters on when to passthrough.

Up to you, if you want to use "BaseActivity"?

nielsvanvelzen:
Yeah BaseActivity is deprecated but this code still relies on it, so either the old code is removed or we use BaseActivity.

Andy2244:
> Yeah BaseActivity is deprecated but this code still relies on it, so either the old code is removed or we use BaseActivity.

Sure, happy to oblige.

Andy2244:
ok all done

app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
```diff
@@ -95,15 +95,14 @@ public static boolean HandleKey(int key, BaseRowItem rowItem, Activity activity)
                                 return true;
                             case Series:
                             case Season:
-                            case BoxSet:
-                                createPlayMenu(rowItem.getBaseItem(), true, false, activity);
+                                // play next, resume or play
+                                PlaybackHelper.playOrPlayNextUp(item, activity);
```
nielsvanvelzen, line 99: I think we should keep the menu for series/seasons/playlists

app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
```diff
@@ -254,8 +253,9 @@ public static PopupMenu createItemMenu(BaseRowItem rowItem, UserItemDataDto user
                         && userData.getUnplayedItemCount() > 0) {
                     menu.getMenu().add(0, MENU_PLAY_FIRST_UNWATCHED, order++, R.string.lbl_play_first_unwatched);
                 }
+                boolean isSeries = item.getBaseItemType() == BaseItemType.Series || item.getBaseItemType() == BaseItemType.Season;
                 menu.getMenu().add(0, MENU_PLAY, order++, item.getIsFolderItem() ? R.string.lbl_play_all : R.string.lbl_play);
-                if (item.getIsFolderItem()) {
+                if (!isSeries && item.getIsFolderItem()) {
```
nielsvanvelzen, line 258: Why disable shuffle for series? I know users have asked for this in the past.

app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
```diff
@@ -380,15 +380,17 @@ public void onError(Exception exception) {
                     query.setSortBy(new String[]{ItemSortBy.SortName});
                     query.setSortOrder(SortOrder.Ascending);
                     query.setLimit(1);
-                    query.setExcludeItemTypes(new String[] {"Series","Season","Folder","MusicAlbum","Playlist","BoxSet"});
+                    // NOTE: exclude filters seem broken?
```
nielsvanvelzen, line 383: What's broken about it?

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
Andy2244: I disagree, if the user wants a menu there is the menu key.
On every music player ever, if you use a explicit "play" action on a playlist it will start play and not show a menu.
On series/seasons the only other meaningful play related action is "play all", which is useless since we also have a queue follow-up settings.
If the user hit "play" on a season/series its pretty clear what the user wants to happen and that's what the new play code does.

We should never show a menu on a "play" action, since we have a "menu" key for it. Play should always default to a actual play action. As noted thats also how the web interface works, if i hit the "play" icon there it plays.

app/src/main/java/org/jellyfin/androidtv/util/apiclient/PlaybackHelper.java
```diff
@@ -257,11 +264,55 @@ public void onError(Exception exception) {
         }
     }
 
+    public static void playOrPlayNextUp(final BaseItemDto item, final Context activity) {
```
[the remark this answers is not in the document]
Andy2244: No, the "MENU_PLAY_FIRST_UNWATCHED" is just a "IsUnplayed" query, not a next-up query.
It also does not resume.

app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
mueslimak3r: Jellyfin web has a shuffle button for series and seasons. I think maintaining feature parity is important.

It is nice to be able to shuffle episodes when rewatching non-linear series.

Andy2244: ok, makes sense

Andy2244: done

PR #1829, opened 2022-07-09
Add new experimental picture viewer

nielsvanvelzen:
I started work on this around 2 months ago but didn't completely finish it yet. It already uses a feature flag which is disabled in releases so it won't hurt to merge it. I've made a new column in my board (https://github.com/jellyfin/jellyfin-androidtv/projects/15#column-19021853) that shows the remaining tasks.

The biggest feature is that it doesn't depend on the media manager anymore so it's sort of a part of the playback rewrite (#1057).

**Changes**
- Add preference to enable new picture viewer
  - Like the playback rewrite; only shows in development builds
- Add KeyHandler DSL to write readable key press handlers
- Add new picture viewer
  - Supports portrait mode
  - Sort of supports touch
  - Supports slideshows
  - Shows controls to make it easy to start slideshows and navigate forward/backward

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/picture/PictureViewerFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/picture/PictureViewerFragment.kt
@@ -1,0 +1,156 @@
+package org.jellyfin.androidtv.ui.picture
+
+import android.os.Bundle
+import android.view.KeyEvent
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.view.animation.Animation
+import android.view.animation.AnimationUtils
+import androidx.core.view.isGone
+import androidx.core.view.isVisible
+import androidx.fragment.app.Fragment
+import androidx.lifecycle.lifecycleScope
+import kotlinx.coroutines.flow.filterNotNull
+import kotlinx.coroutines.launch
+import org.jellyfin.androidtv.R
+import org.jellyfin.androidtv.databinding.FragmentPictureViewerBinding
+import org.jellyfin.androidtv.ui.AsyncImageView
+import org.jellyfin.androidtv.util.createKeyHandler
+import org.jellyfin.sdk.api.client.ApiClient
+import org.jellyfin.sdk.api.client.extensions.imageApi
+import org.jellyfin.sdk.api.client.extensions.libraryApi
+import org.jellyfin.sdk.model.api.BaseItemDto
+import org.jellyfin.sdk.model.api.ImageType
+import org.koin.android.ext.android.inject
+import org.koin.androidx.viewmodel.ext.android.sharedViewModel
+
+class PictureViewerFragment : Fragment(), View.OnKeyListener {
+	companion object {
+		/**
+		 * The download API is used by jellyfin-web when loading images. Unfortunately with our
+		 * current code it doesn't work for the app. Larger files (gifs, large photos etc.) can
+		 * cause the app to go out of memory. This is mostly catched by Glide but it ends up
+		 * never displaying the picture in those cases.
+		 *
+		 * This toggle is left in the code in case we migrate to a different image processor that
+		 * potentially fixes the issue.
+		 */
+		const val USE_DOWNLOAD_API = false
+	}
+
+	private val pictureViewerViewModel by sharedViewModel<PictureViewerViewModel>()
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/picture/PictureViewerFragment.kt
```diff
@@ -0,0 +1,156 @@
...
+import org.jellyfin.sdk.model.api.ImageType
+import org.koin.android.ext.android.inject
+import org.koin.androidx.viewmodel.ext.android.sharedViewModel
+
+class PictureViewerFragment : Fragment(), View.OnKeyListener {
+	companion object {
+		/**
+		 * The download API is used by jellyfin-web when loading images. Unfortunately with our
+		 * current code it doesn't work for the app. Larger files (gifs, large photos etc.) can
+		 * cause the app to go out of memory. This is mostly catched by Glide but it ends up
```
thornbill, line 33: catched -> caught?

app/src/main/java/org/jellyfin/androidtv/ui/picture/PictureViewerViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+	private val _currentItem = MutableStateFlow<BaseItemDto?>(null)
+	val currentItem = _currentItem.asStateFlow()
+
+	suspend fun loadItem(id: UUID) {
+		// Load requested item
+		val itemResponse by api.userLibraryApi.getItem(itemId = id)
+		_currentItem.value = itemResponse
+
+		val albumResponse by api.itemsApi.getItemsByUserId(
+			parentId = itemResponse.albumId,
```
thornbill, line 35: Album refers to music albums, so I think you would want to do something like the following:

```suggestion
			parentId = if (itemResponse.type === BaseItemKind.PHOTO_ALBUM) itemResponse.id else itemResponse.parentId,
```

thornbill (commented):
I'm not sure how perfect you are striving to get this for the first implementation, but I noticed a few issues:

1. There seems to be an issue when changing photos (either manually or via the slideshow) with gifs. The image will change briefly then go back to displaying the previous image.
2. In large photo libraries it seems like the current "position" is getting lost somehow. For example, I started playback on photo 301, click next, and a previous photo is displayed instead of photo 302.

It would probably be a good idea to show the photo name at least when the controls are visible (this would also help diagnose issue 2 further since I have many copies of the same photo in that album).

It might be good to show the "playback" controls by default to help with discoverability and hide them after a short time. I also didn't find pushing "up" to show the controls intuitive... I'm more accustomed to services using the "down" button to display them. That is more of a personal preference probably and there is some inconsistency among other services on that.

app/src/main/java/org/jellyfin/androidtv/ui/picture/PictureViewerViewModel.kt
nielsvanvelzen: In the createIntent() function we check if the given BaseItem is PHOTO so it can never by PHOTO_ALBUM in this place.

nielsvanvelzen:
> There seems to be an issue when changing photos (either manually or via the slideshow) with gifs. The image will change briefly then go back to displaying the previous image.

I haven't experienced this issue but I'd say it's out of scope for this initial PR.

> In large photo libraries it seems like the current "position" is getting lost somehow. For example, I started playback on photo 301, click next, and a previous photo is displayed instead of photo 302.

I'll need to create a bigger album (biggest is about 40 items right now) to test this but I've already made an task for lazy loading which might fix this issue.

> It would probably be a good idea to show the photo name at least when the controls are visible (this would also help diagnose issue 2 further since I have many copies of the same photo in that album).

Agreed, this was one of the things I wanted to add but not in this initial PR. This one too is already a task on the big list, it's not a feature in the current viewer either (except if you count that browse thingy that you can open while viewing).

> It might be good to show the "playback" controls by default to help with discoverability and hide them after a short time. I also didn't find pushing "up" to show the controls intuitive... I'm more accustomed to services using the "down" button to display them. That is more of a personal preference probably and there is some inconsistency among other services on that.

Added support for KEYCODE_DPAD_DOWN to open the controls, I think opening the controls makes sense when the timer to auto-close it is added. I've updated the task to describe this behavior.

PR #1834, opened 2022-07-10
Show notification on server screen when server setup is incomplete

nielsvanvelzen:
**Changes**
- Save setupCompleted boolean for servers
- Show notification when setupCompleted is false

image

**Issues**

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -497,5 +497,5 @@
     <string name="pref_crash_report_logs_enabled">Logs will be included in crash reports</string>
     <string name="pref_crash_report_logs_disabled">Logs will not be included in crash reports</string>
     <string name="crash_report_toast">Oops! Something went wrong, a crash report was send to your Jellyfin server.</string>
-    <string name="server_setup_incomplete">This server has not finished installing. Open your server in the browser to finish setup before signing in.</string>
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -497,4 +497,5 @@
     <string name="pref_crash_report_logs_enabled">Logs will be included in crash reports</string>
     <string name="pref_crash_report_logs_disabled">Logs will not be included in crash reports</string>
     <string name="crash_report_toast">Oops! Something went wrong, a crash report was send to your Jellyfin server.</string>
+    <string name="server_setup_incomplete">This server has not finished installing. Open your server in the browser to finish setup before signing in.</string>
```
nielsvanvelzen, line 500: Open for suggestions for this one, not a big fan of the current message.

sourpatched: "Jellyfin is pending setup, please launch Jellyfin Web GUI from a web browser"
"Sorry, it looks like your Jellyfin server is not completely setup. Login from a web browser before continuing here"

Does this also scan to see if the web gui installed since the Web GUI is optional?

thornbill: Here is my suggestion:

> The setup of this server has not been completed. Open Jellyfin in a web browser to finish setup before signing in.

PR #1839, opened 2022-07-12
Avoid multiple discovery processes running at the same time

nielsvanvelzen:
When switching between the user and server selection screens the process continuously restarted. This causes a lot of logs and network activity. Prevent this by running only one process at a time.

**Changes**
- Add mutex to only run one discovery process at a time

**Issues**


thornbill (approved):
Untested, but looks good

PR #1847, opened 2022-07-16
Always use Quick Connect

nielsvanvelzen:
Issue found during final testing phase for 0.14. When installing the client with no previous data you will likely use the following flow:

1. Add server via discovery
2. Press public user to sign in

This opened the password input instead of the quick connect screen because we can't limit a QC login to just a specific user. This is annoying though because you'd need to input a password.

I changed it to always show the quick connect screen. If you enter the code as a different user everything still works fine, just a bit confusing because you don't end up signed in as the user initially used.

**Changes**

- Always show quick connect screen

**Issues**


mueslimak3r (approved):
looks good!

PR #1860, opened 2022-07-18
Update button order on item details screen

thornbill:
**Changes**
Updates the button ordering on the item details screen to try to be more consistent:

1. Shuffle button is now before the Play Trailers button (this is the order in web and I think shuffle is more likely to be used more frequently)
2. Button order in the "Other Options" menu should better match their order when not in the menu

One oddity in the current code is that the Watched button never moves to the menu even though buttons preceding it do... Should this be updated so it will move to the menu before the others (primarily happens with the Play Trailers button for series)?

**Issues**
N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
@@ -1019,7 +1019,7 @@ public class FullDetailsActivity extends BaseActivity implements RecordingIndica
             }
         }
         //Video versions button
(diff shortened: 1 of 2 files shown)


mueslimak3r (approved):

nielsvanvelzen (approved):

nielsvanvelzen:
> One oddity in the current code is that the Watched button never moves to the menu even though buttons preceding it do... Should this be updated so it will move to the menu before the others (primarily happens with the Play Trailers button for series)?

I think this is because it's a toggle and adding that to the submenu is too much work?

nielsvanvelzen:
Well maybe it's just because legacy code then

PR #1861, opened 2022-07-19
Fix live tv guide sorting

thornbill:
**Changes**
Fixes the sorting options not working in the live tv guide by removing the client side sorting and caching of channel orders... this matches the web implementation that just queries the channels endpoint when the order is changed

**Issues**
I couldn't find one, but it has annoyed me forever...

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/livetv/TvManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/livetv/TvManager.java
@@ -44,7 +44,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.Date;
 import java.util.GregorianCalendar;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/livetv/TvManager.java
```diff
@@ -81,11 +80,6 @@ public static List<ChannelInfoDto> getAllChannels() {
         return allChannels;
     }
 
-    public static void clearCache() {
```
thornbill, line 84: Unused

app/src/main/java/org/jellyfin/androidtv/ui/livetv/TvManager.java
```diff
@@ -143,59 +131,22 @@ public void onResponse(final ChannelInfoDtoResult response) {
...
-                        return -Boolean.compare(channelOneFav, channelTwoFav);
-                    }
-                });
+            channelIds = new String[allChannels.size()];
+            String last = getLastLiveTvChannel();
+            int i = 0;
+            for (ChannelInfoDto channel : allChannels) {
+                channelIds[i++] = channel.getId();
+                if (channel.getId().equals(last)) ndx = i;
+                //TvApp.getApplication().getLogger().Debug("Last played for "+channel.getName()+ " is "+channel.getUserData().getLastPlayedDate());
```
nielsvanvelzen, line 148: This comment can be removed (or changed to a Timber.d call)

thornbill: :+1: All commented logging lines have been removed from this file now.

nielsvanvelzen (approved):

PR #1863, opened 2022-07-19
Remove stateListAnimator from button styles

nielsvanvelzen:
Amazon changes the platform Button widget to scale when pressed, this doesn't work nice with some of our custom button styles and is not consistent with other platforms. This change removes the animation.

**Changes**

- Remove stateListAnimator from button styles

**Issues**


thornbill:
What would be a good example of where this is noticeable?

nielsvanvelzen:
I think the most noticeable is the server button. When pressed it scales (amazon style change) but the ripple stays on top and doesn't scale.

thornbill (approved):

PR #1866, opened 2022-07-20
Fix inconsistent red color for recordings

thornbill:
**Changes**
The recording icons were using two different shades of red. This updates the icons to both use the same (less harsh) red color.

**Issues**
N/A


nielsvanvelzen:
Ideally we'd remove the _red version of the icon and use tinting (like all other places in the app) but this will do for now.

nielsvanvelzen (approved):

PR #1876, opened 2022-07-22
fix series logo scaling in player

mueslimak3r:
**Changes**
* adjust logo/title margins so the logo only has a 2dp bottom margin but the title text still has 8dp
* constrain the logo to a max width of 450dp
* copied some other properties from `mainImage` in `view_row_details.xml` which is used to display an item logo or poster in `FullDetailsActivity`.
* use `fitCenter` scaling to fill the view. (this was useful when testing with SVGs since they needed to be scaled. SVG support is not included in this PR)

**Issues**
* series logos would sometimes be far too large (I've seen some that take up half the screen horizontally)

**Notes**
* I'm not super knowledgable about layouts so hopefully someone with more experience can give feedback on whether they are used correctly here, and if some aren't needed.


nielsvanvelzen (approved):
Layout changes look fine!

PR #1891, opened 2022-07-25
Fix typo in source string

crobibero:
Thanks wife

I tried to edit on weblate but looks like editing the source string is disabled?


nielsvanvelzen (approved):

nielsvanvelzen:
> Thanks wife

Thanks @crobibero's wife

> I tried to edit on weblate but looks like editing the source string is disabled?

Yeah to prevent merge conflicts the source should always be changed via GitHub and the translations always via Weblate. We haven't had any issues with Weblate since those rules.

PR #1924, opened 2022-08-02
Add Quality Selection to Playback Overlay

zkhcohen:
**Changes**

1. Added quality selection (max bitrate) menu to playback overlay.
2. Added PlaybackController.refreshSteam() function to restart a stream when transcoding options are changed.
3. Fixed Mbit/s to Kbit/s conversion factor (100 -> 1000). 
4. Fixed unusual Mbit/s increment (21Mbit/s -> 20Mbit/s).

This PR builds off of https://github.com/jellyfin/jellyfin-androidtv/pull/1302 using my _very_ limited understanding of Kotlin, implementing `VideoQualityController` / `SelectQualityAction`.

As a result of this change, it was necessary to implement the function `PlaybackController.refreshSteam()` in order to restart the stream when the setting is changed, applying the new setting.

During this change, it was also noted that the Mbit/s to Kbit/s conversion factor was incorrect (x100), and an unusual quality increment had been added to the settings menu (21 Mbit/s). These have been fixed.

**Issues**
May satisfy these vague enhancements:
- https://github.com/jellyfin/jellyfin-androidtv/issues/1388
- https://github.com/jellyfin/jellyfin-androidtv/issues/455
- https://github.com/jellyfin/jellyfin-androidtv/issues/43

**Testing**
- Try changing quality settings through the playback menu. You should be able to verify that the media is being transcoded through the logs / dashboard.
- Try playing a different video - the settings should persist.
- Open the settings -> playback menu through the main menu - the settings configured through the overlay should persist there.
- Change the quality in the settings -> playback menu through the main menu, then open both menus in the playback overlay - the settings should persist and you should be able to verify that the media is being transcoded through the logs / dashboard.


zkhcohen:
@mueslimak3r I removed the transcoding settings from https://github.com/jellyfin/jellyfin-androidtv/pull/1406 and built this off of the new v0.14.0 release for Jellyfin 10.8.1.

Everything appears to be working as expected, but the audio transcoding options still fail, so they will be added to a separate branch / PR in the future.

mueslimak3r:
I'll test and review tonight or tomorrow

mueslimak3r (changes requested):
Nice work!
The UI button looks nice.
Quality switching during playback works well.
The addition of `refreshStream()` and its use is good, and it mirrors how a stream is reloaded in a couple other places.
I'm not really qualified to review the kotlin code.

Other than some minor changes it looks good. 🙂

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoQualityController.kt
```diff
@@ -0,0 +1,55 @@
...
+
+import org.jellyfin.androidtv.preference.UserPreferences
+import org.jellyfin.androidtv.preference.UserPreferences.Companion.maxBitrate
+import org.koin.java.KoinJavaComponent.get
+
+class VideoQualityController(
+	private val parentController: PlaybackController
+) {
+
+	enum class QualityProfiles(val quality: String) {
```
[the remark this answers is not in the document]
zkhcohen: Done: https://github.com/jellyfin/jellyfin-androidtv/pull/1924/commits/5bbd58bf327c3d6d1b521d4bcb521a62e5343786

zkhcohen:
Suggestions implemented and tested. 

Clearly I don't understand GitHub PR diff behavior because my rebase has integrated changes into the diff which are already merged with master. It seems like it's using the common ancestor in the diff instead of resetting to the tip of the target branch rebased with.

If this is an issue, let me know how you're handling this.

mueslimak3r:
It's a little hard to tell exactly how it happened but it looks like you merged two copies of your branch together but they have different bases. Assuming the most recent merge commit is responsible, and that the branches are the same (so discarding the "merged-from" one won't lose any code, you could try checking out the commit prior to that and force-pushing. Definitely back up first though since that's just a guess.

If I were to guess what happened, it may be due to not syncing your remote and local branches at some point and then merging them. 🤷

I've attached screenshots from gitkraken that made it a little easier to visualize what's going on.

Screen Shot 2022-08-07 at 5 35 53 PM

zkhcohen:
> If I were to guess what happened, it may be due to not syncing your remote and local branches at some point and then merging them. 🤷

Yes, you're correct.

I fixed this issue and implemented some fixes suggested by the code scanner.

Re-tested after https://github.com/jellyfin/jellyfin-androidtv/pull/1943 was implemented and everything is working as expected.

This should be good to go, pending final approval.

mueslimak3r (approved):

app/src/main/res/values/strings.xml
```diff
@@ -190,6 +190,7 @@
     <string name="lbl_fit">Normal</string>
     <string name="lbl_audio_track">Select audio track</string>
     <string name="lbl_playback_speed">Playback Speed</string>
+    <string name="lbl_quality_profile">Quality Profile</string>
```
nielsvanvelzen, line 193:
```suggestion
    <string name="lbl_quality_profile">Quality profile</string>
```

app/src/main/res/drawable/ic_select_quality.xml
```diff
@@ -0,0 +1,14 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
nielsvanvelzen, line 1: This file is not consistent with the other icon files. See `ic_guide.xml` for example:

- pathData should be on a single line (Android Studio formatter should do this automatically)
- `tint` attribute needs to be added
- fill color should be `@android:color/white`

app/src/main/java/org/jellyfin/androidtv/constant/QualityProfiles.kt
```diff
@@ -0,0 +1,31 @@
+package org.jellyfin.androidtv.constant
+
+enum class QualityProfiles(val quality: String) {
```
nielsvanvelzen, line 3: I feel like this shouldn't be an enum, why not just a setOf() with all values?

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoQualityController.kt
```diff
@@ -0,0 +1,32 @@
+package org.jellyfin.androidtv.ui.playback
+
+import org.jellyfin.androidtv.preference.UserPreferences
+import org.jellyfin.androidtv.preference.UserPreferences.Companion.maxBitrate
+import org.koin.java.KoinJavaComponent.get
+import org.jellyfin.androidtv.constant.QualityProfiles
+
+class VideoQualityController(
+	@Suppress("unused")
+	private val parentController: PlaybackController
```
nielsvanvelzen, line 10: Member appears to be unused, why suppress that instead of removing it?

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoQualityController.kt
```diff
@@ -0,0 +1,32 @@
...
+) {
+	companion object {
+		private var previousQualitySelection = QualityProfiles.fromPreference(
+				get<UserPreferences>(UserPreferences::class.java)
```
nielsvanvelzen, line 14: Dependency should be injected via constructor

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoQualityController.kt
```diff
@@ -0,0 +1,32 @@
...
+						.get(maxBitrate))
+	}
+
+	var currentQuality = previousQualitySelection
+		@Suppress("unused")
```
nielsvanvelzen, line 19: Why suppress a unused fucntion instead of removing it?

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoQualityController.kt
zkhcohen: https://github.com/jellyfin/jellyfin-androidtv/pull/1924/files#diff-58f252ef3b690abba89c1a33e1c5bafe249c3e1dcb26aad7ad76795b9537fd07:~:text=menuItem.itemId%5D.quality)-,qualityController.currentQuality,-%3D%20QualityProfiles.fromPreference

app/src/main/java/org/jellyfin/androidtv/constant/QualityProfiles.kt
zkhcohen: Since the setter for the MaxBitrate property of the UserPreferences class only seems to be accessible via the "bind" method in the OptionsItemMutable abstract class, I needed to write a different method for setting the MaxBitrate in VideoQualityController. 

**Both methods need access to the same set of values when getting/setting the MaxBitrate.**

Excuse my inexperience with Kotlin, but I'm not sure if setOf() is appropriate in that situation?

I could see something similar to OptionsItemMutable being written for this new use-case: settings like MaxBitrate and Audio Transcoding options set both in the Playback UI _and_ the Playback Preferences Screen, but that's a bit beyond my experience or available time commitment at the moment. Deciphering the playback code has been difficult.

nielsvanvelzen: Copying the code in the current PlaybackPreferencesScreen to it's own file and assigning it to a variable should work for this purpose, like so:

```kotlin
val qualityOptions = setOf(
	0, // Disable
	3, 5, 7, // 10<
	10, 20, 30, 60, // 100<
	120, 300
).associate {
	val value = if (it == 0) getString(R.string.lbl_none)
	else TimeUtils.formatSeconds(context, it)
		it.toString() to value
}
```

And then just say `entries = qualityOptions` in the playbackpreferencesscreen and something similar in the video quality controller

zkhcohen: Hey - I see you requested a review for this. I'm on vacation for a few more days.

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoQualityController.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.jellyfin.androidtv.preference.UserPreferences;
+import org.koin.java.KoinJavaComponent;
+
+class VideoQualityController(
+	previousQualitySelection: QualityProfiles
+) {
+
+	var currentQuality = previousQualitySelection
+	set(value) {
+		KoinJavaComponent.get<UserPreferences>(UserPreferences::class.java)[UserPreferences.maxBitrate] =
```
nielsvanvelzen, line 13: inject user preferences via constructor to avoid KoinKavaComponent

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/action/SelectQualityAction.kt
```diff
@@ -0,0 +1,85 @@
...
+import org.koin.java.KoinJavaComponent.get
+
+class SelectQualityAction (
+	context: Context,
+	customPlaybackTransportControlGlue: CustomPlaybackTransportControlGlue,
+	playbackController: PlaybackController
+) : CustomAction(context, customPlaybackTransportControlGlue) {
+
+	private var previousQualitySelection = QualityProfiles.fromPreference(
+		get<UserPreferences>(UserPreferences::class.java)[maxBitrate]
```
nielsvanvelzen, line 25: inject user preferences via constructor to avoid KoinKavaComponent

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoQualityController.kt
zkhcohen: Done.

app/src/main/java/org/jellyfin/androidtv/constant/QualityProfiles.kt
zkhcohen: Done.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/action/SelectQualityAction.kt
zkhcohen: Done.

zkhcohen:
@nielsvanvelzen I can also consolidate the mapping transformation into the following class, but I would need some guidance on where to store it since it would no longer be a constant: 
studio64_olbDTTBZaR

Let me know if you think this is a better solution.

nielsvanvelzen:
Right, if the same mapping is needed in two places it makes sense to keep it in a class. Another option would be to make it a function that receives the context and returns the mapping, that might make more sense actually.

zkhcohen:
> Right, if the same mapping is needed in two places it makes sense to keep it in a class. Another option would be to make it a function that receives the context and returns the mapping, that might make more sense actually.

Okay, I updated it to be a top-level function. I still don't really think that storing it in org.jellyfin.androidtv.constant is necessarily appropriate, but without converting back to my original model of using an enum class and then moving the mapping function to a companion object, I'm not really sure where to put it.

Let me know if you want any further changes.

nielsvanvelzen (approved):
I've formatted the code to be more in line with our other Kotlin code and pushed that in the last commit. Everything looks fine now, thanks for the contribution!

Y0ngg4n:
@nielsvanvelzen for bitrated < 5mbit i always get a player error.

nielsvanvelzen:
> @nielsvanvelzen for bitrates < 5mbit i always get a player error.

Please open an issue so we can track it.

Y0ngg4n:
@nielsvanvelzen but the logs do not trow any errors :(

PR #1936, opened 2022-08-06
Remove item delete functionality

nielsvanvelzen:
The scope Android TV client is to USE the Jellyfin server, not to MANAGE it. While it makes sense for some stuff to have a delete function (like Live TV recordings) it doesn't for actual media. This PR removes all item deletion except for Live TV stuff.

**Changes**
- Remove item delete functionality

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsActivity.java
@@ -872,29 +872,6 @@ public class FullDetailsActivity extends BaseActivity implements RecordingIndica
         mActivity.startActivity(intent);
     }
 
-    private void deleteItem() {
-        new AlertDialog.Builder(mActivity)
-                .setTitle(R.string.lbl_delete)
-                .setMessage("This will PERMANENTLY DELETE " + mBaseItem.getName() + " from your library.  Are you VERY sure?")
-                .setPositiveButton("Delete",
-                        (dialog, whichButton) -> apiClient.getValue().DeleteItem(mBaseItem.getId(), new EmptyResponse() {
-                            @Override
-                            public void onResponse() {
-                                Utils.showToast(mActivity, mBaseItem.getName() + " Deleted");
-                                dataRefreshService.getValue().setLastDeletedItemId(mBaseItem.getId());
-                                finish();
-                            }
-
-                            @Override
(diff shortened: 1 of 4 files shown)


mueslimak3r (approved):

MagnuM2K:
Would it be possible to add back the option to delete, at least for the admin user.

I don't think that  removing it entirely was a good idea. It should be available to admin only accounts and also to be toggle on/off, so you can enable to only the admin, if needed.

Plex has the ability to delete files from the client for the admin user.

It makes more sense to delete a file from the client once a show/movie has been watched then to launch a separate device/pc to browse to the directory and then delete the file. This involves many more steps then it has to be.

Please reconsider.

wlau2929:
> Would it be possible to add back the option to delete, at least for the admin user.
> 
> I don't think that removing it entirely was a good idea. It should be available to admin only accounts and also to be toggle on/off, so you can enable to only the admin, if needed.
> 
> Plex has the ability to delete files from the client for the admin user.
> 
> It makes more sense to delete a file from the client once a show/movie has been watched then to launch a separate device/pc to browse to the directory and then delete the file. This involves many more steps then it has to be.
> 
> Please reconsider.

Yes, please. This is a step backwards. Most other clients/DVR's from cable companies and other media servers still have this. It is basic playback. I'm sure most of us have one admin user who can really manage the server (user settings, privileges, playback settings, updates) but media management is hardly server management. At least add an option to enable deletion from client. What's the point in giving users media deletion privileges if they can't delete the media?

Killerherts:
Can we get maybe a toggle or something to turn this back on? It's a bit of a pain to open another device when a find a messed up file. My old flow was literally just wait 3 mins and the arrs would pick up that I deleted and grab me a new one. Seems crazy I have to pull out a device to manage the file if I am administrator. 

Most other clients can delete it would also be nice to not have the android tv client missing features. We have one the web client as if that is the case. It's better to just cast from your phone so you don't lose features.

nielsvanvelzen:
Please open an issue and explain your use-case why this functionality should be kept. If there's enough demand we can consider adding it back.

Don't reply on old pull requests as those comments often end up in the void. I'm locking this issue to prevent further comments.

nielsvanvelzen:
This PR is linked in a few places, so a quick update. There is now an discussion for this feature: https://github.com/jellyfin/jellyfin-androidtv/discussions/2368. Please upvote the discussion and leave a reply explaining your use case so we can see the demand.

PR #1941, opened 2022-08-06
Support previous button in video playlist

### LINKED ISSUE #1940: No previous button in video playlist

There's a next button but no previous button

kevgrig:
Fixes #1940

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -251,6 +251,14 @@ public class PlaybackController implements PlaybackControllerNotifiable {
         return hasNextItem() ? mItems.get(mCurrentIndex + 1) : null;
     }
 
+    public boolean hasPreviousItem() {
+        return mItems != null && mCurrentIndex - 1 >= 0;
+    }
+
+    public BaseItemDto getPreviousItem() {
+        return hasPreviousItem() ? mItems.get(mCurrentIndex - 1) : null;
+    }
+
     public boolean isPlaying() {
         // since playbackController is so closely tied to videoManager, check if it is playing too since they can fall out of sync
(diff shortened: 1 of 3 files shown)


kevgrig:
@nielsvanvelzen fyi

mueslimak3r:
Does this work even after the current video completes and "next up" is shown?

There's some weird logic used rn that clears old items from the video queue after playback completes, and afaik it's there because of some quirk in how next up works.

Can you test letting a video complete/go to "next up" and see if the "previous" button still works/shows up?

Here's the code I mentioned:

`itemComplete()` -> `endPlayback()` -> `removePreviousQueueItems()`

https://github.com/jellyfin/jellyfin-androidtv/blob/ccb97ac1950a816c71e57a2ea08bbab8e6e36464/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java#L1389

kevgrig:
> Does this work even after the current video completes and "next up" is shown?

I forgot to test this as I had disabled the 'Next Up' function completely as I don't like it. I can confirm that it doesn't work if 'Next Up' is used. I'll check this out...

kevgrig:
> There's some weird logic used rn that clears old items from the video queue after playback completes, and afaik it's there because of some quirk in how next up works.

Would it be okay for now if I only enabled this logic if 'Next Up' is completely disabled?

mueslimak3r:
> Would it be okay for now if I only enabled the previous button if 'Next Up' is completely disabled?

In my opinion, a more sustainable approach would be to fix the quirk in the next up behavior that required clearing the old queue items. I can look into that in the next couple days.

That's just a suggestion for how we might move forward with getting it working. 🙂

mueslimak3r:
#1951 is merged so that should fix the issue with nextup breaking the previous button. Can you rebase and confirm things are good to go?

mueslimak3r (approved):
I tested again with the branch rebased and it works as expected 👍

PR #1939, opened 2022-08-06
Fix 'Play from Here' in video playlists

kevgrig:
Fixes #1712

Signed-off-by: Kevin G <kevin@myplaceonline.com>

PR #1943, opened 2022-08-08
Handle uninitialized current media position

kevgrig:
Likely fix for https://github.com/jellyfin/jellyfin-androidtv/pull/1939#issuecomment-1208485753

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
@@ -124,6 +124,7 @@ public class MediaManager {
             newMutableItems.add(items.get(i));
         }
         mCurrentVideoQueue = newMutableItems;
+        mCurrentMediaPosition = 0;
     }
 
     public List<BaseItemDto> getCurrentVideoQueue() { return mCurrentVideoQueue; }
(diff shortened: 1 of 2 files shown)


mueslimak3r (changes requested):
For this change there should be a catch for mItems being null before checking its size.

This check is good to have but also to resolve the crash I think the current position should be set in mediaManager in setCurrentVideoQueue(), which would resolve the cause of the crash.

https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java

kevgrig:
@mueslimak3r 

> For this change there should be a catch for mItems being null before checking its size.

Done, added the `null` check.

> This check is good to have but also to resolve the crash I think the current position should be set in mediaManager in setCurrentVideoQueue(), which would resolve the cause of the crash.

So always set `mCurrentMediaPosition = 0` when the current queue is updated?

mueslimak3r:
Yeah check if the queue has items in it and set it to 0 if it does.

When creating the queue it only adds items that are videos since it's possible that a playlist for instance will have mixed item types.

There is still an edge case where the queue won't have any items in it, and in that case the index should stay -1.

If the queue is empty the player should exit instead of setting the index to 0 in playbackController and attempting to access the first item when there isn't one.

Another option is to set the queue to null and keep the index as -1 instead if it being an empty list. That may the simplest solution

kevgrig:
@mueslimak3r Updated

nielsvanvelzen (approved):
I was able to reproduce the issue @zkhcohen reported and this PR fixes it for me 👍.

mueslimak3r (approved):
This should be good for now 👍

PR #1958, opened 2022-08-11
Allow app installation on external storage

### LINKED ISSUE #1955: Support installing to external storage

My Sony TV has a pretty paltry amount of internal storage available, it would be nice if this app could support installing to external storage.

Reading through https://developer.android.com/guide/topics/data/install-location seems to indicate it would be a fairly simple change to a manifest file (should just require adding `android:installLocation="auto"` to the right spot from what I can tell), but I'm not familiar with Android development at all so there may be some other gotchas. Compatibility looks like it should be fine, seems to date back to Android 2.2 (API level 8).

nielsvanvelzen:
Not thoroughly tested but I believe this should work without any problems.

**Changes**

- Allow app installation on external storage

**Issues**

Fixes #1955

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
Tested on my Shield by using a thumb drive as external storage. App works fine.

mueslimak3r (approved):

PR #1959, opened 2022-08-11
Migrate DefaultLifecycleObserver setup to androidx.startup

nielsvanvelzen:
This PR changes the way we create the DefaultLifecycleObserver that ends up injecting the activity callbacks (including the one that manages the theme). It now uses it's own initializer. This doesn't change any behavior, just moves the code to make it more managable.

**Changes**
- Migrate DefaultLifecycleObserver setup to androidx.startup

**Issues**

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -88,6 +88,9 @@
             <meta-data
                 android:name="org.jellyfin.androidtv.SessionInitializer"
                 android:value="androidx.startup" />
+            <meta-data
+                android:name="org.jellyfin.androidtv.ProcessLifecycleInitializer"
+                android:value="androidx.startup" />
         </provider>
 
         <provider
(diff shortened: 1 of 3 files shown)


Andy2244:
For me themes seem always broken, if i change the theme + restart the app nothing changes at all?
Can i get a screenshot on what is supposed to change, so i can maybe confirm this fix?

Andy2244:
@nielsvanvelzen btw i just noticed something, while checking full log again.

Why don't we extent from `AppCompatActivity` instead of `FragmentActivity` and also use `Theme.AppCompat.Leanback` instead of `Theme.Leanback`?
This combination should give the best compatibility and we still can use support fragments?

This also gets rid of those errors:
`View class org.jellyfin.androidtv.ui.AsyncImageView is an AppCompat widget that can only be used with a Theme.AppCompat theme (or descendant).`

nielsvanvelzen:
Please keep questions unrelated to a pull request in their own issue/discussion, and if related only questions about the code go in the pull requests if there's an open issue.

crobibero (approved):

PR #1966, opened 2022-08-13
Use AsyncImageView in ClockUserView

nielsvanvelzen:
We prefer AsyncImageView for displaying (networked) images since it allows us to easily set the fallback/blurhash and moves all Glide code to a single place. This PR migrates the ClockUserView.

**Changes**
- Support circle crop in AsyncImageView
- Use AsyncImageView in ClockUserView

**Issues**


nielsvanvelzen:
Just realized after creating the PR that I tested the wrong screen (home doesn't use the ClockUserView.. oops)

crobibero (approved):

thornbill (approved):

PR #1970, opened 2022-08-13
Refactor BrowseGridFragment hierarchy into a single class

nielsvanvelzen:
BrowseGridFragment was the only one using these two 

Before: `BrowseGridFragment extends StdGridFragment extends GridFragment extends Fragment`
After: `BrowseGridFragment Fragment`

Merging these classes removed a lot of boilerplate code and should make it easier to maintain it. Although the code is still quite messy. I've removed all the getter/setter functions that were only used in the class itself and merged some variables that were kept in both the grid and browsegrid fragments.

There are no behavior changes in this PR.

**Changes**
- Merge StdGridFragment into BrowseGridFragment
- Merge GridFragment into BrowseGridFragment

**Issues**


Andy2244:
Can this be merged into master, so i can directly use the cleaned-up classes.
I'm working on a few PR that all depend on this code, so would be really hard to merge.

mueslimak3r (approved):

PR #1979, opened 2022-08-14
Misc crash fixes

nielsvanvelzen:
Migrating code to the SDK can cause some issues

**Changes**
- Fix Live TV guide crashing
- Fix movie items crashing

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
@@ -55,6 +55,7 @@ import org.koin.java.KoinJavaComponent;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.List;
+import java.util.UUID;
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideActivity.java
```diff
@@ -622,6 +623,8 @@ private LinearLayout getProgramRow(List<BaseItemDto> programs, String channelId)
             int slot = 0;
             do {
                 BaseItemDto empty = new BaseItemDto();
+                empty.setId(UUID.randomUUID().toString());
+                empty.setType("FOLDER");
```
nielsvanvelzen, line 627: ^ this might look stupid (because it does) but legacy code uses BaseItemDto's for messaging between activities for some stupid reason. SDK's BaseItemDto requires a valid UUID and type set so we need to add those here.

thornbill (approved):

crobibero (approved):

PR #1983, opened 2022-08-16
Use constants from SDK

nielsvanvelzen:
This is mostly binary compatible so some files only show the import line changing.

**Changes**
- Replace legacy apiclient ItemSortBy, MediaType and CollectionType with SDK supplied constants
- Replace some hardcoded strings with the above mentioned constants
- Remove games collection type usage

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentLatestRow.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentLatestRow.kt
@@ -10,6 +10,7 @@ import org.jellyfin.androidtv.ui.presentation.CardPresenter
 import org.jellyfin.apiclient.model.querying.ItemFields
 import org.jellyfin.apiclient.model.querying.ItemsResult
 import org.jellyfin.apiclient.model.querying.LatestItemsQuery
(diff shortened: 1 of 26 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentLatestRow.kt
```diff
@@ -48,7 +49,12 @@ class HomeFragmentLatestRow(
 
 	companion object {
 		// Collections exclused from latest row based on app support and common sense
-		private val EXCLUDED_COLLECTION_TYPES = arrayOf("playlists", "livetv", "boxsets", "channels", "books")
+		private val EXCLUDED_COLLECTION_TYPES = arrayOf(
```
crobibero, line 52: Was "channels" purposefully removed?

crobibero (approved):

app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentLatestRow.kt
nielsvanvelzen: It doesn't exist in the server

thornbill (approved):

PR #1990, opened 2022-08-19
fix grid sort options

Andy2244:
**Changes**
Some of the grid sort options made no sense, since they all use descending name fall-backs, which just results in confusing ordering. I also added some more common options and special handling for shows.

- fix grid sort options
- add DateLastContentAdded option
- fix out of focus key handling
- rename "Rating" -> "Parental Rating"

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
@@ -64,6 +64,7 @@ import org.jellyfin.sdk.model.api.UserDto;
 import org.jellyfin.sdk.model.constant.ItemSortBy;
 import org.koin.java.KoinJavaComponent;
 
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.GregorianCalendar;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
```diff
@@ -835,14 +843,28 @@ public void addToParentIfResultsReceived() {
...
                 if (response.getItems() != null && response.getItems().length > 0) {
-                    setTotalItems(query.getEnableTotalRecordCount() ? response.getTotalRecordCount() : response.getItems().length);
+                    int totalItems = query.getEnableTotalRecordCount() ? response.getTotalRecordCount() : response.getItems().length;
+                    BaseItemDto[] items = response.getItems();
+                    if (filterPlayed) {
+                      items = Arrays.stream(items).filter(x ->
+                                (x.getUserData() == null) || (!x.getUserData().getPlayed() &&
+                                    x.getUserData().getUnplayedItemCount() != null &&
+                                    x.getUserData().getUnplayedItemCount() > 0)
+                        ).toArray(BaseItemDto[]::new);
```
nielsvanvelzen, line 861: We shouldn't do client side filtering because this is not efficient at all for large libraries.

Andy2244: Sure, but this is just a fix i can do now on the client, which was easy to implement.
I don't have the time to also get into the server and working on fixes there.
So until "someone" pick's-up the issue for the server, i rather have working filters for users.

nielsvanvelzen (commented):
Changes seem to work fine in my testing but for some reason the "new content" sorting is incorrect. It shows stuff I added months ago as the first entry and another series where I added an episode today is the fourth item.
Since it doesn't seem to work (on the server) I don't know if we should include that option.

Andy2244:
Need to retest this, since it seemed to work for me.

nielsvanvelzen:
I think the server isn't updating the `DateLastMediaAdded` field for the series item when an episode is added (probably only updates the season entry) and only updates it when a season is added.

Andy2244:
> I think the server isn't updating the `DateLastMediaAdded` field for the series item when an episode is added (probably only updates the season entry) and only updates it when a season is added.

For me it did update and shows that had just added new episodes, did jump to the front.

Andy2244:
Just retested and for me "new content" is working correctly, i see series that just got new episodes added first.
Whats your server version and OS? I'm on a simple windows portable 10.8.4 server.

PR #1991, opened 2022-08-19
fix missing server.version

Andy2244:
**Changes**
- fix missing server.version

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/auth/apiclient/ApiBinder.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/auth/apiclient/ApiBinder.kt
@@ -41,6 +41,7 @@ class ApiBinder(
 			address = server.address.removeSuffix("/")
 			userId = session.userId.toString()
 			accessToken = session.accessToken
+			version = server.version
 		})
 
 		return true


nielsvanvelzen:
I don't see the point of this change. We don't use the server version from the legacy apiclient so we don't need to set it either.

Andy2244:
I use the server version, so i can handle fixes for the 10.8.x servers.

Otherwise whats the point in having a server.version field that's never set?

nielsvanvelzen:
If you need to check a server version you'll need to get the server id from the current session in the SessionRepository and the version from the server in the ServerRepository. The legacy apiclient is deprecated and should not be used for any new code.

nielsvanvelzen:
See also

https://github.com/jellyfin/jellyfin-androidtv/blob/a8dea85a1b27ab993f247e25188ba4cb9d587e56/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java#L1500-L1517

Andy2244:
So i have to copy&paste all this boilerplate code just to get the server version?
Why is getServerVersionEqualOrGreater() not a static util/api call?

Than getServerInfo() should also be marked deprecated, if we should not use the serverInfo class.

nielsvanvelzen:
I just linked to some example code we temporarily used in 0.13. It's been removed since.

Anyway, closing this PR as we won't be using the old legacy apiclient anymore so this change doesn't do anything.

PR #2002, opened 2022-08-20
Migrate most of KeyProcessor to SDK types

nielsvanvelzen:
There's still some stuff with old apiclient types because they are used in other files and are a PITA to migrate right now.

**Changes**
- Migrate most of KeyProcessor to SDK types

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
@@ -21,16 +21,18 @@ import org.jellyfin.androidtv.ui.playback.AudioNowPlayingActivity;
 import org.jellyfin.androidtv.ui.playback.MediaManager;
 import org.jellyfin.androidtv.ui.playback.PlaybackLauncher;
 import org.jellyfin.androidtv.ui.shared.BaseActivity;
-import org.jellyfin.androidtv.util.apiclient.BaseItemUtils;
 import org.jellyfin.androidtv.util.apiclient.PlaybackHelper;
+import org.jellyfin.androidtv.util.sdk.BaseItemExtensionsKt;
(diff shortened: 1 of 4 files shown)


Andy2244:
Um can you also mark the old api client types as deprecated?

I'm going mad trying to figure out what import is actually valid and which was just auto added.
The issue is, that often its just upper vs lowercase and hard to spot.

nielsvanvelzen:
The legacy apiclient is unmaintained for over a year now and it's not possible to create new versions for it anymore. So no, I can't add those markers.

app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
```diff
@@ -291,17 +293,17 @@ public static PopupMenu createItemMenu(BaseRowItem rowItem, UserItemDataDto user
 
         //Not sure I like this but I either duplicate processing with in-line events or do this and
         // use a single event handler
-        mCurrentItem = item;
+        mCurrentItem = rowItem.getBaseItem();
         mCurrentRowItemNdx = rowItem.getIndex();
-        mCurrentItemId = item.getId();
+        mCurrentItemId = item.getId().toString();
```
crobibero, line 298: Why remove item from L294 but keep it here?

crobibero (approved):

app/src/main/java/org/jellyfin/androidtv/util/KeyProcessor.java
nielsvanvelzen: `rowItem.getBaseItem()` is the old BaseItemDto (from the apiclient) which we're migrating from. So I use the SDK one when I can, in this case I can't use it for mCurrentItem but can for the id.

PR #2038, opened 2022-09-06
Use random device id when initializing QuickConnect

nielsvanvelzen:
Temporary fix until @crobibero creates a new authentication controller (let's call it v2) from scratch, with proper JWT access tokens, refresh tokens and device info included in that access token instead of the stupid header stuff.

**Changes**

- Use random device id when initializing QuickConnect
 - This fixes the issue where existing sessions get destroyed by the server

**Issues**


crobibero (approved):
🖖

PR #2092, opened 2022-09-24
Fix NPE when live tv program is null

nielsvanvelzen:
**Changes**

**Issues**


thornbill (approved):
Looks good. Do you want me to test it on a device?

nielsvanvelzen:
I tested it myself, I have a few radio stations added as live tv so they never have programs :p

PR #2132, opened 2022-10-01
Rewrite navigation

nielsvanvelzen:
After years of work we're finally merging (most) activities into a single activity with proper navigation code.

Some parts are not migrated to the new navigation code:

- Picture viewer - will be migrated when fully replaced with new picture viewer
- Video player & Next Up - will be part of the playback rewrite
- Authentication - not sure how to migrate yet, the existing code for it works fine so it may just stay as-is

Some future changes include:

- ~~Re-enable display mirroring in the SocketListener~~
- ~~Update PlaybackLauncher to return destination instead of a "playbackActivityClass"~~

This PR introduces a major change to the code and it has some known issues:

- Fragments can crash the application if not using lifecycle scope
  - This happens for all legacy apiclient calls when they complete after the fragment is closed
- ~~The back stack from the FragmentManager is not used, this forces fragments to refresh when going back~~
  - ~~**I'm investigating this issue before marking the PR as ready**~~

**Changes**
- Merge most activities into MainActivity
- Add NavigationRepository which can be used to navigate the MainActivity
- Add Destination(s) used for navigating
- Add HomeFragment that combines HomeRowsFragment (previously HomeFragment) and HomeToolbarFragment
- Convert SearchActivity to SearchFragment
- Implement display mirroring

**Issues**


nielsvanvelzen:
I've fixed all issues I could find except for the lifecycle thing (will be done in separate future pull requests). There might still be issues I haven't found myself, it's quite a big change after all.

app/src/main/java/org/jellyfin/androidtv/data/eventhandling/SocketHandler.kt
```diff
@@ -37,6 +41,7 @@ class SocketHandler(
 	private val dataRefreshService: DataRefreshService,
```
thornbill, line 41: There seems to be something off with the alignment of these parameters. :slightly_smiling_face:

app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java
```diff
@@ -863,13 +864,7 @@ public void onClick(View v) {
...
             @Override
             public void onClick(View v) {
-                Intent settingsIntent = new Intent(getActivity(), PreferencesActivity.class);
-                settingsIntent.putExtra(PreferencesActivity.EXTRA_SCREEN, DisplayPreferencesScreen.class.getCanonicalName());
-                Bundle screenArgs = new Bundle();
-                screenArgs.putString(DisplayPreferencesScreen.ARG_PREFERENCES_ID, mFolder.getDisplayPreferencesId());
-                screenArgs.putBoolean(DisplayPreferencesScreen.ARG_ALLOW_VIEW_SELECTION, userViewsRepository.getValue().allowViewSelection(mFolder.getCollectionType()));
-                settingsIntent.putExtra(PreferencesActivity.EXTRA_SCREEN_ARGS, screenArgs);
-                requireActivity().startActivity(settingsIntent);
+                navigationRepository.getValue().navigate(Destinations.INSTANCE.displayPreferences(mFolder.getDisplayPreferencesId(), userViewsRepository.getValue().allowViewSelection(mFolder.getCollectionType())));
```
thornbill, line 867: Not sure what the actual line length is here, but it seems a bit long for me to parse easily.

thornbill (approved):

PR #2209, opened 2022-10-25
Add playback rewrite queue system

nielsvanvelzen:
There we go, the first real code for the playback rewrite! This code was relatively stable (didn't see much changes in my occasional refactoring) so it's good enough to PR upstream.

**Changes**

**playback.core**

- Add Queue interface
  A queue contains the items to play, simple as that. Queues are always lazy loaded to allow for big queues or changing future items on demand.

- Add EmptyQueue implementation
  It's a queue that is empty.

- Add PagedQueue implementation
  This one is kinda cool, it loads (sequential) pages of items and cache them. So when the player wants to know the next queue item it might have (pre-)loaded already so no network request is needed. Most of the Jellyfin specific queue implementations use this one as base.

- Add SequenceQueue implementation
  Like PagedQueue but with a sequence, I used this before creating the PagedQueue. Currently unused but might still have some use cases in the future.

- Add initial QueueManager
  It's a helper to work with queues, this one will probably change a bunch in upcoming PR's.

- Add QueueEntry
  It's an item in the Queue. This interface contains properties that are used by all playback components. Including the QueueEntryMetadata which is a container for all kinds of properties used for mediasessions, media notifications, player metadata, logging etc. etc.

**playback.jellyfin**

- Add BaseItemDtoUserQueueEntry
  Implementation for QueueEntry that contains Jellyfin specific data (like BaseItemDto). It contains a builder function that is used by the Jellyfin specific queue implementations.

- Add AudioAlbumQueue, AudioInstantMixQueue, AudioTrackQueue
  Music specific implementations for the player. As I said before, focus is currently on getting music to work perfect before working on video. These three queues should be sufficient for most music playing scenarios. There might still be some more like an ArtistQueue later.

- Add EpisodeQueue
  Focus is on music but video playback already works.. locally. This one is to play episodes in a season.

**Issues**

Part of #1057

### DIFF
--- a/playback/core/src/main/kotlin/queue/item/QueueEntry.kt
+++ b/playback/core/src/main/kotlin/queue/item/QueueEntry.kt
@@ -1,20 +1,36 @@
 package org.jellyfin.playback.core.queue.item
 
-// Queue Items
-sealed interface QueueEntry {
+/**
+ * The QueueEntry is a single entry in a queue. It can represent any supported media type.
+ * Implementations normally extend one of it's subclasses [BrandedQueueEntry] or [UserQueueEntry].
+ */
+interface QueueEntry {
+	/**
+	 * Whether this entry can be skipped and seeked or not. Normally set to `true`.
+	 */
 	val skippable: Boolean
-	val metadata: QueueEntryMetadata?
+
+	/**
+	 * The metadata for this item.
+	 */
+	val metadata: QueueEntryMetadata
 }
 
 /**
- * Branded queue items are used for intros, trailers, advertisements and other related videos
- * It is not possible to seek or skip these items and are normally invisible when showing a queue's contents
+ * Branded queue entries are used for intros, trailers, advertisements and other related videos. It
+ * is not possible to seek or skip these items and are normally invisible when showing a queue's
+ * contents.
  */
-abstract class BrandedQueueEntry : QueueEntry {
+open class BrandedQueueEntry : QueueEntry {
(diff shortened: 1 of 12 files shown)


playback/core/src/main/kotlin/queue/item/QueueEntry.kt
```diff
@@ -0,0 +1,20 @@
...
+	val metadata: QueueEntryMetadata?
+}
+
+/**
+ * Branded queue items are used for intros, trailers, advertisements and other related videos
+ * It is not possible to seek or skip these items and are normally invisible when showing a queue's contents
+ */
+abstract class BrandedQueueEntry : QueueEntry {
+	override val skippable = false
+}
```
thornbill, line 15: No default `metadata` override?

nielsvanvelzen: This looks like a leftover from an earlier queue concept. We'll probably set the skippable property in the BaseItemDtoUserQueueEntry based on it's BaseItemDto.Type value in the future and remove UserQueueEntry and BrandedQueueEntry entirely.

I'll keep it like this for now and fix change it in an upcoming PR.

nielsvanvelzen: BrandedQueueEntry and UserQueueEntry now behave the same. I've also added some more comments and made some small adjustments.

thornbill (approved):

PR #2226, opened 2022-10-30
Show a Resume from button when series has an episode that can be resumed

sparky3387:
Updated FullDetailsFragment.java to show a resume from button when the series has a resumable episode

**Changes**

**Issues**


nielsvanvelzen:
Thanks for your work on the app. Unfortunately we're not accepting new code that uses the old apiclient code because we're migrating everything to our [Kotlin SDK](https://github.com/jellyfin/jellyfin-sdk-kotlin). Because the FullDetailsFragment is still written in Java and really big it's not easy to migrate it to Kotlin (and thus the SDK) so changing this PR to use the SDK will be a bit of work. The SDK doesn't work well with Java code so it required a bit of work to get going. 

My workflow would be to create a Kotlin file that contains functions for the required API calls. Those functions then use a callback to send the responses back to the Java code.

nielsvanvelzen:
Closing as inactive

PR #2239, opened 2022-11-04
Apply subtitle delay to external subs

### LINKED ISSUE #2200: Subtitle offset doesn't work on libvlc playback
### Describe the bug

I'm using Jellyfin for android TV 0.14.4 and enabled libvlc player
When playing a video with subtitles (in my case .srt files), offseting the subtitles throught the GUI has no effect (even though it seems like the playback is frozen for a fraction of second when changing the value so something seems to happen, but no subtitle offset

### Logs

```shell
I tried adb but did not find anything. I'm not sure how to use it correctly to extract jellyfin logs
```

### Application version

0.14.4

Amazon Appstore

### Device information

Amazon FireTV

### Android version

fireos 5.2.9.1

### Jellyfin server version

10.8.1

### Which video player implementations does this bug apply to?

- [ ] Web player (default)
- [ ] Integrated player (ExoPlayer)
- [ ] External player (VLC, mpv, MX Player)

DenizenB:
**Changes**
Subtract the subtitle delay from the playback position that is used to calculate when to show/hide external subtitles.

I have tested changing the subtitle delay on:
* a video with embedded subtitles
* a video with external subtitles

Both of them behaved as expected in my limited testing.

**Issues**
Fixes #2200

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -1532,7 +1532,7 @@ public class PlaybackController implements PlaybackControllerNotifiable {
                 updateTvProgramInfo();
             }
             if (mFragment != null && finishedInitialSeek)
-                mFragment.updateSubtitles(mCurrentPosition);
+                mFragment.updateSubtitles(mCurrentPosition - getSubtitleDelay());
         }
         if (mFragment != null)
             mFragment.setCurrentTime(mCurrentPosition);


mueslimak3r:
Is it correct that subtitle delay is currently only used for libVLC playback? If that's the case then the delay should only be applied when libVLC is used, and not with exoplayer.

This change may be all that's needed for exoplayer to use subtitle delay though 🤷‍♂️

DenizenB:
> Is it correct that subtitle delay is currently only used for libVLC playback? If that's the case then the delay should only be applied when libVLC is used, and not with exoplayer.

That's how I understand the code, yeah. `VideoManager.setSubtitleDelay` checks if libVLC is used and then calls `org.videolan.libvlc.MediaPlayer.setSpuDelay`.

I don't think libVLC has any knowledge of external subtitles. Even so, the delay for both embedded and external subtitles is currently stored inside `org.videolan.libvlc.MediaPlayer`.

When Exoplayer is used, `VideoManager.setSubtitleDelay` does nothing and `VideoManager.getSubtitleDelay` returns 0.

Perhaps it would make sense to store the subtitle delay in `VideoManager` if:
- Exoplayer is used, or
- libVLC is used with external subtitles

I think that would also allow Exoplayer to support subtitle delays as soon as those buttons are added to the UI. But I'm very new to this codebase, would be cool if someone more experienced than me could weigh in here. :)

nielsvanvelzen (approved):

PR #2245, opened 2022-11-07
Add Video Codec Tag

### LINKED ISSUE #2244: Update existing feature

On the item view page next to the date has tags based on the media file [1080] [cc] [5.1] and so on. Can we add AV1 to that list of tags (If already there, extend to 6 tags and give priority to media codec) and would it be possible to make the 4K tag gold? this would make it stand out. I got lost in the code trying to look for the right area.

[AV1] [4K] [ACC] [EAC3] [ [CC] [MA 15+] (media file that's AV1 with 2 audio tracks and is 4K)

@nielsvanvelzen  You'll be best person to ask about this since you've built most of this app. No rush.

The-Randalorian:
Add Video Codec Tag to Item View Page

**Changes**
Adds an additional tag to Created `addVideoCodecDetails` function to InfoLayoutHelper.java. The function is called by `addRatingAndRes` in the same file to add a tag with the codec name after the resolution tag.

**Issues**
Fixes  #2244

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
@@ -345,7 +345,10 @@ public class InfoLayoutHelper {
                 addBlockText(context, layout, "8K");
             }
 
-            addSpacer(context, layout, "  ");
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (approved):
Works as expected, thanks!

PR #2256, opened 2022-11-13
Improved server unavailability handling during startup

AndreasGB:
**Changes**
This PR changes large parts of the SessionRepository to make its current session an Arrow `Either<A,B>` of either an Error or a valid Session. This can then be used by other components to know why the session is currently unset and allows them to give the user more information on the error. The StartupActivity as well as a modified SplashScreen use this additional information to show the user an information when the server is unavailable.

This draft is in many locations still rough and unfinished (TODOs, ugly draft UI, etc.), but I'm submitting it early to get feedback on whether it's okay to introduce Arrow and the use of its `Either<>` type.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -117,6 +117,10 @@ dependencies {
 	implementation(libs.kotlinx.coroutines)
 	implementation(libs.kotlinx.serialization.json)
 
+	// Arrow
+	implementation(platform("io.arrow-kt:arrow-stack:1.1.2"))
+	implementation("io.arrow-kt:arrow-core")
+
 	// Android(x)
 	implementation(libs.androidx.core)
 	implementation(libs.androidx.activity)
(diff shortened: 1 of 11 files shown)


nielsvanvelzen:
Kotlin already has [`Result<T>`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/) which can be `T` or an exception. If we want the failure type to be something else we can easily write our own `Result<T, E>` class, no need to add a big dependency for that IMO.

Also, you should not use Jetpack Compose for interactive elements (like buttons). It doesn't support focusing properly yet (but Google is working on it).  We only use compose for non interactive UI.

AndreasGB:
I avoided Kotlin's `Result<T>` because it obscures the types of errors that may need to be handled by client code with its acceptance of any `Throwable` for failure, but I agree, adding a full dependency to get `Either` may be overkill. I could also move to e.g. a `SessionResult` sealed class that contains a `Success` data class as well as data classes for each error, would this be preferrable?

To avoid the Jetpack Compose usage, should I build a `ReconnectFragment` using `ConstraintLayout` in addition to the current `SplashScreen` or a single `ConstraintLayout`-based fragment for both?

Aside from these changes, do you think this PR is moving in the right direction design-wise?

nielsvanvelzen:
When the server doesn't respond (and the user won't be authenticated) the login screens opens. We can probably just query the server in that screen (which I believe we already do) and show an error if that specific request fails. That way we get the same behavior but the implementation is much simpler. All session handling code is already complex enough.

nielsvanvelzen:
Closing this one for now as inactive, please re-open (or create a new one) when you pick it up again.

PR #2297, opened 2022-11-27
Remember button focus on summary page

teobaranga:
Fixes remembering the last selected button so that it's reselected when coming back to the buttons row. As far as I can tell from other places in the app and other TV apps in general, this seems to be standard behaviour.

Side note: big fan of Jellyfin, keep up the good work 👍

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/DetailRowView.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/DetailRowView.kt
@@ -13,4 +13,18 @@ class DetailRowView @JvmOverloads constructor(
 	defStyleRes: Int = 0,
 ) : FrameLayout(context, attrs, defStyleAttr, defStyleRes) {
 	val binding = ViewRowDetailsBinding.inflate(LayoutInflater.from(context), this, true)
+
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
This change doesn't appear to do anything for me. I tested it by opening an episode and trying both the "previous episode" button and the "go to series" button and then going back with the system back button again. In both cases the play button (first in the row) was focused.

teobaranga:
Ahh sorry, I wasn't clear in the description. I meant that the button is remembered when navigating up or down on the same screen with the buttons. Eg. when you're on the series details screen and you navigate down to select a season, when coming back up to the buttons row, the last selected button should be reselected. Without these changes, the middle button in that row is always reselected for some reason. Hope that clears it up!

nielsvanvelzen:
Right, I understand it now. It works as expected but I think the the change should be moved to DetailRowView instead since so it manages it's own layout.

teobaranga:
That makes sense, thanks, updated 👍

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks for the PR!

PR #2308, opened 2022-12-04
Fix LegacyImageCardView scaling on non-1080p devices

nielsvanvelzen:
Found this one a few days ago when I created a new emulator to test something on an old Android TV version and I chose a 720p display instead of 1080p.

**Changes**

Properly scale LegacyCardView with the screen density instead of hardcoding 2x. This fixes the scaling for pretty much all cards in app to be a consistent size across different display sizes. Previously only devices that rendered UI at 1080p (like most 4k/1080p devices) showed as intended.

AFAIk this bug was present in all previous app versions and actually makes the app unusable on 720p devices because the cards are way to big. A single episode card in the home screen would cover about 85% of the display.

**Issues**


nielsvanvelzen:
Small example viewing a collection of 3 movies on the 720p emulator. And this one isn't even that bad because it's only 3 items and they are poster sized.

## Before
image

## After
image

thornbill (approved):

Issue #2314, opened 2022-12-06
5.1 surround sound played as stereo in LibVLC


FerronN:
### Describe the bug

I'm using Jellyfin on my Google TV. When I use the 5.1 (Dolby surround or Dolby Digital) sound track it is played as stereo with LibVLC. With Exoplayer it works just fine, but Exoplayer stutters like hell with h265 content. My Android TV has the same problem.

### Application version

0.14.5

### Device information

Google TV

### Android version

12

### Jellyfin server version

10.8.8

FerronN:
I've deployed a debug version of Jellyfin to my Google TV. It shows the same results as the store version. The weird thing is that surround works for Exoplayer but only when the sound is set to "Mix to stereo". Where I would expect this would mix it to stereo and with direct it would do passthrough. 

Jellyfin uses the libvlcjni library for it's VLC player. I've opened an [issue](https://code.videolan.org/videolan/libvlcjni/-/issues/2) there to see if someone has a clue to what can cause this problem to occur.

PR #2318, opened 2022-12-06
#2314 surround works with LibVLC 

FerronN:
Seems like mentioned in a earlier issue (https://github.com/jellyfin/jellyfin-androidtv/issues/786) that specifying 'setAudioOutput' prevents passtrough/auto codec selection. With this line gone 5.1 works with LibVLC in audio direct modus.

I don't know what the line I deleted was supposed to do, but it works great without. Maybe someone else knows what this was supposed to do.


FerronN:
By default it is direct, but setting it to down mix stereo doesn't make a difference.

FerronN:
That's true. Then I'm going to test again, because the debug version works like it should on my android tv, but de store version didn't, but I'm going to double check.

FerronN:
I tested it again on my Google TV and it doesn't work in both modes. I will continue this in the issue #2314

PR #2336, opened 2022-12-10
Revert "Remember button focus on summary page"

nielsvanvelzen:
Reverts jellyfin/jellyfin-androidtv#2297

---

There are two issues with the PR:

1. The global listener isn't removed when the view detaches, this is an easy fix. just store it in a var and add/remove in onAttachedToWindow/onDetachedFromWindow
2. The call `requestFocus()` somehow crashes the app when viewing a person opened via a movie. I have no clue why and I tried to fix it for about 3 hours after giving up. **This is the reason for the revert.**


nielsvanvelzen:
Hey @teobaranga unfortunately I'll need to revert your PR due to the mentioned issue. I'm trying to stabilize 0.15 for release and I was unable to fix it myself. If you're able to fix it you can recreate the original PR with the fix and we'll merge it.

teobaranga (approved):
All good 👍 I'm looking into this and will open a new PR once I have a proper fix, sorry about that

crobibero (approved):

PR #2345, opened 2022-12-14
Remember button focus on details screen

teobaranga:
**Changes**
Reworked #2297 in way that does not cause regressions.

The previous implementation added a global focus listener that worked even if the buttons row was offscreen. The linked bug happened because a "last button" was stored when navigating to the movie screen and then when navigating to the person screen, that same button would request focus even if was offscreen. Detaching the listener when navigating between screens was not sufficient because it would always be called too late, ie. after the incorrect focus request happened. TIL global focus listener is very dangerous.

The rework involves a targeted focus listener for each button which does not operate while offscreen. The last focused button is only stored when navigating away from the buttons row, and only restored when coming back to that same row.

**Issues**
N/A


nielsvanvelzen (approved):

nielsvanvelzen:
Thanks for the quick rewrite, I wasn't able to reproduce the previous issue and everything works as expected!

Issue #2366, opened 2022-12-26
[bug] Movies with FLAC audio are not played.


didaktiik:
### Describe the bug

I keep two versions of the same movie, one for local playback, second for remote playback (aac sound, restricted bitrate etc).
Files for local playback usually have a FLAC audio track. They used play fine (and they still do outside Jellyfin ATV app, like Kodi+JellyCon will play home version fine).

Alien Director's Cut (1979)/Alien Director's Cut (1979) - Home.mkv (FLAC audio)
Alien Director's Cut (1979)/Alien Director's Cut (1979) - Internet.mp4 (AAC audio)

At home, "Home.mkv" should play,  but "Internet.mp4" is played instead.  I have checked other movies with FLAC, and they won't play either. I mean they will play, but instead of playing "home" version, the other (more compatible) versions will be played.

I can see AudioCodecNotSupported in the log and I think it is a bug.

### Logs

[2022-12-26 12:17:35.527 +01:00] [INF] [145] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "192.168.1.1" request
[2022-12-26 12:18:27.392 +01:00] [INF] [121] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "192.168.1.87" request
[2022-12-26 12:18:35.454 +01:00] [INF] [96] Jellyfin.Api.Helpers.MediaInfoHelper: User policy for "didaktiik". EnablePlaybackRemuxing: True EnableVideoPlaybackTranscoding: True EnableAudioPlaybackTranscoding: True
[2022-12-26 12:18:35.455 +01:00] [INF] [96] Jellyfin.Api.Helpers.MediaInfoHelper: StreamBuilder.BuildVideoItem( Profile="AndroidTV-libVLC", Path="/mnt/NAS414/Plex/Movies/Alien Director's Cut (1979)/Alien Director's Cut (1979) - Home.mkv", AudioStreamIndex=null, SubtitleStreamIndex=null ) => ( PlayMethod=DirectPlay, TranscodeReason=0 ) "media:/videos/3444458c-6a59-4812-4bd9-aadf46fc8405/stream.mkv?MediaSourceId=3444458c6a5948124bd9aadf46fc8405&Static=true&VideoCodec=h264&AudioCodec=flac&AudioStreamIndex=1&api_key=<token>&Tag=e56e30f4ad14287f30edcf86a63b57e1"
[...]

PR #2376, opened 2022-12-30
Allow Opus, FLAC audio in video direct play (#2366, #2359)

### LINKED ISSUE #2359: OPUS Support in ExoPlayer

OPUS seems like the best way to compress audio while maintaining quality but when playing OPUS content it transcodes to ac3

Also OPUS is supported in most other clients so it would be nice to just encode to one audio codec for everything

Clubfan22:
**Changes**

I added the Opus and FLAC codecs to the supported audio codecs for directly playable videos.
According to the [Android documentation](https://developer.android.com/guide/topics/media/media-formats#core), Opus decoding is natively supported since Android 5.0, the minimum required version of this app; FLAC since 3.1.

**Issues**

Fixes #2359 #2366

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
@@ -50,7 +50,8 @@ class ExoPlayerProfile(
 		Codec.Audio.TRUEHD,
 		Codec.Audio.PCM_ALAW,
 		Codec.Audio.PCM_MULAW,
-		Codec.Audio.OPUS
+		Codec.Audio.OPUS,
+		Codec.Audio.FLAC
 	)
 
 	init {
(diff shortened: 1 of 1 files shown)


Clubfan22:
I am not entirely sure if adding Opus to ```downmixSupportedAudioCodecs``` would be useful.
As far as I understand it, downmixing is used to play surround sound on stereo speakers or headphones.

thornbill:
We also need to specifically check for Amazon device compatibility as that does not always match what Android supports. Actually I see several older devices there that do not list opus support so that is likely already not working. 

https://www.developer.amazon.com/docs/fire-tv/device-specifications-fire-tv-streaming-media-player.html

Clubfan22:
Thanks, that's a good point. I guess we could ensure Opus support by enabling the corresponding decoder in the ffmpeg extension. I already have a local build doing that, for which I could create a pull request.
Still, it would probably cleaner to build the player profile based on the codecs supported by the device.
If that is a desired feature and could be used by the new playback code in the future, I would try to implement that.

VampiricAlien:
Could we also add Vorbis mkv support since that support android 4+ (mp4 requires Android 10+)

nielsvanvelzen:
@Clubfan22 can you rebase the PR so we can review and potentially merge?

nielsvanvelzen:
Closing as stale, please re-create if you plan to fix the conflict.

PR #2380, opened 2023-01-02
Support fast forward and rewind in audio player

### LINKED ISSUE #2379: Pressing FireTV Fast Forward button goes to next song instead of seeking
### Describe the bug

1. Jellyfin Web } Hamburger } Dashboard
1. Libraries } Add Media Library
1. Content type: Music
1. Add folder with multiple MP3s
1. Ok
1. Jellyfin Android TV } Select Media Library
1. Select first MP3 under Latest and start playing it
1. Click the Fast Forward button
1. Jellyfin advanced to the next MP3 in the library instead of fast forwarding in the currently playing MP3

Video:

[jellyfin_ff.webm](https://user-images.githubusercontent.com/6432587/210183916-ac915e3b-5472-49e6-9264-9a67d7e8c331.webm)

Button pressed:

firetvremote_ff

### Logs

[seek.log](https://github.com/jellyfin/jellyfin-androidtv/files/10329419/seek.log)

### Application version

0.15.z

Sideloaded APK

### Device information

Amazon Fire TV, Fire OS 5.2.7.4

### Android version

5.1.1

### Jellyfin server version

release-10.8.z

kevgrig:
Fixes #2379

I don't know the semantics of `KEYCODE_DPAD_RIGHT` and `KEYCODE_DPAD_LEFT` so I didn't change them.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
@@ -27,6 +27,7 @@ import org.jellyfin.androidtv.constant.QueryType;
 import org.jellyfin.androidtv.data.compat.AudioOptions;
 import org.jellyfin.androidtv.data.compat.StreamInfo;
 import org.jellyfin.androidtv.data.model.DataRefreshService;
+import org.jellyfin.androidtv.preference.UserSettingPreferences;
 import org.jellyfin.androidtv.ui.itemhandling.AudioQueueItem;
 import org.jellyfin.androidtv.ui.itemhandling.BaseRowItem;
 import org.jellyfin.androidtv.ui.itemhandling.ItemRowAdapter;
@@ -967,7 +968,6 @@ public class MediaManager {
         }
     }
 
-
     public void resumeAudio() {
         if (mCurrentAudioItem != null && getIsAudioPlayerInitialized()) {
             ensureAudioFocus();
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
```diff
@@ -985,6 +985,34 @@ public void resumeAudio() {
         }
     }
 
+    public void fastForward() {
+        UserSettingPreferences prefs = KoinJavaComponent.<UserSettingPreferences>get(UserSettingPreferences.class);
```
nielsvanvelzen, line 989: Injection can be moved to a class property (with inject instead of get)

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java
kevgrig: I'm not very familiar with the code base or what CDI is used - do you happen to have an example?

nielsvanvelzen: Just like line 100 in the same file, add a property for the UserSettingPreferences and then use that property to read the preference from.

https://github.com/jellyfin/jellyfin-androidtv/blob/3ad59c8a4de1805f618fbd5c72ae4fd827180fdb/app/src/main/java/org/jellyfin/androidtv/ui/playback/MediaManager.java#L100

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks for the PR!

PR #2381, opened 2023-01-02
Allow all supported audio codecs in ExoPlayer transcoding profile

nielsvanvelzen:
I've only tested this one a single video item that needed transcoding for subs but also transcoded the audio. Not sure why the transcoding profile didn't allow all audio codecs.

**Changes**
- Allow all supported audio codecs in ExoPlayer transcoding profile

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
@@ -39,18 +39,19 @@ class ExoPlayerProfile(
 	 * Returns all audio codecs used commonly in video containers.
 	 * This does not include containers / codecs found in audio files
 	 */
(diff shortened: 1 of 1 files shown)


thornbill:
👀 I wonder if this is the cause of all the reports of unnecessary audio transcoding...

nielsvanvelzen:
I can confirm it fixes that issue. Previously the dashboard said the transcode reason was because of audio, now it says it's because of the subtitles. So far everything seems to work fine for me. But I'm still not sure if it fits for 0.15.1 or if we should put it back to 0.16.0.

nielsvanvelzen:
After tinkering about it for a bit I think we should include this in 0.15.z. I don't expect new issues from this change and we can always revert if a bug appears.

thornbill (approved):
Untested but lgtm!

VampiricAlien:
https://exoplayer.dev/supported-formats.html
Does the app include FFmpeg extension?

PR #2392, opened 2023-01-08
Add AV1 direct play support.

MWisBest:
**Changes**
This adds the ability for ExoPlayer to play AV1 without transcoding on API level 29 and up.

**Issues**
There's a few issues that reference AV1 support, but nothing really specific to ExoPlayer. LibVLC still appears to just black screen attempting to play AV1, but it doesn't appear to be related to the profiles stuff that was modified here.

There may be more/better ways of doing this, in theory ExoPlayer supports AV1 at even lower API levels, this might not correctly detect/advertise 10-bit support, but the only device I have available to test this on works correctly. This PR is bringing attention to this missing feature and to get the ball rolling on it.

Tested on: Chromecast with Google TV **HD** (which has hardware AV1 support; Chromecast with Google TV _4K_ does not do hardware AV1 but it might do software?).

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
@@ -3,10 +3,21 @@ package org.jellyfin.androidtv.util.profile
 import android.media.MediaCodecInfo.CodecProfileLevel
 import android.media.MediaCodecList
 import android.media.MediaFormat
+import android.os.Build
 import timber.log.Timber
 
 class MediaCodecCapabilitiesTest {
 	private val mediaCodecList by lazy { MediaCodecList(MediaCodecList.REGULAR_CODECS) }
+
+	fun supportsAV1(): Boolean = Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q &&
+		hasCodecForMime(MediaFormat.MIMETYPE_VIDEO_AV1)
+
+	fun supportsAV1Main10(): Boolean = Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q &&
+		hasDecoder(
+			MediaFormat.MIMETYPE_VIDEO_AV1,
+			CodecProfileLevel.AV1ProfileMain10,
+			CodecProfileLevel.AV1Level5
+		)
 
 	fun supportsHevc(): Boolean = hasCodecForMime(MediaFormat.MIMETYPE_VIDEO_HEVC)
 
--- a/app/src/main/java/org/jellyfin/androidtv/constant/Codec.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/constant/Codec.kt
@@ -62,6 +62,7 @@ object Codec {
 		const val MPEG2VIDEO = "mpeg2video"
 		const val VP8 = "vp8"
 		const val VP9 = "vp9"
+		const val AV1 = "av1"
 	}
 
 	object Subtitle {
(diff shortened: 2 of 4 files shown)


Kab1r:
Many of my AV1 files now plays, but I am having issues paired with certain audio codecs. I haven't tested these files paired with H264 instead of AV1 yet, so it might not necessarily be AV1 related. I haven't had any trouble with audio codec incompatibility on any of my H264 files before.

This one gives up after a few tries (logs indicate incompatible audio codec?)
```
Audio
Title: Eng - AAC - 6 ch - Default
Language: eng
Codec: AAC
Codec tag: mp4a
Profile: LC
Channels: 6 ch
Bitrate: 384 kbps
Sample rate: 48000 Hz
Default: Yes
Forced: No
External: No
```
But this one plays fine
```
Audio
Title: Eng - AAC - 5.1 - Default
Language: eng
Codec: AAC
Codec tag: mp4a
Profile: LC
Layout: 5.1
Channels: 6 ch
Bitrate: 341 kbps
Sample rate: 96000 Hz
Default: Yes
Forced: No
External: No
```
DTS-HD fails differently: instead of the codec failing, it lags out (maybe because of the high bitrate?)
```
Audio
Title: DTS-HD MA 5.1 - Eng - Default
Language: eng
Codec: DTS
Profile: DTS-HD MA
Layout: 5.1
Channels: 6 ch
Bitrate: 2163 kbps
Sample rate: 48000 Hz
Bit depth: 16 bit
Default: Yes
Forced: No
External: No
```
My AC3 file appears to play fine too.
```
Audio
Title: Eng - Dolby Digital - 5.1 - Default
Language: eng
Codec: AC3
Layout: 5.1
Channels: 6 ch
Bitrate: 384 kbps
Sample rate: 48000 Hz
Default: Yes
Forced: No
External: No
```
If you can, see if you run into similar issues with a variety of audio codecs.

MWisBest:
I haven't had any issues with standard DTS, I will try a DTS HD-MA track when I get a chance. I can tell you that the Chromecasts really do not like high bitrate content, which is why I looked into this in the first place.

Couple things I noticed is your first AAC track lists 6 channels but doesn't have the 5.1 'layout' info the AAC track that works lists, and that DTS HD-MA track is higher bitrate than any of the ones I have in my collection.

I think I'd read a while back that when things were getting transcoded the Android TV build would have both video and audio transcoded when only one or the other was needed, so I'd go back to a build without the AV1 patch and see if that DTS-HD MA file was actually playing as DTS-HD MA or if it was getting dropped to AC3 or something.

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
```diff
@@ -3,11 +3,23 @@ package org.jellyfin.androidtv.util.profile
...
 
+	@RequiresApi(Build.VERSION_CODES.Q)
+	fun supportsAV1(): Boolean = hasCodecForMime(MediaFormat.MIMETYPE_VIDEO_AV1)
+
+	@RequiresApi(Build.VERSION_CODES.Q)
+	fun supportsAV1Main10(): Boolean = hasDecoder(
+		MediaFormat.MIMETYPE_VIDEO_AV1,
+		CodecProfileLevel.AV1ProfileMain10,
+		CodecProfileLevel.AV1Level5
+	)
```
nielsvanvelzen, line 21: I'd prefer it if those functions checked the API version instead of the calling side. Both can just return `false` for anything below Android Q

Kab1r:
> I noticed is your first AAC track lists 6 channels but doesn't have the 5.1 'layout' info the AAC track that works lists

I can confirm that this file fails to play in exactly the same way on the latest release.

> I think I'd read a while back that when things were getting transcoded the Android TV build would have both video and audio transcoded when only one or the other was needed, so I'd go back to a build without the AV1 patch and see if that DTS-HD MA file was actually playing as DTS-HD MA or if it was getting dropped to AC3 or something.

I am unable to get the file to transcode. On the latest release, this files fails similarly to the AAC 6 channel file: it gives up after trying a couple of times.

MWisBest:
Had a git oops there...

Hopefully I've addressed what you asked for @nielsvanvelzen

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks for the PR!

Issue #2396, opened 2023-01-09
Fails to play TrueHD audio media


need2:
### Describe the bug

When attempting to play and media with TrueHD audio tracks on the 0.15.1 client on Nvidia Shield Pro it fails to play the media and falls back to the previous screen after multiple retries.

This was not an issue prior to the 0.15.1 update.

Verified the files can be played on any other client, including web and direct file access.

Verified every media file with TrueHD audio tracks fails to play on the Nvidia Shield Pro with the 0.15.1 version of the app.

Since the app never actually crashes, it does not appear any app logs were uploaded to the server. The app just gracefully falls back to the previous screen.

### FFmpeg logs

```shell
/videos/8f1014c0-3cfc-f89c-b559-f29082dee339/hls1/main/0.ts

[...]

PR #2397, opened 2023-01-09
Remove experimental audio codecs from transcoding profile

nielsvanvelzen:
Regression from #2381

**Changes**
- Remove DCA / TrueHD from ExoPlayer transcoding profile

**Issues**

Client-side fix for #2396


nielsvanvelzen:
10.8.9 will add support for both of those codecs but not everyone updates their server immediately after we release. The client will auto-update for most users so fixing it client-side for now avoids a lot of pain for both us (the developers) and the end-users.

We can revert this PR for 0.16 (which will likely require server >= 10.9)

MrChip53 (approved):
Simple enough. LGTM.

PR #2415, opened 2023-01-14
Add media management preference

nielsvanvelzen:
This new preference shows a delete option for items when the user has permissions to delete the displaying item. This is a rewritten implementation of the previous delete option that was removed in 0.15.

**Changes**
- Add preference "enable media management"
- Add "delete" option to item details screen (when management is enabled)
  - Only shows when the user has permission to delete the item
    - Also checks per-library delete permission now
  - Catches issues and shows a toast for it, to prevent crashes when deletion fails
- Change DataRefreshService.lastDeletedItemId to UUID
- Item deleted toast is now translatable

**Issues**

Fixes #2368

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
@@ -104,6 +104,7 @@ import org.jellyfin.apiclient.model.querying.UpcomingEpisodesQuery;
 import org.jellyfin.sdk.model.api.BaseItemKind;
 import org.jellyfin.sdk.model.api.BaseItemPerson;
 import org.jellyfin.sdk.model.api.SeriesTimerInfoDto;
+import org.jellyfin.sdk.model.api.UserDto;
 import org.jellyfin.sdk.model.constant.ItemSortBy;
 import org.jellyfin.sdk.model.constant.MediaType;
 import org.jellyfin.sdk.model.constant.PersonType;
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
```diff
@@ -891,10 +892,21 @@ private void gotoSeries() {
...
 
+    private void deleteItem() {
+        FullDetailsFragmentHelperKt.deleteItem(
+                this,
+                api.getValue(),
+                ModelCompat.asSdk(mBaseItem),
+                dataRefreshService.getValue(),
+                navigationRepository.getValue()
+        );
+    }
```
neuroid, line 903: The old implementation had a delete confirmation prompt which I found quite useful. Any plans to bring that back?

nielsvanvelzen: I felt like I missed something, it was the dialog. Added it back now.

neuroid: LGTM

app/src/main/res/values/strings.xml
```diff
@@ -495,4 +495,10 @@
     <string name="server_setup_incomplete">The setup of this server has not been completed. Open Jellyfin in a web browser to finish setup before signing in.</string>
     <string name="episode_name_special">special</string>
     <string name="app_notification_beta">App updated to %1$s\nThank you for participating in the Jellyfin beta program</string>
+    <string name="item_deleted">%1$s deleted</string>
+    <string name="item_deletion_failed">Unable to delete %1$s</string>
+    <string name="pref_enable_media_management">Enable media management</string>
+    <string name="pref_enable_media_management_description">Show option to manage media while browsing</string>
+    <string name="item_delete_confirm_title">Delete item</string>
+    <string name="item_delete_confirm_message">Are you sure you want to delete this item? This action cannot be undone.</string>
```
crobibero, line 503: I don't know what the message used to say, but maybe make it clear that this will remove the file on disk?

nielsvanvelzen: Hardcoded string was
`This will PERMANENTLY DELETE " + mBaseItem.getName() + " from your library.  Are you VERY sure?`

crobibero (approved):

app/src/main/res/values/strings.xml
MailYouLater: @nielsvanvelzen & @crobibero:
Sorry, I didn't see these comments before the PR got merged, but what do you think of this?
```suggestion
    <string name="item_delete_confirm_message">Are you sure you want to delete this item from the server? This action cannot be undone.</string>
```

nielsvanvelzen: Yeah that's a good suggestion, made a PR #2477

PR #2422, opened 2023-01-15
Ignore missing notification permission from Glide

teobaranga:
Ignores the missing `android.permission.POST_NOTIFICATIONS` permission caused by some Glide class that's not used in this project. There's a issue already raised but it hasn't been actioned yet: https://github.com/bumptech/glide/issues/4940

This should allow upgrading AGP to 7.4.0 without any issues.


nielsvanvelzen (approved):
Thanks!

PR #2427, opened 2023-01-17
Add ExoPlayer backend to playback rewrite

nielsvanvelzen:
There we go, the third big PR for the playback rewrite! Well actually not that big but it adds an important part. This PR is a follow up on #2401 where I didn't include the ExoPlayer backend intentionally.

**Changes**

- Add ExoPlayerBackend
  - Supports all current features
  - This will be the only backend in the forseeable future, multi-backend support and smart-backend selection are a low priority for now (but will likely be high priority when video support is added)
  - Some features related to video are removed from this PR

**Issues**

Part of #1057

### DIFF
--- a/playback/exoplayer/src/main/kotlin/ExoPlayerBackend.kt
+++ b/playback/exoplayer/src/main/kotlin/ExoPlayerBackend.kt
@@ -1,0 +1,109 @@
+package org.jellyfin.playback.exoplayer
+
+import android.content.Context
+import com.google.android.exoplayer2.C
+import com.google.android.exoplayer2.DefaultRenderersFactory
+import com.google.android.exoplayer2.ExoPlayer
+import com.google.android.exoplayer2.MediaItem
+import com.google.android.exoplayer2.PlaybackException
+import com.google.android.exoplayer2.Player
+import com.google.android.exoplayer2.video.VideoSize
+import org.jellyfin.playback.core.backend.BasePlayerBackend
+import org.jellyfin.playback.core.mediastream.MediaStream
+import org.jellyfin.playback.core.model.PlayState
+import org.jellyfin.playback.core.model.PositionInfo
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.ZERO
+import kotlin.time.Duration.Companion.milliseconds
+
+class ExoPlayerBackend(
+	private val context: Context,
+) : BasePlayerBackend() {
+	override fun getPositionInfo(): PositionInfo = PositionInfo(
+		active = exoPlayer.currentPosition.milliseconds,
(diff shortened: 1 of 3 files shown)


playback/exoplayer/src/main/kotlin/ExoPlayerBackend.kt
```diff
@@ -0,0 +1,109 @@
...
+		}.build()
+
+		if (exoPlayer.mediaItemCount == 0) {
+			exoPlayer.setMediaItem(mediaItem)
+		} else {
+			// Remove old items
+			while (exoPlayer.mediaItemCount > 1) exoPlayer.removeMediaItem(0)
+			// Add new item
+			exoPlayer.addMediaItem(mediaItem)
+		}
```
crobibero, line 74: Is there a benefit to this method over just doing

```
while (exoPlayer.mediaItemCount > 1) exoPlayer.removeMediaItem(0)
exoPlayer.addMediaItem(mediaItem)

**Review by crobibero — APPROVED** [review]

File: playback/exoplayer/src/main/kotlin/ExoPlayerBackend.kt [hunk 2]
```diff
@@ -0,0 +1,109 @@
...
+		val renderersFactory = DefaultRenderersFactory(context).apply {
+			setEnableDecoderFallback(true)
+			setExtensionRendererMode(DefaultRenderersFactory.EXTENSION_RENDERER_MODE_ON)
+		}
+		ExoPlayer.Builder(context, renderersFactory)
+			.build()
+			.also { player -> player.addListener(PlayerListener()) }
+	}
+
+	inner class PlayerListener : Player.Listener {
```
DavidFair, line 38: More of a nit, is it worth pulling derivative classe(s) out into a dedicated file(s) early? 

The existing Exoplayer/VLC code organically grew and turned into a mixture of boilerplate and implementation, hence I'm wondering if it's worth separating before we add to it?

nielsvanvelzen: Right now the ExoPlayer implementation is small enough that it doesn't matter. When the class grows it will likely get split into multiple files. I also believe large files aren't a huge problem, the fact that it isn't related to the file it's in is the problem. In the current player code we have pretty much everything in one file, the "ExoPlayerBackend" only contains the binding to ExoPlayer.

playback/exoplayer/src/main/kotlin/ExoPlayerBackend.kt
nielsvanvelzen: Nop, fixed it

Issue #2321, opened 2022-12-07
Implement Shuffle similarly to Android/Web


nwprince:
### Describe the feature you'd like

Shuffling shows and seasons within Jellyfin was a feature added and documented in this pull request: https://github.com/jellyfin/jellyfin-web/pull/3627

Please implement the same feature for Android TV. We would love the ability to shuffle irrelevant TV shows like Rick & Morty, Bob's Burgers, The Office, Friends, etc. Some shows have no continuous or connected plot (outside a few small exceptions) and these shows are great for having on in the background. Having a shuffle feature would give a good mix of old and new episodes without the need to setup a playlist or handle the shuffle manually by starting in one season then switching to another later. 

See attachment for Android mobile implementation.
![Screenshot_20221206_191306_Jellyfin](https://user-images.githubusercontent.com/17996032/206063336-1bf4e8db-6e57-4ac6-bdb1-94e02b0f032f.jpg)
![Screenshot_20221206_191333_Jellyfin](https://user-images.githubusercontent.com/17996032/206063390-5a0f9cb9-8221-42d6-bd7d-aa482f219fbf.jpg)

tim-vk:
As far as I'm aware this is already implemented. Just go into the show you want to shuffle and it should be available. Could you verify that the button is not there for you? And if it's not: Can you check your folder structure for your show? (as far as I can see this is more-or-less the only thing that determines if shuffle is available or not).

![image](https://user-images.githubusercontent.com/9055206/215603319-c526b923-e23d-4062-97e1-11257e811062.png)

nwprince:
So I have two streaming devices that are running Jellyfin for Android TV. The first is a Chromecast with Google TV, the other is a NVIDIA Shield (tube version). Neither of them render a Shuffle All link. 

Here are some screenshots of what I see and the web client. I also included Metadata and filesystem structures for the show. 
![chromecast](https://user-images.githubusercontent.com/17996032/215618047-9a2217bd-7d51-499c-9227-b51c595ad11c.jpg)
![shield](https://user-images.githubusercontent.com/17996032/215618049-94f3f3d6-c580-4401-8190-69b294e3846a.jpg)
![web_version](https://user-images.githubusercontent.com/17996032/215618050-ba55b523-31b1-4aff-9c02-f733f03a0289.png)
![file_structure](https://user-images.githubusercontent.com/17996032/215618269-089cfc77-ebb6-4324-94aa-95130367b96b.png)
[...]

PR #2459, opened 2023-01-31
adds "shuffle with" button if "always ask" is selected as preferred player.

tim-vk:
**Changes**

Shows the "shuffle" button when video player is set to "always ask". Selecting the shuffle button prompts the player types to be selected.

**Issues**

Implements #2321

**Some questions for the reviewers**

- The button is now labelled: "Shuffle" not "shuffle with". is that desired?
- I'm not used to java: I've created a seperate function: `playWithItemMenu` which is used within shuffleMenu and PlayMenu. Is this the correct way to implement such a feature?
- There are some other functions that are also hidden if "always ask" is selected. Would you like me to implement those too? If yes: should that be in this PR or seperate PR's per "feature"?
- What is the conversion policy for when/how to convert to Kotlin?

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
@@ -981,7 +981,7 @@ public class FullDetailsFragment extends Fragment implements RecordingIndicatorV
             });
             mDetailsOverviewRow.addAction(playButton);
 
-            if (mBaseItem.getIsFolderItem() || baseItem.getType() == BaseItemKind.MUSIC_ARTIST) {
+            if (mBaseItem.getIsFolderItem()) {
                 shuffleButton = TextUnderButton.create(requireContext(), R.drawable.ic_shuffle, buttonSize, 2, getString(R.string.lbl_shuffle_all), new View.OnClickListener() {
                     @Override
                     public void onClick(View view) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
```diff
@@ -1510,40 +1523,52 @@ public boolean onMenuItemClick(MenuItem item) {
...
-                                mediaManager.getValue().playNow(requireContext(), response, false);
-                            } else {
-                                videoQueueManager.getValue().setCurrentVideoQueue(response);
-                                navigationRepository.getValue().navigate(Destinations.INSTANCE.externalPlayer(0));
-                            }
-                        }
-                    });
-                    return true;
+    private boolean playWithItemMenu(MenuItem item, boolean shuffle) {
+            if(item.getItemId() == R.id.play_with_vlc) {
```
nielsvanvelzen, line 1541: Either keep the switch or use else if statements. Also make sure indenting is correct (space after if)

nielsvanvelzen (changes requested):

nielsvanvelzen:
> The button is now labelled: "Shuffle" not "shuffle with". is that desired?

I think that's fine.

> I'm not used to java: I've created a seperate function: playWithItemMenu which is used within shuffleMenu and PlayMenu. Is this the correct way to implement such a feature?

In this case, good enough. This file is a bit messy anyway so I'm not that picky reviewing it.

> There are some other functions that are also hidden if "always ask" is selected. Would you like me to implement those too? If yes: should that be in this PR or seperate PR's per "feature"?

They could probably be in the same PR but tbh it makes sense to hide those options if the user chooses to not have a default player.

> What is the conversion policy for when/how to convert to Kotlin?

New files are Kotlin, existing files _may_ be converted but the left-over Java files is mostly stuff that's slowly getting rewritten so I wouldn't bother with it if I were you.

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
tim-vk: Ok! Fixed to else if. Reason was because android linting apparently does not like switch-cases (i do not really know the details why. But adding code that creates warnings gives me the heebejeebies).

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
```diff
@@ -980,6 +980,19 @@ public void onClick(View view) {
                 }
             });
             mDetailsOverviewRow.addAction(playButton);
+
+            if (mBaseItem.getIsFolderItem() || baseItem.getType() == BaseItemKind.MUSIC_ARTIST) {
```
nielsvanvelzen, line 984: MUSIC_ARTIST can be removed here because music related stuff doesn't use the video player so this would bring up a player selection that won't be used.

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
```diff
@@ -1510,40 +1523,50 @@ public boolean onMenuItemClick(MenuItem item) {
...
-                        }
-                    });
-                    return true;
+    private boolean playWithItemMenu(MenuItem item, boolean shuffle) {
+            if (item.getItemId() == R.id.play_with_vlc) {
+                systemPreferences.getValue().set(SystemPreferences.Companion.getChosenPlayer(), PreferredVideoPlayer.VLC);
+                play(mBaseItem, 0, shuffle);
+                return true;
+            }
+            else if (item.getItemId() == R.id.play_with_exo) {
```
nielsvanvelzen, line 1546:
```suggestion
            } else if (item.getItemId() == R.id.play_with_exo) {
```
Correct formatting

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
```diff
@@ -1510,40 +1523,50 @@ public boolean onMenuItemClick(MenuItem item) {
...
+                systemPreferences.getValue().set(SystemPreferences.Companion.getChosenPlayer(), PreferredVideoPlayer.EXOPLAYER);
+                play(mBaseItem, 0, shuffle);
+                return true;
+            }
+            else if (item.getItemId() == R.id.play_with_external) {
```
nielsvanvelzen, line 1551:
```suggestion
            } else if (item.getItemId() == R.id.play_with_external) {
```

nielsvanvelzen (changes requested):

nielsvanvelzen:
> Luckily it's not just me who finds this messy. Are there any running project/plans to refactor this file? Reason I ask is because I also wanted to take a look at language settings (for subtitles e.g.) before you press play: i think that also touches this file quite a lot.

I'm currently working on rewriting all playback code (#1057) and after that I want to rewrite the UI with Jetpack Compose when it's ready for Android TV.

nielsvanvelzen (approved):
Thanks for the PR!

Issue #2180, opened 2022-10-15
 Marking episode as watched, doesn't refresh next up at home page


JVT038:
### Describe the bug

1. Mark an episode as watched
2. Return to home page
3. The episode disappears from the continue watching / next up row
4. The next up row completely doesn't refresh, so the next episode doesn't appear
5. To fix this, switch users and the home page is refreshed

### Application version

0.14.4

### Where did you install the app from?

Google Play

### Device information

Philips QM163E

### Android version

Android 8.0

### Jellyfin server version

10.8.5

Bandid0:
That happens to me too

VampiricAlien:
I think this is a known bug, same thing happens on android 11

PR #2472, opened 2023-02-03
Fix old items are shown when new filter is applied.

### LINKED ISSUE #2470: Turning on filter only refreshes the first couple of items.
### Describe the bug

1. Go to a library (e.g. tv-shows).
2. Ensure filters are off
3. Turn on the "unwatched only" filter (or the "favorites" filter).
4. Only the first couple of items are refreshed. All other items are still shown.

### Logs

_No response_

### Application version

0.15.3

Google Play

### Device information

Google chromcast with TV

### Android version

Android 11

### Jellyfin server version

10.8.5

tim-vk:
**Changes**

When a filter is applied all current items are removed first. Before updating them.

**Issues**

Fixes #2470, #2180


tim-vk:
I think this is one of the (many) possible solutions: But i hope the problem is clear.

Solves the issue with my testing without introducing new bugs (fingers crossed).

nielsvanvelzen (approved):
Might not be the most efficient solution but this UI is kinda hard to work with so I can't think of anything better myself. Thanks for the PR!

PR #2498, opened 2023-02-14
Rewrite LibraryDreamService with Compose

nielsvanvelzen:
Originally made these changes about 5 months ago but for some reason never made the PR 🤷. Some future plans I have for the screensaver:

- Add now playing info for current playing music
- Add lyrics for current playing music (10.9)
- Replace the current fake screensaver in the NowPlayingFragment with this one

**Changes**
- Add DreamServiceCompat which extends DreamService to support lifecycles and a compose view
- Rewrite LibraryDreamService with compose
  - Looks the same, only change is the ken burns effect which I replaced with a zoom effect because I didn't want to write a custom ken burns implementation
- Remove old daydream layout stuff
- Remove KenBurns dependency

**Issues**


app/src/main/java/org/jellyfin/androidtv/integration/dream/LibraryDreamService.kt
```diff
@@ -26,52 +57,203 @@ import org.jellyfin.sdk.model.api.ImageFormat
 import org.jellyfin.sdk.model.api.ImageType
 import org.jellyfin.sdk.model.constant.ItemSortBy
 import org.koin.android.ext.android.inject
+import org.koin.androidx.compose.get
 import timber.log.Timber
-import java.util.concurrent.ExecutionException
+import kotlin.time.Duration.Companion.seconds
+
+fun Modifier.overscan(): Modifier = then(padding(48.dp, 27.dp))
```
nielsvanvelzen, line 64: Move this to the composable folder so it's reusable for future UI.

app/src/main/java/org/jellyfin/androidtv/integration/dream/LibraryDreamService.kt
```diff
@@ -26,52 +57,203 @@ import org.jellyfin.sdk.model.api.ImageFormat
...
+		)
+
+		// Clock
+		AnimatedVisibility(
+			visible = showClock,
+			enter = fadeIn(),
+			exit = fadeOut(),
+		) {
+			Text(
+				text = DateUtils.formatDateTime(LocalContext.current, System.currentTimeMillis(), DateUtils.FORMAT_SHOW_TIME),
```
nielsvanvelzen, line 102: The clock doesn't update right now, so it stays stuck on the time the daydream started. Fix that and make it a separate composable so it's reusable.

PR #2507, opened 2023-02-16
(DO NOT MERGE) Trigger server 500 when updating display preferences

nielsvanvelzen:
Opening a PR so we get a safe-to-share debug APK to reproduce the server issue :-)

**Changes**

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/PreferencesRepository.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/PreferencesRepository.kt
@@ -1,5 +1,8 @@
 package org.jellyfin.androidtv.preference
 
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.runBlocking
 import org.jellyfin.sdk.api.client.ApiClient
 import kotlin.collections.set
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/preference/PreferencesRepository.kt
```diff
@@ -25,11 +28,13 @@ class PreferencesRepository(
...
-		// Relevant server issue: https://github.com/jellyfin/jellyfin/issues/5261
-		liveTvPreferences.update()
-		userSettingPreferences.update()
+	suspend fun onSessionChanged() = coroutineScope {
+		for (i in 1..100) {
+			awaitAll(
+				async { liveTvPreferences.update() },
+				async { userSettingPreferences.update() },
+			)
+		}
```
nielsvanvelzen, line 37: this change runs the two updates in parallel instead of waterfall. I added the for loop to make sure  the bug is triggered as those requests can respond pretty fast so it doesn't always trigger the bug otherwise. The update function calls the `getDisplayPreferences()` endpoint on the server.

app/src/main/java/org/jellyfin/androidtv/preference/store/DisplayPreferencesStore.kt
```diff
@@ -72,6 +72,9 @@ abstract class DisplayPreferencesStore(
 				displayPreferencesDto = DisplayPreferencesDto.empty()
 			}
 
+			// TMP for reproduction
+			throw err
```
nielsvanvelzen, line 76: This is to make sure the app doesn't ignore the exception but crashes instead

PR #2520, opened 2023-02-20
Revert "Update aboutlibraries to v10.6.0"

nielsvanvelzen:
Reverts jellyfin/jellyfin-androidtv#2509


nielsvanvelzen:
Forgot the description; update caused some issues with local dev environments when no explicit Java version is set and it uses a Java version older then 15.

PR #2534, opened 2023-02-23
Mark LibVLC as experimental

nielsvanvelzen:
There's a lot of crashes when using LibVLC because our implementation is just bad. This PR changes the preference to end with "(experimental)" so users are less likely to use it.

**Changes**
- Mark LibVLC as experimental

**Issues**


crobibero (approved):

thornbill:
> so users are less likely to use it

I don't believe you... 🤔

thornbill (approved):

nielsvanvelzen:
> > so users are less likely to use it
> 
> I don't believe you... 🤔

At the very least I can respond to issues that it's experimental

Issue #1888, opened 2022-07-24
Move logo and clock randomly in the screensaver to prevent burn-ins on OLEDs


EmberLightVFX:
### Describe the feature you'd like

Moving the Jellyfin logo and the clock to different positions when the screensaver is active would help prevent burn-ins for OLED TVs. Ether they could jump to different corners or maybe be possible to dissable them in the settings.

kingp0dd:
+1

kingp0dd:
is there a chance that will make it to the new roadmap? :)

nielsvanvelzen:
Yes this is definitely something we should address, added to the roadmap.

Hukuma1:
The new screensaver is great, but would love the option to disable static elements. (clock + jellyfin logo)

Had a scare because I forgot to turn off the TV the other night. D:

nielsvanvelzen:
Starting with version 0.19 of the app the screensaver will no longer show the Jellyfin logo which fixes this issue.

Passific:
By the PR title and content, I guess the clock is staying put.
That does not completely close the issue then.

kingp0dd:
> By the PR title and content, I guess the clock is staying put. That does not completely close the issue then.

Your are right

Zark247:
You can disable the clock from the gear icon (Preferences):
Customization > Show Clock : Never

PR #2566, opened 2023-03-04
Dim screensaver and static elements

DavidFair:
**Changes**
~Adds an option to remove static elements, specifically the clock and logo, from the screensaver. This should reduce the chances of burn-in on the screensaver.~

Reduces the alpha on various static elements. This not only reduces the burn-in effect on OLED devices, but it also is quite satisfying as the TV feels like it "wakes up" or "dims down" switching into the screen saver. The alpha values have been hand picked to balance between readability and impact. But I'm happy to tweak them based on testing with real devices.

I've also applied it to the Logo backdrop, as any images that fail to load will automatically show that now.

Screenshots:

image

image

**Issues**
Fixes: #1888

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentLibraryShowcase.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentLibraryShowcase.kt
@@ -8,6 +8,7 @@ import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.Shadow
 import androidx.compose.ui.graphics.asImageBitmap
@@ -35,7 +36,7 @@ fun DreamContentLibraryShowcase(
 			contentDescription = null,
 			alignment = Alignment.Center,
 			contentScale = ContentScale.Crop,
-			modifier = Modifier.fillMaxSize()
+			modifier = Modifier.fillMaxSize().alpha(0.7f)
 		)
 	}
 
@@ -55,6 +56,8 @@ fun DreamContentLibraryShowcase(
 					blurRadius = 2f,
 				)
 			),
+			// The alpha needs to be higher than the backdrop so it's still readable
+			modifier = Modifier.alpha(0.85f)
 		)
 	}
 }
--- a/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentLogo.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentLogo.kt
@@ -8,6 +8,7 @@ import androidx.compose.foundation.layout.width
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
@@ -18,6 +19,7 @@ import org.jellyfin.androidtv.R
 fun DreamContentLogo() = Box(
 	modifier = Modifier
 		.fillMaxSize()
+		.alpha(0.5f)
 		.background(Color.Black),
 ) {
 	Image(
--- a/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHeader.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHeader.kt
@@ -13,6 +13,7 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.Shadow
 import androidx.compose.ui.platform.LocalContext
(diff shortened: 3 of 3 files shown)


nielsvanvelzen:
The screensaver should always show the Jellyfin logo, so this change would then only affect the clock which you can already hide. I'd rather create a composable to move the elements around (which we would need for the title anyway) or use transparency so the colors of the backdrop change the pixels.

DavidFair:
> The screensaver should always show the Jellyfin logo, 

But this risks burning in the Jellyfin Logo, unless we place it in completely random locations? Even if we shift it corner-to-corner over years those corners will gear higher wear if a user leaves their TV on often. At least with backdrops it's effectively random and even wear across all pixels.

> I'd rather create a composable to move the elements around (which we would need for the title anyway) 

Agreed, but I can't imagine how we evenly wear an OLED TV whilst having the elements place in sane locations


> or use transparency so the colors of the backdrop change the pixels.

But I imagine the Logo would stay the same colour? This is the one I expect to burn the worst since the text and time changes throughout the day

nielsvanvelzen:
I've looked at various screensavers (mostly via YouTube video's) and I see a lot of them have static  elements but those are pretty much always semi-transparent so the colors of a backdrop come trough. I guess this would be the approach we can keep the logo while not burning in OLED devices. I'm expecting our logo will still show well at 80%-ish transparency, same with the other elements (clock/showcase title).

DavidFair:
> I've looked at various screensavers (mostly via YouTube video's) and I see a lot of them have static elements but those are pretty much always semi-transparent so the colors of a backdrop come trough. I guess this would be the approach we can keep the logo while not burning in OLED devices. I'm expecting our logo will still show well at 80%-ish transparency, same with the other elements (clock/showcase title).

Is it worth dropping the OLED-safe option then and just reducing the opacity for everything by default?

nielsvanvelzen:
I think that would be a good solution yes.

app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentLogo.kt
```diff
@@ -18,6 +19,7 @@ import org.jellyfin.androidtv.R
 fun DreamContentLogo() = Box(
 	modifier = Modifier
 		.fillMaxSize()
+		.alpha(0.5f)
```
nielsvanvelzen, line 22: The background for DreamContentLogo is static so the alpha doesn't make any sense here.

app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentLibraryShowcase.kt
```diff
@@ -35,7 +36,7 @@ fun DreamContentLibraryShowcase(
 			contentDescription = null,
 			alignment = Alignment.Center,
 			contentScale = ContentScale.Crop,
-			modifier = Modifier.fillMaxSize()
+			modifier = Modifier.fillMaxSize().alpha(0.7f)
```
nielsvanvelzen, line 39: Why add alpha to the backdrop?

app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHeader.kt
```diff
@@ -41,11 +44,14 @@ fun DreamHeader(
...
 			exit = fadeOut(),
-			modifier = Modifier.height(41.dp),
+			modifier = Modifier.height(41.dp).alpha(alpha),
 		) {
 			Image(
 				painter = painterResource(R.drawable.app_logo),
 				contentDescription = stringResource(R.string.app_name),
+				// We also set the alpha for the non-header logo, as this can
+				// be displayed as a fallback
+				modifier = Modifier.alpha(alpha)
```
nielsvanvelzen, line 54: Why set the alpha for both the row and the logo? That just multiplies it. Right now the logo is barely readable.

app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentLibraryShowcase.kt
DavidFair: I've done it to  "wakes up" or "dim down" switching into the screen saver, rather than have the screen at full brightness. On a real TV it feels like we're in a dedicated screensaver rather than just a library preview

It's a bit of scope creep, but I thought I'd float it whilst already doing it?

app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHeader.kt
DavidFair: The problem is the row has text elements, such as the name, I want to dim down too so I applied it to the row

Unfortunately, I also wanted to dim the logo too, for the times where we fail to load a backdrop (so the brightness doesn't go up or down), so I unintentionally multiplied these alphas.

I could manually do alpha on each element of the row (excluding the logo), or add a constructor to the Logo to set the alpha to avoid multiplying this into the row? I'm not sure what the best approach would be

app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentLibraryShowcase.kt
nielsvanvelzen: But it is a dedicated screensaver?

app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHeader.kt
nielsvanvelzen: Just apply it to the row only?

nielsvanvelzen:
Closing as stale. Please re-open or create a new pull request if interested in finishing.

placidcasual98:
Please could we have a feature for this turning off the jellyfin logo or moving it around? Definitely going to get screen burn in .

PR #2569, opened 2023-03-05
Refactor CustomPlaybackTransportControlGlue to extract logic into actions

DavidFair:
**Changes**
Refactors the `CustomPlaybackTransportControlGlue` by:
- Removing static dispatch for custom actions and built-in actions
- Encapsulating the logic for each action into the action class

This will make it much easier to unit test actions, because we only need to inject the mocks for the method signature rather than everything the glue class could possibly need.

I've done this with as minimal changes to `videoAdapter` to not interfere with any playback rewrites going on. It's not perfect but (hopefully) it creates so better separation of concerns that will allow us to write tests at a later date.

In the future I want to shift as much logic out of the glue class into the relevant actions, but this diff is already quite large...

**To Test**
- Check all buttons work whilst playing media.
- Double check live TV buttons (as I don't have access to those options locally)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
@@ -305,7 +305,7 @@ public class CustomPlaybackTransportControlGlue extends PlaybackTransportControl
     public void onCustomActionClicked(Action action, View view) {
         // Handle custom action clicks which require a popup menu
         if (action instanceof CustomAction){
-            ((CustomAction) action).handleClickAction(playbackController, getPlayerAdapter(), getPlayerAdapter().getLeanbackOverlayFragment(), getContext(), view);
+            ((CustomAction) action).handleClickAction(playbackController, getPlayerAdapter(), getContext(), view);
         }
 
         if (action == playbackSpeedAction) {
@@ -391,10 +391,10 @@ public class CustomPlaybackTransportControlGlue extends PlaybackTransportControl
         VideoPlayerAdapter playerAdapter = getPlayerAdapter();
 
         if (playerAdapter.hasSubs() && keyCode == KoinJavaComponent.<UserPreferences>get(UserPreferences.class).get(UserPreferences.Companion.getShortcutSubtitleTrack())) {
-            closedCaptionsAction.handleClickAction(playbackController, getPlayerAdapter(), getPlayerAdapter().getLeanbackOverlayFragment(), getContext(), v);
+            closedCaptionsAction.handleClickAction(playbackController, getPlayerAdapter(), getContext(), v);
         }
         if (playerAdapter.hasMultiAudio() && keyCode == KoinJavaComponent.<UserPreferences>get(UserPreferences.class).get(UserPreferences.Companion.getShortcutAudioTrack())) {
-            selectAudioAction.handleClickAction(playbackController, getPlayerAdapter(), getPlayerAdapter().getLeanbackOverlayFragment(), getContext(), v);
+            selectAudioAction.handleClickAction(playbackController, getPlayerAdapter(), getContext(), v);
         }
         return super.onKey(v, keyCode, event);
     }
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
@@ -28,17 +28,23 @@ import org.jellyfin.androidtv.preference.UserPreferences;
 import org.jellyfin.androidtv.preference.constant.ClockBehavior;
 import org.jellyfin.androidtv.ui.playback.PlaybackController;
 import org.jellyfin.androidtv.ui.playback.PlaybackManager;
+import org.jellyfin.androidtv.ui.playback.overlay.action.AndroidAction;
 import org.jellyfin.androidtv.ui.playback.overlay.action.ChannelBarChannelAction;
 import org.jellyfin.androidtv.ui.playback.overlay.action.ChapterAction;
 import org.jellyfin.androidtv.ui.playback.overlay.action.ClosedCaptionsAction;
 import org.jellyfin.androidtv.ui.playback.overlay.action.CustomAction;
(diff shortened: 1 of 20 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
```diff
@@ -446,11 +375,18 @@ public void setInjectedViewsVisibility() {
 
     @Override
     public boolean onKey(View v, int keyCode, KeyEvent event) {
-        if (hasSubs() && event.getAction() == KeyEvent.ACTION_UP && keyCode == KoinJavaComponent.<UserPreferences>get(UserPreferences.class).get(UserPreferences.Companion.getShortcutSubtitleTrack())) {
-            closedCaptionsAction.handleClickAction(playbackController, getPlayerAdapter().getLeanbackOverlayFragment(), getContext(), v);
+        if (event.getAction() != KeyEvent.ACTION_UP){
```
nielsvanvelzen, line 378:
```suggestion
        if (event.getAction() != KeyEvent.ACTION_UP) {
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomPlaybackTransportControlGlue.java
```diff
@@ -281,65 +288,23 @@ void addMediaActions() {
...
-            getPlayerAdapter().getLeanbackOverlayFragment().setFading(false);
-            closedCaptionsAction.handleClickAction(playbackController, getPlayerAdapter().getLeanbackOverlayFragment(), getContext(), view);
-        } else if (action == settingAction) {
-            getPlayerAdapter().getLeanbackOverlayFragment().setFading(false);
-            settingAction.setSubtitlesPresent(hasSubs());
-            settingAction.handleClickAction(playbackController, getPlayerAdapter().getLeanbackOverlayFragment(), getContext(), view);
-        } else if (action == playbackSpeedAction) {
-            getPlayerAdapter().getLeanbackOverlayFragment().setFading(false);
-            playbackSpeedAction.handleClickAction(playbackController, getPlayerAdapter().getLeanbackOverlayFragment(), getContext(), view);
+        if (action instanceof CustomAction){
```
nielsvanvelzen, line 299:
```suggestion
        if (action instanceof CustomAction) {
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/action/AndroidAction.kt
```diff
@@ -0,0 +1,9 @@
+package org.jellyfin.androidtv.ui.playback.overlay.action
+
+import org.jellyfin.androidtv.ui.playback.overlay.VideoPlayerAdapter
+
+interface AndroidAction {
```
nielsvanvelzen, line 5: Would it be possible to also inherit this class from our CustomAction (perhaps merge the two?) so we don't have 2 separate ways to trigger them.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/action/ClosedCaptionsAction.kt
```diff
@@ -21,16 +22,17 @@ class ClosedCaptionsAction(
 
 	override fun handleClickAction(
 		playbackController: PlaybackController,
-		leanbackOverlayFragment: LeanbackOverlayFragment,
+		videoPlayerAdapter: VideoPlayerAdapter,
 		context: Context,
-		view: View,
+		view: View
```
nielsvanvelzen, line 27:
```suggestion
		view: View,
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/action/FastForwardAction.kt
```diff
@@ -0,0 +1,12 @@
+package org.jellyfin.androidtv.ui.playback.overlay.action
+
+import android.content.Context
+import androidx.leanback.widget.PlaybackControlsRow
+import org.jellyfin.androidtv.ui.playback.overlay.VideoPlayerAdapter
+
+class FastForwardAction(context: Context) : PlaybackControlsRow.FastForwardAction(context),
+	AndroidAction {
```
nielsvanvelzen, line 8: I'd format it like this:

```suggestion
class FastForwardAction(
	context: Context
) : PlaybackControlsRow.FastForwardAction(context), AndroidAction {
```

or just a single line

```suggestion
class FastForwardAction(context: Context) : PlaybackControlsRow.FastForwardAction(context), AndroidAction {
```

This also applies to the other actions.

nielsvanvelzen (changes requested):
Overall this is a good change, found some formatting and left a comment about the CustomAction/AndroidAction.

Keep in mind that all of this code is for the current playback UI and part of the rewrite (#1057) is to add new UI from scratch (with Jetpack Compose). So these changes will likely make it to a next release but be removed some time later, if you don't mind that I'll keep reviewing the changes!

DavidFair:
> Keep in mind that all of this code is for the current playback UI and part of the rewrite (#1057) is to add new UI from scratch (with Jetpack Compose). So these changes will likely make it to a next release but be removed some time later, if you don't mind that I'll keep reviewing the changes!

I don't mind.  be happy for you to merge this then delete it a week later, if it meant we got a shiny new playback module.

I'll pick up your review comments in the coming days.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/action/AndroidAction.kt
DavidFair: Since the default Actions need to inherit from a Android base class, and Kotlin doesn't support multiple inheritance, I've restructured it to:

```
CustomAction (interface) -> CustomActionBase (base class) -> Action*
```

DavidFair: Not easily, doing some further digging, we need to have a way to call `onCustomActionClicked` if we get Android actions to use the CustomAction class

Since we can't do multiple inheritance, we either have to inherit the default action (won't derive `onCustomActionClicked`), or reimplement the Android component we no longer derive from

I'm sure there's a clever way to do this, but because we need the view for CustomActions it means they fundamentally require different things

nielsvanvelzen (approved):

PR #2591, opened 2023-03-13
Save accounts to account manager by address

### LINKED ISSUE #2586: Switching between servers forces me to login again every time
### Describe the bug

I have two jellyfin instances at home, one for video and one for music. I have both instances configured in the selection screen on my CCwGTV device. Whenever I switch servers, it forces me to log in again, despite being authenticated in the past.

I would like the client to remember the logins for all servers and not require me to authenticate unless necessary. To be honest, I'm not sure if this is a bug report or a feature request, so feel free to re-classify it any way you'd like.

What would be really cool: On the pick-a-user screen, show a grid for all the saved users for all saved servers with each row containing the users for a particular server (with the servers sorted by last-accessed, descending), and then having the user scroll sideways (as it does currently) to pick a user.

### Logs

_No response_

### Application version

Current beta as of March 10th, 2023

Google Play

### Device information

Chromecast with Google TV

### Android version

12

### Jellyfin server version

10.8.9

nielsvanvelzen:
Previously we would only use the username. Unfortunately this username should be unique per app. So if you have multiple servers with the same username the app would (accidentally) always overwrite the existing account when signing in. This PR updates the name we send to the account manager by appending the server URL.

The changes in this PR do not affect existing logins, it will update the name the next time you authenticate (so not on account use). We don't use those account manager names ourselves because we store our values in the user data.

**Changes**
- Use the combination of username+server address when storing users to the account manager
  image

**Issues**

Fixes #2586

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/auth/store/AccountManagerStore.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/auth/store/AccountManagerStore.kt
@@ -32,6 +32,7 @@ class AccountManagerStore(
 
 	private fun getAccountData(account: Account): AccountManagerAccount = AccountManagerAccount(
 		id = accountManager.getUserData(account, ACCOUNT_DATA_ID).toUUID(),
+		address = account.name,
 		server = accountManager.getUserData(account, ACCOUNT_DATA_SERVER).toUUID(),
 		name = accountManager.getUserData(account, ACCOUNT_DATA_NAME),
 		accessToken = accountManager.peekAuthToken(account, ACCOUNT_ACCESS_TOKEN_TYPE)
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
I've confirmed this issue fixes #2586

app/src/main/java/org/jellyfin/androidtv/auth/store/AccountManagerStore.kt
```diff
@@ -55,9 +56,13 @@ class AccountManagerStore(
 		}
 
 		// Update name
-		if (androidAccount.name != accountManagerAccount.name) {
+		if (androidAccount.name != accountManagerAccount.address) {
 			androidAccount = suspendCoroutine { continuation ->
-				accountManager.renameAccount(androidAccount, accountManagerAccount.name, { continuation.resume(it.result) }, null)
+				accountManager.renameAccount(
+					androidAccount, accountManagerAccount.address,
```
thornbill, line 62: This should be 2 lines right?

nielsvanvelzen: yup

thornbill (approved):

PR #2593, opened 2023-03-15
Add media manager implementation using new playback module

nielsvanvelzen:
This time the changes are slightly less refined and there's a bunch of TODO's in the code.. they'll be solved eventually. With this PR you can actually start using the rewrite for audio playback, although I don't recommend it yet, there's a bunch of rough edges that will be polished in the coming weeks/months.

**Changes**
- Add RewriteMediaManager
  - A (partial) implementation of the MediaManager interface using the playback rewrite
  - Some features like repeat/shuffle/queue modification are still missing and will come soonish
- Enable playback rewrite module app-wide (DI)
- Refactor BackendRepository to BackendService
- Add MediaStreamService to convert queue -> mediastream using MediaStreamResolvers
- Add UniversalAudioMediaStreamResolver that used the universal audio API to get an audio stream using the Jellyfin API
- Add onMediaStreamEnd event to backend
- ExoPlayer improvements
  - Add stream end detection
  - Check if requested operations are allowed (like changing speed etc.) and warn otherwise - useful for debugging

**Issues**

Part of #1057


app/src/main/java/org/jellyfin/androidtv/di/PlaybackModule.kt
```diff
@@ -31,4 +45,31 @@
...
+
+	single { createPlaybackManager() }
+}
+
+fun Scope.createPlaybackManager() = playbackManager {
+	install(exoPlayerPlugin(get()))
+	install(jellyfinPlugin(get(), get()))
+
+	val activityIntent = Intent(get(), MainActivity::class.java)
+	val pendingIntent = PendingIntent.getActivity(get(), 0, activityIntent, PendingIntent.FLAG_IMMUTABLE)
```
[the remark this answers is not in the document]
nielsvanvelzen: ^ This will need to be fixed later but the intent will change before release so I won't bother for now.

app/src/main/java/org/jellyfin/androidtv/ui/playback/rewrite/RewriteMediaManager.kt
```diff
@@ -0,0 +1,282 @@
...
+import org.jellyfin.playback.core.model.PlayState
+import org.jellyfin.playback.core.queue.Queue
+import org.jellyfin.playback.core.queue.item.QueueEntry
+import org.jellyfin.playback.core.queue.queue
+import org.jellyfin.playback.jellyfin.queue.item.BaseItemDtoUserQueueEntry
+import org.jellyfin.sdk.api.client.ApiClient
+import org.jellyfin.sdk.model.api.BaseItemDto
+import kotlin.time.Duration.Companion.milliseconds
+
+@Suppress("TooManyFunctions")
```
DavidFair, line 32: Is there a way we can break this down using composition?

I'm thinking if you had multiple instances such as `AudioManager`, `VideoManager`, `CodecManager` ....etc. then this simply becomes a glue class that delegates to it's instances. It also makes it trivial to structure future code too

Even if it's a single line currently, it means when we inevitably add more and more code it's already encapsulated outside of this class

I'm concerned that whilst this is a good improvement, there's a lot of implementation in a single location still and `Media` turns into a catch-all term

playback/jellyfin/src/main/kotlin/mediastream/UniversalAudioMediaStreamResolver.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.jellyfin.playback.core.queue.item.QueueEntry
+import org.jellyfin.playback.jellyfin.queue.item.BaseItemDtoUserQueueEntry
+import org.jellyfin.sdk.api.client.ApiClient
+import org.jellyfin.sdk.api.client.extensions.universalAudioApi
+import org.jellyfin.sdk.model.api.BaseItemKind
+
+class UniversalAudioMediaStreamResolver(
+	val api: ApiClient,
+) : MediaStreamResolver {
+	override suspend fun getStream(queueEntry: QueueEntry): MediaStream? {
```
DavidFair, line 14: Since this doesn't use anything from Android should it be in utils? A lot of our code for determining streams / codecs is currently there at the moment

app/src/main/java/org/jellyfin/androidtv/ui/playback/rewrite/RewriteMediaManager.kt
nielsvanvelzen: This is just an implementation for an interface that will be removed when we rewrite the UI. Just bridging the new code to the old spaghetti here.

playback/jellyfin/src/main/kotlin/mediastream/UniversalAudioMediaStreamResolver.kt
nielsvanvelzen: Everything related to the jellyfin API is in the playback/jellyfin module Media stream resolver implementations are one of those things. I don't see why that would be moved to the utils package in the app module.

PR #2622, opened 2023-03-25
Add main preferences screen in Compose

teobaranga:
I noticed there is a [task](https://github.com/jellyfin/jellyfin-androidtv/projects/15#card-82973332) to convert the preferences screens to Compose as an experiment towards rewriting more of the app. I gave this a shot and converted the main preferences screen just to get an idea of what people think.

**Changes**
- Replaced the Compose dependency bundle with a dependency on the [BOM](https://developer.android.com/jetpack/compose/bom/bom) as this should make it easier to add new compose libs.
- Added a Compose variant of the main preferences screen that matches the original almost perfectly.
  - Can only be turned on in development builds by setting `shouldUseCompose` to true.
  - Added the core UI components used to build preferences. They match their leanback counterparts as closely as I could get it.
  - Uses the latest alpha version of the [Compose TV library](https://developer.android.com/jetpack/androidx/releases/tv).

I thought this would be a fun experiment to modernize the app a bit more so please let me know what you think. If people are happy with this, I'm definitely keen on continuing this refactoring.

**Screenshot**

Screenshot_1679770789

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -133,7 +133,11 @@ dependencies {
 	implementation(libs.androidx.window)
 	implementation(libs.androidx.cardview)
 	implementation(libs.androidx.startup)
-	implementation(libs.bundles.androidx.compose)
+	val composeBom = platform(libs.androidx.compose.bom)
+	implementation(composeBom)
+	implementation(libs.androidx.compose.material)
+	implementation(libs.androidx.compose.ui.tooling.preview)
+	debugImplementation(libs.androidx.compose.ui.tooling)
 
 	// Dependency Injection
 	implementation(libs.bundles.koin)
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -133,9 +133,13 @@ dependencies {
 	implementation(libs.androidx.window)
 	implementation(libs.androidx.cardview)
 	implementation(libs.androidx.startup)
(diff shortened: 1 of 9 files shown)


nielsvanvelzen (commented):
Cool! This has been on my wish list for a bit but there was a reason I didn't start it yet. I've said this before in a few places but we won't be using Compose for interactive UI (so stuff that uses focus handling like buttons etc.) because of some issues with Compose. The TV support is still in early stages and regularly crashes. When those issues are fixed we will add compose in a few small places in a release so we can see how many issues pop up in the wild. Only after that will we actively start creating new UI with compose. So you may go ahead with this PR but it probably won't be reviewed for a while.

Regarding the dependencies; we don't use bill of materials because we already have a version catalog that is maintained by Renovate. We should also not use material & material3 at the same time (pick one, not both).

teobaranga:
Thank for the review! I found the branch you mentioned in #2193 and will have a look and get a bit more context around this focus issue.

I agree that mat3 and mat should not be used at the same time. The reason I had to add mat3 is because the navigation drawer in mat3 offered enough flexibility to make it match the leanback implementation.

In any case, sounds like this is not the right time for this PR as there need to be some things to work out first, including figuring out the focus / waiting for Google to fix it, waiting for a stable TV lib, as well as migrating to material3. Will revisit this once there's more progress around those, cheers!

PR #2641, opened 2023-04-05
Use ItemRowAdapterHelperKt.setItems for NextUpQuery in ItemRowAdapter

nielsvanvelzen:
**Changes**
- Use ItemRowAdapterHelperKt.setItems for NextUpQuery in ItemRowAdapter

This will prevent the selected item to reset when you go back to the home screen after watching an item.

**Issues**


crobibero (approved):
Changes look so much cleaner with whitespace ignored

PR #2644, opened 2023-04-05
Add (experimental) reactive homepage

nielsvanvelzen:
**Changes**
- Add (experimental) reactive homepage

When enabled, the home fragment will watch for updates from the server and reload **all** rows. The rows that use the ItemRowAdapterHelper will update their items while others will flicker. This functionality will do stuff like "live" reload the "next up" row when an episode is added to the server or watched on another device.

**Issues**


nielsvanvelzen:
Without #2641 this functionality is basically worthless.

crobibero (approved):

PR #2667, opened 2023-04-14
Add separate buttons for artists and album artists in music smart screen

nielsvanvelzen:
**Changes**
- Add separate buttons for artists and album artists in music smart screen
- Old "Artists" is renamed to "Album artists"
- Newly added "Artists" shows artists

Also adds some garbage because we couldn't retrieve artists from the ItemRowAdapter yet.

**Issues**


nielsvanvelzen:
image

Open for icon suggestions, otherwise it will be like this for now.

thornbill (approved):

PR #2670, opened 2023-04-16
Rewrite GridButton with Compose

nielsvanvelzen:
**Changes**
- Rewrite GridButton with Compose
- Add image-less grid button
- Use image-less grid button for everything except the "favorites" playlist in the music smart screen

Unfortunately the "active" effect cannot be changed for specific items without rewriting a lot of code so for now it uses the zoom effect instead of our regular background color change with buttons. This can be fixed when we convert even more stuff to compose.

**Issues**

**Screenshots**
image

top: music favorite button
bottom: image-less grid button


thornbill:
I always hated the way the icon card things looked. Definitely prefer the plain button styling! 💜

Issue #2614, opened 2023-03-22
Video transcodes with VobSub and PGS subtitles


branches78:
### Describe the bug

VobSub and PGS subtitles are not playing directly, that is, video is transcoding without audio, logs are attached.

[log_20230322.log](https://github.com/jellyfin/jellyfin-androidtv/files/11045314/log_20230322.log)

[VobSub.log](https://github.com/jellyfin/jellyfin-androidtv/files/11045315/VobSub.log)

[PGS.log](https://github.com/jellyfin/jellyfin-androidtv/files/11045316/PGS.log)

### Application version

0.15.5

### Device information

Homatics Box R 4K Plus and Realme 4K Stick

### Android version

Both on Android 11, security patches november 2022 and january 2023

### Jellyfin server version

10.8.9

branches78:
You should have not closed this ticket until I was able to test and report, although PGS subtitles now play directly without transcoding, I am getting playback errors when trying to play videos with VobSub subtitles, please check the attached log.

[VobSub.log.log](https://github.com/jellyfin/jellyfin-androidtv/files/11305568/VobSub.log.log)

Please reopen this ticket.

nielsvanvelzen:
We did not add support for VobSub which is a different format from PGS, please open a separate issue for that.

branches78:
> please open a separate issue for that.

Thank you for the clarification, I will open a new one at https://features.jellyfin.org/.

branches78:
I just created one:

https://features.jellyfin.org/posts/2151/please-add-support-for-vobsub-subtitles

PR #2690, opened 2023-04-22
Enable PGS subtitles in ExoPlayer

nielsvanvelzen:
ExoPlayer doesn't mention it in their documentation but they have an implementation for PGSSUB: https://github.com/google/ExoPlayer/blob/release-v2/library/core/src/main/java/com/google/android/exoplayer2/text/SubtitleDecoderFactory.java#L119-L120

**Changes**
- Enable PGS subtitles in ExoPlayer

**Issues**


crobibero (approved):

mikewesten:
Are there plans to support vobsub (sub/idx) without the need for transcoding as well?

branches78:
> Are there plans to support vobsub (sub/idx) without the need for transcoding as well?

I wonder the same, I just uploaded a log at ticket #2614.

nielsvanvelzen:
It doesn't look like vobsub is supported by ExoPlayer but please open an issue so we don't forget the demand.

mikewesten:
> It doesn't look like vobsub is supported by ExoPlayer but please open an issue so we don't forget the demand.

Okay, I'll make a feature request here.

Issue #2726, opened 2023-04-30
Search does not cancel when typing slow


polson:
### Describe the bug

1. Click search
2. Type a word like "Hello" slowly (ie 1 second between letters)

Observe that the search results are refreshed for every letter.  So the search results will be refreshed 5 times, even though we have fully typed out Hello.  This may not be a problem if the server is fast enough, but with a slow server the user will see the list update 5 times for each letter:

"H"
"He"
"Hel"
"Hell"
"Hello"

If the server takes a long time to respond, the user has to sit and wait for every one of these results to come back.

It would be better if the search call itself was cancelled each time a new letter is input.

This will not happen if the user types fast enough, because we are in fact debouncing input on the search box.  But if the user types slowly, this will become a problem.

I am willing to fix this myself and will put up a PR soon.

### Application version

0.15.8

### Where did you install the app from?

Google Play

### Device information

Nvidia Shield Pro

### Android version

Android 11

### Jellyfin server version

Any

nielsvanvelzen:
Should be fixed in 0.16.

PR #2727, opened 2023-04-30
Migrate search to SDK to fix cancellation issues

polson:
**Changes**
~~* Extends ItemItemRowAdapter and provides a suspending function for the search logic~~
~~* Suspending function is canceled when a new search happens~~
See comment below

The issue was caused by the old API client using Volley for HTTP requests, which launches a thread.  It does not use a coroutine / suspending function, so even though the coroutine is being canceled with new search input, the api call is not.

**Issues**
https://github.com/jellyfin/jellyfin-androidtv/issues/2726


polson:
I think I would like to change this by doing a small refactor:
* Share a viewmodel between the two search fragments. 
* Make the search call in the viewmodel.
* Pass the `SearchHint`s to the `ItemRowAdapter` directly

Then we can move the networking out of the ItemRowAdapter entirely.  I noticed we're already doing this in a few places.  There's `ItemRowAdapter` constructors that take in a list of `ChapterItemInfo` for example.

One annoying thing about this however is if we try to pass in a list of `SearchHint` it will conflict with that constructor because of type erasure.  So we might have to pass in a generic item list and do type checking on it.

nielsvanvelzen:
If we're migrating the search code to the SDK we should use the items API with the `searchTerm` parameter for the query instead. This API returns normal BaseItemDto models instead of search hints. The web client also uses this approach for searching.

nielsvanvelzen:
If we're migrating the search code to the SDK we should use the items API with the `searchTerm` parameter for the query instead. This API returns normal BaseItemDto models instead of search hints. The web client also uses this approach for searching.

polson:
Updated this to use the new search api

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchViewModel.kt
```diff
@@ -0,0 +1,70 @@
...
+		searchJob = viewModelScope.launch {
+			delay(debounceMs)
+			val deferredResults = groups.map { (stringRes, itemKinds) ->
+				async {
+					val result = SearchRepository.search(query, itemKinds).waitForSuccess()
+					val items = result?.content?.items ?: emptyList()
+					SearchHintGroup(stringRes, items)
+				}
+			}
+			val results = deferredResults.awaitAll()
```
polson, line 48: This will run all search calls in parallel, which is what the web client appears to be doing

app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
```diff
@@ -827,6 +812,11 @@ private void loadStaticAudioItems() {
         notifyRetrieveFinished();
     }
 
+    private void loadSearchItems() {
+        loadStaticItems();
+        addToParentIfResultsReceived();
+    }
```
nielsvanvelzen, line 818: This can be inlined

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchRepository.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.jellyfin.androidtv.auth.repository.UserRepository
+import org.jellyfin.androidtv.util.apiclient.AsyncState
+import org.jellyfin.sdk.api.client.ApiClient
+import org.jellyfin.sdk.api.client.extensions.itemsApi
+import org.jellyfin.sdk.model.api.BaseItemKind
+import org.jellyfin.sdk.model.api.ItemFields
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.inject
+
+object SearchRepository: KoinComponent {
```
nielsvanvelzen, line 14: This is not how our repositories work. We generally have an interface+implementation and do dependency injection via constructor.

app/src/main/java/org/jellyfin/androidtv/ui/search/LeanbackSearchFragment.kt
```diff
@@ -1,35 +1,50 @@
...
-import org.koin.java.KoinJavaComponent.inject
+import androidx.lifecycle.lifecycleScope
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
 
-class LeanbackSearchFragment : SearchSupportFragment() {
-	private val backgroundService = inject<BackgroundService>(BackgroundService::class.java)
+class LeanbackSearchFragment :
+	SearchSupportFragment(),
+	SearchSupportFragment.SearchResultProvider {
```
nielsvanvelzen, line 13:
```suggestion
class LeanbackSearchFragment : SearchSupportFragment(), SearchSupportFragment.SearchResultProvider {
```

app/src/main/java/org/jellyfin/androidtv/ui/search/LeanbackSearchFragment.kt
```diff
@@ -1,35 +1,50 @@
...
+		super.onViewCreated(view, savedInstanceState)
+		observeSearchResults()
+	}
+
+	private fun observeSearchResults() {
+		viewModel.searchResultsFlow
+			.onEach {
+				searchFragmentDelegate.showResults(it)
+			}.launchIn(lifecycleScope)
+	}
```
nielsvanvelzen, line 40: Can be inlined

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchViewModel.kt
```diff
@@ -0,0 +1,70 @@
...
+					SearchResultGroup(stringRes, items)
+				}
+			}
+			val results = deferredResults.awaitAll()
+			_searchResultsFlow.value = results
+		}
+		return true
+	}
+
+	companion object {
```
nielsvanvelzen, line 54: We normally keep our companion objects in the top of a class

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchViewModel.kt
```diff
@@ -0,0 +1,70 @@
...
+class SearchViewModel : ViewModel() {
+
+	private var searchJob: Job? = null
+
+	private var previousQuery: String? = null
+
+	private val _searchResultsFlow = MutableStateFlow<List<SearchResultGroup>>(emptyList())
+	val searchResultsFlow = _searchResultsFlow.asStateFlow()
+
+	fun search(query: String) = searchDebounced(query, 0.milliseconds)
```
nielsvanvelzen, line 27: Maybe call this one "searchImmediatly" so it's more obvious what it does

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchResultGroup.kt
```diff
@@ -0,0 +1,9 @@
+package org.jellyfin.androidtv.ui.search
+
+import androidx.annotation.StringRes
+import org.jellyfin.sdk.model.api.BaseItemDto
+
+data class SearchResultGroup(
+	@StringRes val labelRes: Int,
+	val items: List<BaseItemDto>
```
nielsvanvelzen, line 8:
```suggestion
	val items: Collection<BaseItemDto>,
```

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchRepository.kt
```diff
@@ -0,0 +1,43 @@
...
+
+	private const val QUERY_LIMIT = 25
+
+	private val apiClient: ApiClient by inject()
+
+	suspend fun search(
```
nielsvanvelzen, line 20: Just have this return `Result<BaseItemDtoQueryResult>` instead of the AsyncState. Introducing that to our repositories might make sense but would need to be a separate pull request.

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchRepository.kt
```diff
@@ -0,0 +1,43 @@
...
+			searchTerm = searchTerm,
+			limit = QUERY_LIMIT,
+			imageTypeLimit = 1,
+			includeItemTypes = itemTypes,
+			fields = listOf(
+				ItemFields.PRIMARY_IMAGE_ASPECT_RATIO,
+				ItemFields.CAN_DELETE,
+				ItemFields.BASIC_SYNC_INFO,
+				ItemFields.MEDIA_SOURCE_COUNT
+			),
```
nielsvanvelzen, line 35: We can probably use `StdItemQuery.defaultFields` here, since those are also rendered by the same presenter.

app/src/main/java/org/jellyfin/androidtv/ui/search/TextSearchFragment.kt
```diff
@@ -3,78 +3,92 @@ package org.jellyfin.androidtv.ui.search
...
 
-	override fun onCreate(savedInstanceState: Bundle?) {
-		super.onCreate(savedInstanceState)
+	private val viewModel by viewModels<SearchViewModel>()
 
-		// Create provider
-		searchProvider = SearchProvider(requireContext(), lifecycle)
-	}
+	private var _binding: FragmentSearchTextBinding? = null
+	private val binding get() = _binding!!
```
nielsvanvelzen, line 24:
```suggestion
	private lateinit var binding: FragmentSearchTextBinding
```
We normally do it like this

app/src/main/java/org/jellyfin/androidtv/ui/search/TextSearchFragment.kt
```diff
@@ -3,78 +3,92 @@ package org.jellyfin.androidtv.ui.search
...
+	override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+		_binding = FragmentSearchTextBinding.inflate(inflater, container, false)
+		return binding.root
+	}
 
 	override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
 		super.onViewCreated(view, savedInstanceState)
+		setupSearchBar()
+		setupResultFragment()
+		observeSearchResults()
```
nielsvanvelzen, line 37: These three functions can be inlined

app/src/main/java/org/jellyfin/androidtv/ui/search/TextSearchFragment.kt
```diff
@@ -3,78 +3,92 @@ package org.jellyfin.androidtv.ui.search
...
+			.onEach {
+				searchFragmentDelegate.showResults(it)
+			}.launchIn(lifecycleScope)
+	}
 
-	override fun afterTextChanged(editable: Editable) {
-		searchProvider.onQueryTextChange(editable.toString())
+	override fun onDestroyView() {
+		_binding = null
+		super.onDestroyView()
```
nielsvanvelzen, line 70: This shouldn't be necessary

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchFragmentDelegate.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.jellyfin.androidtv.ui.itemhandling.ItemLauncher
+import org.jellyfin.androidtv.ui.itemhandling.ItemRowAdapter
+import org.jellyfin.androidtv.ui.presentation.CardPresenter
+import org.jellyfin.androidtv.ui.presentation.CustomListRowPresenter
+import org.jellyfin.androidtv.ui.presentation.MutableObjectAdapter
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.inject
+
+class SearchFragmentDelegate(
+	private val fragment: Fragment
```
nielsvanvelzen, line 21: You can use a context here instead of the fragment.

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchFragmentDelegate.kt
```diff
@@ -0,0 +1,56 @@
...
+) : KoinComponent {
```
nielsvanvelzen, line 22: Use dependency injection via constructor instead

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchFragmentDelegate.kt
nielsvanvelzen: Looks like ItemLauncher still requires an activity, I'll open a PR to fix that so you can rebase on top of master once that is merged

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchFragmentDelegate.kt
```diff
@@ -0,0 +1,56 @@
...
+	val onItemViewClickedListener =
+		OnItemViewClickedListener { _, item, _, row ->
+			if (item !is BaseRowItem) return@OnItemViewClickedListener
+			row as ListRow
+			val adapter = row.adapter as ItemRowAdapter
+			ItemLauncher.launch(item as BaseRowItem?, adapter, item.index, fragment.requireActivity())
+		}
+
+	val onItemViewSelectedListener = OnItemViewSelectedListener { _, item, _, _ ->
+		val searchHint = item?.let { (item as BaseRowItem).searchHint }
```
nielsvanvelzen, line 49: This line needs to be updated to not use search hint (because we're not using it anymore)

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/search/TextSearchFragment.kt
polson: According to google

```
Note: Fragments outlive their views. Make sure you clean up any references to the binding class instance in the fragment's [onDestroyView()](https://developer.android.com/reference/kotlin/androidx/fragment/app/Fragment#ondestroyview) method.
```

https://developer.android.com/topic/libraries/view-binding#fragments

But if you still want to change it I will

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchRepository.kt
polson: No problem.  At some point I would like to add a loading spinner somewhere.  Will handle that elsewhere.

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchRepository.kt
polson: It seems `StdItemQuery.defaultFields` uses `org.jellyfin.apiclient.model.querying`

But the api call requires `org.jellyfin.sdk.model.api`

I don't see a nice way to convert between them

app/src/main/java/org/jellyfin/androidtv/ui/search/TextSearchFragment.kt
nielsvanvelzen: No thanks for linking, we should probably change our fragments to mirror this approach. I believe this wasn't in the Android docs when we started using view binding.

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchRepository.kt
nielsvanvelzen: Ok nvm then, let's keep it like this for now 👍

app/src/main/java/org/jellyfin/androidtv/ui/search/TextSearchFragment.kt
polson: FWIW I've never seen a memory leak happen because the `viewBinding` wasn't cleared, but I guess it's possible somehow

app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapter.java
polson: Done

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchRepository.kt
polson: Updated, please ensure I did this correctly

app/src/main/java/org/jellyfin/androidtv/ui/search/LeanbackSearchFragment.kt
polson: Done

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchViewModel.kt
polson: Done

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchViewModel.kt
polson: Changed

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchRepository.kt
polson: Updated

polson:
Addressed comments

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchRepository.kt
```diff
@@ -0,0 +1,45 @@
...
+                            ItemFields.PRIMARY_IMAGE_ASPECT_RATIO,
+                            ItemFields.CAN_DELETE,
+                            ItemFields.BASIC_SYNC_INFO,
+                            ItemFields.MEDIA_SOURCE_COUNT
+                    ),
+                    recursive = true,
+                    enableTotalRecordCount = false,
+            )
+            result.content.items ?: emptyList()
+        } catch (e: IOException) {
```
polson, line 41: Need to double check that IOException is what gets thrown on an http error

Otherwise we can catch `Exception`, if you'd rather just do that

polson: Nevermind, looks like `ApiClientException` is what we want

app/src/main/java/org/jellyfin/androidtv/di/SearchModule.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.jellyfin.playback.core.playbackManager
+import org.jellyfin.playback.exoplayer.exoPlayerPlugin
+import org.jellyfin.playback.jellyfin.jellyfinPlugin
+import org.koin.core.scope.Scope
+import org.koin.dsl.module
+import org.jellyfin.androidtv.ui.playback.PlaybackManager as LegacyPlaybackManager
+
+val searchModule = module {
+	factory { (context: Context) ->
+		SearchFragmentDelegate(context, get())
```
nielsvanvelzen, line 32: Let's put this in the app module for now, adding a module for a single instance doesn't seem worth it.

app/src/main/java/org/jellyfin/androidtv/ui/search/LeanbackSearchFragment.kt
```diff
@@ -1,35 +1,47 @@
...
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+import org.koin.android.ext.android.inject
+import org.koin.androidx.viewmodel.ext.android.viewModel
+import org.koin.core.parameter.parametersOf
 
-class LeanbackSearchFragment : SearchSupportFragment() {
-	private val backgroundService = inject<BackgroundService>(BackgroundService::class.java)
+class LeanbackSearchFragment : SearchSupportFragment(),
+	SearchSupportFragment.SearchResultProvider {
```
nielsvanvelzen, line 14:
```suggestion
class LeanbackSearchFragment : SearchSupportFragment(), SearchSupportFragment.SearchResultProvider {
```

app/src/main/java/org/jellyfin/androidtv/ui/search/LeanbackSearchFragment.kt
```diff
@@ -1,35 +1,47 @@
...
 
-		// Create provider
-		val searchProvider = SearchProvider(requireContext(), lifecycle)
-		setSearchResultProvider(searchProvider)
+	override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+		super.onViewCreated(view, savedInstanceState)
+		viewModel.searchResultsFlow
+				.onEach {
+					searchFragmentDelegate.showResults(it)
+				}.launchIn(lifecycleScope)
```
nielsvanvelzen, line 36:
```suggestion
				.onEach { searchFragmentDelegate.showResults(it) }
				.launchIn(lifecycleScope)
```

app/src/main/java/org/jellyfin/androidtv/ui/search/TextSearchFragment.kt
```diff
@@ -3,78 +3,81 @@ package org.jellyfin.androidtv.ui.search
...
+				false
+			}
+		}
+	}
 
-		// Return "false" to automatically close keyboard
-		return false
+	private fun EditText.onTextChanged(onTextChanged: (String) -> Unit) {
+		val textWatcher = object : TextWatcher {
+			override fun afterTextChanged(s: Editable?) {
```
nielsvanvelzen, line 75:
```suggestion
			override fun afterTextChanged(s: Editable) {
```

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchRepository.kt
```diff
@@ -0,0 +1,46 @@
...
+        private val apiClient: ApiClient
+) : SearchRepository {
+    companion object {
+        private const val QUERY_LIMIT = 25
+    }
+
+    override suspend fun search(
+            searchTerm: String,
+            itemTypes: Collection<BaseItemKind>
+    ): List<BaseItemDto> {
```
nielsvanvelzen, line 25: If we use Result<> here instead of emptyList on failure we can deal with errors in the usage side. Right now errors are ignored (not even logged). The interface would use `Result<List<BaseItemDto>>` as return type and this function would use either 
`Result.success(result.content.items.orEmpty())` or `Result.failure(e)` (in the catch)

nielsvanvelzen (changes requested):
Some files are indented with spaces instead of tabs, and a lot of places use (like constructors) use 2x the amount of indent compared to existing code.

polson:
Sorry about the whitespace changes, I tend to use the AS formatter quite frequently so that's why it moved things around

app/src/main/java/org/jellyfin/androidtv/ui/search/SearchRepository.kt
polson: I'll look into this, at first glance I didn't see a way to do that with `org.jellyfin.sdk.api.client.Response`

nielsvanvelzen:
Hmm that's weird, the default Android Studio config + our editorconfig should configure the formatter properly. I use it all the time. Maybe it broke in a newer version? I'll look into it.

nielsvanvelzen (approved):
I've made some additional changes;
- Updates formatting
- Moved `searchJob?.cancel()` to before the query.isBlank check to prevent a race condition where items wouldn't be cleared when emptying the input.

Thanks a lot for the pull request!

PR #2743, opened 2023-05-07
Fix loading issues with embedded SUBRIP subtitles while using Exoplayer

### LINKED ISSUE #145: [Bug] Subtitles Don't Always Load
Subtitles don't always load, tested on Nvidia Shield TV and a Mibox.
Some films they do and others they don't.

On the films they don't, this usually occurs with me selecting the subtitle I want, then it says "Loading subtitles", I let it sit for a minute or two, sometimes it just lingers and most times it ends up saying "Subtitles could not be loaded"

(Unrelated but another issue I've come across with subtitles is sometimes I would click "No subtitles" to disable them, they go away but then about 30sec later they turn back on, I go back and manually disable them again (The "no subtitles" is still selected which is odd, I just click it again) and then it's fine throughout the rest of the show.

nathkn:
Currently, embedded SRT/SUBRIP subtitles can take a long time to load because it requires the server to use ffmpeg to extract the embedded subtitles out of the video stream in order to return a JSON file of the subtitles so they can be used by the custom overlay. A longer explanation is [here](https://github.com/jellyfin/jellyfin-androidtv/issues/145#issuecomment-1121735121).  For large videos (such as 4K), this can take minutes to load a single subtitle. After these fixes, they load near instantaneously by Exoplayer.

**Changes**
* Enables the SRT/SUBRIP embedded subtitle profiles, which means they will be displayed directly by Exoplayer instead of being shown through the custom overlay fragment
* In order to keep parity with the custom overlay's ability to change the formatting of the subtitles based on user preferences, this PR also implements subtitle formatting within Exoplayer, since otherwise those settings would be lost when using Exoplayer directly for embedded subtitles
* Some of these code changes have been influenced by this previous attempt to remedy the issue: https://github.com/jellyfin/jellyfin-androidtv/pull/1949

**Issues**
Fixes #145

(Note: I'm not a Java developer and haven't contributed to this project before, so any and all suggestions for how to implement this better are welcomed!)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
@@ -2,7 +2,9 @@ package org.jellyfin.androidtv.ui.playback;
 
 import android.content.Context;
 import android.content.res.Configuration;
+import android.graphics.Color;
 import android.graphics.PixelFormat;
+import android.graphics.Point;
 import android.media.audiofx.DynamicsProcessing;
 import android.media.audiofx.DynamicsProcessing.Limiter;
 import android.media.audiofx.Equalizer;
@@ -15,6 +17,7 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.FrameLayout;
 
+import androidx.annotation.Dimension;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (commented):
I think it's fine to replace our custom subtitle rendering with the one provided by ExoPlayer (in fact, this is one of my planned changes with #1057), but we should do that for all (text-)subtitles to make the behavior consistent. That means both embedded and external subtitles should use it and our custom subtitle rendering code must be removed.

nathkn:
@nielsvanvelzen I definitely agree with that - this PR was more of a quick fix until without needing modify a ton of code that was going to be overhauled with the playback rewrite anyway. 

I could take a shot at removing the custom subtitle rendering and have ExoPlayer handle external and embedded subtitles if you wanted? Could also do that after this PR is merged too - let me know how you'd prefer I proceed

nielsvanvelzen:
You can update this PR to completely remove the custom subtitle stuff if you want. I won't merge it as-is.

nielsvanvelzen:
Hey @nathkn, are you still working on this PR/planning to finish it?

nielsvanvelzen:
Closing this PR as abandoned.

PR #2745, opened 2023-05-09
Migrate video player & next up screen to fragments

nielsvanvelzen:
This PR migrates the video player and next up screen to fragments. It required some additional tweaks to the navigation (add replace option), clock (don't use activity-detection), external player (drop next up screen support) and of course the video player.

**Changes**
- Migrate video player & next up screen to fragments
- Support replacement during navigation
- Remove NextUp screen from ExternalPlayerActivity

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -428,34 +431,66 @@ public void handleOnBackPressed() {
...
+                mPlaybackController.play(0);
+                return true;
+            } else if (keyCode == KeyEvent.KEYCODE_MEDIA_PAUSE) {
+                mPlaybackController.pause();
+                return true;
+            } else if (keyCode == KeyEvent.KEYCODE_MEDIA_PLAY_PAUSE) {
+                mPlaybackController.playPause();
+                return true;
+            } else if (keyCode == KeyEvent.KEYCODE_MEDIA_FAST_FORWARD || keyCode == KeyEvent.KEYCODE_BUTTON_R1 || keyCode == KeyEvent.KEYCODE_BUTTON_R2) {
+                mPlaybackController.skip(30000);
```
crobibero, line 485: I assume this was pulled from somewhere but to me it just looks like magic :)

crobibero (approved):

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
nielsvanvelzen: Keyhandling code was moved from the activity to the fragment. so this is pretty much copy-paste.

PR #2783, opened 2023-05-20
Add playback order support to playback rewrite

nielsvanvelzen:
**Changes**

This change adds a new "playbackOrder" property to the playback state. The currently supported values are:

- default - this plays back items in the default order of the queue
- random - this plays random items, infinitely
- shuffle - this plays items in a shuffled order, this is the same as random but every item in the queue can only be played once

Changing the order property only affects the upcoming items in a queue. This means you can play the first 3 items using default order and then switch to shuffling. The queue service keeps track of this and manages the ordering.

To implement this ordering we now need to know the size of a queue, for this a new "size" property is added to the queue interface. I've implemented this properly for the queue used by the RewriteMediaManager but the other queue implementations that aren't used yet have implementations that may not work (expose size=0 until the first getItem call which will never happen because size=0). This will be fixed later, when we start using those queues.

The current playback UI only allows switching between default & shuffle. The random option is not used.

**Issues**

Part of #1057

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/rewrite/RewriteMediaManager.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/rewrite/RewriteMediaManager.kt
@@ -21,6 +21,7 @@ import org.jellyfin.androidtv.ui.playback.PlaybackController
 import org.jellyfin.androidtv.ui.presentation.CardPresenter
 import org.jellyfin.playback.core.PlaybackManager
 import org.jellyfin.playback.core.model.PlayState
+import org.jellyfin.playback.core.model.PlaybackOrder
 import org.jellyfin.playback.core.queue.Queue
 import org.jellyfin.playback.core.queue.item.QueueEntry
 import org.jellyfin.playback.core.queue.queue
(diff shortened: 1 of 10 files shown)


neopc10:
For videos can this be applied only to 'unwatched/played' items?

nielsvanvelzen:
> For videos can this be applied only to 'unwatched/played' items?

This code has nothing to do with the watched state of items or even videos at all. This code only applies to music playback right now.

app/src/main/java/org/jellyfin/androidtv/ui/playback/rewrite/RewriteMediaManager.kt
```diff
@@ -69,8 +70,8 @@ class RewriteMediaManager(
 		private set
 
 	override val isAudioPlayerInitialized: Boolean = true
-	override var isShuffleMode: Boolean = false
-		private set
+	override val isShuffleMode: Boolean
+		get() = playbackManager.state.playbackOrder.value == PlaybackOrder.RANDOM
```
thornbill, line 74: Should this be shuffle or random?

nielsvanvelzen: Should be
```suggestion
		get() = playbackManager.state.playbackOrder.value != PlaybackOrder.DEFAULT
```

crobibero (approved):

PR #2788, opened 2023-05-23
Fix navigating back from an actor page should keep focus on the actors row 

### LINKED ISSUE #2713: Navigating back from an actor page should keep focus on the actors row
### Describe the bug

1. Navigate to a movie or show
2. Scroll down to the actors' row and click on an actor.
3. Click 'back'
4. Notice that the focus has reverted to the top of the movie / show page and is no longer on the actors' row.

I know this one is not a very serious defect, but it annoys me each time I read about the various actors in a movie and have to come back down to the actors' row **every** time.

In general, I think that if a specific control is used to enter a view, clicking back from that view should end with focus on same control, unless there's a good reason not to have this be the case.

### Logs

_No response_

### Application version

0.15.7

Google Play

### Device information

Chromecast with Google TV

### Android version

12

### Jellyfin server version

10.8.10

fedesenmartin:
Changes
- Use childfragmentManager instead of supportfragmentmanger at `FullDetailsFragment.java` and `AudioNowPlayingFragment.java`
- indentation for `FullDetailsFragment.java`

Issues:
Fixes #2713


nielsvanvelzen:
It looks like the same issue also exists in the AudioNowPlayingFragment. Perhaps you can fix that file in this same pull request.

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
```diff
@@ -445,7 +448,9 @@ protected MyDetailsOverviewRow doInBackground(BaseItemDto... params) {
             Double aspect = ImageUtils.getImageAspectRatio(ModelCompat.asSdk(item), false);
             posterHeight = aspect > 1 ? Utils.convertDpToPixel(requireContext(), 160) : Utils.convertDpToPixel(requireContext(), ModelCompat.asSdk(item).getType() == BaseItemKind.PERSON || ModelCompat.asSdk(item).getType() == BaseItemKind.MUSIC_ARTIST ? 300 : 200);
 
-            mDetailsOverviewRow = new MyDetailsOverviewRow(ModelCompat.asSdk(item));
+            if(mDetailsOverviewRow == null){
```
nielsvanvelzen, line 451:
```suggestion
            if (mDetailsOverviewRow == null) {
```

app/src/main/java/org/jellyfin/androidtv/ui/presentation/MyDetailsOverviewRowPresenter.kt
```diff
@@ -59,7 +59,11 @@ class MyDetailsOverviewRowPresenter(
 			}
 
 			binding.fdButtonRow.removeAllViews()
-			for (button in row.actions) binding.fdButtonRow.addView(button)
+			for (button in row.actions) {
+				if (button.parent != null) {
```
nielsvanvelzen, line 63: This check makes the buttons never show at all.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/presentation/MyDetailsOverviewRowPresenter.kt
fedesenmartin: removed from "MyDetailsOverViewRow" all views; but without this check `if (button.parent != null)` I cant get rid off  -> `java.lang.IllegalStateException: The specified child already has a parent. You must call removeView() on the child's parent first.`

bentoti (approved):

app/src/main/java/org/jellyfin/androidtv/ui/presentation/MyDetailsOverviewRowPresenter.kt
nielsvanvelzen: You're storing the views in the presenter instead of the viewholder right now. That means views are still re-used. You need to make sure to never re-use the views and scope them to the viewholder.

fedesenmartin:
I am not actively working on the fix. I have an idea on how to solve it, but it requires rewriting the entire view. I will continue it when I have a moment, unless there is a significant change in the code. I will leave it as a draft if that's okay with you, in case someone wants to continue it. Does that sound okay to you? @nielsvanvelzen

nielsvanvelzen:
While fixing a crash in the initial 0.16 beta I made similar changes as this PR (#2939, #2940 and #2943). The mentioned issue (#2713) might be fixed now in the master branch. I've not tested it myself though.

nielsvanvelzen:
Closing this PR as stale.

PR #2803, opened 2023-05-30
Rewrite media session code with androidx.media3

nielsvanvelzen:
The current media session implementation had a few issues:

- Rare crashes due to the threading not implemented correctly
- Missing a lot of features
- Messy
- Uses androidx.media2 which almost nobody does (because Google documentation uses appcompat which is weird..)

In this PR the implementation is rewritten with androidx.media3 and most missing features are added. It works great in my testing (emulator, phone, shield).

**Changes**
- Rewrite media session code with androidx.media3
- Update QueueEntryMetadata to be based on media3 properties
- Add missing features to media session

**Issues**

Part of #1057


playback/jellyfin/src/main/kotlin/queue/item/BaseItemDtoUserQueueEntry.kt
```diff
@@ -21,23 +20,29 @@ class BaseItemDtoUserQueueEntry private constructor(
...
 			val metadata = QueueEntryMetadata(
 				mediaId = baseItem.id.toString(),
+				duration = baseItem.runTimeTicks?.ticks,
 				title = baseItem.name,
 				artist = baseItem.albumArtist,
-				album = baseItem.album,
-				date = baseItem.dateCreated,
-				displayDescription = baseItem.overview,
+				albumTitle = baseItem.album,
+				albumArtist = baseItem.albumArtists?.mapNotNull { it.name }?.joinToString(","),
```
crobibero, line 27: Do we want this to be `", "` with a space like L45

Issue #2811, opened 2023-06-01
InfoLayoutHelper not displaying resolution and video codec tags when external subtitles are available


StubbyB:
### Describe the bug

On Android TV some tags are shown such as 'R', '4K' 'HEVC', etc.
Titles which have external subtitles, are missing the resolution and video codec tags (e.g. '1080' and 'H264').

### Application version

0.15.9

### Where did you install the app from?

Google Play

### Device information

Nvidia Shield

### Android version

Android 9

### Jellyfin server version

10.8.4

PR #2812, opened 2023-06-01
Fix InfoLayoutHelper not using first video stream, update details when selecting version

StubbyB:
On Android TV some tags are shown such as 'R', '4K' 'HEVC', etc.
Titles which have external subtitles, are missing the resolution and video codec tags (e.g. '1080' and 'H264').

The source of the issue is `org.jellyfin.androidtv.util.InfoLayoutHelper.addRatingAndRes` getting all media streams, and assuming that the video stream will be the first one, when in fact external subtitles are first (also verifiable by looking at the debug info at the bottom).

**Changes**
Instead of assuming the video is the first stream, take the first stream tagged as `MediaStreamType.VIDEO`.

**Issues**
#2811

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/MyDetailsOverviewRow.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/MyDetailsOverviewRow.kt
@@ -14,7 +14,7 @@ class MyDetailsOverviewRow @JvmOverloads constructor(
 	var infoItem1: InfoItem? = null,
 	var infoItem2: InfoItem? = null,
 	var infoItem3: InfoItem? = null,
-	var selectedMediaSourceIndex: Int = 0
+	var selectedMediaSourceIndex: Int = 0,
 ) : Row() {
 	private val _actions = mutableListOf<TextUnderButton>()
 	val actions get() = _actions.toList()
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/MyDetailsOverviewRow.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/MyDetailsOverviewRow.kt
@@ -14,6 +14,7 @@ class MyDetailsOverviewRow @JvmOverloads constructor(
 	var infoItem1: InfoItem? = null,
 	var infoItem2: InfoItem? = null,
 	var infoItem3: InfoItem? = null,
+	var selectedMediaSourceIndex: Int = 0
 ) : Row() {
 	private val _actions = mutableListOf<TextUnderButton>()
 	val actions get() = _actions.toList()
--- a/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
@@ -45,6 +45,7 @@ public class InfoLayoutHelper {
                 break;
         }
     }
+
     public static void addInfoRow(Context context, BaseItemDto item, int mediaSourceIndex, LinearLayout layout, boolean includeRuntime, boolean includeEndTime) {
         layout.removeAllViews();
         if (item.getId() != null) {
@@ -336,7 +337,7 @@ public class InfoLayoutHelper {
 
         MediaStream videoStream = StreamHelper.getFirstVideoStream(item, mediaSourceIndex);
 
-        if(videoStream != null && videoStream.getWidth() != null && videoStream.getHeight() != null) {
+        if (videoStream != null && videoStream.getWidth() != null && videoStream.getHeight() != null) {
             int width = videoStream.getWidth();
             int height = videoStream.getHeight();
             if (width <= 960 && height <= 576) {
--- a/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
@@ -336,26 +336,24 @@ public class InfoLayoutHelper {
 
         MediaStream videoStream = StreamHelper.getFirstVideoStream(item, selectedVersionPopupIndex);
 
-        if (videoStream != null) {
-            if(videoStream.getWidth() != null && videoStream.getHeight() != null) {
-                int width = videoStream.getWidth();
-                int height = videoStream.getHeight();
-                if (width <= 960 && height <= 576) {
-                    addBlockText(context, layout, context.getString(R.string.lbl_sd));
-                } else if (width <= 1280 && height <= 962) {
-                    addBlockText(context, layout, "720");
-                } else if (width <= 1920 && height <= 1440) {
-                    addBlockText(context, layout, "1080");
-                } else if (width <= 4096 && height <= 3072) {
-                    addBlockText(context, layout, "4K");
-                } else {
-                    addBlockText(context, layout, "8K");
-                }
-
-                addSpacer(context, layout, " ");
-
-                addVideoCodecDetails(context, layout, videoStream);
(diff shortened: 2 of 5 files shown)


nielsvanvelzen (changes requested):
We already have a `StreamHelper.getFirstAudioStream` function so it would probably make sense to add a `StreamHelper.getFirstVideoStream` or maybe even merge them to a `StreamHelper.getFirstStreamOfType()`.

StubbyB:
> We already have a `StreamHelper.getFirstAudioStream` function so it would probably make sense to add a `StreamHelper.getFirstVideoStream` or maybe even merge them to a `StreamHelper.getFirstStreamOfType()`.

Moved logic to `StreamHelper`.
Also, since the order of the video versions in the selection menu is the order of the media sources, I added the functionality of settings the tags given a selected media source index, with 0 being the default (preserving existing behavior).
Had to add function signatures accordingly.

StubbyB:
made a few cosmetic changes, done modifying until reviewed :) @nielsvanvelzen

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/MyDetailsOverviewRow.kt
```diff
@@ -14,6 +14,7 @@ class MyDetailsOverviewRow @JvmOverloads constructor(
 	var infoItem1: InfoItem? = null,
 	var infoItem2: InfoItem? = null,
 	var infoItem3: InfoItem? = null,
+	var selectedMediaSourceIndex: Int = 0
```
nielsvanvelzen, line 17:
```suggestion
	var selectedMediaSourceIndex: Int = 0,
```

app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
```diff
@@ -45,17 +45,20 @@ public static void addInfoRow(Context context, BaseRowItem item, LinearLayout la
                 break;
         }
     }
-
-    public static void addInfoRow(Context context, BaseItemDto item, LinearLayout layout, boolean includeRuntime, boolean includeEndTime) {
+    public static void addInfoRow(Context context, BaseItemDto item, int mediaSourceIndex, LinearLayout layout, boolean includeRuntime, boolean includeEndTime) {
```
nielsvanvelzen, line 48: (missing the empty line before the function)
```suggestion

    public static void addInfoRow(Context context, BaseItemDto item, int mediaSourceIndex, LinearLayout layout, boolean includeRuntime, boolean includeEndTime) {
```

app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
```diff
@@ -325,14 +328,17 @@ private static void addDate(@NonNull Context context, BaseItemDto item, LinearLa
...
             addBlockText(context, layout, item.getOfficialRating());
             addSpacer(context, layout, "  ");
         }
-        if (item.getMediaStreams() != null && item.getMediaStreams().size() > 0 && item.getMediaStreams().get(0).getWidth() != null && item.getMediaStreams().get(0).getHeight() != null) {
-            int width = item.getMediaStreams().get(0).getWidth();
-            int height = item.getMediaStreams().get(0).getHeight();
+
+        MediaStream videoStream = StreamHelper.getFirstVideoStream(item, mediaSourceIndex);
+
+        if(videoStream != null && videoStream.getWidth() != null && videoStream.getHeight() != null) {
```
nielsvanvelzen, line 339:
```suggestion
        if (videoStream != null && videoStream.getWidth() != null && videoStream.getHeight() != null) {
```

app/src/main/java/org/jellyfin/androidtv/util/apiclient/StreamHelper.java
```diff
@@ -26,9 +26,33 @@ public static List<MediaStream> getAudioStreams(MediaSourceInfo mediaSource) {
...
+        return getStreams(mediaSource, MediaStreamType.VIDEO);
+    }
+
     public static MediaStream getFirstAudioStream(BaseItemDto item) {
-        if (item.getMediaSources() == null || item.getMediaSources().size() < 1) return null;
-        List<MediaStream> streams = getAudioStreams(item.getMediaSources().get(0));
+        return getFirstAudioStream(item, 0);
+    }
+
+    public static MediaStream getFirstAudioStream(BaseItemDto item, int mediaSourceIndex) {
```
nielsvanvelzen, line 37:
```suggestion
    public static MediaStream getAudioStream(BaseItemDto item, int mediaSourceIndex) {
```
It's not the first if we explicitly tell it the index.

app/src/main/java/org/jellyfin/androidtv/util/apiclient/StreamHelper.java
```diff
@@ -26,9 +26,33 @@ public static List<MediaStream> getAudioStreams(MediaSourceInfo mediaSource) {
...
+        return getFirstStreamOfType(item, MediaStreamType.AUDIO, mediaSourceIndex);
+    }
+
+    public static MediaStream getFirstVideoStream(BaseItemDto item) {
+        return getFirstVideoStream(item, 0);
+    }
+
+    public static MediaStream getFirstVideoStream(BaseItemDto item, int mediaSourceIndex) {
```
nielsvanvelzen, line 45:
```suggestion
    public static MediaStream getVideoStream(BaseItemDto item, int mediaSourceIndex) {
```

app/src/main/java/org/jellyfin/androidtv/util/apiclient/StreamHelper.java
```diff
@@ -26,9 +26,33 @@ public static List<MediaStream> getAudioStreams(MediaSourceInfo mediaSource) {
...
+        return getFirstStreamOfType(item, MediaStreamType.VIDEO, mediaSourceIndex);
+    }
+
+    public static MediaStream getFirstStreamOfType(BaseItemDto item, MediaStreamType streamType) {
+        return getFirstStreamOfType(item, streamType, 0);
+    }
+
+    public static MediaStream getFirstStreamOfType(BaseItemDto item, MediaStreamType streamType, int mediaSourceIndex) {
```
nielsvanvelzen, line 53:
```suggestion
    public static MediaStream getStreamOfType(BaseItemDto item, MediaStreamType streamType, int mediaSourceIndex) {
```

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/apiclient/StreamHelper.java
StubbyB: It is the first matching `MediaStream` for the given `MediaSource`.

The idea was to allow the existing functionality of retrieving the first overall media stream, and also an overload allowing to choose a specific source.

Also `getFirstAudioStream` is an existing public util which I was reluctant to change.

Would you rather it be a separate function signature named something like:
`getFirstStreamOfSource(BaseItemDto item, int mediaSourceIndex)`
`getFirstStreamOfSource(BaseItemDto item, int mediaSourceIndex, MediaStreamType streamType)`?

StubbyB: This applies to all other comments ofc

nielsvanvelzen: Aah you're right, I missed that part. This naming is fine 👍.

nielsvanvelzen (approved):

nielsvanvelzen:
Thank you!

PR #2826, opened 2023-06-06
Replace ExoPlayer with androidx.media3

nielsvanvelzen:
**Changes**

Replace our usages of ExoPlayer with androidx.media3. This affects the old playback code and the rewritten code. It's pretty much only package name changes because I fixed the breaking changes in earlier PR's.

**To do**

- ~~Migrate `exoplayer-ffmpeg-extension` to media3~~

**Issues**

Also a part of #1057 I guess

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -139,8 +139,10 @@ dependencies {
 	implementation(libs.bundles.koin)
 
 	// Media players
-	implementation(libs.exoplayer)
-	implementation(libs.jellyfin.exoplayer.ffmpegextension)
+	implementation(libs.androidx.media3.exoplayer)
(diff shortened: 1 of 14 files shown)


nielsvanvelzen:
This is now ready. I'll likely add it to the 0.16 release branch later.

crobibero (approved):

masterflitzer:
does **backportable** mean it'll land in v0.16.3?

nielsvanvelzen:
No necessarily. The backportable label means it is a candidate for backporting. In this specific case I need to test the media3 migration more thoroughly first before sending it to half a million users. So it might be in 0.16.3 but could also be 0.16.4 or 0.17.0.

masterflitzer:
ok thank you for clarifying 👍

PR #2864, opened 2023-07-01
Show UI mode notification on startup screen

nielsvanvelzen:
This PR shows the UI mode warning on the startup screen. The login UI is not really usable on phones so we can't rely on the notification after sign in anymore.

**Changes**
- Add "public" option to AppNotification
  - Currently only used for the UI mode warning
- Show public notification on server selection screen
  - This is the initial screen that shows on a fresh install
- Sneak in a little build script warning fix

image

**Issues**


nielsvanvelzen:
Note: already backported to the release branch with some slight changes for the older library versions.

crobibero (approved):

PR #2893, opened 2023-07-13
Improve codec info

mohd-akram:
**Changes**
- Move CC ~~after audio codec info~~ before format info
- Specify interlaced or progressive in resolution
- Specify if Dolby Vision or other HDR
- Specify if Dolby Atmos or DTS:X
- Specify if DD+ or DTS-HD, rename Dolby to DD

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
@@ -335,17 +335,23 @@ public class InfoLayoutHelper {
             addSpacer(context, layout, "  ");
         }
 
+        if (Utils.isTrue(item.getHasSubtitles())) {
+            addBlockText(context, layout, "CC");
+            addSpacer(context, layout, "  ");
+        }
+
         MediaStream videoStream = StreamHelper.getFirstVideoStream(item, mediaSourceIndex);
 
         if (videoStream != null && videoStream.getWidth() != null && videoStream.getHeight() != null) {
             int width = videoStream.getWidth();
             int height = videoStream.getHeight();
+            Boolean isInterlaced = videoStream.isInterlaced();
             if (width <= 960 && height <= 576) {
                 addBlockText(context, layout, context.getString(R.string.lbl_sd));
             } else if (width <= 1280 && height <= 962) {
-                addBlockText(context, layout, "720");
+                addBlockText(context, layout, "720" + (isInterlaced == null || !isInterlaced ? "p" : "i"));
             } else if (width <= 1920 && height <= 1440) {
-                addBlockText(context, layout, "1080");
+                addBlockText(context, layout, "1080" + (isInterlaced == null || !isInterlaced ? "p" : "i"));
             } else if (width <= 4096 && height <= 3072) {
                 addBlockText(context, layout, "4K");
             } else {
@@ -355,11 +361,6 @@ public class InfoLayoutHelper {
             addSpacer(context, layout, " ");
 
             addVideoCodecDetails(context, layout, videoStream);
-        }
-        if (Utils.isTrue(item.getHasSubtitles())) {
-            addBlockText(context, layout, "CC");
-            addSpacer(context, layout, "  ");
-
         }
     }
 
@@ -374,7 +375,13 @@ public class InfoLayoutHelper {
 
     private static void addVideoCodecDetails(Context context, LinearLayout layout, MediaStream stream) {
         if (stream != null) {
-            if (stream.getCodec() != null && stream.getCodec().trim().length() > 0) {
+            if (stream.getVideoDoViTitle() != null && stream.getVideoDoViTitle().trim().length() > 0) {
+                addBlockText(context, layout, "VISION");
+                addSpacer(context, layout, "  ");
+            } else if (stream.getVideoRangeType() != null && !stream.getVideoRangeType().equals("SDR")) {
+                addBlockText(context, layout, stream.getVideoRangeType().toUpperCase());
+                addSpacer(context, layout, "  ");
+            } else if (stream.getCodec() != null && stream.getCodec().trim().length() > 0) {
                 String codec = stream.getCodec().toUpperCase();
                 addBlockText(context, layout, codec);
                 addSpacer(context, layout, "  ");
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
```diff
@@ -340,12 +344,13 @@ private static void addRatingAndRes(Context context, BaseItemDto item, int media
...
             int height = videoStream.getHeight();
+            Boolean isInterlaced = videoStream.isInterlaced();
             if (width <= 960 && height <= 576) {
                 addBlockText(context, layout, context.getString(R.string.lbl_sd));
             } else if (width <= 1280 && height <= 962) {
-                addBlockText(context, layout, "720");
+                addBlockText(context, layout, "720" + (isInterlaced == null ? "" : isInterlaced ? "i" : "p"));
             } else if (width <= 1920 && height <= 1440) {
-                addBlockText(context, layout, "1080");
+                addBlockText(context, layout, "1080" + (isInterlaced == null ? "" : isInterlaced ? "i" : "p"));
```
nielsvanvelzen, line 353: When isInterlaced is null we should assume it's false, so always show either an `i` for interlaced or `p`.

app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
```diff
@@ -106,6 +106,10 @@ public static void addInfoRow(Context context, BaseItemDto item, int mediaSource
         addSeriesStatus(context, item, layout);
         addRatingAndRes(context, item, mediaSourceIndex, layout);
         addMediaDetails(context, audioStream, layout);
+        if (Utils.isTrue(item.getHasSubtitles())) {
+            addBlockText(context, layout, "CC");
+            addSpacer(context, layout, "  ");
+        }
```
nielsvanvelzen, line 112: We should prioritize the subtitle information before the audio information. That's the way the web client also does it.

app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
```diff
@@ -385,8 +391,19 @@ private static void addVideoCodecDetails(Context context, LinearLayout layout, M
...
+            } else if (stream.getProfile() != null && stream.getProfile().contains("DTS:X")) {
+                addBlockText(context, layout, "DTS:X");
+                addSpacer(context, layout, " ");
+            } else if (stream.getCodec() != null && stream.getCodec().trim().length() > 0) {
+                String codec =
+                    stream.getProfile() != null && stream.getProfile().contains("DTS-HD") ? "DTS-HD" :
+                    stream.getCodec().equals("dca") || stream.getCodec().equals("DCA") ? "DTS" :
+                    stream.getCodec().equals("eac3") || stream.getCodec().equals("EAC3") ? "DD+" :
+                    stream.getCodec().equals("ac3") || stream.getCodec().equals("AC3") ? "DD" :
+                    stream.getCodec().toUpperCase();
```
nielsvanvelzen, line 406: This should be replaced with a switch statement or continuing the if-else statements

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
mohd-akram: Apple puts it [after](https://help.apple.com/assets/6414D1E8F0A05F36C35A15D0/6414D1EBF0A05F36C35A15DA/en_GB/9b152618525e549cad25615d6619e2b4.png) the audio. Amazon puts it [before](https://i0.wp.com/www.aftvnews.com/wp-content/uploads/2022/09/The-Lord-of-the-Rings-The-Rings-of-Power-on-Prime-Video.jpg). It looks very awkward sandwiched between the two. It's also the only optional element so the layout is not consistent for different items - that's why I put it at the end. A/V info should be next to each other IMO so it's easy to read.

mohd-akram: I moved it before the video information. It looks better next to the rating before all the technical stuff.

app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
```diff
@@ -374,7 +375,13 @@ private static void addSeriesStatus(Context context, BaseItemDto item, LinearLay
 
     private static void addVideoCodecDetails(Context context, LinearLayout layout, MediaStream stream) {
         if (stream != null) {
-            if (stream.getCodec() != null && stream.getCodec().trim().length() > 0) {
+            if (stream.getVideoDoViTitle() != null && stream.getVideoDoViTitle().trim().length() > 0) {
+                addBlockText(context, layout, "VISION");
+                addSpacer(context, layout, "  ");
```
nielsvanvelzen, line 380: I believe you can use videoRangeType to check for Dolby Vision, the value should be "DOVI".

app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
```diff
@@ -385,10 +392,28 @@ private static void addVideoCodecDetails(Context context, LinearLayout layout, M
     private static void addMediaDetails(Context context, MediaStream stream, LinearLayout layout) {
 
         if (stream != null) {
-            if (stream.getCodec() != null && stream.getCodec().trim().length() > 0) {
-                String codec = stream.getCodec().equals("dca") || stream.getCodec().equals("DCA") ? "DTS" : stream.getCodec().equals("ac3") || stream.getCodec().equals("AC3") ? "Dolby" : stream.getCodec().toUpperCase();
-                addBlockText(context, layout, codec);
+            if (stream.getProfile() != null && stream.getProfile().contains("Dolby Atmos")) {
```
nielsvanvelzen, line 395: Is it not possible to find out if atmos/dts:x/dts-hd are used based on the codec?

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
mohd-akram: It only outputs DOVI for Dolby Vision profile 5, which is not backwards compatible with any other range. Otherwise, it will mention the backwards compatible range eg. HDR10 for profile 8.1.

app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
mohd-akram: Atmos and DTS:X are spatial formats on top of the codec. DTS-HD is a codec, but FFmpeg only mentions it in the profile (it lists the codec as just `dts`). The intention was to show both the spatial format and the codec, but it takes too much space and gets cut off in the details view.

nielsvanvelzen (approved):
Thanks for the pull request (and the explanations on my comments)!

nielsvanvelzen:
If the video is not Dolby Vision it should display HDR

nielsvanvelzen:
Hmm maybe we need to detect HDR in a different way, can you create an issue for it?

PR #2957, opened 2023-08-09
Use Android 14 SDK (API 34)

nielsvanvelzen:
**Changes**
- Use Android 14 SDK (API 34)
  - Also removes the targets from library modules as those are deprecated
  - Fix one compile issue related to a signature change in the new SDK
- Fix Koin deprecation warnings when using `get<>()` in Compose code

**Issues**

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -10,7 +10,7 @@ apply("baselineWorkaround.gradle")
 
 android {
 	namespace = "org.jellyfin.androidtv"
-	compileSdk = 33
+	compileSdk = 34
 
 	defaultConfig {
 		minSdk = 21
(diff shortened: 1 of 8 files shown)


nielsvanvelzen:
Leaving this as draft _for now_ as this SDK appears to be released but there hasn't been any announcement yet. At the same time the Compose update released today (1.5.0) indirectly depends on this SDK 🤷

nielsvanvelzen:
Because androidx.core now needs API 34 too I've updated the PR to compile with API 34 but target API 33 so we don't get the Android 14 behavior changes and everything should be fine. The fact that this SDK is not officially stable is still weird though.

crobibero (approved):

PR #2964, opened 2023-08-12
Add margin to media grid

DoggoOfSpeed:
**Changes**
image

It has always annoyed me how the grid and the title are not aligned so this PR matches their margin values

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java
@@ -6,7 +6,6 @@ import android.app.Activity;
 import android.os.Bundle;
 import android.os.Handler;
 import android.util.DisplayMetrics;
-import android.util.Log;
 import android.view.Gravity;
 import android.view.KeyEvent;
 import android.view.LayoutInflater;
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
Thanks for the pull request! There are some issues that need fixing:

- When scrolling the grid the items now have a weird cutoff at the margin instead of the edge of the screen
- This breaks the vertical grid direction (now needs scroll both horizontally and vertically)
- You should use display pixels (dp) in layout and scalable pixels (sp) for text

DoggoOfSpeed:
Yay! Now it works! I made it so that the start of the grid is aligned to the title and the end to the clock, so the value is no longer hard coded. 

Also, there is no cutoff anymore.

Also also all the code changes are in the horizontal grid logic so there should be no problems in the vertical one.

Also also also I moved the title a bit to the left and moved the options buttons so that they are aligned with the clock.

nielsvanvelzen:
> maybe the code changes should be moved to setAutoCardGridValues? I'm a baby when it comes to Android programming ;/

Unfortunately the code for the grid view is already messy so there's not really a "best" place for these small changes until everything is cleaned first.

nielsvanvelzen (approved):
Appears to work fine now! The changed alignment of the title also looks a lot better. Thanks for the PR :)

PR #2975, opened 2023-08-15
Fix blurry banners and thumbnails 

### LINKED ISSUE #2974: Banners and thumbnails fail to load required images
### Describe the bug

Set the image type to thumbnail/banner in the media grid settings;
The resulting cards have the images from the Poster setting, just in a different aspect ratio;

### Logs

_No response_

### Application version

0.16 beta 2

Sideloaded APK

### Device information

Reproducible both on an emulator and a Sony TV

### Android version

Android 9

### Jellyfin server version

10.8.10

DoggoOfSpeed:
**Changes**

Now when a banner/thumbnail is not found the image height used in retrieving an appropriate image from the server will be changed to account for the increased width.

Sadly this results in some performance issues as an image required to fill a banned can become quite large but I'm not sure how to fix that.

**Issues**
Fixes #2974

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/BaseRowItem.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/BaseRowItem.kt
@@ -152,27 +152,27 @@ open class BaseRowItem protected constructor(
 		BaseItemKind.MUSIC_ARTIST
 	).contains(baseItem?.type)
 
-	fun getImageUrl(context: Context, imageType: ImageType, requiredWidth: Int) = when (baseRowType) {
+	fun getImageUrl(context: Context, imageType: ImageType, preferedWidth: Int) = when (baseRowType) {
 		BaseRowType.BaseItem,
 		BaseRowType.LiveTvProgram,
 		BaseRowType.LiveTvRecording -> {
 			val apiClient = get<LegacyApiClient>()
 			when (imageType) {
-				ImageType.BANNER -> ImageUtils.getBannerImageUrl(baseItem, apiClient, requiredWidth)
-				ImageType.THUMB -> ImageUtils.getThumbImageUrl(baseItem, apiClient, requiredWidth)
-				else -> getPrimaryImageUrl(context, requiredWidth)
+				ImageType.BANNER -> ImageUtils.getBannerImageUrl(baseItem, apiClient, preferedWidth)
+				ImageType.THUMB -> ImageUtils.getThumbImageUrl(baseItem, apiClient, preferedWidth)
+				else -> getPrimaryImageUrl(context, preferedWidth)
 			}
 		}
 
-		else -> getPrimaryImageUrl(context, requiredWidth)
+		else -> getPrimaryImageUrl(context, preferedWidth)
 	}
 
-	fun getPrimaryImageUrl(context: Context, requiredWidth: Int) = when (baseRowType) {
+	fun getPrimaryImageUrl(context: Context, preferedWidth: Int) = when (baseRowType) {
 		BaseRowType.BaseItem,
 		BaseRowType.LiveTvProgram,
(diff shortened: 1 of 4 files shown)


nielsvanvelzen (changes requested):
Instead of using the maxHeight we should probably set a maxWidth in the request so the returned image works for the different width of the thumb/banner views instead of the height.

This would require some more changes though. Possibly by calling `ImageHelper.getPrimaryImageUrl(item, maxWidth, null)` to get the primary image with a max width.

DoggoOfSpeed:
I was afraid to break anything so kept the changes as minimal as possible w/o refactoring, but if that's ok I'll update the code to accomodate maxWidth instead.

DoggoOfSpeed:
I only moved the parts relevant to banners/thumbnails/posters while things everything else still uses `maxHeight` as they have no issues. Should move them as well just for internal consistency or is this fine?

P.S I used `preferedWidth` instead of `maxWidth` as it's slightly clearer in my opinion.

app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/BaseRowItem.kt
```diff
@@ -152,27 +152,27 @@ open class BaseRowItem protected constructor(
...
 		BaseRowType.BaseItem,
 		BaseRowType.LiveTvProgram,
 		BaseRowType.LiveTvRecording -> {
 			val apiClient = get<LegacyApiClient>()
 			when (imageType) {
-				ImageType.BANNER -> ImageUtils.getBannerImageUrl(baseItem, apiClient, maxHeight)
-				ImageType.THUMB -> ImageUtils.getThumbImageUrl(baseItem, apiClient, maxHeight)
-				else -> getPrimaryImageUrl(context, maxHeight)
+				ImageType.BANNER -> ImageUtils.getBannerImageUrl(baseItem, apiClient, preferredWidth)
+				ImageType.THUMB -> ImageUtils.getThumbImageUrl(baseItem, apiClient, preferredWidth)
```
nielsvanvelzen, line 162: We should only use the preferredWidth for the banner/thumb (as their aspect ratio is above 1), for the poster we should keep the preferredHeight. Also make sure only those two use the preferredWidth so we don't introduce changes for other places that use images, as its quite hard to check every part of the app for regression.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/BaseRowItem.kt
DoggoOfSpeed: Wouldn't the first commit be enough to fix the issue then? As it's a simple one-liner that calculates the required height of the image needed to fill the banner, it's entirely self-contained and the only shortcoming it has is that it doesn't call maxWidth by name, which I don't think matters as long as the results are the same...

nielsvanvelzen: No, the first commit still relies on sending a max height and hacks that by multiplying the requested value

DoggoOfSpeed: Right, but it's not multiplied randomly. It calculates the exact size of the poster image required to fill the card based on the aspect ratios.

thornbill:
Should this be using `fillWidth`? https://github.com/jellyfin/jellyfin-web/issues/2191

DoggoOfSpeed:
> Should this be using `fillWidth`? [jellyfin/jellyfin-web#2191](https://github.com/jellyfin/jellyfin-web/issues/2191)

Yeah, that could be used, but I just checked and the images you get by querying `maxWidth=*` vs `fillWidth=*` are identical.

thornbill:
Ah right it probably only makes a difference if you request width and height 😅

PR #2979, opened 2023-08-18
Folder improvements

DoggoOfSpeed:
**Changes:**
Moved the folder icon to the top left for everything else (Saves space for the name)
Changed purple theme's indicator to match all other themes (Although honestly I prefer the rounded rectangle over the circle, but it was the odd one out)


app/src/main/res/layout/view_card_legacy_image.xml
```diff
@@ -64,6 +64,40 @@
             android:visibility="gone"
             tools:visibility="visible" />
 
+        <androidx.constraintlayout.widget.ConstraintLayout
```
nielsvanvelzen, line 67: I think a FrameLayout can be used here to achieve the same result but more efficiently

app/src/main/res/values/attrs.xml
```diff
@@ -23,7 +23,8 @@
         <attr name="inputDefaultStrokeWidth" format="dimension" />
         <attr name="inputDefaultStrokeColor" format="reference|color" />
         <attr name="popupMenuBackground" format="reference|color" />
-        <attr name="accentShape" format="reference" />
+        <attr name="accentCircle" format="reference" />
+        <attr name="backgroundCircle" format="reference" />
```
nielsvanvelzen, line 27: Those attributes are not changed by any theme (except fort the default) so they can be omitted.

nielsvanvelzen (changes requested):
The icon looks a lot better on the top-left. It does look like the padding of the text is bigger on the left side compared to the right side though.

DoggoOfSpeed:
So that's a pretty big commit 😅. Here's what changed:

Watched indicator changes:
Moved the logic that decides how/if to show the blue watched indicator to setUnwatchedCount();
Now the watched indicator is only shown on shows as it's not very useful on folders;
The settings that decide when to show the indicator have been renamed to be clearer;
The setting `When the show has been watched` (previously `Hide unwatched count`) now works.

Folder card changes:
As per your suggestion I converted both the watched indicator and the item type icon into FrameLayout;
The text margins to the side of the card have been made identical;
The number of items in the folder is now semi-transparent to help with readability as under the right circumstances a user may confuse the number as part of the folder name.

nielsvanvelzen:
That's indeed a big commit. Unfortunately we cannot introduce string changes now if we want to land the changes in the 0.16 release. I'm also not a fan of the unrelated changes you added in the last commits (like changing the behavior of the watch indicator), those things should be in a separate pull request.

nielsvanvelzen:
The changed color for folder item count is barely readable when the item doesn't have a dark background.

DoggoOfSpeed:
Yeah, I can see that being an issue 😅. I'll see what I can do.

DoggoOfSpeed:
Alright, so in the end I decided to just force some distance between the title and the item count as:
1. Moving it just sounded dumb. A third circle? Really?
2. Changing the color to non-white just didn't look very good.
3. Changing the size of the text was also rather weird.

Now the contrast for the counter should be around the same as before the PR while the folder title is actually whiter as I removed the subtle transparency.

nielsvanvelzen (approved):

nielsvanvelzen:
Looks fine now, thanks!

PR #2995, opened 2023-08-25
Alignment changes to audio player

DoggoOfSpeed:
**Changes**
The progress bar item is now aligned with the buttons,
There is now a lot more space for the song title


nielsvanvelzen (approved):
I wasn't entirely sure about the centering at first but after asking for feedback in our UI/UX room at Matrix/Discord everyone preferred it so center it is!

PR #2996, opened 2023-08-25
Synochronise music timers

DoggoOfSpeed:
**Changes**
Right now the timers are updated in such a way that, depending on the song, they may update up to half a second apart.
This PR rounds the current time value to seconds so that the rounding that happens when the milliseconds are displayed doesn't cause the timers to be out of sync.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingFragment.java
@@ -273,7 +273,8 @@ public class AudioNowPlayingFragment extends Fragment implements View.OnKeyListe
 
         @Override
         public void onProgress(long pos) {
-            setCurrentTime(pos);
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/AudioNowPlayingFragment.java
```diff
@@ -273,7 +273,8 @@ public void onPlaybackStateChange(@NonNull PlaybackController.PlaybackState newS
 
         @Override
         public void onProgress(long pos) {
-            setCurrentTime(pos);
+            // Round the current time as otherwise the time played and time remaining will not be in sync
+            setCurrentTime(Math.round(pos / 1000) * 1000);
```
nielsvanvelzen, line 277: It would be better to move the rounding to inside the setCurrentTime function to avoid having the same issue if the function were called from some other place

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

PR #2997, opened 2023-08-25
Watched indicator settings string clarity improvements

DoggoOfSpeed:
**Changes**
The settings strings are now clearer in what they mean.

For reference, the setting is titled `Show watched indicators`

`Hide unwatched count` -> `When the show has been watched`;
The new string follows up the setting to create sentence `Show watched indicator when the show has been watched`. The previous string was very hard to understand as the word "Hide" opposes the word "Show" and "Watched indicators" and "Unwatched count" are referring to the same thing.

`On episodes only` -> `When there are episodes remaining`
In this case the original is not clear in whether it only changes the behavior when there are episodes you haven't seen yet or whether it shows on all forms of media that have episodes.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -344,7 +344,7 @@
     <string name="login_help_title">Need help?</string>
     <string name="login_help_description">Jellyfin requires a server to connect with. Visit our documentation at docs.jellyfin.org or scan the QR code to get started with Jellyfin.</string>
     <string name="pref_watched_indicator">Show watched indicators</string>
-    <string name="lbl_hide_unwatched_count">When the show has been watched</string>
+    <string name="lbl_hide_unwatched_count">When the item is watched</string>
     <string name="lbl_hide_watched_checkmark">When there are episodes remaining</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -344,8 +344,8 @@
     <string name="login_help_title">Need help?</string>
     <string name="login_help_description">Jellyfin requires a server to connect with. Visit our documentation at docs.jellyfin.org or scan the QR code to get started with Jellyfin.</string>
     <string name="pref_watched_indicator">Show watched indicators</string>
-    <string name="lbl_hide_unwatched_count">Hide unwatched count</string>
-    <string name="lbl_episode_marks">On episodes only</string>
+    <string name="lbl_hide_unwatched_count">When the show has been watched</string>
```
nielsvanvelzen, line 347: Using "item" instead of "show" likely makes more sense, as it also applies for other item types like movies.

nielsvanvelzen, line 347:
```suggestion
    <string name="lbl_hide_unwatched_count">When the item is watched</string>
```

nielsvanvelzen (approved):

PR #2998, opened 2023-08-25
Watched indicator fixes

DoggoOfSpeed:
**Changes**
Move indicator logic away from setItem() to setUnwatchedCount() so now there is no need for magic numbers like -1 and everything is generally simpler;
Implement the `HIDE_UNWATCHED` setting;
Also fixes the `EPISODES_ONLY` setting;
Fixes checkmark not appearing (broken in the .16 betas);
Make the indicator hidden by default. Currently I only enabled it on TV shows, but perhaps I forgot about some other places where they're useful;


nielsvanvelzen:
image

Doesn't seem to work right 😄. This is for an item in the "recently added" home section, it shows me that I've watched everything and still have 15 to watch at the same time. I think this is because the card views are re-used while scrolling, so the 15 and checkmark came from another item.

DoggoOfSpeed:
Uh oh 😭 It's another case of "Works on my machine".

Have you pulled all the commits? I just rebased and that should fix your issue as this PR depends on one of my previous PRs which has been merged already.

nielsvanvelzen:
I think it is actually because you removed the `.setVisibility(INVISIBLE);` statements

DoggoOfSpeed:
Yes, that's correct. As I said, I made the indicators invisible by default, so turning them invisible a second time is unnecessary.

nielsvanvelzen:
Once again, the default in the XML will not be magically set when the view is reused in the adapter. You need to make sure that all changes made in the Java code are reverted when the items change.

DoggoOfSpeed:
Welp, I can only hope that fixes that as I'm unable to replicate the issue :(.

Still don't understand why that's happening to you as the default in the XML _is_ set to `GONE`, but if it works that's good.

nielsvanvelzen:
Closing due to inactivity and unsolved review comments.

PR #3005, opened 2023-08-28
Fix user image being too cropped in

DoggoOfSpeed:
**Changes**
On first open of the user select menu the transition between the placeholder and the user image forced the image to use the dimensions of the placeholder. This resulted in the image being far too cropped in.

**Before**
Screenshot 2023-08-28 133137

**After**
Screenshot 2023-08-28 133100

This is _kinda_ a hack as I simply converted the placeholder image into a square, but it works and I'm not sure what else could be done.

### DIFF
--- a/app/src/main/res/drawable/tile_square_user.xml
+++ b/app/src/main/res/drawable/tile_square_user.xml
@@ -3,16 +3,16 @@
     <item>
         <shape android:shape="rectangle">
             <size
-                android:width="256dp"
+                android:width="384dp"
                 android:height="384dp" />
             <solid android:color="?attr/tile_port_user_bg" />
         </shape>
     </item>
 
     <item
-        android:bottom="124dp"
         android:drawable="@drawable/ic_user"
-        android:left="60dp"
-        android:right="60dp"
-        android:top="124dp" />
+        android:left="124dp"
+        android:right="124dp"
+        android:top="124dp"
+        android:bottom="124dp"/>
 </layer-list>
--- a/app/src/main/res/layout/fragment_server.xml
+++ b/app/src/main/res/layout/fragment_server.xml
@@ -40,7 +40,7 @@
     <androidx.leanback.widget.HorizontalGridView
         android:id="@+id/users"
         android:layout_width="match_parent"
-        android:layout_height="153dp"
+        android:layout_height="wrap_content"
         android:clipChildren="false"
         android:clipToPadding="false"
         android:scrollbars="none"
@@ -49,7 +49,6 @@
         app:layout_constraintTop_toBottomOf="@+id/title"
         app:rowHeight="153dp"
         tools:itemCount="4"
-        tools:layout_height="120dp"
         tools:listitem="@layout/view_card_default" />
 
     <TextView
--- a/app/src/main/res/layout/fragment_server.xml
+++ b/app/src/main/res/layout/fragment_server.xml
@@ -40,7 +40,7 @@
     <androidx.leanback.widget.HorizontalGridView
         android:id="@+id/users"
         android:layout_width="match_parent"
-        android:layout_height="wrap_content"
+        android:layout_height="153dp"
         android:clipChildren="false"
         android:clipToPadding="false"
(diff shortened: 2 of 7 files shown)


DoggoOfSpeed:
Also, for some reason I'm only able to replicate the bug in 0.16 betas and on master. On the .15 release the image is loaded correctly with no issues.

nielsvanvelzen:
We switched the image library from Glide to Coil (#2889) for 0.16 so this is likely a regression from that change.

thornbill:
I noticed this also in the 0.16 branch fwiw

app/src/main/res/drawable/tile_square_user.xml
```diff
@@ -3,16 +3,16 @@
     <item>
         <shape android:shape="rectangle">
             <size
-                android:width="256dp"
+                android:width="384dp"
                 android:height="384dp" />
```
nielsvanvelzen, line 7: File should be named `tile_user.xml` (like all other square tiles) and size should be 256x256 dp

app/src/main/res/drawable/tile_square_user.xml
```diff
@@ -3,16 +3,16 @@
...
     <item
-        android:bottom="124dp"
         android:drawable="@drawable/ic_user"
-        android:left="60dp"
-        android:right="60dp"
-        android:top="124dp" />
+        android:left="124dp"
+        android:right="124dp"
+        android:top="124dp"
+        android:bottom="124dp"/>
```
nielsvanvelzen, line 17: These can _likely_ be all 60dp like the other square tiles

app/src/main/res/drawable/tile_square_user.xml
```diff
@@ -3,16 +3,16 @@
     <item>
         <shape android:shape="rectangle">
             <size
-                android:width="256dp"
+                android:width="384dp"
                 android:height="384dp" />
             <solid android:color="?attr/tile_port_user_bg" />
```
nielsvanvelzen, line 8: Attribute should be renamed to reflect new tile name

app/src/main/res/layout/fragment_server.xml
```diff
@@ -40,7 +40,7 @@
     <androidx.leanback.widget.HorizontalGridView
         android:id="@+id/users"
         android:layout_width="match_parent"
-        android:layout_height="153dp"
+        android:layout_height="wrap_content"
```
nielsvanvelzen, line 43: Why change this?

nielsvanvelzen (changes requested):

nielsvanvelzen:
Issue seems to be caused by https://github.com/coil-kt/coil/issues/1505, changing the placeholder to be square should be a good fix to workaround the issue.

app/src/main/res/layout/fragment_server.xml
DoggoOfSpeed: Ouch 😅 I think I was playing with the ui before making the changes and this seeped in.

app/src/main/res/layout/fragment_server.xml
```diff
@@ -49,7 +49,6 @@
         app:layout_constraintTop_toBottomOf="@+id/title"
         app:rowHeight="153dp"
         tools:itemCount="4"
-        tools:layout_height="120dp"
```
nielsvanvelzen, line 52: I guess this change was also accidental?

nielsvanvelzen (approved):
Looks fine except for the one comment.

app/src/main/res/layout/fragment_server.xml
DoggoOfSpeed: Well yes and no. I included it as imo it's a change not worthy of a PR as it's dev-facing only. 

Removing that lines allows the preview to actually reflect the final UI and I have no idea why that value was there to begin with.

nielsvanvelzen: You're right, just checked and removing this line does actually improve the preview.

PR #3020, opened 2023-09-02
Grab background images no larger than the screen

DoggoOfSpeed:
**Changes**
Fix the app grabbing 4k background images even when the screen is just 1080p.

As a side-effect that also saves a lot of cache space :)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
@@ -1,6 +1,7 @@
 package org.jellyfin.androidtv.data.service
 
 import android.content.Context
+import android.content.res.Resources
 import androidx.compose.ui.graphics.ImageBitmap
 import androidx.compose.ui.graphics.asImageBitmap
 import androidx.core.graphics.drawable.toBitmap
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
I'm not sure about this one. Yes it reduces the image size on smaller screens which leads to smaller cache but at the same time it will now make the server resize all images and cache them there which makes the (first) requests slower.

DoggoOfSpeed:
Well for me personally that still makes some sense.

Sure, it takes time to process the images, but in turn that means the images require less bandwith and are quicker to download.

The same can kinda be applied to storage space as the server, in all likelyhood, is much less storage limited whereas many tvs and set-top boxes don't even have 8gb.

nielsvanvelzen:
Right that's true, should be fine to change this then.

app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
```diff
@@ -60,6 +61,7 @@ class BackgroundService(
 				imageType = ImageType.BACKDROP,
 				tag = tag,
 				imageIndex = index,
+				maxWidth = Resources.getSystem().getDisplayMetrics().widthPixels,
```
nielsvanvelzen, line 64: I believe we can use `context.resources` here

nielsvanvelzen: Yeah you should be able to do `context.resources.displayMetrics.widthPixels`

nielsvanvelzen:
Looking a bit further we should use `fillWidth` instead of `maxWidth` and also add `fillHeight`.

nielsvanvelzen (approved):

PR #3024, opened 2023-09-05
Fix sync issue in ItemRowAdapterHelper

### LINKED ISSUE #3021: TV Show grid shows duplicate when a show is removed and filtered
### Describe the bug

1. Set the TV Show filter to only show unwatched shows/seasons
2. Watch a TV Show where there is only one new episode
3. Observe that the last (aka, bottom-right) TV Show icon is now duplicated

For example, my grid of shows has "Foo" in the upper-left with 1 episode in it, a whole bunch of TV shows, and then "Bar" at the bottom-right of the grid.  When I finish watching "Foo", the TV Show disappears from the grid (correctly, because I have the display filtering only unwatched shows), but the grid keeps the same number of TV Shows listed.

It should shrink by one, but instead it shrinks by one (aka, removing "Foo"), but shows "Bar" and "Bar" as the last two.

### Logs

_No response_

### Application version

0.16.beta2

Google Play

### Device information

NVIDIA Shield TV Pro

### Android version

10

### Jellyfin server version

10.8.10

nielsvanvelzen:
**Changes**
- Fix sync issue in ItemRowAdapterHelper
  Basically the item row adapter is garbage and my helper failed with an edge case because the code is garbage
- Add some changes to experimental reactive home to have it reload faster
  Previous fix introduced a new (or actually it now happened more often..) race condition where having two retrievals running for the same row would duplicate items because once again the item row adapter is garbage

**Issues**

Fixes #3021

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeRowsFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeRowsFragment.kt
@@ -168,7 +168,7 @@ class HomeRowsFragment : RowsSupportFragment(), AudioEventListener, View.OnKeyLi
 				if (userPreferences[UserPreferences.homeReactive]) {
 					val listener = socketInstance.addGlobalListener {
 						if (it is UserDataChangedMessage || it is LibraryChangedMessage) {
-							refreshRows(force = true)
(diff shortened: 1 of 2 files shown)


thornbill (approved):
Embrace the garbage!
image

PR #3049, opened 2023-09-14
Migrate SuggestedMoviesFragment to SDK

nielsvanvelzen:
**Changes**
- Migrate SuggestedMoviesFragment to Kotlin
- Migrate SuggestedMoviesFragment to SDK
  - Except for the ItemRowAdapter thingy
- Use string template for the "because you watched x" string

**Issues**


nielsvanvelzen:
I think this is the last SDK migration for a bit (again). Left over code is either playback related or the item row adapter which will be a pain to do.

crobibero (approved):

PR #3057, opened 2023-09-19
Fix progress bar appearing on fallback poster

DoggoOfSpeed:
**Changes**
Fix this:
Screenshot 2023-09-19 201711


nielsvanvelzen:
I've made a proper fix in #3058

DoggoOfSpeed:
Yeah, that certainly makes it look better but I also don't see the reason for it to be there at all.

It's not like the progress bar appears when there the actual logo is found. As well as that, the specific time the person has watched is indicated under the play button which makes the bar kinda redundant :)

nielsvanvelzen:
#3058 restores it to how is was in 0.15. I agree that this progress bar is not great though. We should find a different place for the progress bar so we can always show it.

DoggoOfSpeed:
Maybe play button could be turned into a kind of a progress bar? Apart from that I'm not sure where else it could be put...

PR #3062, opened 2023-09-21
Add initial playback capability testing

nielsvanvelzen:
**Changes**

This PR adds a few things:
- Media information in MediaStream (container/tracks)
  - This only contains container/audio tracks right now with the others coming later (0.17)
  - The data model is far from finished
- Add `supportsStream` function to backend
  - This uses the media information to query the backend for playback support
  - Currently implemented for ExoPlayer audio testing only
- Test is a given stream is supported and warn to the logs if it is not
  - Future PR (0.16 release blocker) will add the behavior to fallback to remux/transcode when a stream is not supported

Some future changes that are likely to happen:
- Add MediaInfo class to contain the container/tracks and use that for the `supportsStream` function instead of a full MediaStream instance. This way we can query the backend before building the stream URL etc.
- Add video/subtitle tracks
- Add more properties/info to container/audio track
- Extend codec->mime mapping
- Move codec->mime mapping to core (could be useful for other backends)

**Issues**

Important part of #1057


crobibero (approved):
mapping will be ..fun to populate

PR #3085, opened 2023-09-28
Use layer-list for app_banner drawable

nielsvanvelzen:
Google told me I had to make sure we have square icons and pointed to [this](https://developer.android.com/design/ui/tv/guides/system/tv-app-icon-guidelines) documentation. We do have a square icon but we did not have an adaptive banner so I started work on that only to find out there is no such thing, the documentation is wrong. So I removed the v26 app_banner xml and kept all other changes.

**Changes**
- Use layer-list for app_banner drawable

**Issues**


crobibero (approved):
Thanks elgoog

Issue #3089, opened 2023-09-30
embedded subtitles not working when subtitle filter on library


horizon1250:
### Describe the bug

1. Set supported embedded subtitles on library to image only.
![image](https://github.com/jellyfin/jellyfin-androidtv/assets/146419076/993d7684-a2ac-4c29-8dd1-0853d192cda8)

2. Add external subtitle files in same directory of movie 

3. Load embedded image based subtitle on android tv, a toast appears which says "Unable to load subtitle" . The external subtitles still work. This error only happens when enabling the previously mentioned filter on the library.

### Application version

v0.15.12, v0.16.0-beta.4 and latest master.

### Where did you install the app from?

Google Play

### Device information

Nvidia shield pro (2019)

### Android version

Android 11

### Jellyfin server version

10.8.10

horizon1250:
I have a fix for this issue and will open it in a PR :)

PR #3090, opened 2023-09-30
Fix embedded subtitles not loading when subtitle filter on library

horizon1250:
**Changes**
In `setExoPlayerTrack`, filter for the correct index instead of addressing the index directly. This ensures the correct subtitle steam is selected, even when some tracks are filtered out.

**Issues**
#3089


nielsvanvelzen (approved):

nielsvanvelzen:
Thanks!

PR #3108, opened 2023-10-07
Enable support for Opus and FLAC in ExoPlayer

reacocard:
**Changes**
Add Opus and FLAC to ExoPlayer's list of codecs so they can direct-play.

Remake of https://github.com/jellyfin/jellyfin-androidtv/pull/2376 since the original author seems to be absent.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-androidtv/issues/2359 https://github.com/jellyfin/jellyfin-androidtv/issues/2366


nielsvanvelzen (approved):
Thanks for the PR!

PR #3113, opened 2023-10-10
Support remux/transcode fallback in new music player

nielsvanvelzen:
Add automatic transcoding support to new playback code based on the capability testing from #3062. This is the final release blocker, unless something else comes up.
I'm not entirely happy with how it works but unfortunately the server doesn't give us much to work with. Direct play & remux can be tested client side because we know what the media information will look like but a proposed transcode from the server does not give us any information except for the container & protocol which is too little to properly do capability testing with.

**Changes**
- Add IMediaStream interface for the base definition of a media stream
  - Implemented by MediaStream and PlayableMediaStream
  - What we previously called MediaStream is now PlayableMediaStream - it adds the queueEntry & url propertyes on top of the interface
  - Use IMediaStream for play report testing magic
- Refactor MediaStreamState
  - Move responsibility of generating a compatible stream to the resolvers
  - Skip to next entry if no stream is found
- Rewrite the Jellyfin AudioMediaStreamResolver
  - Now does some magic to find a direct play -> remux -> transcoding stream (in order)

**Issues**

:) #1057

### DIFF
--- a/playback/jellyfin/src/main/kotlin/JellyfinPlugin.kt
+++ b/playback/jellyfin/src/main/kotlin/JellyfinPlugin.kt
@@ -7,6 +7,9 @@ import org.jellyfin.playback.jellyfin.playsession.PlaySessionSocketService
 import org.jellyfin.sdk.api.client.ApiClient
 import org.jellyfin.sdk.api.sockets.SocketInstance
 import org.jellyfin.sdk.model.api.DeviceProfile
+import org.jellyfin.sdk.model.api.DlnaProfileType
+import org.jellyfin.sdk.model.api.EncodingContext
+import org.jellyfin.sdk.model.api.TranscodingProfile
 
 fun jellyfinPlugin(
 	api: ApiClient,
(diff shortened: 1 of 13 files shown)


playback/core/src/main/kotlin/mediastream/PlayableMediaStream.kt
thornbill, line 1: Nit: `MediaStream` might be better as the file name since it contains different types of media streams

playback/jellyfin/src/main/kotlin/JellyfinPlugin.kt
```diff
@@ -20,13 +23,23 @@ fun jellyfinPlugin(
 		responseProfiles = emptyList(),
 		subtitleProfiles = emptyList(),
 		supportedMediaTypes = "",
-		transcodingProfiles = emptyList(),
+		// Add at least one transcoding profile so the server returns a value
+		// for "SupportsTranscoding" based on the user policy
+		// We don't actually the profile
```
thornbill, line 28: We don't actually _use?_ the profile

playback/core/src/main/kotlin/mediastream/PlayableMediaStream.kt
nielsvanvelzen: This change was not intentional, I think the IDE automatically renamed the file when I renamed the class.

crobibero (approved):

PR #3135, opened 2023-10-21
Fixed random sort not available in grid view

### LINKED ISSUE #2309: Random sort is not available in grid view
### Describe the bug

1. Go to any Grid View (like Music Albums)
2. Open the Sort options
3. "Random" option is missing

The server and web interface already support this, so I think that all that needs to be changed is another sort option added to the interface here:

https://github.com/jellyfin/jellyfin-androidtv/blob/2bcd2b9482895b04e372c818671ffee6244ece50/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java#L156

Something like this: (The `lbl_random` localization key doesn't exist right now.)

```java
sortOptions.put(7, new SortOption(getString(R.string.lbl_random), ItemSortBy.Random, SortOrder.ASCENDING));
```

I really like random sort in the web interface, so I really miss it in the TV interface.

### Logs

_No response_

### Application version

0.14.5 release

Google Play

### Device information

NVIDIA Shield Android TV

### Android version

Shield 9.1.1 (33.2.0.157)

### Jellyfin server version

10.8.5

tsarrabayrouse:
**Changes**
Added a new sort option (Sort by Random) and created a "Random" label in the strings.xml file.

**Issues**
Fixes #2309

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java
@@ -154,6 +154,7 @@ public class BrowseGridFragment extends Fragment implements View.OnKeyListener {
             sortOptions.put(4, new SortOption(getString(R.string.lbl_community_rating), ItemSortBy.CommunityRating + "," + ItemSortBy.SortName, SortOrder.DESCENDING));
             sortOptions.put(5, new SortOption(getString(R.string.lbl_critic_rating), ItemSortBy.CriticRating + "," + ItemSortBy.SortName, SortOrder.DESCENDING));
             sortOptions.put(6, new SortOption(getString(R.string.lbl_last_played), ItemSortBy.DatePlayed + "," + ItemSortBy.SortName, SortOrder.DESCENDING));
+            sortOptions.put(7, new SortOption(getString(R.string.lbl_random), ItemSortBy.Random, SortOrder.ASCENDING));
         }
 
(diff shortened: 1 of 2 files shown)


hmbqzidixiwc:
This PR will cause crashing on my device.

Conditions to replicate
Go to Movies, items listed is 634
Scroll down to anywhere that causes the listing to scroll
Select a movie (item 40 was far enough down)
Go back to listing - (back key) without playing or any time the listing is returned to

Top items to not seem to be affected, appears to have to do with scrolling.

Will cause the app to crash and send a crashlog to the server.

That crashlog indicates a failure to set a selected item from what I can determine.

java.lang.IndexOutOfBoundsException: Invalid item position -1(-1). Item count:69 androidx.leanback.widget.VerticalGridView{74ca94e VFE...... ......ID 38,0-1882,821 #7f0b0091 app:id/browse_grid}, adapter:androidx.leanback.widget.VerticalGridPresenter$VerticalGridItemBridgeAdapter@e0b0a4, layout:androidx.leanback.widget.GridLayoutManager@78e0a6f, context:org.jellyfin.androidtv.ui.browsing.MainActivity@c5af60c

I would like this feature but it needs more code to make it work properly.

Regards, UNKNOWN

tsarrabayrouse:
Hi @hmbqzidixiwc, thank you for the comment. Repeating the same steps you listed, I also get a crash. This seems to happen only when scrolling down if the Grid direction is set to vertical.

PR #3210, opened 2023-12-09
util: Don't use TZ offset for media Premiere Date

### LINKED ISSUE #3190: Firestick app shows wrong airdates
### Describe the bug

My Jellyfin player on Firestick seems to display most or all release/air dates for episodes and films one day prior to the actual release/air date.

Other jellyfin clients including iOS and desktop client show the correct date, and the correct date is set in the metadata.

My server OS, Jellyfin server container and my Firestick are all on the same Timezone location. I don't see a timezone or location setting within the Jellyfin client app. Given that "release date" is a fixed attribute, I don't know how the Firestick app is somehow pulling the wrong date.

### Logs

_No response_

### Application version

0.15.12

Amazon Appstore

### Device information

Amazon FireTV  4K Max 1st Gen (2021) (AFTKA)

### Android version

Fire OS 7.6.6.4

### Jellyfin server version

10.8.12

fice-t:
I'm fairly new to Java/Kotlin/Android, so I'm not sure if this approach is the best, or if it should be used in other instances (such as in `TimeUtils`). It at least appears to fix the linked issue.

Is it reasonable to cut out usage of `Date` and stick with `LocalDateTime`, here and perhaps in `TimeUtils`?

Also: the "Premiere" date for TV seasons inside `InfoLayoutHelper::addDate` is in UTC, but the "Premiere"/"Born" date for actors is... odd. For example, in UTC-6: `1950-01-01T06:00`, and UTC+3: `1950-01-01T15:00`. Is this coming from the server, or was it changed in the client? Perhaps it acts oddly when the client is set to a different TZ than the server?

**Changes**

`TimeUtils.getDate` internally calls `Date.format`, which translates the UTC timestamp into local TZ timestamp. If a user is in a TZ with negative offset, then the Premiere Date will be presented as a day earlier.

Example with TZ of UTC-6:

item.getPremiereDate()                    => 2023-12-09T00:00
TimeUtils.getDate(item.getPremiereDate()) => Sat Dec 08 18:00:00 CST 2023

**Issues**

Fixes #3190.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
@@ -24,6 +24,8 @@ import org.jellyfin.sdk.model.api.SeriesStatus;
 import org.koin.java.KoinJavaComponent;
 
 import java.text.NumberFormat;
+import java.time.format.DateTimeFormatter;
+import java.time.format.FormatStyle;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (approved):
The old apiclient used Date, everything with the new SDK uses java.time. For conversion we have the ModelCompat helpers that will convert Date->java.time. I thought I fixed most of the timezone issues but apparently not 😄.

Issue #2717, opened 2023-04-27
Icon or keypress to go to Home screen on Firestick


TheHYPO35:
### Describe the feature you'd like

I use the Jellyfin app via Firestick. As far as I can tell, there is no "home" button anywhere on screen in the app. There is also no remote keypress that seems to go to the Jellyfin home screen (I've tried pressing and holding buttons, etc.)

I'm quite sure when I first got the app, I could hit "home" (which takes you to the Firestick home screen) and then go back to Jellyfin and it would put me at the Jellyfin home screen, but for the last little while, when I do this, it takes me back to whatever screen I was on. So if I've watched 5 episodes of something, I have to hit "back" 5 times to go back to each episode page, then back to home. There should be a single remote button or at least an icon on screen to return to the Jellyfin home screen.

polson:
Going back to the screen you were at after homing out is normal Android behavior.  You probably want a home button in the app to do that.

PR #3262, opened 2023-12-27
Add home button when away from home view

### LINKED ISSUE #2053: Global back button

Currently, the back button on the remote is kinda broken.

If watching a show with say 8 episodes back to back (more than one episode really), upon completion, pressing the back button on your remote will not take you back to the home screen. There is basically no way to go back to the home screen unless you quit the app and restart it.

So what happens is that you will have to keep pressing back on the remote repeatedly, which will just scroll you through all the watched episodes of that series until you get back to the home screen.

A better solution would be to allow a long press back button that will just take you straight back to the Jelyfin home page. And this long pressback should work regardless of where in the app you are in.

VizuaaLOG:
**Changes**
Added a home button to the top toolbar when away from the home view, allowing for a quick "back to home" feature. The Jellyfin web UI and android mobile app has this. Having this button provides a quick way to leave a long navigation chain without having to press the back button multiple times.

The home button has been added to the `ClockUserView` as this appears to already be shown on all views although it does make the name of the view a little misleading.

**Issues**
Closes #2053 and #2717

#2053 does mention that new navigation features may provide a resolution for this in future. However, given the nice quality of life this provides I decided to tackle it as my first PR, as I also have frustrations with this.


nielsvanvelzen:
Did you test these changes? Running this in the emulator and a lot of screens have issues, either not being able to focus the button or the button being behind the screen title.

VizuaaLOG:
I did both in an emulator and on my Philips TV 🤔 . I will look into it

VizuaaLOG:
@nielsvanvelzen Testing in a 4k and 1080p emulator.

The examples I have found are both when using the smart screen mode (at least this is how I found the buttons) and I navigate to the "By Letter" or "Genres" screen for each, the library name overlaps the home button, this is also the screens where I cannot focus either.

Are there other instances you have seen that I am missing?

I'm also thinking that the title on those screens should be left aligned? As shorter titles puts "Movies", for example, on the right which seems a little odd 🤔

VizuaaLOG:
I have had an attempt at resolving this. Android development isn't my expertise, so please do let me know if there is a better way to accomplish what I have done.

I believe the issue with the overlap/no focus was because the clock and user icon view was being added to the browse frame and then positioned to appear in the title area. After lots of research on how Leanback works it appears to recommendation, to allow for focusing, is to create a custom `TitleView` / `BrowseTitleView` and use the theme to replace it. This will then mean in terms of hierarchy the clock is now in the title, and so focusing it by pressing up would be possible.

This change has meant the `BrowseFolderFragment` no longer needs to add the clock/rearrange the title in the `onStart` call, instead this is now being handled by a normal view and the `TitleView` class.

From my research this appears to only affect the "ByLetter" and "ByGenre" screens, other screens in my testing appear to work okay and are focusable and do not have issues with titles overlapping.

nielsvanvelzen:
Another round of testing:

- In the "now playing" screen for music, moving the focus to the home icon introduces some weird scroll behavior
- The spacing between the title and the home icon is too little in the folder views (e.g. genre view)
- The video player should not have the home icon
- Using the home icon doesn't clear navigation history properly. When I open the app, navigate to somewhere, then use the home icon and after that use the back button on my remote it doesn't close the app

VizuaaLOG:
> In the "now playing" screen for music, moving the focus to the home icon introduces some weird scroll behavior

I think I have fixed this. Was caused by the focus handler not being aware the home button is also a focusable button now, not just the queue row

> The spacing between the title and the home icon is too little in the folder views (e.g. genre view)

Increased the spacing slightly - added a 16dp margin - let me know if you would prefer more or less

> The video player should not have the home icon

I did think about this when adding the button initially and I actually thought it could be useful to be there. Given an instance of finishing a movie or episode of a show, if the user would like to watch, listen, or view something else entirely they're about to just go home straight from the video player. Saving them from needing to press back first, small saving granted. If you would prefer it removed I can though.

> Using the home icon doesn't clear navigation history properly.

This one took some digging to find. I think I have found the issue, however, I'm not sure how to resolve it, so any ideas would be appreciated!

The issue appears to be caused by the Navigation Repository's `reset` method not clearing the back stack in the `supportFragmentManager`. I see reset is called in some other instances and while I'm not 100% sure how to trigger them I think they may have a similar issue too? The deeper you are into the app, thus the higher that count in the back stack, the more you have to press back when on the home screen - as the back handler pops the stack. Once it hits zero, it'll exit the app.

I tried to see if there was a way to just reset the stack but I had trouble finding a method for that. I suppose an option would be too loop over the stack and pop it until empty but that didn't feel like the best solution so I thought I'd see if you have any ideas?

nielsvanvelzen:
- In the music player, the title can overlap with the home icon

> I did think about this when adding the button initially and I actually thought it could be useful to be there. Given an instance of finishing a movie or episode of a show, if the user would like to watch, listen, or view something else entirely they're able to just go home straight from the video player. Saving them from needing to press back first, small saving granted. If you would prefer it removed I can though.

Navigation is a part of browsing, the video player is a completely separate view. So yes it should be removed.

> I tried to see if there was a way to just reset the stack but I had trouble finding a method for that. I suppose an option would be too loop over the stack and pop it until empty but that didn't feel like the best solution so I thought I'd see if you have any ideas?

The reset function is called by the MainActivity when it's fragment history is reset, not the other way around. So you'll need to add new behavior to the navigation code to deal with this. I don't have any suggestions for that.

VizuaaLOG:
Apologies for the delay!

* The home button is now hidden from the video player.
* The artist title in the music player has been fixed so shouldn't overlap now.
* I have also added a new GoHome navigation action which clears the back stack to fix the issue with the back button not exiting the app when using the home button.

Thanks!

app/src/main/java/org/jellyfin/androidtv/ui/browsing/MainActivity.kt
```diff
@@ -158,6 +160,21 @@ class MainActivity : FragmentActivity() {
 
 		NavigationAction.GoBack -> supportFragmentManager.popBackStack()
 
+		NavigationAction.GoHome -> {
```
nielsvanvelzen, line 165: It might be better to re-use the NavigateFragment action but add a new "clear" flag to it instead of adding an entire new action.

nielsvanvelzen:
Thank you for the pull request! I'm sure a lot of users will love this feature.

Instead of commenting on the small details I've added an additional commit that fixes some styling issues and merged the "goHome" function into the existing "reset" function in NavigationRepository.

nielsvanvelzen (approved):

VizuaaLOG:
> Thank you for the pull request! I'm sure a lot of users will love this feature.
> 
> Instead of commenting on the small details I've added an additional commit that fixes some styling issues and merged the "goHome" function into the existing "reset" function in NavigationRepository.

I know I'll love it when I can't decide what to watch!

PR #3271, opened 2023-12-31
Add MediaSession for exo player

domenipavec:
**Changes**
Create android media session for exo player.

**Issues**


nielsvanvelzen:
We will not add support for media sessions in the old playback code. The metadata of items is incomplete and it will not work when using the LibVLC backend in the implementation of this PR.

Besides, media session support is already a thing with the new backend which is actively being worked on.

PR #3281, opened 2024-01-04
Widen support for HEVC Main 10

lawadr:
Previously the HEVC codec profile only flagged support for HEVC Main 10 when MediaCodec advertised its support for level 5 (main tier) or above. This meant devices that support only levels 4 or 4.1 (such as the Chromecast with Google TV HD) would transcode HEVC Main 10 content even when supported.

This change lowers the requirement from level 5 to level 4 and also adds additional profiles specifically for HEVC Main and Main 10 with conditions on the level so that levels beyond what is supported by MediaCodec get transcoded.

The result is that HEVC Main 10 level 5 videos get transcoded on devices supporting only 4.1, but HEVC Main 10 level 4.1 content plays back directly.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
@@ -24,7 +24,15 @@ class MediaCodecCapabilitiesTest {
 	fun supportsHevcMain10(): Boolean = hasDecoder(
 		MediaFormat.MIMETYPE_VIDEO_HEVC,
 		CodecProfileLevel.HEVCProfileMain10,
-		CodecProfileLevel.HEVCMainTierLevel5
+		CodecProfileLevel.HEVCMainTierLevel4
+	)
+
+	fun getHevcMainLevel(): String = getHevcLevelString(
+		CodecProfileLevel.HEVCProfileMain
+	)
+
+	fun getHevcMain10Level(): String = getHevcLevelString(
+		CodecProfileLevel.HEVCProfileMain10
 	)
 
 	fun supportsAVCHigh10(): Boolean = hasDecoder(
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
```diff
@@ -33,6 +41,48 @@ class MediaCodecCapabilitiesTest {
 		CodecProfileLevel.AVCLevel4
 	)
 
+	private fun getHevcLevelString(profile: Int) : String {
```
nielsvanvelzen, line 44:
```suggestion
	private fun getHevcLevelString(profile: Int): String {
```

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
```diff
@@ -33,6 +41,48 @@ class MediaCodecCapabilitiesTest {
...
+			level < CodecProfileLevel.HEVCMainTierLevel31 -> return "90"
+			level < CodecProfileLevel.HEVCMainTierLevel4 -> return "93"
+			level < CodecProfileLevel.HEVCMainTierLevel41 -> return "120"
+			level < CodecProfileLevel.HEVCMainTierLevel5 -> return "123"
+			level < CodecProfileLevel.HEVCMainTierLevel51 -> return "150"
+			level < CodecProfileLevel.HEVCMainTierLevel52 -> return "153"
+			level < CodecProfileLevel.HEVCMainTierLevel6 -> return "156"
+			level < CodecProfileLevel.HEVCMainTierLevel61 -> return "180"
+			level < CodecProfileLevel.HEVCMainTierLevel62 -> return "183"
+			else -> return "186"
```
nielsvanvelzen, line 61: Where do these numbers come from? It might be better if we add constants for them.

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
```diff
@@ -33,6 +41,48 @@ class MediaCodecCapabilitiesTest {
...
+
+			try {
+				val capabilities = info.getCapabilitiesForType(mime)
+				for (profileLevel in capabilities.profileLevels) {
+					if (profileLevel.profile == profile) {
+						maxLevel = maxOf(maxLevel, profileLevel.level)
+					}
+				}
+			} catch (e: IllegalArgumentException) {
+				Timber.w(e)
```
nielsvanvelzen, line 79: Add a message that explain why we ignore the exception

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
lawadr: It's copied from the below function, which has no comment either. I can add a comment here though.

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
lawadr: These are the numbers ffprobe uses and therefore the Jellyfin server. Each number is the level to one decimal place, with the decimal point removed, then multiplied by 3. So level 4.1 becomes 41*3=123 and level 5 becomes 50*3=150.

I'll add this explanation to a comment and make them constants too.

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
nielsvanvelzen: Yeah it can be really simple like `Timber.w("Unsupported codec $mime", e)`, but it helps a lot when tracing exceptions found in client logs.

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
nielsvanvelzen: That would be really helpful!

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
lawadr: > Yeah it can be really simple like `Timber.w("Unsupported codec $mime", e)`, but it helps a lot when tracing exceptions found in client logs.

This exception isn't for an unsupported codec. It just happens for each codec we aren't currently looking for. I'd expect it to happen 20+ times for each time these two functions are called. 

They should both probably be debug and not warning messages.

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
```diff
@@ -33,6 +58,35 @@ class MediaCodecCapabilitiesTest {
...
 
+	private fun getHevcLevelString(profile: Int): String {
+		val level = getDecoderLevel(MediaFormat.MIMETYPE_VIDEO_HEVC, profile)
+
+		return hevcLevelStrings.asReversed().find { item: Pair<Int, String> ->
+			level >= item.first
+		}?.second ?: "0"
+	}
+
+	private fun getDecoderLevel(mime: String, profile: Int) : Int {
```
nielsvanvelzen, line 69:
```suggestion
	private fun getDecoderLevel(mime: String, profile: Int): Int {
```

nielsvanvelzen (approved):

nielsvanvelzen:
Thank you!

PR #3288, opened 2024-01-05
Support setting audio and subtitle streams via remote control

3l0w:
**Changes**
Implemented SET_AUDIO_STREAM_INDEX and SET_SUBTITLE_STREAM_INDEX commands by calling associated method on PlaybackController.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/eventhandling/SocketHandler.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/eventhandling/SocketHandler.kt
@@ -57,6 +57,8 @@ class SocketHandler(
 				supportsMediaControl = true,
 				supportedCommands = buildList {
 					add(GeneralCommandType.DISPLAY_CONTENT)
+					add(GeneralCommandType.SET_SUBTITLE_STREAM_INDEX)
+					add(GeneralCommandType.SET_AUDIO_STREAM_INDEX)
 
 					add(GeneralCommandType.DISPLAY_MESSAGE)
 					add(GeneralCommandType.SEND_STRING)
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/data/eventhandling/SocketHandler.kt
```diff
@@ -92,6 +94,20 @@ class SocketHandler(
 			addListener<PlayMessage> { message -> onPlayMessage(message) }
 			addListener<PlayStateMessage> { message -> onPlayStateMessage(message) }
 
+			addGeneralCommandsListener(setOf(GeneralCommandType.SET_SUBTITLE_STREAM_INDEX)) { message ->
+				val index by message
+				coroutineScope.launch(Dispatchers.Main) {
+					playbackControllerContainer.playbackController?.switchSubtitleStream(index!!.toInt())
+				}
```
nielsvanvelzen, line 101: General command messages are currently not typed safely in the server, so we need to check for null and for a valid int.

So this would be something like:
```kotlin
val index = message["index"]?.toIntOrNull()
if (index != null) {
// magic
}
```

nielsvanvelzen (changes requested):
Thanks!

app/src/main/java/org/jellyfin/androidtv/data/eventhandling/SocketHandler.kt
3l0w: Oh i didn't know that, didn't notice that the "by" keyword return a nullable.
Good catch !

3l0w:
Fixed and rebased from master

nielsvanvelzen (approved):

PR #3309, opened 2024-01-12
Prevent usage of Jellyfin 10.9 to prevent crashes due to incompatibility

nielsvanvelzen:
We already have a minimum supported server version, this PR adds a minimum unsupported version (the first version we know doesn't work). Right now the app and 10.9 are incompatible. Depending on how many issue reports are going to come in I'll decide if we really need this or not (in which case the PR will suddenly be gone)

**Changes**
- Block login when server is using a version too new and marked as unsupported (currently set to >=10.9)
- Added additional documentation to the versions specified

**Issues**


nielsvanvelzen:
Closing as I don't see much use for this anymore. It would also require us to keep bumping the unsupported version periodically so it doesn't create false-positives and I'm sure I'm going to forget.

Issue #3257, opened 2023-12-25
RTL Subtitles (Puncitions)


Obegg:
### Describe the bug

1. Download subtitles that it's language is RTL such as Hebrew, here's an example of `.srt` subtitle:
```
1
00:00:16,975 --> 00:00:18,435
.את חייבת לספר לי

```
2. Play the video file with the RTL subtitles on the Google Chromecast device
3. Puncitions are shown LTR instead of RTL:
```
1
00:00:16,975 --> 00:00:18,435
את חייבת לספר לי.

```

Note: Plex has the same issue.

### Application version

0.16.1

### Where did you install the app from?

Google Play

### Device information

Google Chromecast 4K (GA01919-US)

### Android version

Android 12

### Jellyfin server version

10.8.13

nielsvanvelzen:
I've made a possible fix but am unable to test it. Please try the app build from #3267.

To test:
- Go to the build artifacts (https://github.com/jellyfin/jellyfin-androidtv/actions/runs/7351258722?pr=3267)
- Scroll down and click on the "build-artifacts" link in the "Artifacts" section
- Unzip the file
- Sideload the .apk file to your device

It will install as separate app (Called Jellyfin Debug, with an orange logo)

Shasoosh:
Are there any updates on this? 
I would appreciate it being addressed, similar to the resolution in https://github.com/jellyfin/jellyfin-web/pull/4895

PR #3332, opened 2024-01-26
Use LTR for default subtitle direction

MichaelRUSF:
**Changes**
Similar to  [#4895](https://github.com/jellyfin/jellyfin-web/pull/4895) setting the text direction LTR will place the punctuations in the correct direction for RTL text. 

**Issues**
Fixes https://github.com/jellyfin/jellyfin-androidtv/issues/3257

### DIFF
--- a/app/src/main/res/layout/vlc_player_interface.xml
+++ b/app/src/main/res/layout/vlc_player_interface.xml
@@ -55,6 +55,7 @@
             android:gravity="center"
             android:textColor="@color/white"
             android:textSize="28sp"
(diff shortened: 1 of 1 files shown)


MichaelRUSF:
Before: 


After: 
RTL-Fix

Test srt [file](https://github.com/jellyfin/jellyfin-androidtv/files/14059795/Tears.of.Steel.2012.srt.txt)

nielsvanvelzen (approved):
Thanks!

PR #3336, opened 2024-01-31
Fix media details tag for multiple versions support

MichaelRUSF:
**Changes**
Changes the media details tag to check the audio stream based on the media source index.  

**Fixes**
Currently, the media details tag only reflects the first movie regardless of the version selected.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
+++ b/app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
@@ -107,7 +107,7 @@ public class InfoLayoutHelper {
         if (includeRuntime) addRuntime(context, item, layout, includeEndTime);
         addSeriesStatus(context, item, layout);
         addRatingAndRes(context, item, mediaSourceIndex, layout);
-        addMediaDetails(context, audioStream, layout);
+        addMediaDetails(context, item, mediaSourceIndex, layout);
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
```diff
@@ -107,7 +107,7 @@ public static void addInfoRow(Context context, BaseItemDto item, int mediaSource
         if (includeRuntime) addRuntime(context, item, layout, includeEndTime);
         addSeriesStatus(context, item, layout);
         addRatingAndRes(context, item, mediaSourceIndex, layout);
-        addMediaDetails(context, audioStream, layout);
+        addMediaDetails(context, item, mediaSourceIndex, layout);
```
nielsvanvelzen, line 110: Looks like the `audioStream` parameter is now unused so it can be removed (line 64)

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/InfoLayoutHelper.java
MichaelRUSF: Removing it breaks line 54 and if removed it ends up with the same parameters as line 51.

nielsvanvelzen (approved):

nielsvanvelzen:
Awesome, thanks!

PR #3348, opened 2024-02-04
android tv hls playback direct play fix

mnsrulz:
**Changes**
Allow m3u8 links embedded in the .strm files to be direct played

**Issues**
Two fixes mainly, one was that the devices are not fully recognized as appropriate for h264 levels which is forcing the transcoding. Second is to allow the m3u8 links to be direct play instead of rencode.


mnsrulz:
this is essentially making android player inline with roku

nielsvanvelzen (approved):
Thanks!

PR #3368, opened 2024-02-11
Dim static elements in screensaver

abjorck:
I'd like to "reopen" pull #2566 to dim the white static elements on the screensaver

**Changes**

Set alpha on logo, clock, and content title.

I also did a miniscule change to the fade since I think the snap-out was sometimes off and gave a janky feeling when the old image disappeared, not sure why, maybe a frame timing thing. Seems hard to record and report but I can try making it a seperate issue if you dont want it here

**Issues**


nielsvanvelzen (changes requested):
Somewhat dimming the text is fine but the Jellyfin logo should not be made semi-transparent.

nielsvanvelzen:
Closing as stale

PR #3372, opened 2024-02-14
Remove "auto" max bit rate and add additional values

nielsvanvelzen:
**Changes**
- Remove "auto" max bitrate preference
  - A value of "0" will now default to the default value of 100Mbit/s
- Add new values for max bitrate: 140Mbit/s, 180Mbit/s and 200Mbit/s
- Require UserPreferences as parameter in Utils.getMaxBitrate

**Issues**


crobibero (approved):

arthurmelton:
Is there any way the "auto" setting could be restored but dynamically change the bitrate depending on how much buffer is left? For example, it could target a 15-second buffer: if the buffer reaches 25 seconds (without direct play/the highest bitrate possible), it would increase the bitrate; if it drops to 5 seconds, it would decrease the bitrate.

PR #3379, opened 2024-02-17
Rewrite BaseItemInfoRow with Compose

nielsvanvelzen:
This is an old branch of mine that I started about a year ago but never finished... I've finally completed it now lol.

The info row is what is shown below the title while browsing, almost all screens use it and thus it is a bit messy. This pull request aims to rewrite it in Kotlin with Compose, not necessarily to clean it up.

The info row items shown are mostly the same. There are some tweaks to the design, improvements to datetime formatting and in some cases some items are removed/reordered or added.

This also adds a lot of proper plurals to the strings, so stuff like "1 episodes" should not happen in this row anymore.

**Changes**
- Replace most of InfoLayoutHelper with new BaseItemInfoRow

**Before**
91400002c3

**After**
90700002c2

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/browsing/composable/inforow/InfoRowColors.kt
```diff
@@ -0,0 +1,12 @@
+package org.jellyfin.androidtv.ui.browsing.composable.inforow
+
+import androidx.compose.ui.graphics.Color
+
+/**
+ * Colors used inm the [BaseItemInfoRow].
```
crobibero, line 6:
```suggestion
 * Colors used in the [BaseItemInfoRow].
```

crobibero (approved):

PR #3384, opened 2024-02-19
Change background colour to transparent

### LINKED ISSUE #2967: Grey bars when playing HDR content on Sony Bravia
### Describe the bug

1. Play HDR content
2. Profit

Was previously fixed in #369, but apparently this bug is back.

### Logs

_No response_

### Application version

0.16.0-beta2

Google Play

### Device information

Sony Bravia X900F

### Android version

Android 9

### Jellyfin server version

10.8.10

viggy96:
**Changes**
Changed background colour to transparent in layouts. Video background is now completely black during HDR video playback on Sony Bravia TVs. Tested on my Sony Bravia X900F.

**Issues**
Fixes #2967

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
@@ -1,6 +1,8 @@
 package org.jellyfin.androidtv.data.service
 
 import android.content.Context
+import android.graphics.Color
+import android.graphics.drawable.ColorDrawable
 import androidx.compose.ui.graphics.ImageBitmap
 import androidx.compose.ui.graphics.asImageBitmap
 import androidx.core.graphics.drawable.toBitmap
@@ -125,6 +127,13 @@ class BackgroundService(
 		if (_backgrounds.isEmpty()) return
 
 		_backgrounds = emptyList()
+		update()
+	}
+
+	fun removeBackgrounds() {
+		loadBackgroundsJob?.cancel()
+
+		_backgrounds = listOf(ColorDrawable(Color.TRANSPARENT).toBitmap(1, 1).asImageBitmap())
 		update()
 	}
 
--- a/app/src/main/java/org/jellyfin/androidtv/ui/background/AppBackground.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/background/AppBackground.kt
@@ -18,7 +18,6 @@ import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.BlendMode
-import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.ColorFilter
 import androidx.compose.ui.graphics.asImageBitmap
 import androidx.compose.ui.layout.ContentScale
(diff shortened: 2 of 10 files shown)


nielsvanvelzen (changes requested):
This change does not only affect the background of the video player but also of all other screens which is not desired behavior.

viggy96:
I didn't see any unintented adverse effects when running the app however.

nielsvanvelzen:
Immediately after starting the app with the changes in this PR I can see a different background color on the home screen. It doesn't show when you have backdrops enabled and the item in focus has one. A good method to test is to change your theme to emerald and disable backdrops from showing, that will make it obvious.

viggy96:
Hmm, I don't see anything in that case either. I set the theme to emerald, disabled backdrops and the app restarts, and the home screen background is just black.

viggy96:
I haven't found where to change just the background for just the video player. From what I've seen, ExoPlayer is only responsible for the area taken up by the video, and bars above and below are done by the app.

https://github.com/google/ExoPlayer/issues/8803

nielsvanvelzen:
> Hmm, I don't see anything in that case either. I set the theme to emerald, disabled backdrops and the app restarts, and the home screen background is just black. I guess its not "not_quite_black" anymore, but I don't see that as an issue per say.

The background of the emerald theme is not black though.

> Are you okay with the workaround only applying to Sony devices? That way your concern does not show up on other devices, and the theme thing can be marked as a known limitation.

No. We do not add device specific changes as they are hard to test. The only thing we allow is testing for the API level or platform features.

viggy96:
Why can't the background be black though?

viggy96:
I understand that the emerald theme currently does not have a black background, my question is why it can't be black?

nielsvanvelzen:
> I understand that the emerald theme currently does not have a black background, my question is why it can't be black?

Because that is how the theme is supposed to look.

> Isn't the background service cleared of all backgrounds before playback begins?

Yeah, we ask it to "forget" all the backdrops so it will change to the default background, which is actually a copy of the window background made when the BackgroundService initialized.

viggy96:
Hmm, as an experiment I added a function to set the background to black in the background service, and I set the default theme window background to transparent. 
I can tell the window background takes effect, because the background while browsing the home screen or on the page for a particular item the background is pitch black instead of "not quite black". However, when starting playback, the grey bars appear, which makes me believe that the background service isn't changing the background itself, but rather drawing the background over the underlying window background of the app.

viggy96:
A simple solution would be to have just one theme that actually has a transparent background.

nielsvanvelzen:
> which makes me believe that the background service isn't changing the background itself, but rather drawing the background over the underlying window background of the app.

Yes that is exactly how it works.

> A simple solution would be to have just one theme that actually has a transparent background.

Not a solution

viggy96:
>Yes that is exactly how it works.

But that would imply that it doesn't matter what's done in the BackgroundService, because it can't actually change the underlying background.

nielsvanvelzen:
Thanks for your work on this, your insights helped me get to a solution that should work and keep working in the future when more UI changes are made. I've made a different fix that is similar to what you have right now in #3411. I've updated the "Muted Purple" theme to have a very obvious pink background. Can you test for me:

1. The pink background showing while browsing etc.
2. The background being properly black during video playback

If this fix work I'll make sure it will be in the next 0.16 release.

Issue #3401, opened 2024-02-23
Some devices do not report AVC High10 capabilities even though playback is supported, causing decoder checks to fail


MichaelRUSF:
### Describe the bug

The ```hasDecoder```  function in MediaCodecCapabilitiesTest does not work as expected.  It always returns false for: 
```supportsAVCHigh10```  
```supportsHevcMain10 ``` 

It works correctly if you test: 
```supportsHevcMain```  		
```supportsAV1Main10```  		
*```supportsAVCProfileMain``` 	

 @lawadr Also,  the ```getDecoderLevel``` which is based off of  the ```hasDecoder``` does not work  for: 
```getHevcMain10Level ``` 
*```getAVCHigh10Level``` 

It works correctly if you test:
``` getHevcMainLevel``` 
*```getAVCMainLevel```

This is why H264 High10 videos will always transcode. Also, why 4k HEVC content gets remuxed instead of being transcoded on devices that only support 1080p. I'm sure there are other bugs, but these are the two that I could test.  

*Not implemented

### Application version

0.16.5

### Where did you install the app from?

None

### Device information

~

### Android version

~

### Jellyfin server version

10.8.3

**Comment by lawadr** [comment]
You haven't listed your device. Maybe it doesn't support those profiles?

**Comment by MichaelRUSF** [comment]
I tested out the code on Andriod Studio with the emulator. Lots of ways to test it, but I'll try to explain my method.

If this is true (MediaCodecCapabilitiesTest)
```kotlin
fun supportsAVCHigh10(): Boolean = hasDecoder(
		MediaFormat.MIMETYPE_VIDEO_AVC,
		CodecProfileLevel.AVCProfileHigh10,
		CodecProfileLevel.AVCLevel4
	)
``` 
Then because of this (ProfileHelper) AVC High 10 is supposed to direct play given no other factors, e.g., subtitles, level support, bitrate, etc.. 
```kotlin
	val h264VideoProfileCondition by lazy {
		ProfileCondition(
			ProfileConditionType.EqualsAny,
			ProfileConditionValue.VideoProfile,
			listOfNotNull(
				"high",
				"main",
				"baseline",
				"constrained baseline",
				if (MediaTest.supportsAVCHigh10()) "high 10" else null
			).joinToString("|")
		)
	}
``` 
Using current build, play  AVC High 10 L4.2 media, no subtitles or audio to eliminate other factors. Notice the video will transcode, reason should be unsupported profile.


Change this

[...]

PR #3406, opened 2024-02-27
Add a preference option to bypass AVC High10 checks

MichaelRUSF:
Allow users to bypass AVC High10 checks by adding a preference option. Enabling this option will automatically add AVC High10 as a supported profile, as long as the device supports AVC Main. The same AVC Main levels will be used for AVC High10.

**Changes**

- Add user preference to bypass AVC High10 check
- Add a new profile condition in ProfileHelper to be used if bypass is enabled
- Update ExoPlayerProfile  based on the new preference

**Issues**
Workaround for #3401

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -53,6 +53,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 
 		/* Playback - General*/
 		/**
+		 * Preferred behavior for audio streaming.
+		 */
+		var avcHigh10Bypass = enumPreference("avc_high10_bypass", false)
+
+		/**
 		 * Maximum bitrate in megabit for playback.
 		 */
 		var maxBitrate = stringPreference("pref_max_bitrate", "100")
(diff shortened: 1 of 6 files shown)


MichaelRUSF:
Agreed that this should be fixed. I see this as a stop-gap solution until a proper solution can be implemented. There are already so many Amazon Fire TV users on the troubleshooting forum with issues and the solution is often to use LibVLC. And given that LibVLC support will be temporarily removed, this might help alleviate some of those issues. Also, an HEVC bypass could be implemented for more compatibility issues. 

In the long run, I don’t see this as a big issue b/c most AVC High10 media contain subtitles that trigger transcoding anyway. Probably the reason this has slipped under the radar. 

As for a fix, I hope that someone can come up with something. How do you test for a condition when the device doesn’t report that condition, even though the device supports it? 

You can go ahead and close this. I think I gave myself like 5% odds that you would accept it 😊.

lawadr:
I've done some thinking into handling AVC and HEVC profile levels more accurately and I think the way HEVC is done now, combined with per-device raising or lowering of the maximum profile levels and/or supported profiles is the _only_ way to go.

The way I see it, if AVC or HEVC struggles to play on a device, upon some investigation it should be quite easy to determine if a profile needs removing for that device or a profile level reduced.

For existing devices, if things start to transcode that weren't transcoding before these recent changes, it should be quite easy to bump up a profile level or add a new profile.

For future devices with incorrect MediaCodec reporting, I guess there will be no regression so maybe no one will care that some hypothetical HEVC Main10 6.1 file fails to direct play on a device that reports up to 6.0.

Ultimately, you can either hard code the exact codec support of every single device out there (which is incredibly difficult due to the number of devices and the lack of information out there for even the most popular devices), or you rely on what the device reports through MediaCodec and massage the results to fit some edge cases as and when they get reported.

Issue #3152, opened 2023-10-31
Jellyfin 0.16.0 beta 5 plays 7.1 Opus audio as stereo


ryannathans:
### Describe the bug

1. play content with 7.1 opus audio
2. comes out only in stereo, and in my case, dialogue on the right with music on the left

### Media info of the file

```shell
General
Complete name                            : video recording.mkv
Format                                   : Matroska
Format version                           : Version 4
File size                                : 13.3 GiB
Duration                                 : 3 h 12 min
Overall bit rate                         : 9 881 kb/s
Frame rate                               : 23.976 FPS
Movie name                               : Video Recording (2021)
Encoded date                             : 2023-06-27 18:55:00 UTC
Writing application                      : mkvmerge v63.0.0 ('Everything') 64-bit
Writing library                          : libebml v1.4.2 + libmatroska v1.6.4
Attachments                              : av1an-06-27-2023-01-42-06.txt
Video
ID                                       : 1
Format                                   : AV1
Format/Info                              : AOMedia Video 1
Format profile                           : Main@L5.0
HDR format                               : SMPTE ST 2086, HDR10 compatible
Codec ID                                 : V_AV1
Duration                                 : 3 h 12 min
Bit rate                                 : 9 496 kb/s
Width                                    : 3 840 pixels
Height                                   : 2 076 pixels
Display aspect ratio                     : 1.85:1
Frame rate mode                          : Constant
Frame rate                               : 23.976 (24000/1001) FPS
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Bit depth                                : 10 bits
Bits/(Pixel*Frame)                       : 0.050
Stream size                              : 12.8 GiB (96%)
Title                                    : 2160p HDR AV1
Default                                  : Yes
Forced                                   : No
Color range                              : Limited
Color primaries                          : BT.2020
Transfer characteristics                 : PQ
Matrix coefficients                      : BT.2020 non-constant
Mastering display color primaries        : R: x=0.708000 y=0.292000, G: x=0.170000 y=0.797000, B: x=0.046000 y=0.131000, White point: x=0.329000 y=0.312700
Mastering display luminance              : min: 0.0050 cd/m2, max: 1000 cd/m2
Maximum Content Light Level              : 0
Maximum Frame-Average Light Level        : 0
Audio
ID                                       : 2
Format                                   : Opus
Codec ID                                 : A_OPUS
Duration                                 : 3 h 12 min
[...]

Issue #520, opened 2020-07-07
AAC LC audio always outputs 2.0 PCM. DTS/AC3 passes through correctly.


michaelkrieger:
AAC LC formatted audio always outputs 2.0ch PCM 48kHz irrespective of what passthrough settings are selected on Android TV.
DTS/AC3 correctly plays 5.1 audio through Jellyfin
No transcoding is occurring as per Jellyfin's logs.

Possible solutions:
 * Mimic what Kodi does with the options below to correctly send AAC to a receiver
 * If not possible, add ability to force Jellyfin (server) to transcode the audio from AAC to DTS/AC3

**Checks**
- Issue appears to happen on the Android TV Jellyfin app.
- Issue seems to occur with Automatic, LibVLC and ExoPlayer as well
- Issue occurs with VLC (external)
- Issue does NOT occur with Kodi (external) with the settings I have below.
- When playing the same file on an iPad or PC, it correctly shows 6 channel audio.

**Media Info of the file**
> General
> Complete name                            : S:\Movies\Movie1.mp4
> Format                                   : MPEG-4
> Format profile                           : Base Media
> Codec ID                                 : isom (isom/iso2/avc1/mp41)
> File size                                : 1.85 GiB
> Duration                                 : 1 h 37 min
> Overall bit rate                         : 2 732 kb/s
> Movie name                               : Movie1
> Encoded date                             : UTC 2020-06-29 11:26:31
> Tagged date                              : UTC 2020-06-29 11:26:31
> Writing application                      : Lavf58.20.100
> Comment                                  : Movie1
> 
> Video
> ID                                       : 1
> Format                                   : AVC
> Format/Info                              : Advanced Video Codec
> Format profile                           : High@L4.1
> Format settings                          : CABAC / 4 Ref Frames
> Format settings, CABAC                   : Yes
> Format settings, Reference frames        : 4 frames
> Codec ID                                 : avc1
> Codec ID/Info                            : Advanced Video Coding
> Duration                                 : 1 h 37 min
> Bit rate                                 : 2 500 kb/s
> Width                                    : 1 920 pixels
> Height                                   : 804 pixels
> Display aspect ratio                     : 2.40:1
> Frame rate mode                          : Constant
> Frame rate                               : 23.976 (23976/1000) FPS
> Color space                              : YUV
> Chroma subsampling                       : 4:2:0
> Bit depth                                : 8 bits
> Scan type                                : Progressive
[...]

PR #3422, opened 2024-03-01
Fix pcm multichannel audio for Tvs

### LINKED ISSUE #2991: Blacklist specific audio formats

It would be useful to be able to have the player NOT advertise support for specific audio formats.

For example, in some configurations multi-channel AAC may be "supported" by the Android TV device (e.g., NVIDIA Shield TV which supports most everything) and hence the Jellyfin client says it supports it, however the Android->TV->Soundbar path may force that audio to get downgraded to PCM stereo.  

If we could have the Jellyfin client not advertise support for that format, and therefore force the server to transcode to another format, the overall end-user audio experience would be better.

MichaelRUSF:
**Changes**
Android apps running on Smart TVs are limited to PCM 2.0 Audio, while DD(+) and DTS have full support. This causes multichannel formats such as AAC and Opus to output stereo audio when playing directly from a TV to a receiver. This change identifies what is a TV based on feature testing and will transcode non-DD (+)/DTS multichannel audio if the app is running on a TV.

Also, add a condition for enabling/disabling DTS. The bitstreaming option in the preference menu isn't tied to any action. 

**Issues**

Fixes  #2991  #3294  #3152  #520

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -570,7 +570,7 @@ public class PlaybackController implements PlaybackControllerNotifiable {
         DeviceProfile internalProfile = new ExoPlayerProfile(
                 mFragment.getContext(),
                 isLiveTv && !userPreferences.getValue().get(UserPreferences.Companion.getLiveTvDirectPlayEnabled()),
-                userPreferences.getValue().get(UserPreferences.Companion.getAc3Enabled()), (hasSurroundAudio)
+                userPreferences.getValue().get(UserPreferences.Companion.getAc3Enabled()), hasSurroundAudio
         );
         internalOptions.setProfile(internalProfile);
         return internalOptions;
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -561,6 +561,7 @@ public class PlaybackController implements PlaybackControllerNotifiable {
         internalOptions.setItemId(item.getId().toString());
         internalOptions.setMediaSources(item.getMediaSources());
         internalOptions.setMaxBitrate(maxBitrate);
+        boolean hasSurroundAudio = getNumberOfChannels() > 2;
         if (exoErrorEncountered || (isLiveTv && !directStreamLiveTv))
             internalOptions.setEnableDirectStream(false);
         internalOptions.setMaxAudioChannels(Utils.downMixAudio(mFragment.getContext()) ? 2 : null); //have to downmix at server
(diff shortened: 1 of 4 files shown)


nielsvanvelzen (changes requested):
We should not introduce new behavior that relies on the manufacturer/model of the device. Use feature testing instead.

MichaelRUSF:
New tests were added based on features.

nielsvanvelzen:
If you want to know whether an audio codec is supported or not you should not test for various system features but for that specific audio codec.

MichaelRUSF:
Testing the audio codec will not work for this issue. The issue is not the lack of codec support from the TV, it’s the fact that TV apps are not allowed to output multichannel PCM, which is what Jellyfin sends for Non-DD (+)/DTS media. 

If the app is running directly from the TV, testing for Non-DD (+)/DTS codecs would come back as true if supported, but if the media is multichannel then only stereo audio will be played b/c of these limitations.

TVs can passthrough multichannel PCM from external devices plugged in via HDMI. So this is not an issue for any connected devices, e.g., Shields, Fire TVs, Chromecast, etc.

There are only two ways to solve this. Either detect if an app running from a TV and transcode multichannel PCM formats to DD, which this does. Or give users the option to bypass manually. Because automatic codec detection will not work for this issue.

MichaelRUSF:
> However all internal apps including Plex do not support lossless audio (like Dolby TrueHD) passthru, only external HDMI devices can use eArc to passthru the lossless audio formats - so those require an external media player that supports those formats. 

>There is a hardware limitation that prevents full eArc support of lossless audio codecs from the internal TV apps, this affects all current TVs not just the Sonys. eArc is only fully supported on current TVs with external HDMI devices. Internal app audio support is based on using them with streaming services that use compressed DD+ with atmos audio at maximum.

[Source](https://www.avsforum.com/threads/2021-sony-bravia-xr-a80j-oled-owners-thread-no-price-talk.3207724/page-570#post-61296846)


>If you select PCM even though you have a home theater system or soundbar connected, the sound system will only receive 2.0 channel sound and the result will not be multi-channel surround sound, even if the sound system is capable of multi-channel PCM audio.

[Source](https://www.samsung.com/us/support/answer/ANS00085244/)

entropicdrifter:
Maybe it would make more sense to just have a setting in the menu, something like a toggle to turn on transcoding of all surround (that's not already DD, DD+, or DTS) into Dolby or DTS. That'll fix AAC, OGG, TrueHD and DTS-HD MA all at once on the devices in question. 

People who don't bother to turn on the setting are probably the same people who wouldn't notice/care about the issue, right?

nextlooper42:
Same in Kodi, there is a toggle to transcode all unsupported audio to AC3. Works like charm.

nielsvanvelzen:
What's the current state of this change? Do we still need it or has it been fixed with one of the other profile changes already?

MichaelRUSF:
Closing with new proposed PR #3936 to add a S/PDIF/ARC audio output mode. Which corresponds with the capabilities of those standards:

image png 5886fa9dfdbdb66ab015970b4d8447a9
Source: HDMI Licensing
[Source](https://www.rtings.com/soundbar/learn/opitical-vs-hdmi)

PR #3435, opened 2024-03-07
Add WoL-capabilities

### LINKED ISSUE #2823: Wake server on lan when client is booting

To limit unnecessary power consumption, make a configuration that enables the user to use WOL on an attached server, enabling that the server can be available only when needed.

newhinton:
**Changes**
This adds a small util that can send a WoL-Packet. This allows the client to autostart a sleeping server.
This PR is missing the user interface to actually set mac&broadcast-ip. I need help implementing this.

**Issues**
Fixes #2823


nielsvanvelzen:
Thank you for the pull request but we have no interest in wake on lan functionality in any of our clients. It does not fit the scope of Jellyfin overall, this was decided years ago.

newhinton:
Ahh, that's sad. But understandable.

Thank you!

crusoe01:
Not understandable. The jellyfin client's has should be a plugin what send a magic pocket for the sleeping server. It should have be basic function.

PR #3438, opened 2024-03-08
Recreate icon/banner resources

nielsvanvelzen:
Google is currently rolling out an update to the Google TV launcher that will use app icons instead of banners. The icon we currently have has too little padding causing it to be cut off.
This PR not only fixes that but replaces all the icons/banners exposed in the manifest (so not the ones used inside of the app) to adhere to the current recommendations from Google, additionally the banner is based on our new logo font. Credits to @erikbucik for creating all the resources in the correct sizes.

**Changes**

- Drop all drawable icon/banner resources
- Drop all debug icon/banner resources (in-app logo still has it)
- Add new mipmap resources for icon and banner
  - Icon also uses adaptive icon
  - Banner does **not** use adaptive icon because the OS automatically adds a rounding to the corner which breaks most launchers (including the stock Android TV launcher)
- Update fastlane icon/featureGraphic/tvBanner


nielsvanvelzen:
image

In the Google TV emulator

crobibero (approved):
👀 new branding

PR #3443, opened 2024-03-11
Update app_logo to use new logo font

nielsvanvelzen:
**Changes**
- Update app_logo to use new logo font
- Update logo sizing in splash screen and screensaver
  - This was needed because previously the new app_logo has a different size

I'm not going to update the app screenshots for the logo change, will do that once we make some actual UI improvements.

**Screenshots**

image
screensaver

image
search


helptiger:
https://github.com/jellyfin/jellyfin-androidtv/commit/47282c09b002ce132a3c1111490acdb0b979df60#commitcomment-140549715

PR #3472, opened 2024-03-26
Update to Kotlin SDK 1.5.0-beta.1 (Jellyfin 10.9)

nielsvanvelzen:
Jellyfin 10.9 is here! While our current release supports both 10.8.1>= and 10.9.0, it does not use any of the new features. This will obviously change in the coming weeks. We need a new SDK version that allows us to do so. This is the initial PR that bumps the SDK and migrates all broken code. I've been running this branch on my own device for the past few weeks, keeping up-to-date with API changes.

With this change the minimum server version is now 10.9.0. It is no longer possible to use 10.8.z with the app.

**Changes**
- Update to jellyfin-sdk-kotlin version 1.5.0-beta.1
- Migrate code

**Issues**


app/src/main/java/org/jellyfin/androidtv/util/RecommendedServerIssueExtensions.kt
```diff
@@ -39,6 +39,9 @@ fun RecommendedServerIssue.getFriendlyMessage(context: Context): String = when (
 	)
 
 	is RecommendedServerIssue.SlowResponse -> context.getString(R.string.server_issue_timeout)
+
+	is RecommendedServerIssue.SecureConnectionFailed -> TODO()
+	is RecommendedServerIssue.ServerUnreachable -> TODO()
```
nielsvanvelzen, line 44: Reminder for self to fix this

app/src/main/java/org/jellyfin/androidtv/util/RecommendedServerIssueExtensions.kt
```diff
@@ -47,12 +50,15 @@ fun RecommendedServerIssue.getFriendlyMessage(context: Context): String = when (
...
-		is RecommendedServerIssue.UnsupportedServerVersion -> 4
-		is RecommendedServerIssue.OutdatedServerVersion -> 3
-		is RecommendedServerIssue.SlowResponse -> 2
+		is RecommendedServerIssue.MissingSystemInfo -> 7
+		is RecommendedServerIssue.UnsupportedServerVersion -> 6
+		is RecommendedServerIssue.OutdatedServerVersion -> 5
+		is RecommendedServerIssue.SecureConnectionFailed -> 4
+		is RecommendedServerIssue.SlowResponse -> 3
+		is RecommendedServerIssue.ServerUnreachable -> 2
 		RecommendedServerIssue.MissingVersion -> 1
```
crobibero, line 62: How does this work without the `is`?

crobibero (approved):

app/src/main/java/org/jellyfin/androidtv/util/RecommendedServerIssueExtensions.kt
nielsvanvelzen: MissingVersion is an object so we can use an equals check. The others are data classes so we need to check if it is an instance of that class.

nielsvanvelzen: See https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/master/jellyfin-core/src/commonMain/kotlin/org/jellyfin/sdk/discovery/RecommendedServerIssue.kt#L34

PR #3557, opened 2024-05-07
Initial video support in playback rewrite

nielsvanvelzen:
These are the initial changes to get video playback working. Currently only direct-play is actually functional and you need to make sure your media can play yourself as codec detection is not working.

When media does play it already integrates with remote control and media sessions. Playback progress is also properly tracked. There are also funny bugs, like the screensaver showing the video your playing because we don't filter for music only.

Also, no subtitles at all right now.

**Changes**

- Add PlayerSurfaceView to display video output
- Check for mediaType in AudioMediaStreamResolver
- Fix PlaybackLauncher not converting item id to string
- Add hls dependency to media3 module
- Fix typo in AudioMediaStreamResolver
- Update PlaybackForwardingActivity to show video output
- Add VideoMediaStreamResolver

**Issues**

Part of #1057


crobibero (approved):
Should this be behind a "Use new video player backend" flag?

nielsvanvelzen:
#2455 👀 

image

PR #3558, opened 2024-05-07
Fix screensaver and media manager displaying video entries

nielsvanvelzen:
**Changes**
- Fix the screensaver showing video entries, it should only show audio entries (music)
- Fix the "now playing" thingy while browsing showing video entries

**Issues**

Part of #1057


app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHost.kt
```diff
@@ -55,7 +56,7 @@ fun DreamHost() {
 
 	DreamView(
 		content = when {
-			mediaItem != null -> DreamContent.NowPlaying(mediaItem)
+			mediaItem?.mediaType == MediaType.Audio -> DreamContent.NowPlaying(mediaItem)
```
crobibero, line 59: But my audio books

nielsvanvelzen: We don't currently support audiobooks in the app. When we do this will need to be updated 😉

PR #3582, opened 2024-05-16
Fix "Add server" button not visible when there are too many servers

nielsvanvelzen:
The average user has 1 Jellyfin server configured. That's why I have 7, because I'm above average.

**Changes**
- Fix the visibility of the "Enter server address" button in the server selection screen when there are 5+ servers (Will be 5px tall with 6 and gone with 7 or more).
- This effectively moves the "enter server address" button down to the bottom of the screen when you have only 1 server configured. It's something I can live with though.

**Issues**


crobibero (approved):
ok mr "above average"

PR #3592, opened 2024-05-20
Add user configuration of player menu background and fade out

Draegert:
Added support for user configuration of player menu background colors.
This allows the user to select between background colors (transparent, light (default), and dark) in the playback settings under the new 'Player menu'-section.
The player menu is shown when a video is paused i.a. and always has slight darkening of the video, until now.

Added support for user configuration of player menu fade out timer.
This allows the user to set the fade out timer (between 0 and 6 seconds) in the playback settings under the new 'Player menu'-section.
The player menu takes 6 seconds to fade out after a video is resumed, until now, where the user can adjust the timer as they desire. By setting the timer to 0, the play menu won't show when using skips but still when you pause.

I found these behaviours bothersome, so I fixed it and now share the result.

Keep up the good work. It's much appreciated.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
@@ -3,9 +3,7 @@ package org.jellyfin.androidtv.ui.itemdetail;
 import static org.koin.java.KoinJavaComponent.inject;
 
 import android.app.AlertDialog;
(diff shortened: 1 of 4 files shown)


nielsvanvelzen:
I'm not a fan of introducing options for everything the app can do. In my opinion we should make sure we have sane defaults.

Changing the overlay hide timer to a different value is fine, but not with an option. Just change the value. What would your preference be here?

The option to change the background color I don't really understand. Why would you want that?

PR #3593, opened 2024-05-20
Add user configuration for using DPAD left and right to skip in videos

Draegert:
Added support for user configuration of DPAD left and right actions.
This allows the user to toggle, in the playback settings under the new 'Playback controls'-section, if DPAD left and right should skip backwards and forwards when playing a video. It defaults to disabled as is practice today.

Added support for user configuration of minimum delay between skips.
This allows the user to set the delay, in the playback settings under the existing 'Advanced'-section. It defaults to 800ms as is default today.
Waiting 800ms after a skip is performed seems (for some) unnecessarily long.
It is not the most intuitive setting as is thus placed in the Advanced-section. 

I use the Nvidia Shield remote in Jellyfin and found I would like to use the DPAD left and right for skipping in videos.

Keep up the good work. It's much appreciated.


nielsvanvelzen:
I think it's fine to add this as a default behavior, no need for more preferences.

nielsvanvelzen:
Closing as stale. Please re-open or create a new pull request if interested in finishing.

PR #3596, opened 2024-05-20
Upgrade ConnectHelpAlertFragment to Jetpack Compose

pascalwei:
Change the XML implementation of ConnectHelpAlertFragment to Jetpack Compose and integrate Material TV as a basis for better theming options in the future.

This is just a trial balloon activity to see if a switch to the new Material design is welcomed or not.

Before:

before

After:

after


nielsvanvelzen:
This looks really good (better then my 1-1 rewrite in https://github.com/jellyfin/jellyfin-androidtv/compare/master...nielsvanvelzen:jellyfin-androidtv:compose-connect-help-alert). Unfortunately there's one major problem, we can't use compose yet for anything interactive as there are still focus issues. Additionally I'm not sure if I want to use tv-material or create our own theme. We've had extremely bad experience with using the androidx.leanback libraries.

From my regular testing the focus issues are solved in the [current beta](https://developer.android.com/jetpack/androidx/releases/compose-ui#1.7.0-beta01) for compose-ui. We don't use alpha/beta releases of the androidx libraries in our app though (except for tvprovider as google stopped updating that library). I'm expecting the new compose-ui version to release as stable in the coming months, we can probably merge this at that point in time.

nielsvanvelzen:
After playing around with the androidx.tv.material library in my playground app, I've decided we can use it for now. In #3611 I migrated the existing code to use it too.

Additionally, I think it's fine to remove the "ok" button if we replace it with a guide on how to close. Similar to how our rich text popup works (see bottom-left):



A lot of our users are not too familiar with smart-tv remotes so providing a small hint to press the back button will help them.

pascalwei:
Sure, when our users expect a button we should include one. I don't think there is a significant difference between a guide button in the lower left and a normal button though, as we have the space below the text anyways. I added the previous button back in another commit.

Thank you for taking a look at the library as well. I'll see what else I can port over to Compose.

nielsvanvelzen:
We've now updated to compose 1.7 and all focus related issues I encountered before are resolved. This means we should be able to go forward with this pull request! Can you rebase the pull request and revert the custom theme? That one can be added later.

pascalwei:
Thank you for the heads-up. Without the focus requester on the button it still didn't work to go back in one click, so I left it in there for now. I've removed the theme as you requested, however keep in mind that tv-material3 doesn't share the android-material3 or leanback themes, meaning until we add our own theme it will be inconsistent full white:
image

nielsvanvelzen:
The initial commits hardcoded the colors to use using `colorResource()`, which is fine to use for now.

PR #3611, opened 2024-05-27
Migrate to androidx.tv.material

nielsvanvelzen:
The androidx.compose.material library is for mobile/tablet. The androidx.tv.material library is for televisions. This PR switches our current compose over to the tv library, although it doesn't really affect much as we didn't really use material components (yey).

**Changes**
- Remove `androidx.compose.material`
- Add `androidx.tv.material`
  - Replace LinearProgressIndicator with Box + draw instructions
  - Replace Card with Box (the TV card is not usable for a message box)

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentNowPlaying.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentNowPlaying.kt
@@ -13,8 +13,6 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.shape.RoundedCornerShape
-import androidx.compose.material.LinearProgressIndicator
-import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentNowPlaying.kt
```diff
@@ -116,7 +116,9 @@ fun DreamContentNowPlaying(
 				text = item.run {
 					if (!artists.isNullOrEmpty()) return@run artists?.joinToString(", ")
 					val albumArtistNames = albumArtists?.mapNotNull { it.name }
-					if (!albumArtistNames.isNullOrEmpty()) return@run albumArtistNames.joinToString(", ")
+					if (!albumArtistNames.isNullOrEmpty()) return@run albumArtistNames.joinToString(
+						", "
+					)
```
thornbill, line 121: This formatting is kind of weird... what about brackets for the `if` instead?

nielsvanvelzen: Good catch, file was auto-formatted when I updated the import. I've rewritten the code here to make it more readable (while producing the same output).

thornbill (approved):

PR #3615, opened 2024-05-28
Fix app crash in EditServerScreen

pascalwei:
The java.text.DateFormat can't handle java.time.Instant resulting in a crash when loading the EditServerScreen.

Trigger crash:
1. Login to a server
2. Preferences > Login
3. Press server under "manage servers"


nielsvanvelzen (approved):
Good catch, most likely a side-effect of my many refactoring rounds the past few weeks.

PR #3039, opened 2023-09-12
Remove LibVLC

nielsvanvelzen:
The next big release (0.17) will use the new playback code for video playback. The old code will be completely removed. I need to use this old code as a reference though and the current exo+vlc mix is just a pain to look around in. This pull request removes the VLC implementation to make it easier to navigate around. **This does not mean we will not use LibVLC again!** The focus for now is to make the ExoPlayer backend work as best as we can in the rewrite and when that's done we can consider new backends (VLC/MPV/etc).

**Changes**
- Remove the LibVLC implementation from the old playback code

**Issues**

Part of #1057

PR #3639, opened 2024-06-02
Remove LibVLC

### LINKED ISSUE #3623: CVE-2023-5217 - update libvlc
### Describe the bug

ref: https://www.openwall.com/lists/oss-security/2023/09/28/5

ref: https://github.com/jellyfin/jellyfin-androidtv/blob/v0.16.11/gradle/libs.versions.toml#L40

### Logs

_No response_

### Application version

-

None

### Device information

-

### Android version

-

### Jellyfin server version

-

nielsvanvelzen:
The last stable release of libvlc-jni is version 3.5.1, released August 11th 2022. There have been multiple "early access preview" releases since but those are not stable releases. Unfortunately the members of the F-Droid team made us aware that the latest stable does contain at least one unpatched  CVE (see #3623).

Our legacy video player implementation has always been a pain to work with due to the mix of ExoPlayer and LibVLC code. The LibVLC implementation in particular could cause memory issues and could sometimes be active when using ExoPlayer, resulting in poorer performance. Additionally, the library caused the APK/AAB size to increase significantly. Because of this we've marked the implementation as experimental some time ago, moved the default over to ExoPlayer and started discouraging the use of LibVLC.

The new playback implementation will not support a LibVLC backend at launch. I can't say for sure whether it will return, it may be possible we choose a different engine (like MPV/FFmpeg) or only support ExoPlayer with extensions.

For the above reasons (no new releases, CVE, messy code) I'm removing the implementation now for 0.17. This makes it slightly easier to maintain old playback code (and increases reliability/performance) while work continues on the new implementation.

**Changes**
- Remove LibVLC implementation

**Issues**

This is a rebase of #3039.
Fixes #3623


nielsvanvelzen:
| File | Size on master branch | Size on this branch |
| --- |  --- |  --- |
| build-artifacts.zip | 107MB | 24MB |
| jellyfin-androidtv-v0.0.0-dev.1-debug.apk | 110MB | 26MB |

crobibero (approved):

dinosmm:
The reasons for removing LibVLC are understandable. 
Is there anything in the pipeline to re-add DTS:X (and -HD, -HD MA) audio support to the Android TV client (as LibVLC was the only way to play these movies)?

nielsvanvelzen:
DTS should already be supported according to our device profile. If you run into any issues with it, open an GitHub issue. Don't reply to old pull requests..

PR #3654, opened 2024-06-08
Replace ItemQuery with GetItemsRequest

nielsvanvelzen:
There are probably some bugs in this one... 

**Changes**
- Replace ItemQuery with GetItemsRequest
- Rewrite GenericFolderFragment and CollectionFragment in Kotlin
- Fully remove ApiClient from ItemRowAdapter 🎉 
- Fix item update magic code clearing everything
- Optimize item updating by using collection instead of array
- Fix crash when opening Live TV recordings user view
- Remove StdItemQuery

Behavior of the "New premieres" row is slightly changed to make it more efficient and consume less API resources. It still does some client side filtering that can be removed once Jellyfin 10.10 is released (requires jellyfin/jellyfin#12027).

**Issues**


nielsvanvelzen:
With this the only places left using the legacy ApiClient are:

- Old playback code
- Most Live TV related UI
- A few browsing related places (like KeyProcessor)

Additionally we still use some of the utility classes like `Response<T>` that needs cleaning.

Issue #3389, opened 2024-02-21
Next season is not autoplayed on Android TV client


rmatte:
### Describe the bug

On the Android TV client when a season ends it doesn't automatically progress to the next one like it should. In the web interface and in the Roku client it does, so clearly this is a bug. It's annoying having to manually play the next season when binge watching shows, especially in cases where each season only has a handful of episodes in it, it gets tiresome after a while. It would be nice to at least have an option that could be toggled on/off for this (on by default would make sense).

I'm currently using the beta version of the Android TV client but this was the case with the non-beta version as well.

Also see this other report of the same issue 8 months ago: https://forum.jellyfin.org/t-autoplay-not-working-at-end-of-season-android-tv

### Logs

Not applicable

### Application version

0.16.5

### Where did you install the app from?

None

### Device information

Sony Android TV

### Android version

Android 13

### Jellyfin server version

10.8.13

nielsvanvelzen:
This is intended behavior so I've changed this from bug to feature request.

rmatte:
Not sure how this is intended behavior when it works differently in the roku app and when viewing via the web interface (both of those advance to the next season like anyone would expect them to).

rmatte:
Not sure how this is intended behavior when it works differently in the roku app and when viewing via the web interface (both of those automatically advance to the next season like anyone would expect them to). I don't think it's at all unreasonable to consider this a bug when there's such an obvious disparity in functionality between the other clients and this one.

nielsvanvelzen:
The code is written to work in a specific way, so it is intended behavior. That does not mean it is desired behavior. The way something works in a different app doesn't matter.

ConnorS1110:
I am working on a PR for this. There is absolutely no queuing settings on the web interface other than playing the next episode automatically or not (which also does not exist on the android tv version). 
As I have been thinking about this. I think this setting on the android TV version IS supposed to be the `Play next episode automatically` setting and maybe the two diverged at some point or whoever developed this had a different concept of what that would entail. I am working on basically changing the language of this playback setting and removing the season parameter when it gets the items.
Behavior will be preserved when this setting is turned off for consistency, and also is what should be happening with the web interface (on the web interface when an episode finishes with this setting off it just stays on the player, but maybe it makes more sense to go back to the details back for the episode like it does on the android tv version). I hope to have this up soon.

CodeMonkeyUK:
[...]

PR #3658, opened 2024-06-09
Change 'Queue all subsequent episodes' setting to queue all episodes in a series

ConnorS1110:
**Changes**
This setting I think was intended to match the `Play next episode automatically` setting like in the web interface. I updated the language to match that setting, and it not will queue the entire series, vs just the current season. Behavior is maintained when the setting is disabled.

**Issues**
Addresses #3389


nielsvanvelzen (approved):
Thanks!

Issue #2269, opened 2022-11-16
Crush when 'Pass Direct Path for external player' is enabled and starting a video with VLC


miberecz:
### Describe the bug

Since the update to v15 whenever I open a media file I get 'Ops something went wrong..." error. 
After some debugging I found out, if I turn off the _**Preferences -> Playback -> Pass Direct Path for external player**_ the error goes away.
 

### Logs

---
client: Jellyfin for Android TV
client_version: 0.15.0-beta.1
client_repository: https://github.com/jellyfin/jellyfin-androidtv
type: crash_report
format: markdown
---
### Logs

***Stack Trace***: 
```log
java.lang.RuntimeException: Unable to start activity ComponentInfo{org.jellyfin.androidtv/org.jellyfin.androidtv.ui.playback.ExternalPlayerActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.List org.jellyfin.sdk.model.api.MediaSourceInfo.getMediaStreams()' on a null object reference
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2914)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3049)
	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:78)
	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1809)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loop(Looper.java:193)
	at android.app.ActivityThread.main(ActivityThread.java:6680)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.List org.jellyfin.sdk.model.api.MediaSourceInfo.getMediaStreams()' on a null object reference
	at org.jellyfin.androidtv.data.compat.StreamInfo.getSubtitleProfiles(StreamInfo.java:301)
[...]

PR #3663, opened 2024-06-11
Simplify video player preference

nielsvanvelzen:
**Changes**
- Remove preferred video player preference
- Remove preferred Live TV player preference
- Add "use external player" preference (used for all video related playback)
- Remove "choose player" menu

**Issues**


neopc10:
Can we have something like "long press play button in episode details to use external player?"

nielsvanvelzen:
No, you either have external player enabled for everything or you don't.

crobibero (approved):

gulabbisht:
@nielsvanvelzen, is there any way to choose which external player will be used, like vlc?

nielsvanvelzen:
The external player functionality uses the video player configured as default app in the OS.

kingp0dd:
> Can we have something like "long press play button in episode details to use external player?"

I can see a use case for this too. I use exoplayer as default for the resume functionality, but there are videos that do not play with exoplayer, which i have to open with Kodi. So what i do is go to settings to choose "Ask every time" as default video player and then revert the setting again after watching the unplayable video

jaum20:
Why was libvlc option removed?

nielsvanvelzen:
Please stop commenting on pull requests, especially when they're unrelated to your "issue". This pull request did **not** remove LibVLC.

sheharyarn:
Was there a reason for removing "choose player" option? It allowed customizing the playback experience, and I really liked it.

miberecz:
Its getting ridiculous. Every year a very useful feature is deprecated. There were https://github.com/jellyfin/jellyfin-androidtv/issues/2269 which was immensely useful, now this. 
The beauty of this client was that it could handle almost every type of media. If for some reason not, then I still had the choice to circumvent the given issue. Now its less and less. 
Are there any alternative clients for jellyfin on android TV?

PR #3674, opened 2024-06-13
Add a blur to backdrop images

### LINKED ISSUE #3642: Enable blurred backdrop

Option to enable blurred backdrop. When blurred background image, it creates better contrast with text, less distraction when reading and a modern look. A slow fade between backdrops when navigating is welcomed to reduce distraction.

Thank you. I love Jellyfin project.

haydenzeller:
**Changes**

Add a checkbox in the Theme customization view to blur the backdrop images. Add blurBackdropEnable to UserPreferences. Add _blur to BackgroundService. Add conditional modifier in AppBackground Image() to blur using androidx.compose.ui.draw.blur.

**Issues**

fix #3642


nielsvanvelzen:
I don't mind adding a blur to the background, that most likely improves the design. But adding an option for it is entirely unnecessary.

haydenzeller:
Yes I agree, another menu item is rather cumbersome. I was just unsure if you wanted this feature by default or by user pref so I added it just in case. I will update my branch.

nielsvanvelzen (changes requested):
Aside from the invalid formatting that needs to be fixed I found one issue, the server splash screen when authenticating should not be blurred.

app/src/main/java/org/jellyfin/androidtv/ui/background/AppBackground.kt
```diff
@@ -62,8 +64,9 @@ private fun AppThemeBackground() {
 fun AppBackground() {
 	val backgroundService = koinInject<BackgroundService>()
 	val currentBackground by backgroundService.currentBackground.collectAsState()
+	val blurBackground by backgroundService.blurBackground.collectAsState()
 	val enabled by backgroundService.enabled.collectAsState()
-
+
```
nielsvanvelzen, line 69: fix formatting

app/src/main/java/org/jellyfin/androidtv/ui/background/AppBackground.kt
```diff
@@ -80,7 +83,9 @@ fun AppBackground() {
 					alignment = Alignment.Center,
 					contentScale = ContentScale.Crop,
 					colorFilter = ColorFilter.tint(colorResource(R.color.background_filter), BlendMode.SrcAtop),
-					modifier = Modifier.fillMaxSize()
+					modifier = Modifier.fillMaxSize().then(
+						if (blurBackground) Modifier.blur(10.dp) else Modifier
+					)
```
nielsvanvelzen, line 88:
```suggestion
					modifier = Modifier
						.fillMaxSize()
						.then(if (blurBackground) Modifier.blur(10.dp) else Modifier)
```

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):
Awesome!

acordtag:
Hello, where does the 'enable' option appear within the Android TV interface? - cannot be seen on the 'Customization' section; it only appear the "Show backdrops" option only.

Currently running Server 10.9.9 and AndroidTV Jellyfin

nielsvanvelzen:
Background blur is always enabled on supported devices (Android 12 and newer).

Antonio1637:
Hello. Is it possible to disable this features. I have a headache while reading text and scrolling. Blur effect is across whole screen and it does not like nice.

nielsvanvelzen:
Please open an issue for that. If there's enough demand we may add an option for it.

shahribms:
@nielsvanvelzen this should count as demand enough. referencing both issues https://github.com/jellyfin/jellyfin-androidtv/issues/4716

PR #3680, opened 2024-06-15
Add media logo to screensaver

### LINKED ISSUE #3679: Display show/movie title logos instead of plain text on screen saver

Currently, the screen saver cycles through shows and movies, and shows the title as text in the bottom corner.

It would be a nice, "official" look if the plain title text for the displayed media could be replaced with the same title logo that displays on the media's page and on the pause screen.

haydenzeller:
**Changes**
Replace the media title text with the media logo on the screensaver. 
- Modify DreamHost to fetch item logo 
- add logo to DreamContent LibraryShowcase
- Replace Text() with Image() in DreamContentLibraryShowcase 

**Issues**

fix #3679


app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHost.kt
```diff
@@ -107,7 +107,20 @@ private suspend fun getRandomLibraryShowcase(
...
+			imageType = ImageType.LOGO,
+			tag = tag,
+			format = ImageFormat.WEBP,
+		)
+
+		val logo = withContext(Dispatchers.IO) {
+			imageLoader.execute(
+				request = ImageRequest.Builder(context).data(logoUrl).build()
+			).drawable?.toBitmap()
+		} ?: return null
```
nielsvanvelzen, line 121: Ideally we preload the backdrop & logo simultaneously

nielsvanvelzen, line 121: We must support items without a logo and fallback to just showing the title like before. 

A lot of the items in my library do not have a logo set, so this would effectively make the screensaver useless.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentLibraryShowcase.kt
```diff
@@ -47,12 +49,21 @@ fun DreamContentLibraryShowcase(
...
-			text = content.item.name.orEmpty(),
-			style = TextStyle(
-				color = Color.White,
-				fontSize = 32.sp
-			),
-		)
+		if (content.logo != null) {
+			Image(
+				bitmap = content.logo.asImageBitmap(),
+				contentDescription = null,
```
nielsvanvelzen, line 55: Content description could be set to the item name

app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHost.kt
```diff
@@ -101,13 +103,34 @@ private suspend fun getRandomLibraryShowcase(
...
 
-		val backdrop = withContext(Dispatchers.IO) {
-			imageLoader.execute(
-				request = ImageRequest.Builder(context).data(backdropUrl).build()
-			).drawable?.toBitmap()
-		} ?: return null
+		val logoUrl = api.imageApi.getItemImageUrl(
+			itemId = item.id,
+			imageType = ImageType.LOGO,
+			tag = tag,
```
nielsvanvelzen, line 109: This is the tag for the backdrop, not the logo

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHost.kt
haydenzeller: Is a tag required for the logo? if so does something like this seem correct?

```kotlin
val logoTag = item.parentLogoImageTag!!.randomOrNull()
     ?: item.imageTags?.get(ImageType.LOGO)
```

app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentLibraryShowcase.kt
```diff
@@ -47,12 +49,21 @@ fun DreamContentLibraryShowcase(
...
-			text = content.item.name.orEmpty(),
-			style = TextStyle(
-				color = Color.White,
-				fontSize = 32.sp
-			),
-		)
+		if (content.logo != null) {
+			Image(
+				bitmap = content.logo.asImageBitmap(),
+				contentDescription = content.item.name.orEmpty(),
```
nielsvanvelzen, line 55: content description should be null if we don't have a name

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):
Looks great, thanks for the PR!

PR #3718, opened 2024-07-03
Respect the user's DisplayMissingEpisodes setting in TV Show libraries

### LINKED ISSUE #3717: Respect the user's DisplayMissingEpisodes setting in TV Show libraries
### Describe the bug

1. Set the correct settings for the user and library metadata:
1.1: user -> Display -> disable "Display missing episodes within seasons"
image
1.2: library -> Disable "Missing Episode Fetcher"
image
1.3: plugins -> TheTVDB -> Disable "Include missing specials"
image
2. Reindex the library (replace all metadata)
3. In the web/android clients missing episodes do not show up any more; these settings work on those platforms. On the TV client however, no amount of repetitions get rid of these.

As a first step, I started with wireshark to see if the server or the client is supposed to filter these out. The web client do not receive missing episode information, while the TV client does. The API calls are also different for the same functionality in this client than in the web client. After reading the API documentation, comparing the two client's web traffic, and trying out modified calls to the API using `curl`, here are my findings:
- the web client always sends the `userId` parameter; the TV client never does (in my testing)
- adding the `userId` parameter to the API calls sometimes fixes the issue (the server correctly filters these out from the response list)
- adding the `isMissing=false` query parameter almost never works (I don't know if this is intended or not, might be a server bug)
- adding the `excludeItemTypes=Virtual` query parameter always works, but this would require client awareness of this setting, while the other clients rely exclusively on the server for this functionality.

There are also several cases, where the main list has this filter applied, but the aggregated response fields do not. For example, during the Items query for the episodes of a season, I got only existing episodes, but the `ChildCount` was wrong for this list; it included the missing episodes as well.

Side note 1: During the exploratory rounds, I found this dependency in one of the Gradle files, and it seems suspiciously outdated: `com.github.jellyfin.jellyfin-sdk-kotlin:android`. Is this okay?
Side note 2: I have never worked on a Gradle or an Android project, so please bear with me.

### Logs

_No response_

### Application version

0.16+

Google Play

### Device information

Nvidia Shield TV Pro

### Android version

Android 11?

### Jellyfin server version

10.9.7

fpeter8:
**Changes**
- Add missing `userId` query parameter to API calls where it fixes the problem
- Replace API calls with one or more API calls that do respect this user setting on the server side otherwise

**Issues**
Fixes #3717

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/constant/QueryType.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/constant/QueryType.kt
@@ -26,4 +26,5 @@ enum class QueryType {
 	SeriesTimer,
 	Premieres,
 	Resume,
+	Episodes,
 }
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseRowDef.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseRowDef.java
@@ -7,6 +7,7 @@ import org.jellyfin.androidtv.data.querying.GetSpecialsRequest;
 import org.jellyfin.androidtv.data.querying.GetUserViewsRequest;
 import org.jellyfin.sdk.model.api.request.GetAlbumArtistsRequest;
 import org.jellyfin.sdk.model.api.request.GetArtistsRequest;
+import org.jellyfin.sdk.model.api.request.GetEpisodesRequest;
 import org.jellyfin.sdk.model.api.request.GetItemsRequest;
 import org.jellyfin.sdk.model.api.request.GetLatestMediaRequest;
 import org.jellyfin.sdk.model.api.request.GetLiveTvChannelsRequest;
@@ -31,6 +32,7 @@ public class BrowseRowDef {
     private GetAlbumArtistsRequest albumArtistsQuery;
     private GetResumeItemsRequest resumeQuery;
     private GetSpecialsRequest specialsQuery;
+    private GetEpisodesRequest episodesQuery;
     private QueryType queryType;
 
     private int chunkSize = 0;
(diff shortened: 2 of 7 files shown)


nielsvanvelzen:
why did you add the userId everywhere?

fpeter8:
For some, it actually fixes the problems listed in the issue. For others, the parameter exists, I was there, and it doesn't seem to do anything harmful. When I compared the API requests made by the web and the androidtv version, it seemed to me that the web version always fills this parameter, so I followed that.

nielsvanvelzen:
We went through a big effort to remove the requirement for the userId in all API's. It's no longer needed and we won't be adding it back. Just adding the user id should not change any behavior.

fpeter8:
I've reverted the userId parameter changes. The bot comments relate to lines I did not change, or require annotations not present anywhere else in the file, so I'll ignore those for now.

nielsvanvelzen:
> The bot comments relate to lines I did not change, or require annotations not present anywhere else in the file, so I'll ignore those for now.

You changed a lot of formatting and whitespace in those files making the workflows think you made changes to these lines. Please undo them.

fpeter8:
I've used the default formatter in android studio. where can I find the project format files?

fpeter8:
I've tried to revert all whitespace changes, let me know if I missed anything. The issue unfortunately is not fully fixed, as the server itself sometimes returns virtual episodes when it's not supposed to.

nielsvanvelzen (changes requested):
I'd rather not introduce another query type because I'm trying to remove all of them. The current state is just temporary.

From my understanding the server sets `IsMissing` to false when `DisplayMissingEpisodes` is true. So we could probably just do that client-side? We should already have the user configuration locally.

nielsvanvelzen:
Hey @fpeter8, are you still interested in finishing this pull request? it's currently waiting for you to address my comment above.

PR #3753, opened 2024-07-14
Improve BlurHash decoding performance

cbeyls:
**Changes**
Replace the Kotlin Multiplatform BlurHash library with a new local implementation of the decoder which is at least 2 times faster while using less memory (no cache is needed).

For details about the optimizations and some benchmarks, see https://github.com/woltapp/blurhash/pull/254.


nielsvanvelzen (commented):
Thanks for the pull request. I'm not against using our own implementation as it won't need much maintenance (it's just a single file).

It should use our package name (probably somewhere in utils) and follow our code style (use tabs).

Additionally, the Detekt warnings can be suppressed for this file as the magic numbers are to be expected.

nielsvanvelzen (approved):

nielsvanvelzen:
Awesome, thanks for the contribution! I've tweaked the styling a bit more (adding new lines, moving the CHARS const to the top and using when statements instead of if statements).

PR #3754, opened 2024-07-14
Add ability to remember audio and subtitle selection settings

ConnorS1110:
**Changes**
Adds two new settings under advanced playback settings to remember the previously selected audio and subtitle track. If these settings are enabled, the player will find the audio and/or subtitle track that matches what was selected in the previous episode. If the audio/subtitle tracks don't align 1-to-1, it makes a best effort guess by looking through the display name and language value for the media streams and makes a selection.

**Issues**
Fixes my own frustrations with this discrepancy between the web and Android TV versions of the app

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -49,6 +49,7 @@ import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 import java.util.List;
+import java.util.Objects;
 
 import kotlin.Lazy;
 import timber.log.Timber;
@@ -81,6 +82,7 @@ public class PlaybackController implements PlaybackControllerNotifiable {
     private Boolean spinnerOff = false;
 
     private VideoOptions mCurrentOptions;
+    private VideoOptions mPrevOptions;
     private int mDefaultSubIndex = -1;
     private int mDefaultAudioIndex = -1;
     private boolean burningSubs = false;
@@ -107,6 +109,8 @@ public class PlaybackController implements PlaybackControllerNotifiable {
 
     private Display.Mode[] mDisplayModes;
     private RefreshRateSwitchingBehavior refreshRateSwitchingBehavior = RefreshRateSwitchingBehavior.DISABLED;
+    private Boolean rememberPreviousAudioTrack = false;
+    private Boolean rememberPreviousSubtitleTrack = false;
 
     public PlaybackController(List<BaseItemDto> items, CustomPlaybackOverlayFragment fragment) {
         this(items, fragment, 0);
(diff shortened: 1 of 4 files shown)


ConnorS1110:
This is like 99% done, but I have one issue I have spent days trying to solve I can't figure out and want to turn to others for help at this point. This successfully saves the user setting, and when you move to a new episode it will remember and correctly selects the tracks it should, but the initial playback just keeps looping endlessly. It starts out black, then the first frame appears, I get a toast notification saying it is burning in subtitles, and then the player just restarts. It's black for a second, the first frame appears, the toast appears, and the cycle keeps repeating. I have stuck break points all over the place stepping one step at a time and I cannot see anything that would be causing this loop. I am hoping someone with a better understanding of how playback works on android tv can see what I have changed and offer ideas what I need to change for this.

nielsvanvelzen:
I don't think we need preferences for this. Could just be default behavior.

ConnorS1110:
I will update that. Do you have any suggestions about the playback thing I mentioned in my comment

nielsvanvelzen:
No sorry, the old playback code is not my favorite place so I don't touch it when possible. It's full of these odd behaviors.

phidix:
Hi,
I can only say that I'm really interested in having this feature working in Android TV client for Jellyfin. Fingers crossed someone from the community will help address @ConnorS1110 issues described regarding "endless loop".

Thanks!

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -719,12 +768,15 @@ public void switchAudioStream(int index) {
...
+        Integer currOptionsAudioIndex = mCurrentOptions.getAudioStreamIndex();
+        Integer prevOptionsAudioIndex = mPrevOptions.getAudioStreamIndex();
+        int audioIndexToUse = rememberPreviousAudioTrack && prevOptionsAudioIndex != null ? prevOptionsAudioIndex : index;
+        Timber.d("trying to switch audio stream from %s to %s", currAudioIndex, audioIndexToUse);
+        if (currAudioIndex == audioIndexToUse || (prevOptionsAudioIndex != null && prevOptionsAudioIndex == audioIndexToUse)) {
+            Timber.d("skipping setting audio stream, already set to requested index %s", audioIndexToUse);
+            if ((currOptionsAudioIndex == null || currOptionsAudioIndex != audioIndexToUse) || (prevOptionsAudioIndex == null || prevOptionsAudioIndex != audioIndexToUse)) {
+                Timber.d("setting mCurrentOptions audio stream index from %s to %s", currOptionsAudioIndex, audioIndexToUse);
+                mCurrentOptions.setAudioStreamIndex(audioIndexToUse);
             }
```
entropicdrifter, line 780: It looks like you wanted these two conditions to be mutually exclusive. Current behavior would only attempt to change the audio stream index when the skip condition is true. That's probably what's leading to your infinite loop.
```suggestion
        if (currAudioIndex == audioIndexToUse || (prevOptionsAudioIndex != null && prevOptionsAudioIndex == audioIndexToUse)) {
            Timber.d("skipping setting audio stream, already set to requested index %s", audioIndexToUse);
        }
        else if ((currOptionsAudioIndex == null || currOptionsAudioIndex != audioIndexToUse) || (prevOptionsAudioIndex == null || prevOptionsAudioIndex != audioIndexToUse)) {
            Timber.d("setting mCurrentOptions audio stream index from %s to %s", currOptionsAudioIndex, audioIndexToUse);
            mCurrentOptions.setAudioStreamIndex(audioIndexToUse);
        }
```

entropicdrifter: @ConnorS1110 forgot to tag

mo-st:
Hi @ConnorS1110, are you still having the same issue/are you still working on this and could I help you with something? To be transparent I'm a novice in this codebase but I would love to help however possible. I could start by testing out if the suggested change by entropicdrifter works for example. in case you didn't get around to doing that yet.

ConnorS1110:
Closing because there's another PR out there with a better approach

ethaldeman:
@ConnorS1110 Since https://github.com/jellyfin/jellyfin-androidtv/pull/4514 was closed without merging, any chance you want to continue this pull request?

Issue #3033, opened 2023-09-07
Seek values are different for D-Pad and Fast Forward/Rewind buttons


silentTeee:
### Describe the bug

Not sure if this was intentional, but the directional arrow buttons on my television's remote seem to have very different seek behaviors than the traditional fast-forward and rewind buttons. 

When I use the dedicated Fast Forward and Rewind buttons on my remote, they seem to honor the step values set on my Jellyfin server, and will increment exactly to the current position +/- the step. But when I use the remote's D-Pad Left and Right buttons, they seem to seek to the nearest 30 second interval (if I am at position 00:12:45 and I push Left, it will jump back to 00:12:30, and if I push left again it will jump to 00:12:00).

Since not all TV remotes offer dedicated playback buttons and sometimes just map them to the D-Pad (with the 'Select' button usually inheriting Play/Pause behavior), I am wondering if it makes sense to have behavior be consistent across both button sets.

### Application version

0.15.12

### Device information

HiSmartTV A4

### Android version

Android 9

### Jellyfin server version

10.8.10

nielsvanvelzen:
This is intended behavior for now. It might be revisited when the video player is redesigned as a part of #1057.

silentTeee:
Understood, should I leave this issue open then, or would you prefer it closed?

nielsvanvelzen:
We can keep the issue open

rawkhopper:
I have a TCL TV and the remote only has d pads. I would like them to be 10 seconds as well. Maybe everyone would be happy if it were configurable? Then each person could choose 10sec or 30sec? I totally agree that 30secs is way too long when you miss a sentence or two.  I tend to forget what I was listening for by the time 30secs plays back :). Then I end up doing it again!

FunctionDJ:
I frequently rewind shows or movies, e.g. because i didn't understand some dialogue.

It would be incredibly helpful if there were settings to change the seek/jump length and the seek/jump type (prev/next interval vs. relative).

Also a setting to toggle autoresume after seek/jump would be useful. I think it pauses for the purpose of not lagging while seeking, but resume could simply be debounced with something like 500ms or 1s of delay.

ericleb010:
[...]

PR #3794, opened 2024-07-23
Add option to customize seek time

kabel2:
**Changes**
Adds a slider in the playback settings to set the fast-forward/rewind times. Minimum 1 second, maximum 60 seconds. Default 15 seconds.
TODO:
I think the naming of the setting needs to be changed. It may also be necessary to change the slider to two dropdowns (as in the Android smartphone app). 

**Issues**
Fixes my own frustration of not being able to rewind a short dialog.


fice-t:
This should ideally take into account the user-specific settings made on the Jellyfin server, i.e. the `Skip {forward,back} length` values set via the web client's `menu->User->Settings`.

These settings are already partially used in the Android TV client, as [seen here](https://github.com/jellyfin/jellyfin-androidtv/blob/dcbf87488cd035a02e198e36eac7c39bcb04d446/app/src/main/java/org/jellyfin/androidtv/preference/UserSettingPreferences.kt#L17-L18) and in [this issue](https://github.com/jellyfin/jellyfin-androidtv/issues/3033).

Having an override specifically for the TV client can still make sense (a user may want different seeking behaviour between clients), but I think that it should still default to the server-side settings.

Also, perhaps the seek time should be cached somewhere so the preferences aren't being hit for each `getSeekPositions`? Could the seek time itself just be fed into `CustomSeekProvider`?

kabel2:
I have already seen the settings for forward and back. However, the SeekProvider does not allow a distinction between forward and back, so there is only one value. 

The setting is therefore different from the Android app/web app. I think using only the value of “back” would not be much better than creating a new (local) value (seek time). Maybe this would be a temporary solution until there is a rewrite of the playback.

The seektime can of course be given directly to the CustomSeekProvider, would be no problem to change that.

fice-t:
> However, the SeekProvider does not allow a distinction between forward and back, so there is only one value.

Right, that does complicate things. Luckily, the ATV playback rewrite [allows such a distinction](https://github.com/jellyfin/jellyfin-androidtv/blob/b12e4a69781bdb481a838d327cc977dd7be0fffc/playback/media3/session/src/main/kotlin/MediaSessionPlayer.kt#L132-L133).

Since there is the rewrite coming up, I would propose the following:

### Legacy code
- Use one local setting (`UserPreferences`)
  - Do not fall back to either server setting (`UserSettingPreferences`) to avoid possible confusion
  - Always enabled in the UI

### Rewrite code
- Use two local settings, one each for back/forward
  - Fall back to the respective server setting if not changed locally
  - Only enable each local setting in the UI if requested (perhaps gated by a checkbox?)
    - Either way, the UI should present each server-side value if the local setting is not enabled.

This PR already accomplishes the legacy part, and could be considered self-contained. I wrote some thoughts about how this PR could consider the transition between legacy/rewrite, but came to the conclusion that it doesn't matter much on the legacy side. Perhaps the names could be changed from `*seek_time*` -> `*seek_time_legacy*`, since in the above scheme the keys/descriptions would need to be different between legacy/rewrite. Also, maybe `duration` is better than `time` for the names?

So perhaps the rewrite side could be punted off for later. FWIW, I was previously planning on working on custom seek times once the rewrite was completed if no one else got to it first.

@nielsvanvelzen Thoughts on all this?

nielsvanvelzen:
We should re-use the skip forward/backward preferences already available in the user settings. Not add a new one.

nielsvanvelzen:
Hey @kabel2, are you still interested in finishing this pull request? it's currently waiting for you to address my comment above.

kabel2:
Sorry for the late reply.
Without the rewrite of the player it is not possible to distinguish between forward and backwards. 
So should we use the mean value of forward and backwards or how should two settings be used for one value?

nielsvanvelzen:
We can use the "forward" preference for now and update it later when we redo the player design.

suchtie664:
Okay, I'll take a look at the necessary changes tomorrow :)

kabel2:
Okay, I'll take a look at the necessary changes tomorrow :)

kabel2:
Is now adjusted. 
But there is still a problem, the minimum skip length is just 0, although I have set it to `min = 5_000`.

app/src/main/java/org/jellyfin/androidtv/preference/UserSettingPreferences.kt
```diff
@@ -15,7 +15,7 @@ class UserSettingPreferences(
 ) {
 	companion object {
 		val skipBackLength = intPreference("skipBackLength", 10_000)
-		val skipForwardLength = intPreference("skipForwardLength", 30_000)
+		val skipForwardLength = intPreference("skipForwardLength", 10_000)
```
nielsvanvelzen, line 18: The default value needs to be consistent with jellyfin-web

https://github.com/jellyfin/jellyfin-web/blob/9fbc37fbf30e986451964d3cd8fd178d1a3895be/src/scripts/settings/userSettings.js#L366

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/LeanbackOverlayFragment.java
```diff
@@ -18,6 +20,7 @@ public class LeanbackOverlayFragment extends PlaybackSupportFragment {
     private VideoPlayerAdapter playerAdapter;
     private boolean shouldShowOverlay = true;
     private Lazy<PlaybackControllerContainer> playbackControllerContainer = inject(PlaybackControllerContainer.class);
+    private final Integer skipForwardLength = KoinJavaComponent.<UserSettingPreferences>get(UserSettingPreferences.class).get(UserSettingPreferences.Companion.getSkipForwardLength());
```
nielsvanvelzen, line 23: Move the reading of the setting to the actual usage and just inject the class here.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/preference/UserSettingPreferences.kt
kabel2: i changed the value to 30 seconds

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/LeanbackOverlayFragment.java
kabel2: I have created an extra variable for the skipForwardLength. Or should the whole UserSettingPreferences be passed to the CustomSeekProvider?

nielsvanvelzen:
The tests are currently failing

nielsvanvelzen (approved):
I've made 2 final changes:

1. Fix the issue where you can change the value below the minimum - this wasn't your mistake but a bug in our UI code
2. Use the long type in CustomSeekProvider and do the casting from LeanbackOverlayFragment (where we initialize the provider)

Thanks for the pull request!

Issue #3492, opened 2024-04-08
Subtitles get out of sync when resuming a movie, or the audio/video is out of sync with the subtitles.


joost00719:
### Describe the bug

Reproduce steps:

1. Watch a movie until like 50% of it.
2. Notice that the subtitles are (probably) in sync.
3. Shut down jellyfin app.
4. Start Jellyfin again, and try to resume the movie.
5. Notice that the subtitles are out of sync. However this doesn't happen at every movie.

I also noticed something else odd, which could explain why this bug happens.

If I go to timestamp 1:00:00 on the androidtv client (chromecast), and I go to 1:00:00 on the Windows 10 client, it seems like the androidtv client is like 5-10 seconds behind the Windows 10 client. The subtitles are 100% exactly on the same timestamp, so those appear at the identical time if you sync the two clients up, but the video and audio on the androidtv client seems to be 5-10 seconds behind, which could explain the subtitle sync issue.
Long story short, it looks like the audio/video get's out of sync with the time, but the subtitles are actually correct.

For me this is a big issue since I have trouble hearing voices with background noise, especially if it's not my native tongue.

If any more info is needed, I am able to (privately) provide a test-account to my JF server to a developer, with sftp access to my library.

### Logs

```
[2024-04-08 13:03:05.430 +02:00] [INF] [74] Jellyfin.Api.Helpers.MediaInfoHelper: User policy for "joost". EnablePlaybackRemuxing: True EnableVideoPlaybackTranscoding: True EnableAudioPlaybackTranscoding: True
[...]

PR #3812, opened 2024-07-26
Add Option to Disable HLS

pierfied:
**Changes**
Adds an option in settings to disable HLS with exoplayer.

**Issues**
This is a workaround for subtitles getting out of sync when skipping around/resuming certain video files when using HLS
(issue https://github.com/jellyfin/jellyfin-androidtv/issues/3492).

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -92,6 +92,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 */
 		var refreshRateSwitchingBehavior = enumPreference("refresh_rate_switching_behavior", RefreshRateSwitchingBehavior.DISABLED)
 
+		/**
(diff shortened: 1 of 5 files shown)


nielsvanvelzen:
The app relies on HLS for all streams when remuxing/transcoding. I'm not strictly opposed to change that but it should be substantiated why another approach is better and properly tested.

I'm not a fan of a preference for it though, that only makes testing harder. We should do everything over HLS or nothing.

viown:
Subtitles going out of sync should only occur during transcoding (not while direct playing/streaming), so maybe the client can request burn-in when the video has to be transcoded?

pierfied:
In retrospect, yes, I think trying to burn in subtitles when transcoding is probably a simpler and safer solution. If that would be a preferred route, we can close this PR and open a new one.

nielsvanvelzen:
Can you try #3825 (or master branch since it's already merged) and let me know if that fixes those sync issues for you?

pierfied:
From a quick test it looks like it does fix the sync issue! I will go ahead and close this PR.

PR #3818, opened 2024-07-30
Prefer extension renderers to core renderers

### LINKED ISSUE #3816: Player error when a FLAC audio track is selected
### Describe the bug

The player crashes when I play a movie with a FLAC audio track. The same movie works fine instead when I play it through a USB stick using the TV's integrated media player.

Similar to #3782, but the exception is different.

This error happens only if I set the "Audio output" option to "Direct" in the client's settings. If I set it to "Downmix to stereo", instead, no error occurs (also because when downmixing the audio codec changes to aac instead of preserving flac).

### Logs

```
Audio codec error
androidx.media3.decoder.DecoderInputBuffer$InsufficientCapacityException: Buffer too small (32768 < 36143)
    at androidx.media3.decoder.DecoderInputBuffer.createReplacementByteBuffer(DecoderInputBuffer.java:244)
    at androidx.media3.decoder.DecoderInputBuffer.ensureSpaceForWrite(DecoderInputBuffer.java:195)
    at androidx.media3.exoplayer.source.SampleDataQueue.readSampleData(SampleDataQueue.java:292)
    at androidx.media3.exoplayer.source.SampleDataQueue.readToBuffer(SampleDataQueue.java:130)
    at androidx.media3.exoplayer.source.SampleQueue.read(SampleQueue.java:440)
    at androidx.media3.exoplayer.source.ProgressiveMediaPeriod.readData(ProgressiveMediaPeriod.java:506)
    at androidx.media3.exoplayer.source.ProgressiveMediaPeriod$SampleStreamImpl.readData(ProgressiveMediaPeriod.java:985)
    at androidx.media3.exoplayer.BaseRenderer.readSource(BaseRenderer.java:524)
    at androidx.media3.exoplayer.mediacodec.MediaCodecRenderer.feedInputBuffer(MediaCodecRenderer.java:1362)
    at androidx.media3.exoplayer.mediacodec.MediaCodecRenderer.render(MediaCodecRenderer.java:875)
    at androidx.media3.exoplayer.ExoPlayerImplInternal.doSomeWork(ExoPlayerImplInternal.java:1136)
    at androidx.media3.exoplayer.ExoPlayerImplInternal.handleMessage(ExoPlayerImplInternal.java:561)
    at android.os.Handler.dispatchMessage(Handler.java:102)
    at android.os.Looper.loop(Looper.java:223)
    at android.os.HandlerThread.run(HandlerThread.java:67)
```

### FFmpeg logs

_No response_

### Media info of the file

```shell
Stream #0:0(eng): Video: h264 (High), yuv420p(progressive), 1920x1080 [SAR 1:1 DAR 16:9], 23.98 fps, 23.98 tbr, 1k tbn
Stream #0:1(eng): Audio: flac, 48000 Hz, 7.1, s32 (24 bit) (default)
```

### Application version

0.17.0-beta5

### Device information

Panasonic MZ800

### Android version

Android 11

### Jellyfin server version

10.9.8

rickysixx:
**Changes**

Prefer extension renderers for ExoPlayer to core renderers.

Extension renderes like `LibflacAudioRenderer` and `LibopusAudioRenderer` seem to work better than the core ones.

**Issues**

Fixes #3816


nielsvanvelzen:
We should generally prefer the media3 implementation of codecs as those can use platform decoders, including ones made by vendors. This will, at times, be much more efficient then using the ffmpeg renderer. Media3 should only use the ffmpeg renderer when there is no platform decoder or the platform decoder is not compatible.

rickysixx:
@nielsvanvelzen Unfortunately it seems that media3 fallback to ffmpeg renderers does not (always) work properly, at least in my case (see #3816).

I have no experience with media playback on Android so I cannot tell if this PR is the proper fix to #3816, but I have tested it on different movies/shows and it seem to work fine. I still feel that this is somewhat a quick & dirty hack, but I was not able to come up with a different non-hacky solution.

Maybe we could add an advanced preference to allow the user to switch between `EXTENSION_RENDERER_MODE_ON` and `EXTENSION_RENDERER_MODE_PREFER` (the default value would be `EXTENSION_RENDERER_MODE_ON`, as it is right now). In this way the user can at least try to play the media anyway instead of being stuck due to playback errors with media3 renderers.

nielsvanvelzen:
The linked issue does not contain full logs so I can't really tell much about it. Normally the player should fallback to the next decoder when something is wrong.

zkhcohen:
> The linked issue does not contain full logs so I can't really tell much about it. Normally the player should fallback to the next decoder when something is wrong.

Could you provide a little more insight about how this decoder-fallback mechanism is implemented? I think it might help with troubleshooting the behavior I'm seeing in the following comment, specifically how media may begin to play after a few minutes:

https://github.com/jellyfin/jellyfin-androidtv/issues/3830#issuecomment-2276891040

nielsvanvelzen (changes requested):
I'd be fine introducing this behavior under a developer preference. It would be called something like "Prefer FFmpeg for audio playback".

Additionally it should also propagate to the new playback engine.

rickysixx:
I have added the new preference.

I have also added a check to activate [ExoPlayer debug logs](https://developer.android.com/media/media3/exoplayer/debug-logging) which help to understand which decoder is being used.

I am however facing some troubles to propagate this change to the new playback engine (e.g. `org.jellyfin.playback.media3.exoplayer.ExoPlayerBackend`, if I am not wrong). Using the `UserPreferences` class there would introduce a circular dependency between modules `jellyfin-androidtv.app.main` and `jellyfin-androidtv.playback.media3.exoplayer.main`.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -195,6 +195,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Whether items shown in the screensaver are required to have an age rating set.
 		 */
 		var screensaverAgeRatingRequired = booleanPreference("screensaver_agerating_required", true)
+
+		/**
+		 * Whether to prefer extension renderers (i.e. FFmpeg) to core ones.
+		 */
+		var preferExtensionRenderers = booleanPreference("prefer_extension_renderers", defaultValue = false)
```
nielsvanvelzen, line 202:
```suggestion
		var preferExoPlayerFfmpeg = booleanPreference("exoplayere_prefer_ffmpeg", false)
```

nielsvanvelzen: Also move it up to after refreshRateSwitchingBehavior so it's grouped with the other playback preferences.

app/src/main/res/values/strings.xml
```diff
@@ -514,6 +514,8 @@
     <string name="past_week">Past week</string>
     <string name="scheduled_in_next_24_hours">Scheduled in next 24 hours</string>
     <string name="past_24_hours">Past 24 hours</string>
+    <string name="pref_prefer_extension_renderers_label">Prefer FFmpeg for audio playback</string>
+    <string name="pref_prefer_extension_renderers_description">May fix some playback errors and avoid audio transcoding.</string>
```
nielsvanvelzen, line 518:
```suggestion
    <string name="pref_prefer_extension_renderers_description">Use FFMpeg to decode audio codecs when platform codecs are available.</string>
```

nielsvanvelzen (changes requested):
To propagate the preference to the ExoPlayerBackend you'll have to add a new boolean to ExoPlayerOptions class and populate it from the PlaybackModule where we initialize the playback code.

This does mean it requires an app restart when the preference changes, but that's fine for now.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
rickysixx: done

rickysixx:
Requested changes applied.

Thank you for your help with the code!

nielsvanvelzen (approved):
Thanks! I've pushed a final commit with some small changes.

Issue #3451, opened 2024-03-16
Chromecast HD cannot play 4k HEVC, but chooses direct-play anyways


tlambertz:
### Describe the bug

My Chromecast HD is not capable of smoothly playing 4k HEVC, but requests direct-play anyways.
Audio works smoothly, but I am getting around 5 FPS, with the audio front-running the video.

My video file is somewhat atypical: 4k HEVC, but SDR and only 2 mbps bitrate.
When playing back a more 'typical' HEVC 4k file, 19mbps and HDR10, it always transcodes correctly.

AFAIK, there is no workaround for this issue available right now, except for choosing a veery low bitrate limit to force transcoding, which impacts quality too much. For this particular file, I have to lower the bitrate limit to 1 mbps, which looks quite blocky.

One possible fix would be to add the Chromecast HD to the list of devices that don't support 4k video at all:
https://github.com/jellyfin/jellyfin-androidtv/blob/be826c3a47e78415a7c9db14cff39809accdbf43/app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.kt#L49-L51
That might unnecessarily force transcoding for non-hevc 4k content, though. I have NOT verified yet if the chromecast can playback non-hevc 4k content, but I assume it can since this issue hasn't surfaced earlier? I am also not sure how prevalent such content is, or how many folks rely on direct-play there. For my use-case, blacklisting 4k content from direct-play would be perfect.

I see PR https://github.com/jellyfin/jellyfin-androidtv/pull/3110 is intended to make 4k-direct-playback configurable, but having a good working default for known devices would improve user experience.

### Media info of the file

~5 fps Playback for 'atypical' file:

```
Video
Title: 4K HEVC SDR
Codec: HEVC
Codec tag: hvc1
Profile: Main
Level: 150
Resolution: 3840x2024
Aspect ratio: 480:253
Interlaced: No
Framerate: 25
Bitrate: 2092 kbps
Bit depth: 8 bit
Video range: SDR
Video range type: SDR
Color space: bt709
Color transfer: bt709
Color primaries: bt709
Pixel format: yuv420p
Ref frames: 1
```

Chromecast correctly chooses to transcodes following file:

```
Video
Title: 4K HEVC HDR
Codec: HEVC
Profile: Main 10
Level: 150
Resolution: 3840x2160
Aspect ratio: 16:9
Interlaced: No
Framerate: 23.976025
Bitrate: 18599 kbps
[...]

Issue #3388, opened 2024-02-20
4k hevc content fails to play on 1080p devices


MichaelRUSF:
### Describe the bug

~~Side effect of #3281,~~ 4k content gets remuxed on devices that only support 1080p. This will lead to the player failing. 

Tested on Fire TV Stick (2nd Gen), see logs for details. 

[Fire TV Stick Gen 2 Logcat.txt](https://github.com/jellyfin/jellyfin-androidtv/files/14351773/Fire.TV.Stick.Gen.2.Logcat.txt)
[FFmpeg.Remux.log](https://github.com/jellyfin/jellyfin-androidtv/files/14351775/FFmpeg.Remux.log)

Tested on prior release,16.4, and had the same results, 4k videos get remuxed. 

### Application version

0.16.5

### Where did you install the app from?

Sideloaded APK

### Device information

Fire TV Stick (2nd Gen) 2016

### Android version

Android Level 22 (Android 5.1)

### Jellyfin server version

10.8.13

PR #3820, opened 2024-08-02
Add Google Chromecast HD limitation

majmongoose:
**Changes**
Added the Google Chromecast HD to the list of devices unable to support 4K, as it reports it is capable of downscaling but is incapable of doing so in realtime. Change has been discussed with Niels in the discord.

**Note**
There still may be a bug on the jellyfin server, as the tv app is telling it to send a maximum of 1920x1080 (verified via tcpdump [TCPDump Data](https://pastebin.com/SdWGmyzP)) but it is still streaming 4k.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.kt
@@ -27,7 +27,7 @@ object DeviceUtils {
 	private const val SHIELD_TV_MODEL = "SHIELD Android TV"
 
 	// Google Chromecast HD Model
-	private const val GOOGLE_CHROMECAST_HD_MODEL = "Chromecast HD" // Reports to support 4K - Cannot downscale in device in realtime
+	private const val GOOGLE_CHROMECAST_HD_MODEL = "Chromecast HD"
 
 	@JvmStatic
 	val isFireTv: Boolean = Build.MODEL.startsWith(FIRE_TV_PREFIX)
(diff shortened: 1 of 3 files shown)


MichaelRUSF:
Does seem like a partial fix [#3451](https://github.com/jellyfin/jellyfin-androidtv/issues/3451). Would likely be affected by [#3388](https://github.com/jellyfin/jellyfin-androidtv/issues/3388).

app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.kt
```diff
@@ -26,6 +26,9 @@ object DeviceUtils {
 	// Nvidia Shield TV Model
 	private const val SHIELD_TV_MODEL = "SHIELD Android TV"
 
+	// Google Chromecast HD Model
+	private const val GOOGLE_CHROMECAST_HD_MODEL = "Chromecast HD" // Reports to support 4K - Cannot downscale in device in realtime
```
nielsvanvelzen, line 30:
```suggestion
	private const val GOOGLE_CHROMECAST_HD_MODEL = "Chromecast HD"
```

These are just constants for names, we don't specify why they exist. That's on the usage part.

app/src/main/java/org/jellyfin/androidtv/util/DeviceUtils.kt
```diff
@@ -45,6 +48,9 @@ object DeviceUtils {
 	@JvmStatic
 	val isShieldTv: Boolean = Build.MODEL == SHIELD_TV_MODEL
 
+	@JvmStatic
+	val isChromecastHD: Boolean = Build.MODEL == GOOGLE_CHROMECAST_HD_MODEL
```
nielsvanvelzen, line 52: Doesn't look like this is used

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):
Thanks for the PR!

Issue #2855, opened 2023-06-25
Set buffer limit and read rate


charredchar:
### Describe the feature you'd like

Please add options for tuning the buffer. Kodi allows this (only through advanced settings) and it is extremely helpful when dealing with unreliable internet.

The options would be for total buffer size and read-rate. Most Android devices should have spare RAM for buffering but it would be advisable to allow limits for how much is used.
The other is how fast ahead it should read. For Kodi this is a multiple of the video bitrate. ie 1Mbps video reads the video at 2Mbps.

Together these can be used to fill a larger buffer for the instances where internet can drop out or slow down. This is very important for mobile connections.

theinfguy:
This would be an excellent feature and would greatly enhance the AndroidTV Client playback experience. 

I experience pausing/buffering with higher bitrate 4K content despite having an internet connection of over 150Mbps. 

Kodi allows me to set an aggressive ‘read ahead’ and larger buffer size via advanced settings. Something similar would be ideal.

rTiGd2:
There is little point having 'remote features' without either good frame exhaustion detection and/or the ability to tune the buffer size yourself.

PR #3821, opened 2024-08-02
Add preferences to configure buffering behaviour

rickysixx:
**Changes**

Add 4 new preferences to configure parameters of [`DefaultLoadControl.Builder#setBufferDurationsMs()`](https://developer.android.com/reference/androidx/media3/exoplayer/DefaultLoadControl.Builder?hl=en#setBufferDurationsMs(int,int,int,int))

**Issues**

May be helpful for #2855 (although this PR only configures buffering durations, not the size).

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -195,6 +195,40 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Whether items shown in the screensaver are required to have an age rating set.
 		 */
 		var screensaverAgeRatingRequired = booleanPreference("screensaver_agerating_required", true)
+
+		/**
+		 * Enable reactive homepage
+		 */
+		var homeReactive = booleanPreference("home_reactive", false)
+
+		/**
+		 * Minimum duration of media that the player will attempt to ensure is buffered at all times.
+		 *
+		 * Default value in androidx-media3 v1.4.0 is 50 seconds.
+		 */
+		var minBufferMs = intPreference("min_buffer_ms", 50_000)
+
+		/**
+		 * Maximum duration of media that the player will attempt to buffer
+		 *
+		 * Default value in androidx-media3 v1.4.0 is 50 seconds.
+		 */
+		var maxBufferMs = intPreference("max_buffer_ms", 50_000)
+
+		/**
+		 * Duration of media that must be buffered for playback to start or resume following a user action such as a seek.
+		 *
+		 * Default value in androidx-media3 v1.4.0 is 2.5 seconds.
+		 */
+		var bufferForPlaybackMs = intPreference("buffer_for_playback_ms", 2_500)
+
+		/**
+		 * duration of media that must be buffered for playback to resume after a rebuffer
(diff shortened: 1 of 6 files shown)


nielsvanvelzen:
I don't see a use-case for those preferences. Please explain why we need 4 (!!) new preferences to customize the buffer behavior.

rickysixx:
@nielsvanvelzen Being able to change those parameters allows the user to tweak buffering behaviour to their specific needs, e.g. in case of unstable/not-so-fast internet connection it may be beneficial to exceed the default buffer duration of 50 seconds.

Reducing these durations may be beneficial too, e.g. when listening music lower buffering durations avoid waste of bandwidth in case of frequent skips.

Yes, these are 4 new preferences, but I do not see any problem with that. User is not forced to change these settings if they do not understand what they are doing (or they simply don't care about them) and they still have the library default values for those parameters.

nielsvanvelzen:
Sorry but I don't see the need to fix those problems by introducing additional preferences. Besides, this change doesn't affect the playback code used for music at all.

viown:
I believe we should keep just one preference for controlling the buffer size, in this case minimum buffer (or max buffer depending on how it behaves), but instead of asking the user to specify a duration in seconds, I think we should show selections (e.g Small / Default / Large / Very Large) each with a pre-defined number of seconds or bytes.

As for the other two preferences, I think it just adds unnecessary complexity to the app as I can't think of a scenario where a user would want to change them from the defaults.

rickysixx:
> As for the other two preferences, I think it just adds unnecessary complexity to the app as I can't think of a scenario where a user would want to change them from the defaults.

A user may prefer to wait a little more at the beginning of a movie (i.e. changing `bufferForPlaybackMs` to 20-30 seconds) if this avoids further rebuffering. The same applies to `bufferForPlaybackAfterRebufferMs`.

However, if 4 new preferences are too many, we can keep only one, as suggested by @viown, to control both `minBufferMs` and `maxBufferMs` (I think a duration in seconds is more intuitive for the user than a byte size). A single preference will force the user to have the same value for both of these durations, but I think it's fine (it's what ExoPlayer does already and there are no real advantages in having them separated).

> instead of asking the user to specify a duration in seconds, I think we should show selections (e.g Small / Default / Large / Very Large) each with a pre-defined number of seconds or bytes.

I think a seekbar a user can change freely is more appropriate instead (like it is e.g. for subtitle font size), since every device has different memory capabilities and every user has a different internet speed. Maybe we can come to a compromise here by changing the seekbar increment from the current 1 second to 5/10/30 seconds.

nielsvanvelzen:
If you could update the pull request to use a single preference I will do some additional testing with that and see if it's worth adding to the app. Right now I still don't see a reason to add such a preference though.

nielsvanvelzen:
Hey @rickysixx, are you still interested in finishing this pull request? it's currently waiting for you to address my comment above.

rickysixx:
Hey @nielsvanvelzen, sorry for my late reply.

After using the app for some time I now agree with you that this change is unnecessary, therefore I prefer to focus on more urgent issues.

nielsvanvelzen:
No problem, let's close this for now. We can always revisit it in the future if we do believe it's useful.

PR #3825, opened 2024-08-04
Remove custom subtitle renderer

### LINKED ISSUE #145: [Bug] Subtitles Don't Always Load

nielsvanvelzen:
Our custom subtitle renderer is causing a lot of issues, like timeouts when waiting for subtitle extraction or issues with the rendering because someone decided to put SSA tags in SRT subs (please don't do this). There really isn't much benefit for the current custom renderer anyway.

While we could put some effort in fixing the issues in the current renderer, it's much easier to rely on media3 to render it (for now), this would also make it easier for the initial subtitle support in the playback rewrite. We can always re-introduce our own renderer in the future.

While I did test these changes, there may be issues with track selection or subtitles that are not embedded. As it's nearly impossible to test all edge-cases.

**Changes**
- Remove custom subtitle renderer
- Remove requirement for external subs (support embedded subtitles)
- Redo subtitle preferences
  - Includes _some_ migrations
  - Default style is now: no background with white text and black outline

**Issues**

Should fix #145 and the many duplicates of that issue


crobibero (approved):
only a few magic numbers

pbasov:
Does media3 support subtitle position offset? Tested this commit, subtitles finally load fine, but I really miss the ability to move the subtitles a bit lower on screen.

PR #3836, opened 2024-08-05
Remove context requirement from ExoPlayerProfile

nielsvanvelzen:
**Changes**
- Remove usages of utils/deviceutils from ExoPlayerProfile and move all behavior changing variables to constructor parameters. This will make it easier to read/maintain the legacy profile.
- Remove `utils.downMixAudio` and use preference directly

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -516,16 +518,16 @@ private VideoOptions buildExoPlayerOptions(@Nullable Integer forcedSubtitleIndex
         internalOptions.setMaxBitrate(maxBitrate);
         if (exoErrorEncountered || (isLiveTv && !directStreamLiveTv))
             internalOptions.setEnableDirectStream(false);
-        internalOptions.setMaxAudioChannels(Utils.downMixAudio(mFragment.getContext()) ? 2 : null); //have to downmix at server
```
nielsvanvelzen, line 519: ^ removed this one because we already set it in the profile

crobibero (approved):

PR #3863, opened 2024-08-08
Graduate reactive homepage to stable

nielsvanvelzen:
The reactive homepage option subscribes to websocket messages for library updates and refreshes the homepage accordingly. This way, whenever a new episode is added to the server the next up section will update without re-opening the app. It also updates after something is watched etc.

This option was unstable in 0.16 but with the many SDK refactorings and item row improvements it has become stable in 0.17 and I somewhat forgot about it because it never caused issues.

A WebSocket connection is required for it to work, otherwise it will just do nothing.

**Changes**
- Remove experimental "reactive homepage" developer option
- Always enable reactive homepage

**Issues**


crobibero (approved):
👨‍🎓 happy graduation

PR #3878, opened 2024-08-10
Display length of show/movie in hour minute format (like web client)

### LINKED ISSUE #2806: Display length of show/movie in hour minute format

Currently the length of shows/movies show the length of the media in terms of minutes only. For example: a movie that is an hour and forty five minutes long is displayed as: "105m". What I would like instead is for the time remaining to display as "1h 45m". This would be consistent with the web player/desktop player. Seems like a relatively simple fix. Okay, thanks.

noahvogt:
# Display length of show/movie in hour minute format

**Changes**
Currently the length of shows/movies shown in FullDetailsFragment.java is in minutes only, or to say as a regex `[0-9]+ min`. With my changes, it now shows in the format `[0-9]+h [0-9][0-9]?m`. Note that the minutes are not zeropadded (displaying eg `2h 02m` instead of `2h 2m`) as this is the way the web client does it as well.

For more consistency with the main web client,  I decided to just change it instead of adding an option so that the user can switch between both display methods. But if that's prefered, I can change it before resubmitting.

**Issues**
Fixes #2806

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
@@ -677,8 +677,18 @@ public class FullDetailsFragment extends Fragment implements RecordingIndicatorV
     }
 
     private String getRunTime() {
-        Long runtime = Utils.getSafeValue(mBaseItem.getRunTimeTicks(), mBaseItem.getRunTimeTicks());
-        return runtime != null && runtime > 0 ? String.format("%d%s", (int) Math.ceil((double) runtime / 600000000), getString(R.string.lbl_min)) : "";
+        Long runtime_ticks = Utils.getSafeValue(mBaseItem.getRunTimeTicks(), mBaseItem.getRunTimeTicks());
+        if (runtime_ticks != null && runtime_ticks > 0) {
+            long runtime_minutes = runtime_ticks / 600000000;
+            if (runtime_minutes >= 60) {
+                long runtime_hours_only = runtime_minutes / 60;
+                long runtime_minutes_only = runtime_minutes - 60 * runtime_hours_only;
+                return runtime_hours_only + getString(R.string.lbl_hours_shorthands) + " " + runtime_minutes_only +
+                        getString(R.string.lbl_minutes_shorthand);
+            }
+            return runtime_minutes + getString(R.string.lbl_minutes_shorthand);
+        } else
+            return "";
     }
 
     private String getEndTime() {
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
@@ -677,16 +677,16 @@ public class FullDetailsFragment extends Fragment implements RecordingIndicatorV
     }
 
     private String getRunTime() {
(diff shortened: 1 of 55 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
```diff
@@ -677,8 +677,18 @@ public void setTitle(String title) {
     }
 
     private String getRunTime() {
-        Long runtime = Utils.getSafeValue(mBaseItem.getRunTimeTicks(), mBaseItem.getRunTimeTicks());
-        return runtime != null && runtime > 0 ? String.format("%d%s", (int) Math.ceil((double) runtime / 600000000), getString(R.string.lbl_min)) : "";
+        Long runtime_ticks = Utils.getSafeValue(mBaseItem.getRunTimeTicks(), mBaseItem.getRunTimeTicks());
```
nielsvanvelzen, line 680: Could probbaly use `TimeUtils.formatMillis` here, which is used in other places.

nielsvanvelzen, line 680: Variables need to be camelCase, so:

```suggestion
        Long runtimeTicks = Utils.getSafeValue(mBaseItem.getRunTimeTicks(), mBaseItem.getRunTimeTicks());
```

etc.

nielsvanvelzen (changes requested):
Don't modify translation files (values-xx folders) directly. Those are automatically managed by our translation tool.

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
noahvogt: Fixed in [88ad0e0](https://github.com/jellyfin/jellyfin-androidtv/pull/3878/commits/88ad0e07b99fa034edd2dafb160d54bff0667815).

noahvogt: Yeah, not really. It displays time in the format `10:05:55` instead of the format the jellyfin web client uses.  I would need to heavily modify this function to be able to use it.

noahvogt:
As requested, in [88ad0e0](https://github.com/jellyfin/jellyfin-androidtv/pull/3878/commits/88ad0e07b99fa034edd2dafb160d54bff0667815) I reverted the translation files, so this PR won't change them directly.

app/src/main/res/values/strings.xml
```diff
@@ -64,7 +64,8 @@
     <string name="lbl_watched">Watched</string>
     <string name="lbl_sd">SD</string>
     <string name="lbl_runs">"Runs"</string>
-    <string name="lbl_min">" min"</string>
+    <string name="lbl_minutes_shorthand">"m"</string>
+    <string name="lbl_hours_shorthands">"h"</string>
```
nielsvanvelzen, line 68:
```suggestion
    <string name="minutes_shorthand">m</string>
    <string name="hours_shorthand">h</string>
```

nielsvanvelzen (changes requested):

nielsvanvelzen:
Hey @noahvogt, are you still interested in finishing this pull request? it's currently waiting for you to address my comment above.

nielsvanvelzen:
Closing as stale. If you (or somebody else) is still interested in pursuing these changes please open a new pull request and link to this one.

PR #3932, opened 2024-08-25
Fix transcoding issues on devices without HEVC Main 10 support

MichaelRUSF:
Files played on devices that lacked HEVC Main 10 support were not being transcoded, leading to potential playback issues. This issue was identified in the `ProfileHelper` class and has been addressed with the following changes:

**Changes**

1. Added support for HEVC Main 10 check:
   - Introduced a new property `supportsHevcMain10` to determine HEVC Main 10 support, mirroring the approach used for other codecs.

2. Updated `deviceHevcCodecProfile`:
   - Modified the `conditions` property to use `ProfileConditionType.EqualsAny` when HEVC is supported. This ensures that profiles for both "Main" and "Main 10" are considered based on support availability, preventing issues with files not transcoding on devices lacking HEVC Main 10 support.

3. Refactored `deviceHevcLevelCodecProfiles`

4. Adjusted logging messages and comments.

**Issues**
Fixes #3923

PR #3933, opened 2024-08-25
Fix capitalization in video profiles causing invalid transcodes

nielsvanvelzen:
After merging #3932 some videos stopped playing for me with this error:

```
[hevc_nvenc @ 0000023043dc9f40] [Eval @ 00000039cbdfe4b0] Undefined constant or missing '(' in 'Main'
[hevc_nvenc @ 0000023043dc9f40] Unable to parse option value "Main"
[hevc_nvenc @ 0000023043dc9f40] Error setting option profile to value Main.
[vost#0:0/hevc_nvenc @ 000002303caf3700] Error initializing output stream: Error while opening encoder for output stream #0:0 - maybe incorrect parameters such as bit_rate, rate, width or height
[libfdk_aac @ 000002303cae4500] 2 frames left in the queue on closing
Conversion failed!
```

Turns out, our server doesn't sanitize shit and de nvenc encoder expects fully lowercase profiles. The server itself compares those strings everywhere ignoring casing though. Ideally this must be fixed serverside too.

**Changes**
- Fix capitalization in video profiles causing invalid transcodes

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
@@ -38,7 +38,7 @@ object ProfileHelper {
 						ProfileCondition(
 							ProfileConditionType.NotEquals,
 							ProfileConditionValue.VideoProfile,
-							"Main 10"
+							"main 10"
 						)
 					)
 				}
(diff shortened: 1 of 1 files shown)


crobibero (approved):
Wack

MichaelRUSF:
I had checked what the server was taking, and saw it was ignoring case. And the profiles were already a mixture up of lower and uppercase entries. All being the same I decided that uppercase looked better, so I went with that. Go figure.

nielsvanvelzen:
Yeah I also wouldn't expect it to cause problems, but when it did I checked the other clients and all send the profiles in lowercase.

Issue #3900, opened 2024-08-15
AAC 6/8-channel audio only outputs as Stereo PCM


salty-sweet:
### Describe the bug

**EXPECTED OUTCOME:**
- AV Receiver indicator should show any of the following: `DD`, `DD+`, `Signal Info: 3/2/0.1`
- All of the Speaker Indicators should light up. _(refer to [this manual](https://www.manua.ls/yamaha/htr-2071/manual?p=8))_

**ACTUAL OUTCOME:**
- AV Receiver indicator shows the following: `PCM`, `Signal Info: 2/-/---`
- Only `L`, `R`, and `SW` Speaker Indicators light up. _(subwoofer is always used by the AVR)_

> My AV Receiver is capable to receive 6 or 8 channel audio streams albeit physically limited to 5.1 audio output. Nearly all of my Jellyfin media contains 6/8-channel audio tracks. They vary in audio codecs, but only `DTS-HD MA`, `E-AC3`, and `AC3` audio codecs work. **The rest contain `AAC` 6/8-channel audio tracks, which only gets played as Stereo PCM.** I have given MediaInfo for the media file (that has AAC 5.1 audio) that I used to test and get ADB Logcat output.
> 
> - Jellyfin Server dashboard shows that my TV is "Direct Playing" and is not transcoding any of my media.
> 
> - Jellyfin AndroidTV client **does not** crash as it still plays the media, so there are no crash files from the server that I can give. The logs I have is ADB Logcat output from: `./adb logcat | Select-String -Pattern "(org.jellyfin.androidtv)|(jellyfin)|(codec)"`
> 
> - This only happened when I updated to `v0.17.x`. All media plays with surround sound back in `v0.16.11`.
>
> - Any other 6/8-channel audio codecs I have not mentioned above, I have not tested yet. 
> - Also audio with Atmos/DTS:X metadata I have not tested, and I assume they wouldn't work anyway on my AVR.

### Logs

08-15 18:21:57.400  1170  1170 I org.jellyfin.sdk.api.ktor.KtorClient: GET http://192.168.137.1:8096/Items/b0b94e1d-c2c8-183d-6e7f-bbdbe8fa1cd7/Intros
[...]

Issue #3903, opened 2024-08-17
Crackling sound on AC3 sound playback through Optical out


lockheed:
### Describe the bug

When playing through Optical Out (into an Ayiyama D03 amp), the sound from my Jellyfin on TLC TV is replaced by constant, loud crackling, as per this recording: https://www.youtube.com/watch?v=RkTO26mK4hs

So far, I noticed it only in videos that use sound in AC3 format, while AAC and MP3 work just fine. 
Also, there's no problem when using TV's built-in speakers.

### Application version

Latest from google store

### Device information

TCL 55C645

### Android version

Google TV

### Jellyfin server version

10.9.7

rickysixx:
1. have you checked "Bitstream Dolby Digital audio" in the application's settings?

2. have you tried to play the same media on your TV + amp outside of the Jellyfin app?
  You can put the media on a USB drive and check if that crackling is still there if you play the media using your TV's internal player.

lockheed:
1. I haven't seen such option in app settings and I spend some time looking through settings.
2. No, and unfortunately I won't be able to for a few weeks/months.

rickysixx:
> I haven't seen such option in app settings and I spend some time looking through settings.

You should found it at the bottom of Settings > Playback preferences > Advanced playback preferences.

PR #3936, opened 2024-08-26
Add ARC audio output mode with updated codec profiles and transcoding

MichaelRUSF:
This PR adds an HDMI ARC audio output mode to address playback issues with high-resolution and uncompressed audio formats over an ARC connection in Jellyfin. Updates to the `ExoPlayerProfile` class and incorporates a new audio output preference. The PR ensures compatibility with PCM and TrueHD formats by transcoding unsupported surround sound formats to AC3, maintaining surround sound capabilities.

**Changes**

1. Preference Option:
    - Added `HDMI_ARC_OUTPUT` option in `AudioBehavior` to allow users to enable HDMI ARC audio output.

2. New Parameter: 
   - Added `enableARCAudio` parameter to `ExoPlayerProfile`.
   - Added `AudioBehavior.HDMI_ARC_OUTPUT` condition to `PlaybackController`.

5. Updated Audio Codecs:
   - Added `downmixARCSupportedAudioCodecs` for ARC audio.
   - Adjusted `allSupportedAudioCodecs` to include/exclude codecs based on `enableARCAudio`.
   - Updated audio codec selection for transcoding and direct play profiles based on new settings.
   - Added maximum audio channel profiles specifically for ARC audio.
  

**Issues**
 Fixes: #3900 #3903 #2991

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
@@ -27,10 +27,18 @@ import org.jellyfin.apiclient.model.dlna.TranscodingProfile
 class ExoPlayerProfile(
 	disableVideoDirectPlay: Boolean,
 	isAC3Enabled: Boolean,
-	downMixAudio: Boolean
+	downMixAudio: Boolean,
+	enableARCAudio: Boolean
 ) : DeviceProfile() {
 	private val downmixSupportedAudioCodecs = arrayOf(
 		Codec.Audio.AAC,
+		Codec.Audio.MP3,
+		Codec.Audio.MP2
+	)
+
+	private val downmixARCSupportedAudioCodecs = arrayOf(
+		// Bypass isAC3Enabled check for HDMI ARC mode
+		Codec.Audio.AC3,
 		Codec.Audio.MP3,
 		Codec.Audio.MP2
 	)
@@ -41,14 +49,18 @@ class ExoPlayerProfile(
 	 */
 	private val allSupportedAudioCodecs = buildList {
 		addAll(downmixSupportedAudioCodecs)
+		if ((isAC3Enabled) || (enableARCAudio)) {
+			add(Codec.Audio.AC3)
+			add(Codec.Audio.EAC3)
+		}
+		if (!enableARCAudio) {
+			add(Codec.Audio.MLP)
+			add(Codec.Audio.TRUEHD)
+		}
 		add(Codec.Audio.AAC_LATM)
 		add(Codec.Audio.ALAC)
-		if (isAC3Enabled) add(Codec.Audio.AC3)
-		if (isAC3Enabled) add(Codec.Audio.EAC3)
 		add(Codec.Audio.DCA)
 		add(Codec.Audio.DTS)
-		add(Codec.Audio.MLP)
-		add(Codec.Audio.TRUEHD)
 		add(Codec.Audio.PCM_ALAW)
 		add(Codec.Audio.PCM_MULAW)
 		add(Codec.Audio.PCM_S16LE)
@@ -69,6 +81,18 @@ class ExoPlayerProfile(
 		maxStreamingBitrate = 20_000_000 // 20 mbps
 		maxStaticBitrate = 10_000_0000 // 10 mbps
 
+		// Determine audio codecs for profiles
+		val audioCodecForTranscodingProfile = when {
+			downMixAudio -> downmixSupportedAudioCodecs
+			enableARCAudio -> downmixARCSupportedAudioCodecs
+			else -> allSupportedAudioCodecsWithoutFFmpegExperimental
+		}.joinToString(",")
+
+		val audioCodecForDirectPlayProfile = when {
+			downMixAudio -> downmixSupportedAudioCodecs
+			else -> allSupportedAudioCodecsWithoutFFmpegExperimental
+		}.joinToString(",")
+
 		transcodingProfiles = arrayOf(
 			// TS video profile
 			TranscodingProfile().apply {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
```diff
@@ -208,6 +227,24 @@ class ExoPlayerProfile(
...
+					Codec.Audio.ALAC,
+					Codec.Audio.PCM_ALAW,
+					Codec.Audio.PCM_MULAW,
+					Codec.Audio.PCM_S16LE,
+					Codec.Audio.PCM_S20LE,
+					Codec.Audio.PCM_S24LE,
+					Codec.Audio.OPUS,
+					Codec.Audio.FLAC,
+					Codec.Audio.VORBIS
+				).forEach { codec ->
```
nielsvanvelzen, line 244: It should be possible to use a comma delimited list for codec in a single codec profile instead of adding separate profiles.

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -297,6 +297,20 @@ object ProfileHelper {
 			)
 		}
 
+	fun maxSPDIFAudioChannelProfile(codec: String, maxChannels: Int): CodecProfile {
```
nielsvanvelzen, line 300: Please inline into the usage side.

nielsvanvelzen (changes requested):

svemonix:
Thanks for the contribution and this is a very good step in the right direction. However I believe the multichannel audio issue will only be partially fixed just like on the Plex client.

Basically the use case not covered is the following. Many Android TV devices can't output multichannel PCM which means that multichannel AAC will be output as 2.0 PCM on a receiver plugged in via ARC. Now, you could activate the S/PDIF mode which would solve the multichannel AAC playback by transcoding to AC3.
The issue here is that it won't allow DD+ playback including DD+ Atmos even though the device and receiver supports it via ARC. It would also limit to 5.1.

So far the only client that can handle this properly is Kodi since it has more advanced settings. Maybe there should be a DD+ toggle together with the S/PDIF & ARC one.

MichaelRUSF:
> The issue here is that it won't allow DD+ playback including DD+ Atmos even though the device and receiver supports it via ARC. It would also limit to 5.1.

That statement is incorrect. Did you test a debug build, or are you making assumptions based on what you believe? AC3 (Dolby Digital) and EAC3 (Dolby Digital Plus) audio are not limited and behave the same as they would in direct play.

svemonix:
> > The issue here is that it won't allow DD+ playback including DD+ Atmos even though the device and receiver supports it via ARC. It would also limit to 5.1.
> 
> That statement is incorrect. Did you test a debug build, or are you making assumptions based on what you believe? AC3 (Dolby Digital) and EAC3 (Dolby Digital Plus) audio are not limited and behave the same as they would in direct play.

As stated it was an assumption but looking at the details maybe there is an issue the other way around.

`		if ((isAC3Enabled) || (enableSPDIF)) {
			add(Codec.Audio.AC3)
			add(Codec.Audio.EAC3)`

Isn't it gonna try to direct play and passthrough EAC3 even if using SPDIF (e.g. optical)? SPDIF doesn't support EAC3 only ARC.

salty-sweet:
> SPDIF doesn't support EAC3 only ARC.

I'd like to say that EAC3 can run through SPDIF up to 5.1 channels only. To temporarily work around the inconvenience caused by #3900, I am currently using a SPDIF connection for audio. Dolby Digital Plus 5.1 channel (and also limited to 1024kbps because more than that, FFMPEG-transcoded DDP/EAC3 audio doesn't work on my AVR) does work.

There could be some channel check in there as well to allow 5.1 EAC3, and filter out 7.1 over SPDIF.

MichaelRUSF:
This PR is aimed at ARC rather than Optical audio, despite the name of the option. The name "SPDIF" was chosen because it is more commonly used across other platforms compared to eARC vs. ARC. However, if this naming is causing confusion, it can be changed.

Targeting Optical audio would result in the loss of EAC3/JOC (DD+ with Dolby Atmos) and limit DD to 5.1, which is not the goal of this PR. If issues arise for optical users after merging this PR, they can be addressed. My objective is to keep this process straightforward and get the PR merged.

app/src/main/res/values/strings.xml
```diff
@@ -516,6 +516,7 @@
     <string name="past_24_hours">Past 24 hours</string>
     <string name="prefer_exoplayer_ffmpeg">Prefer FFmpeg for audio playback</string>
     <string name="prefer_exoplayer_ffmpeg_content">Use FFmpeg to decode audio, even if platform codecs are available.</string>
+    <string name="audio_spdif">SPDIF</string>
```
MichaelRUSF, line 519: Other name options to better describe this output mode:
ARC
ARC Audio
HDMI Audio

svemonix: Thanks for clarifying the intent of the PR.

To avoid confusion, I'd suggest ARC or maybe even better HDMI ARC.

HDMI Audio is probably too vague as it doesn't differentiate ARC and eARC and would just sound like it means HDMI audio passthrough.

ARC Audio seems a bit redundant as ARC already contains Audio.

MichaelRUSF: HDMI ARC seems like the better option, will update  to reflect this.

WortelSoep:
Sorry to intrude, but it would be nice to get this merged. Is there anything else to do other than fix the merge requests? (If so happy to do that)

nextlooper42:
Somebody please merge this so we can have fixed sound finally in next beta/stable.

znidaxp:
When will this be implemented?

thechitowncubs:
Does this fix Android 9 limitation allowing pass through for 5.1 channels? I'm only getting PCM 2.0

n8llcaster:
I merged this to v0.18.11 and it's working without problems or any conflicts, that could not be resolved in seconds. It completely fix audio passthrough and AAC 5.1 problems on Sony TV's (Android 9) and HDMI connected receiver. 

Only major complains from Android studio are about old libraries used in v0.18.11 :D

Can we please move on with this or we just gonna ignore major playback problem? Is it proper fix? Propably not, but does fix problem without breaking anything? Yes. At least we can have this until someone finds better solution, instead of... **nothing**.

nielsvanvelzen:
Closing as stale/abandoned. If you're interested in continuing these changes please open a new pull request.

PR #3953, opened 2024-08-31
Add resolution codec check

MichaelRUSF:
This PR introduces a new check to verify resolution support, ensuring compatibility with 1080p devices. The update involves changes to `MediaCodecCapabilitiesTest`, `ProfileHelper`, `ExoPlayerProfile`, and `PlaybackController`, allowing playback options to be configured according to the codec's resolution capabilities of the device.

**Changes**

1. MediaCodecCapabilitiesTest:
   - Added `getMaxResolution` method to return the maximum supported width and height.
2. ProfileHelper:
   - Added `maxResolutionCodecProfile`
     - Dynamically use the device’s maximum supported resolution obtained from `MediaTest.getMaxResolution`
   - Removed `max1080pProfileConditions`
3. ExoPlayerProfile:
   - Updates the `ExoPlayerProfile` to reflect changes in the handling of 1080p profiles and removes the `disable4KVideo `parameter.
5. PlaybackController:
   - Remove `!DeviceUtils.has4kVideoSupport` condition.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
@@ -166,7 +166,10 @@ class MediaCodecCapabilitiesTest {
 		return false
 	}
 
-	fun supports4KResolution(): Boolean {
+	fun getMaxResolution(): Pair<Int, Int> {
+		var maxWidth = 0
+		var maxHeight = 0
+
 		for (info in mediaCodecList.codecInfos) {
 			if (info.isEncoder) continue
 
@@ -175,21 +178,17 @@ class MediaCodecCapabilitiesTest {
 				val videoCapabilities = capabilities.videoCapabilities
 
 				if (videoCapabilities != null) {
-					// Get the supported width and height ranges
-					val maxWidth = videoCapabilities.supportedWidths?.upper ?: 0
-					val maxHeight = videoCapabilities.supportedHeights?.upper ?: 0
+					val supportedWidth = videoCapabilities.supportedWidths?.upper ?: 0
+					val supportedHeight = videoCapabilities.supportedHeights?.upper ?: 0
 
-					// Check for 4K resolution support
-					if (maxWidth >= 3840 && maxHeight >= 2160) {
-						Timber.i("4K resolution is supported by codec %s", info.name)
-						return true
-					}
+					maxWidth = maxOf(maxWidth, supportedWidth)
+					maxHeight = maxOf(maxHeight, supportedHeight)
 				}
 			} catch (e: IllegalArgumentException) {
 				Timber.d(e, "Codec %s does not support video capabilities", info.name)
 			}
 		}
 
-		return false
+		return Pair(maxWidth, maxHeight)
 	}
 }
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
@@ -4,6 +4,7 @@ import android.media.MediaCodecInfo.CodecProfileLevel
 import android.media.MediaCodecList
 import android.media.MediaFormat
 import android.os.Build
+import android.util.Size
 import timber.log.Timber
 
 class MediaCodecCapabilitiesTest {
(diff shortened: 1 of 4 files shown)


nielsvanvelzen (changes requested):
It would be even better to use the highest supported width/height values directly in the profile instead.

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
```diff
@@ -165,4 +165,30 @@ class MediaCodecCapabilitiesTest {
 
 		return false
 	}
+
+	fun getMaxResolution(): Pair<Int, Int> {
```
nielsvanvelzen, line 169: Use the android.util.Size class instead of a pair

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
```diff
@@ -165,4 +165,30 @@ class MediaCodecCapabilitiesTest {
...
+		var maxWidth = 0
+		var maxHeight = 0
+
+		for (info in mediaCodecList.codecInfos) {
+			if (info.isEncoder) continue
+
+			try {
+				val capabilities = info.getCapabilitiesForType(MediaFormat.MIMETYPE_VIDEO_AVC)
```
nielsvanvelzen, line 177: I'd move the media type to a parameter so the function is re-usable for different codecs

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
```diff
@@ -165,4 +165,30 @@ class MediaCodecCapabilitiesTest {
...
+				val videoCapabilities = capabilities.videoCapabilities
+
+				if (videoCapabilities != null) {
+					val supportedWidth = videoCapabilities.supportedWidths?.upper ?: 0
+					val supportedHeight = videoCapabilities.supportedHeights?.upper ?: 0
```
nielsvanvelzen, line 182:
```suggestion
				val videoCapabilities = capabilities.videoCapabilities ?: continue
				val supportedWidth = videoCapabilities.supportedWidths?.upper ?: continue
				val supportedHeight = videoCapabilities.supportedHeights?.upper ?: continue
```

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -15,6 +15,34 @@ import timber.log.Timber
 object ProfileHelper {
 	private val MediaTest by lazy { MediaCodecCapabilitiesTest() }
 
+	private val supports4K by lazy {
+		val (maxWidth, maxHeight) = MediaTest.getMaxResolution()
+		maxWidth >= 3840 && maxHeight >= 2160
+	}
+
+	val max1080pCodecProfile by lazy {
```
nielsvanvelzen, line 23: Quoting from my previous review:

> It would be even better to use the highest supported width/height values directly in the profile instead.

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
```diff
@@ -15,6 +15,34 @@ import timber.log.Timber
...
+		CodecProfile().apply {
+			type = CodecType.Video
+			conditions = if (supports4K) {
+				// No restriction if 4K is supported
+				arrayOf()
```
nielsvanvelzen, line 28: We should never add empty codec profiles

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
MichaelRUSF: Added to support up to 4k resolutions instead.

app/src/main/java/org/jellyfin/androidtv/util/profile/ProfileHelper.kt
MichaelRUSF: I integrated it into the `max1080pCodecProfile`.

nielsvanvelzen: No you did not? It still hardcodes the width and height based on the result of the getMaxResolution function.

MichaelRUSF: Should be removed now.

nielsvanvelzen (approved):

PR #3969, opened 2024-09-07
Add subtitile position offset option for media3 renderer

pbasov:
This PR adds back subtitle position control after client lost it when sub custom renderer has been removed in https://github.com/jellyfin/jellyfin-androidtv/pull/3825

**Changes**

- Adjusted VideoManager.java to apply the offset value to the subtitle view using `setBottomPaddingFraction`.
- Added a seekbar for configuring subtitle position offset in the `PlaybackPreferencesScreen`.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -182,6 +182,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		var subtitlesTextSize = floatPreference("subtitles_text_size", 1f)
 
 		/**
+		 * Subtitles offset
+		 */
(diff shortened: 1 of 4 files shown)


nielsvanvelzen (changes requested):
While this does work, I'm not sure about the values. Using -10 or 100 will not display the subtitles at all so those don't feel like safe limits. The seekbar doesn't go over 52 for some reason so I had to use my mouse in the emulator (which obviously doesn't work on a real device). Also, there is no unit specified. So I can place my subtitles at "-5", -5 what? pixels? meters? dollars?

nielsvanvelzen:
Hey @pbasov, are you still interested in finishing this pull request? it's currently waiting for you to address my comment above.

nielsvanvelzen:
Closing as stale. If you (or somebody else) is still interested in pursuing these changes please open a new pull request and link to this one.

PR #4020, opened 2024-09-24
Add preference for default zoom mode

### LINKED ISSUE #1981: Remember Aspect ratio (like Playback speed)

First of all, thanks to all the contributors who are continuously improving the client!

Just like how Playback speed is remembered (until the app is restarted or manually changed), the app should similarly remember the aspect ratio changes. I believe the aspect ratio feature is currently only available when using ExoPlayer.

An alternative, probably requiring more effort, would be to add an option in Settings (perhaps in Playback > Video section) to set the default aspect ratio and use that. Until more media players are support, this could be visible when ExoPlayer is selected.

JagerSprinkles:
**Changes**

This change will add a setting under advanced playback preferences to set the default zoom mode. This mode will be used every time a video is started as the setting is not saved if adjusted during playback.

**Issues**

Possibly fixes #1981


app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -200,6 +200,12 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Whether items shown in the screensaver are required to have an age rating set.
 		 */
 		var screensaverAgeRatingRequired = booleanPreference("screensaver_agerating_required", true)
+
+		/**
+		 * Preferred behavior for zoom mode.
+		 */
+
+		var zoomMode = intPreference("zoom_mode", 0)
```
nielsvanvelzen, line 208: Please use an enumPreference with a (new) enum type instead. It's probably best to change the VideoManager to use the enum too.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
JagerSprinkles: I have updated to use a new enum. I am still getting familiar with this language so please let me know if additional changes are required.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -200,6 +201,12 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Whether items shown in the screensaver are required to have an age rating set.
 		 */
 		var screensaverAgeRatingRequired = booleanPreference("screensaver_agerating_required", true)
+
+		/**
+		 * Preferred behavior for zoom mode.
+		 */
+
+		var zoomMode = enumPreference("zoom_mode", ZoomMode.ZOOM_FIT)
```
nielsvanvelzen, line 209:
```suggestion
		var playerZoomMode = enumPreference("player_zoom_mode", ZoomMode.ZOOM_FIT)
```

app/src/main/java/org/jellyfin/androidtv/ui/preference/screen/PlaybackAdvancedPreferencesScreen.kt
```diff
@@ -64,6 +65,11 @@ class PlaybackAdvancedPreferencesScreen : OptionsFragment() {
 				setTitle(R.string.pref_external_player)
 				bind(userPreferences, UserPreferences.useExternalPlayer)
 			}
+
+			enum<ZoomMode> {
+				setTitle(R.string.lbl_zoom)
```
nielsvanvelzen, line 70: We may want to add a new string resource that is more descriptive instead of this one

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/action/ZoomAction.kt
```diff
@@ -28,39 +33,41 @@ class ZoomAction(
 		videoPlayerAdapter.leanbackOverlayFragment.setFading(false)
 		return PopupMenu(context, view, Gravity.END).apply {
```
nielsvanvelzen, line 34: If you migrate this code to use our MenuBuilder you don't need to convert the enum members to int and back.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/preference/screen/PlaybackAdvancedPreferencesScreen.kt
JagerSprinkles: Perhaps "Aspect Ratio" is more descriptive? I have made a new string resource called pref_aspect_ratio with this text.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/action/ZoomAction.kt
JagerSprinkles: I was looking for something like this when creating the enum, I must have missed it previously. Thanks for pointing me in the right direction!

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
JagerSprinkles: I have updated this as requested.

app/src/main/java/org/jellyfin/androidtv/ui/preference/screen/PlaybackAdvancedPreferencesScreen.kt
JagerSprinkles: @nielsvanvelzen do you think "Aspect ratio" is descriptive enough? If not I can come up with another string that is more descriptive for this setting.

nielsvanvelzen: Maybe something like "Default video zoom"?

JagerSprinkles: "Default video zoom" sounds good to me. One other suggestion could be "Default zoom mode" as the setting is under the Video category in the settings menu so not sure if the word "video" would be redundant. I think either could work but its hard for me to tell which one aligns better with the rest of settings.

nielsvanvelzen:
Thanks for the pull request! I've pushed some final changes to fix some formatting, rename an enum member, changed the preference order. I also fixed a missing setter for `mZoomMode` causing the popup in the video player to never show the current selection and changed the preference to "Default video zoom" like you suggested.

nielsvanvelzen (approved):

JagerSprinkles:
Thanks for the small fixes at the end! I'm happy to contribute to the project and look forward to submitting more PRs in the future!

PR #4025, opened 2024-09-27
Add preference to delay start of video

JagerSprinkles:
**Changes**

This updates the delay when starting the video player to be user configurable. Having this as configurable will help people who have displays that are slow to change resolution or refresh rate. I have two displays and a soundbar that take longer than the 0.75 seconds default to change format or mode. With this setting the first few seconds of the video and audio are not missed.

**Issues**

None I can find.


app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -200,6 +200,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Whether items shown in the screensaver are required to have an age rating set.
 		 */
 		var screensaverAgeRatingRequired = booleanPreference("screensaver_agerating_required", true)
+
+		/**
+		 * Delay when starting video playback after loading the video player.
+		 */
+		var videoStartDelay = longPreference("video_start_delay", 750)
```
nielsvanvelzen, line 207: Could default to just `0`. There really is no reason to delay playback by default.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -646,7 +646,7 @@ public void run() {
                 if (mVideoManager != null)
                     mVideoManager.start();
             }
-        }, 750);
+        },  userPreferences.getValue().get(UserPreferences.Companion.getVideoStartDelay()));
```
nielsvanvelzen, line 649: The comment above (about why this is delayed) is inaccurate and could be removed. Additionally a fast-path when there is no delay would be nice to have.

app/src/main/res/values/strings.xml
```diff
@@ -516,6 +516,7 @@
     <string name="past_24_hours">Past 24 hours</string>
     <string name="prefer_exoplayer_ffmpeg">Prefer FFmpeg for audio playback</string>
     <string name="prefer_exoplayer_ffmpeg_content">Use FFmpeg to decode audio, even if platform codecs are available.</string>
+    <string name="lbl_video_start_delay">Video Start Delay</string>
```
nielsvanvelzen, line 519:
```suggestion
    <string name="lbl_video_start_delay">Video start delay</string>
```

nielsvanvelzen: Also the `lbl_` prefix isn't needed

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
JagerSprinkles: This has been updated to 0 now.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
JagerSprinkles: By fast-path I assumed you meant to add a condition to skip the mHandler.postDelayed logic if the delay is set to 0. I have added a check for this.

app/src/main/res/values/strings.xml
JagerSprinkles: I have updated the string name and value as requested.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -639,14 +639,20 @@ private void startItem(BaseItemDto item, long position, StreamInfo response) {
...
 
-        //wait a beat before attempting to start so the player surface is fully initialized and video is ready
-        mHandler.postDelayed(new Runnable() {
-            @Override
-            public void run() {
-                if (mVideoManager != null)
-                    mVideoManager.start();
+        //set video start delay
+        long videoStartDelay = userPreferences.getValue().get(UserPreferences.Companion.getVideoStartDelay());
+        if (mVideoManager != null) {
```
nielsvanvelzen, line 644: This null cehck should only happen in the delayed function as we need to verify the manager is not disposed. We already know it is not null when calling it directly.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
nielsvanvelzen: Yup that's what I meant!

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
JagerSprinkles: I have updated the null check for only when a delay has been set.

nielsvanvelzen (approved):
Awesome, thank you! I've pushed some final changes (see last commit).

JagerSprinkles:
Thanks for the little correction in that last commit!

PR #4037, opened 2024-09-30
Limit audio channels to 6 in video transcoding profile

MichaelRUSF:
This PR modifies the ExoPlayerProfile to limit the maximum number of audio channels to 6 in the video transcoding profile. Remuxing 7.1 audio with transcoded video  leads to low frame rates, resulting in poor playback quality and player errors. This change aims to optimize compatibility by ensuring that 7.1 audio is transcoded when the video stream is transcoded.<br>
<br>

**Changes**

Added `maxAudioChannels`: Set to "6" in the TS video profile for improved audio  handling when transcoding.

**Issues**
 Addresses issues brought up on the forum.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
@@ -86,6 +86,7 @@ class ExoPlayerProfile(
 				protocol = "hls"
 				copyTimestamps = false
 				enableSubtitlesInManifest = true
+				maxAudioChannels = "6"
 			},
 			// MP3 audio profile
 			TranscodingProfile().apply {


app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
```diff
@@ -86,6 +86,7 @@ class ExoPlayerProfile(
 				protocol = "hls"
 				copyTimestamps = false
 				enableSubtitlesInManifest = true
+				maxAudioChannels = "6"
```
nielsvanvelzen, line 89: I don't think this is a proper fix. This will limit **all** transcodes to 6 channels, even when it's just the video codec that is not supported or subtitles need to be burned. While with direct play it will still go to up to 8 channels according to the `downMixAudio` preference.

nielsvanvelzen (changes requested):
I don't think this is a proper fix. This will limit **all** transcodes to 6 channels, even when it's just the video codec that is not supported or subtitles need to be burned.

app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt
MichaelRUSF: The sever  already [limits](https://github.com/jellyfin/jellyfin/blob/7cc3f9506eafe37bc2df263645753c75b6ed3441/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L117) transcodes to 6 channels. This is saying if the video for 8 channel media is being transcoded, then the audio should also be transcoded. There is no issue with direct play so that is the desired effect. I haven't considered brunt subtitles, will test that out.

MichaelRUSF: Burnt in subtitles had the same effect as lowering the bitrate, the video playback was unwatchable due to low framerates on 8 channel EAC3 media and playback failure on 8 channel AAC media.

For TrueHD and DTS (HD/MA/X) 7.1 the server automatically transcodes the audio when the video is transcoded  by either lowering the bitrate or by burning in subs.
```
Stream #0:1(eng): Audio: truehd (Dolby TrueHD + Dolby Atmos), 48000 Hz, 7.1, s32 (24 bit)
Stream mapping:
  Stream #0:0 (hevc) -> setparams:default (graph 0)
  format:default (graph 0) -> Stream #0:0 (hevc_amf)
  Stream #0:1 -> #0:1 (truehd (native) -> aac (libfdk_aac))
 
  Stream #0:0: Video: hevc (hvc1 / 0x31637668), d3d11(tv, bt709, progressive)...
  Stream #0:1: Audio: aac, 48000 Hz, 5.1, s16, 640 kb/s
```

```
Stream #0:1(eng): Audio: dts (DTS-HD MA + DTS:X IMAX), 48000 Hz, 7.1, s32p (24 bit) (default)
Stream mapping:
  Stream #0:0 (hevc) -> setparams:default (graph 0)
  format:default (graph 0) -> Stream #0:0 (hevc_amf)
  Stream #0:1 -> #0:1 (dts (dca) -> aac (libfdk_aac))

  Stream #0:0: Video: hevc (hvc1 / 0x31637668), d3d11(tv, bt709, progressive)...
  Stream #0:1: Audio: aac, 48000 Hz, 5.1, s16, 640 kb/s (default)
```
So this PR would only change the behavior for transcoded AAC and EAC3 8 channel media and it would align with what the server is already doing to TrueHD and DTS (HD/MA/X) media.

rxbrad: Concur with @MichaelRUSF on embedded subs.  (Just came to report my testing, but he beat me to it)

I just tested a AAC 7.1 + VOBSUB file.  It could *only* be played when transcoding both video & audio.  Direct Play of video+audio fails due to sub burn-in.

MichaelRUSF:
I also checked JMP, and it also has [maxAudioChannels](https://github.com/jellyfin/jellyfin-media-player/blob/abcd3d903d6f4da2a8ff118329c8599ef59890e1/native/nativeshell.js#L191) set to 6 in their transcoding profile. So this would align with JMP's settings.

nielsvanvelzen:
Closing as stale & unwanted. We do not want to arbitrarily limit the maximum audio channels to 6.

PR #4065, opened 2024-10-09
Restore forced aspect ratio for poster types on video items

### LINKED ISSUE #4062: Different aspect ratios in the library view
### Describe the bug

Since version 0.17.6, the library view shows cards with different aspect ratios when the movie has no poster and a thumbnail is used instead, even if the image type is set to poster. This causes columns to have different and changing widths as titles with a thumbnail appear and disappear from view.

### Logs

_No response_

### Application version

0.17.6

Google Play

### Device information

Philips 65 OLED 707/77

### Android version

Android TV 11

### Jellyfin server version

10.9.11

nielsvanvelzen:
Found an issue where the image of season 1 of Pioneer One on our demo server wouldn't show properly while trying to reproduce #4062, not sure if it fixes that issue too though. The image logic is a mess.

**Changes**
- Restore forced aspect ratio for poster types on video items

**Issues**

Maybe fixes #4062
Reverts a part of #4015 (@Bond-009 hopefully this doesn't break you images again)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
@@ -184,6 +184,8 @@ public class CardPresenter extends Presenter {
                         case VIDEO:
                             mDefaultCardImage = ContextCompat.getDrawable(mCardView.getContext(), R.drawable.tile_port_video);
                             showProgress = true;
+                            if (imageType.equals(ImageType.POSTER))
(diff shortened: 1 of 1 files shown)


Bond-009:
Yeah, this will break it again

crobibero (approved):

nielsvanvelzen:
Unfortunately we'll have to choose our poison and I rather have special show as poster instead of all items without a primary (poster) image using the wrong size, which breaks the browsing grids.

I tried adding a special case just for specials but the part of the code that shows the UI unfortunately doesn't have any way of knowing if the item is a special.

PR #4080, opened 2024-10-15
Use Java 21 in jvm toolchain

nielsvanvelzen:
The current Android Studio release bundles JDK 21, so using that will make it easier to get started with development as there is no need to manually install a JDK.

**Changes**
- Update toolchain to Java 21
- Refactor toolchain config to use version catalog

**Issues**


crobibero (approved):

nielsvanvelzen:
Ah builds failed.. exactly what I expected!

(need to update the workflows to use JDK21 too, oops)

PR #4085, opened 2024-10-16
Add option to adjust screensaver image display duration

Havarius:
**Changes**
Add settings to show screensaver images for different amount of times.
Updated logo handling to prevent stretching across the full screen.
Add production year under logo, great for larger movie collections.
Changed movie title to uppercase when no logo is present, aligning with standard logo styling.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/integration/dream/DreamViewModel.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/dream/DreamViewModel.kt
@@ -75,6 +75,7 @@ class DreamViewModel(
 	private val _libraryContent = flow {
 		// Load first library item after 2 seconds
 		// to force the logo at the start of the screensaver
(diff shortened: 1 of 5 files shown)


nielsvanvelzen (changes requested):
We won't be adding unnecessary preferences, the duration each item is shown in the screensaver is a static value and we'll keep it that way. The capitalization of item names should not be changed by the client.

Havarius:
Why would it be unnecessary? Most people have a larger library and the default takes 30 seconds to show a single movie/series, so you will never see most of your library.
And if something like this is unnecessary why can I set the screensaver timeout? Why is it here necessary to change it and there is no default of 5 minutes?

nielsvanvelzen:
The 30 seconds per item highlight was chosen in combination with the various animations that play and and is how I decided the user experience should be. The app is designed in a "less is more" fashion and we try to minimize the amount of thinks that can be tweaked to make sure all users have a similar experience and testing the app won't get impossible.

PR #4094, opened 2024-10-19
Rewrite all subtitle behavior

nielsvanvelzen:
In #3825 I made the initial changes to no longer require external subtitles. This came with a few issues; there was no longer support for external subtitles, switching subtitles didn't work a lot of times and the app didn't track which subtitles were playing properly. This pull request aims to fix all the issues by rewriting the subtitle logic.

The new logic relies on the server providing the delivery method for subs. For example, when transcoding the delivery method is set to HLS but on direct-play it is Embedded or External. The client will then decide the appropriate method to loading subtitles based on that information.

In my testing everything appears to work correctly but I'd appreciate some additional testing for:

- Playback starts with the correct subtitles when they are embedded/external/baked
- Switching subtitles from baking to external/embedded

There should be sufficient logging to troubleshoot issues.

**Changes**
- Rewrite subtitle switching
- Support external subtitles
 - Uses the URL provided by server instead of manually constructing it
- Unsupport "sub" codec as I couldn't figure out which media type in ExoPlayer it belongs to so it probably never worked before

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
@@ -41,11 +41,18 @@ import androidx.media3.ui.CaptionStyleCompat;
 import androidx.media3.ui.PlayerView;
 
 import org.jellyfin.androidtv.R;
+import org.jellyfin.androidtv.data.compat.StreamInfo;
 import org.jellyfin.androidtv.preference.UserPreferences;
 import org.jellyfin.androidtv.preference.constant.ZoomMode;
+import org.jellyfin.playback.media3.exoplayer.mapping.SubtitleKt;
+import org.jellyfin.sdk.api.client.ApiClient;
 import org.jellyfin.sdk.model.api.MediaStream;
+import org.jellyfin.sdk.model.api.MediaStreamType;
+import org.jellyfin.sdk.model.api.SubtitleDeliveryMethod;
 import org.koin.java.KoinJavaComponent;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
@@ -97,7 +104,7 @@ public class VideoManager {
                 strokeColor,
                 null
         );
-        mExoPlayerView.getSubtitleView().setFractionalTextSize(0.0533f *  userPreferences.get(UserPreferences.Companion.getSubtitlesTextSize()));
+        mExoPlayerView.getSubtitleView().setFractionalTextSize(0.0533f * userPreferences.get(UserPreferences.Companion.getSubtitlesTextSize()));
         mExoPlayerView.getSubtitleView().setStyle(subtitleStyle);
         mExoPlayer.addListener(new Player.Listener() {
             @Override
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -314,15 +321,47 @@ public long seekTo(long pos) {
...
+                    Uri subtitleUri = Uri.parse(api.createUrl(mediaStream.getDeliveryUrl(), Collections.emptyMap(), Collections.emptyMap(), true));
+                    MediaItem.SubtitleConfiguration subtitleConfiguration = new MediaItem.SubtitleConfiguration.Builder(subtitleUri)
+                            .setId("JF_EXTERNAL:" + String.valueOf(mediaStream.getIndex()))
+                            .setMimeType(SubtitleKt.getFfmpegSubtitleMimeType(mediaStream.getCodec()))
+                            .setLanguage(mediaStream.getLanguage())
+                            .setLabel(mediaStream.getDisplayTitle())
+                            .setSelectionFlags(getSubtitleSelectionFlags(mediaStream))
+                            .build();
+                    subtitleConfigurations.add(subtitleConfiguration);
+                    subtitleConfigurations.add(subtitleConfiguration);
```
crobibero, line 355: Did you mean to add it twice?

nielsvanvelzen: Definitely not, that was for debugging. Good catch

crobibero (approved):

MichaelRUSF:
Findings

Server: 10.10.0
Commit: b3da2ff8f494c91fb19779d64db47fff521a155e

srt: 
embedded – no issue
external – no issue
no subtitles – no issue

pgs: 
embedded – no issue
external – player exited, to many errors  (in version 0.17, it would indicate "unable to load subtitle" and continue playback)
no subtitles – no issue

ass:
embedded – burnt in
external – burnt in
no subtitles – subtitles always appear in a plain, unstyled font, without burn-in from the server

nielsvanvelzen:
External PGS should be fixed now. The codec->mime mapping failed because the server returns `PGSSUB` (fullcaps) as codec and our mapping was case sensitive. Not sure why this is the only codec that happens with.

ASS/SSA should also be fixed. When the subtitle track was default/foced and the video was direct playing the subtitle track would be embedded, I missed a case where the ExoPlayer needed to be disabled. (there as a check to only force-update the ExoPlayer subtitle track when a subtitle was selected).

Thanks for testing!

MichaelRUSF:
New findings: 
ass:
If "none" is set as the default option, the plain unstyled font will appear during the first playback (direct play). If you switch to a subtitle track while playing (burn-in) and then revert to "none," the subtitle then turns off.

pgs:
I tested a file with two embedded tracks and one external track. Only the first embedded subtitle track functions; selecting a secondary or external track results in no display.

nielsvanvelzen:
First one fixed with #4098. For the second one it feels like there is an issue with androidx.media3 that I need some more time to investigate. It wouldn't show subtitles for me whenever there was more then one pgs track.

MichaelRUSF:
I just noticed that the index is off for PGS. 

In the file:
Track 1 - non-sdh pgs
Track 2 - sdh pgs

When I select track 1, it displays the sdh pgs track (track 2). Selecting track 2 leads to no subtitles being displayed.

PR #4137, opened 2024-11-01
Add preference to set payback UI fade time

JagerSprinkles:
**Changes**
This change adds a preference to adjust how long the playback UI is shown on the screen after pressing a button. This was previously hard coded to 6 seconds. This allows the user to adjust this time between 1 and 10 seconds in steps of 0.5 seconds. The seekbar "min" has a maximum value of 100 so I have set it to use between 100 and 1000 while adjusting the value when it is saved into the UserPreferences.playbackUiFadeTime.

**Issues**
None that I could find


nielsvanvelzen (changes requested):
I don't see any need for such a niche user preference.

JagerSprinkles:
> I don't see any need for such a niche user preference.

I had created this as rewinding for the default of 10 seconds has the screen covered for 6 seconds afterwards by the UI. I felt this gives the user more control over what is covering their screens at times. 
If you don't feel this is needed then we can close this out, I have no issues with that.

PR #4154, opened 2024-11-03
Fix audio track selection when baking subtitles

nielsvanvelzen:
Hopefully this doesn't introduce a bunch of other issues again, didn't really have the time to test all cases.

**Changes**

- Fix current audio track selection not set when baking subtitles

**Issues**


crobibero (approved):
#yolo

PR #4164, opened 2024-11-05
Fix exo track id parsing for format change.

reacocard:
Without this changing to other audio tracks fails, it just keeps playing the first one in the file. I was tipped off to the source of the problem by lines like this in logcat:

11-04 19:26:01.244 11669 11669 D VideoManager: failed to parse track ID [0:2]

It appears exoplayer changed from a simple number for trackid to two numbers separated by a colon. Haven't had any luck tracking down why, but the latter of the two appears to match the old single number so, just pull that out and use it.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
@@ -415,7 +415,11 @@ public class VideoManager {
                         if (trackFormat.id != null) {
                             int id;
                             try {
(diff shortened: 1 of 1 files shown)


bin101:
Can confirm that this also fixes my audio track switching issues.

nielsvanvelzen (approved):
When using external subtitles the media source becomes a MergingMediaSource with the original item as first media source and all subtitle tracks after that. This will change the track ids to be prefixed with their source index (so `1:` for the item).

Ideally we should stop parsing the track ids entirely but that might not be possible here.

PR #4171, opened 2024-11-05
Catch errors when reporting playback state

nielsvanvelzen:
Ideally the stop event should notify the user too but that can be done later ©️ 

**Changes**
- Catch errors when reporting playback state

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/apiclient/ReportingHelper.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/apiclient/ReportingHelper.kt
@@ -53,7 +53,9 @@ class ReportingHelper(
 
 		lifecycleOwner.lifecycleScope.launch {
 			Timber.i("Reporting ${item.name} playback started at $position")
-			api.playStateApi.reportPlaybackStart(info)
+			runCatching {
+				api.playStateApi.reportPlaybackStart(info)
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/util/apiclient/ReportingHelper.kt
```diff
@@ -87,7 +89,9 @@ class ReportingHelper(
 
 		lifecycleOwner.lifecycleScope.launch {
 			Timber.d("Reporting ${item.name} playback progress at $position")
-			api.playStateApi.reportPlaybackProgress(info)
+			runCatching {
+				api.playStateApi.reportPlaybackProgress(info)
+			}.onFailure { error -> Timber.w(error, "Failed to report playback progress") }
```
crobibero, line 94: This one isn't as excited as the others

crobibero (approved):

app/src/main/java/org/jellyfin/androidtv/util/apiclient/ReportingHelper.kt
nielsvanvelzen: Correct, if the start/stop fails it can screw with activity / playback reporting etc. A missed progress update can be easily recovered with the next update.

PR #4172, opened 2024-11-05
Add experimental TrickPlay implementation

### LINKED ISSUE #3585: Scrubbing

So now that scrubbing is a built in Jellyfin feature and not a plugin, any plan to support it?

nielsvanvelzen:
Add an implementation of TrickPlay images using the thumbnail support in the Leanback PlaybackSeekDataProvider. We have no control over the UI and I'm not a fan of how it looks, therefor this feature is **disabled by default**. Enabling the feature can be done in the developer settings (it is a public developer option, available in release builds).

**Changes**
- Add new "trickPlayEnabled" preference - disabled by default
  - Needs to be enabled in developer options
- Provide TrickPlay images in CustomSeekProvider

**Screenshots**

**Issues**

Fixes #3585


nielsvanvelzen:
As a side-note; it is possible that the TrickPlay images don't show up when starting playback from specific screens. the app must request the TrickPlay metadata when retrieving items and I might have forgotten to add that to a few places. If you find one of these spots; please open an issue to let me know and I'll fix it.

crobibero (approved):

PR #4203, opened 2024-11-12
Fix player UI not hiding automatically

nielsvanvelzen:
Somewhat surprised nobody complained about this, this was the only known issue to me since the first beta 😄 

**Changes**
- Fix player UI not hiding automatically

**Issues**
Fix regression from #4119


crobibero (approved):

MulverineX:
> Somewhat surprised nobody complained about this

A lot of people are using Kodi while we wait for the video playback rewrite to be complete. Its why I'm sponsoring and I don't really care about anything else being worked on right now.

nielsvanvelzen:
A lot of changes that go into the old player code are in a way trials for the rewrite. And I've already had a lot of feedback and bug reports on other parts in the 0.18 beta that are way harder to hit. So that's why I was somewhat surprised nobody complained about this one, as it was quite easy to notice :p

PR #4211, opened 2024-11-14
Fix background of album in now playing screen when lyrics are displaying

### LINKED ISSUE #4207: Lyrics can be hard to read

May want to wait for other feedback but I find the lyrics hard to read when layered over album artwork and artist background

Image

nielsvanvelzen:
bit of a hack but it's legacy views

**Changes**
- Fix background of album in now playing screen when lyrics are displaying

**Issues**
Fixes #4207


crobibero (approved):
Looking great, thanks!
screen_fixed

PR #4253, opened 2024-11-30
Display media runtime in hour minute format

dtrexler:
# Display media runtime in hour minute format

**Changes**
Updates the runtime display in the FullDetailsFragment to a ```[0-9]+h [0-9][0-9]?m``` format in order to be consistent with the web client.

**Issues**
Fixes issue https://github.com/jellyfin/jellyfin-androidtv/issues/2806
Supersedes pull request https://github.com/jellyfin/jellyfin-androidtv/pull/3878


app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
```diff
@@ -678,7 +678,25 @@ public void setTitle(String title) {
...
+
+        int totalMinutes = (int) Math.ceil((double) runtime / 600000000);
+
+        int hours = totalMinutes / 60;
+        int minutes = totalMinutes % 60;
+
+        String runtimeString = "";
+
+        if (hours > 0) {
+            runtimeString += hours + getString(R.string.hours_shorthand) + " ";
```
nielsvanvelzen, line 694: Use string formatting instead so translators can place the digit behind the shorthand if their language needs that. Something like `%1$dh %1$dm` for the hour+minute string should work.

nielsvanvelzen (changes requested):

app/src/main/res/values/strings.xml
```diff
@@ -64,7 +64,8 @@
     <string name="lbl_watched">Watched</string>
     <string name="lbl_sd">SD</string>
     <string name="lbl_runs">"Runs"</string>
-    <string name="lbl_min">" min"</string>
+    <string name="minutes_shorthand">"m"</string>
+    <string name="hours_shorthand">"h"</string>
```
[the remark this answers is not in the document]
dtrexler: Done. New strings do not have quotes

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
dtrexler: Ah, I see. Thanks for the feedback. Changes completed as requested.

nielsvanvelzen (approved):
Thanks for the PR! I'll merge it once development for 0.19.z starts (somewhere in the coming days after 0.18.3 is released). No need to do any merges with master.

PR #4271, opened 2024-12-05
Always call notifyInteraction on navigation

### LINKED ISSUE #4262: "visited pages" opening behind screensaver - jellyfin remote session
### Describe the bug

1. Cast to your Jellyfin Session.
2. Wait for the Screensaver of the Session to activate
3. Browse arround the library on the Remote Session
--> normaly, the TV session would show what you are currently looking at, but its not because thats behind the screensaver
4. Use your TV Remote and pres "back" 
--> You will find out that all the pages you have viewed, where actualy opened and you now need to press "back" to get back to the player

Not a huge issue, but defenetly inconvinient.
Here is a Video - which is also used to show another issue.
Its clear to see that I needed to press "back" quite often to go back to that "player", even though we never saw the opened things
https://github.com/user-attachments/assets/257a7f7e-85e5-482b-a6ee-b5ed240fa70f

### Logs

_No response_

### Application version

0.18.0

Sideloaded APK

### Device information

Chromecast with Google TV HD

### Android version

Android 12

### Jellyfin server version

10.10.3

nielsvanvelzen:
Navigating does not cancel the screensaver when using remote control. This is not the intended behavior, it should always cancel the screensaver.

**Changes**
- Always call notifyInteraction on navigation

**Issues**
Fixes #4262

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/MainActivity.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/MainActivity.kt
@@ -111,6 +111,8 @@ class MainActivity : FragmentActivity() {
 	}
 
 	private fun handleNavigationAction(action: NavigationAction) {
+		screensaverViewModel.notifyInteraction(true)
+
 		when (action) {
 			// DestinationFragmentView actions
 			is NavigationAction.NavigateFragment -> binding.contentView.navigate(action)


crobibero (approved):

nielsvanvelzen:
Wanted to further test some edge-cases this could've introduced but those don't appear to exist so good to include in next .z release 👍

PR #4298, opened 2024-12-14
Fix watch next row ordering, Change watch next row meta data

hdweiss:
**Changes**
Various fixes to the "play next"/"watch next" TV row on the homescreen.

- episodes now have meta data in line with other streaming apps
- watch next is now correctly ordered
- movies to resume will now be inserted correctly

**Issues**
- #4340


nielsvanvelzen:
> watch next is now correctly ordered

What is wrong with the order? Is there an open issue for that?

hdweiss:
I have created issue #4340 to describe why this an issue for me

nielsvanvelzen:
I understand you want this behavior to be different but we'll have to comply with the guidelines[^1][^2][^3]. The engagement time must be set different for certain scenarios and the way it's implemented right now is how it must be. We cannot change that.

[^1]: https://developer.android.com/training/tv/discovery/guidelines-app-developers
[^2]: https://developer.android.com/training/tv/discovery/watch-next-programs
[^3]: https://developer.android.com/reference/androidx/tvprovider/media/tv/TvContractCompat.WatchNextPrograms#COLUMN_LAST_ENGAGEMENT_TIME_UTC_MILLIS()

hdweiss:
In the current implementation, the timestamp for WATCH_TYPE_WATCH_NEXT elements is never set correctly. It takes the creation timestamp of the episode, instead of the last engagement time of the last watched episode. If there currently was a way to retrieve the last engagement time for a show from the Jellyfin API, this could be solved more straightforward. I do agree that the other three watch types are ordered correctly.

My PR changes the following. Episodes are inserted a single time and will only be deleted once they are removed from the Jellyfin "up next"/"continue watching". Continue watching episodes have an egangement timestamp, which will be continued to be used. For the WATCH_TYPE_WATCH_NEXT episodes, we must find the engagement timestamp of the last episode watched and not the current episodes engangement timestamp. I did not find a clean way to retrieve the last engangement timestamp on a TV show via the API, even though I did discover that this information is available. If there was an easy way to retrieve this via the API, I have missed it, apologies.

https://github.com/jellyfin/jellyfin/blob/5774b601f59762fb3a02a1555d2aed5803ac6693/Emby.Server.Implementations/Data/SqliteItemRepository.cs#L2864

When a WATCH_TYPE_WATCH_NEXT episode is missing from the row, it must mean that the egangement with a tv show must have happened after last time the LeanbackChannelWorker had been invoked. Because I did not have the ability to retrieve the exact engagement timestamp, I default to the current time/date. This will put all new WATCH_TYPE_WATCH_NEXT in the front of the row. Once a WATCH_TYPE_WATCH_NEXT episode is inserted, it will keep that timestamp indefinitely. I understand that strictly speaking the "correct" timestamp is somewhere in between the last time the LeanbackWorker was called and the current date/time, but the current behaviour never leads to any incorrect ordering in accordance to the Android TV guidelines.

I have been daily driving my branch for more than one week and it always leads to the desired behaviour, which is also in line with the guidelines. This means that upon finishing an episode, the new WATCH_TYPE_WATCH_NEXT episode will be added to the front of the watch next row, not the very back.

hdweiss:
I did change the WATCH_NEXT_TYPE_NEW to current date/time, which is incorrect. This should be the release date of the episode. I have pushed that change to the PR branch.

hdweiss:
@nielsvanvelzen could you please take a look at what I wrote above?

nielsvanvelzen:
I don't think we should ever "make up" timestamps for the "watch next".

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -425,39 +465,45 @@ class LeanbackChannelWorker(
...
-
 			when {
 				// User has started playing the episode
 				(item.userData?.playbackPositionTicks ?: 0) > 0 -> {
 					setWatchNextType(WatchNextPrograms.WATCH_NEXT_TYPE_CONTINUE)
 					setLastPlaybackPositionMillis(item.userData!!.playbackPositionTicks.ticks.inWholeMilliseconds.toInt())
 					// Use last played date to prioritize
-					engagement = item.userData?.lastPlayedDate
+
+					setLastEngagementTimeUtcMillis(item.userData?.lastPlayedDate?.atZone(ZoneId.systemDefault())?.toInstant()?.toEpochMilli()
```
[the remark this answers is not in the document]
hewe-saxo: This one had been bothering me as well.

Bravia TV: getprop persist.sys.timezone -> returns Europe/Copenhagen
Jellyfin running in docker: echo $TZ -> returns Europe/Copenhagen

I have done some extensive testing, and doing the conversion to the local timezone (Europe/Copenhagen) is the only way I can get the correct timestamp on my end. If I don't do this, the jellyfin watch next item is offset by one hour compared with all the other streaming services' watch next items. Is there anywhere in the android client or the jellyfin server i can check for correct timezones?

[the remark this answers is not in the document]
hdweiss: This one had been bothering me as well.

Bravia TV: getprop persist.sys.timezone -> returns Europe/Copenhagen
Jellyfin running in docker: echo $TZ -> returns Europe/Copenhagen

I have done some extensive testing, and doing the conversion to the local timezone (Europe/Copenhagen) is the only way I can get the correct timestamp on my end. If I don't do this, the jellyfin watch next item is offset by one hour compared with all the other streaming services' watch next items. Is there anywhere in the android client or the jellyfin server i can check for correct timezones?

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -469,10 +474,16 @@ class LeanbackChannelWorker(
...
-				setEpisodeNumber(item.indexNumber ?: 0)
-				setSeasonNumber(item.parentIndexNumber ?: 0)
+
+				if ((item.indexNumber ?: 0) > 0)
+					setEpisodeNumber(item.indexNumber!!)
+
+				if ((item.parentIndexNumber ?: 0) > 0)
+					setSeasonNumber(item.parentIndexNumber!!)
+			}
+			else {
```
nielsvanvelzen, line 484:
```suggestion
			} else {
```

app/src/main/java/org/jellyfin/androidtv/integration/LeanbackChannelWorker.kt
```diff
@@ -469,10 +474,16 @@ class LeanbackChannelWorker(
 			if (item.seriesName != null) {
 				setTitle(item.seriesName)
 				setEpisodeTitle(item.name)
-				setEpisodeNumber(item.indexNumber ?: 0)
-				setSeasonNumber(item.parentIndexNumber ?: 0)
+
+				if ((item.indexNumber ?: 0) > 0)
+					setEpisodeNumber(item.indexNumber!!)
```
nielsvanvelzen, line 479: This could be written as

```kt
item.indexNumber?.takeIf { it > 0 }?.let { setEpisodeNumber(it) }
```

Same for the parentIndexNumber below

nielsvanvelzen: Also similar if statements on line 377-380

nielsvanvelzen (changes requested):

PR #4322, opened 2024-12-20
Fix UI audio queue order when shuffling

nielsvanvelzen:
When shuffling the queue stays the same data structure and the playback order is used to return items based on the shuffled indices. The updateAdapter function used the internal list of the queue supplier so playback ordering was never used, causing the wrong items to display during shuffle.

**Changes**
- Use the actual queue to provide items instead, limited to the next 100 items (including current)
- Refresh adapter when playback order changes
- Compare items by id (performance gain)

**Issues**


BotBlake:
I love you Niels 🗿 🫶

thornbill (approved):

Issue #3139, opened 2023-10-24
Next Up on homescreen not updating correctly


Harteex:
### Describe the bug

Next Up in the home screen (Android TV Launcher) doesn't show the correct episode after you've watched it.

1. Open Jellyfin and play an episode of a series you're watching, for example ep3.
2. Back out of the video when the credits roll, ensure it says Watched is checked.
3. Back out to the main screen of the jellyfin app, ensure it says ep4 is next.
4. Now back out to the home screen
5. Here you'll see that Next up in the home screen shows the episode you just watched, ep3.
6. Try restarting your device, no difference

To fix it, you have to open Jellyfin again, and exit it. Now it will show the correct episode on the home screen, ep4.

### Application version

0.15.12

### Where did you install the app from?

Google Play

### Device information

Nvidia Shield TV (2019)

### Android version

Android 11

### Jellyfin server version

10.8.9

nielsvanvelzen:
The channels in the Android launcher are updates every 60 minutes and on app start. Up-to-date information is available in the app.
We're working on making the app "smarter" and having it know when items changed, when this is in a more mature stage we should be able to update the channels in the launcher when needed.

Harteex:
Seems confirmed above, and I've not seen anything addressing the issue yet.

GlassedSilver:
> The channels in the Android launcher are updates every 60 minutes and on app start. Up-to-date information is available in the app. We're working on making the app "smarter" and having it know when items changed, when this is in a more mature stage we should be able to update the channels in the launcher when needed.

Is there an issue for this so we can track it?

nielsvanvelzen:
This issue

GlassedSilver:
> This issue

wow, massive derp right there haha

GlassedSilver:
Still very much an issue and not an edge case either. This should affect basically everyone who uses Next Up.

allmazz:
+1

PR #4347, opened 2024-12-27
Run LeanbackHelper onStop() to refresh home screen

hdweiss:
**Changes**
Run LeanbackHelper onPause() to refresh home screen. When the app is paused, it is probably because the user navigated to the home screen.

**Issues**
#3139

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/MainActivity.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/MainActivity.kt
@@ -101,14 +101,14 @@ class MainActivity : FragmentActivity() {
 	override fun onPause() {
 		super.onPause()
 
-		val workManager by inject<WorkManager>()
-		workManager.enqueue(OneTimeWorkRequestBuilder<LeanbackChannelWorker>().build())
-
 		screensaverViewModel.activityPaused = true
 	}
 
 	override fun onStop() {
 		super.onStop()
+
+		val workManager by inject<WorkManager>()
+		workManager.enqueue(OneTimeWorkRequestBuilder<LeanbackChannelWorker>().build())
 
 		lifecycleScope.launch {
 			Timber.d("MainActivity stopped")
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
This is not a proper solution to the linked issue. The MainActivity will get paused a lot more often, like when playing a video or when opening the preferences. The worker code does quite a lot of API calls so we shouldn't spam it.

hdweiss:
That is a good point. Is there any other hook we could use or is there a way to throttle the worker?

My only idea would be to move the worker call to onStop(), so that the home row either gets updated by the periodic worker within one hour or when the TV clears the app from memory. Would that be an acceptable solution?

nielsvanvelzen:
We could maybe add an additional function to update the "next up" row only once (video-)playback finishes. The other rows shouldn't change that often so I don't think anyone will notice if those update later.

hdweiss:
According to the Android guidelines, the refresh for TV shows should happen when 1. "The user exits the app on Android TV." and 2. "The user pauses or stops playback for a piece of content for more than 5 minutes."

I have tried moving the worker call into the onStop() callback of the MainActivity, which would satisfy the first point. I have tested this and it is working perfectly on my TV.

For the second point, would this be done by listening on the PlaybackControllerNotifiable:onCompletion event?

hdweiss:
@nielsvanvelzen any chance this can get merged? While not solving the issue 100%, this small change works well enough and in practice solves it for most users.

app/src/main/java/org/jellyfin/androidtv/ui/browsing/MainActivity.kt
```diff
@@ -104,6 +107,9 @@ class MainActivity : FragmentActivity() {
 	override fun onStop() {
 		super.onStop()
 
+		val workManager by inject<WorkManager>()
```
nielsvanvelzen, line 110: Use a class property instead

nielsvanvelzen (changes requested):
I'm fine with adding it to onStop. Please also update the PR title.

hdweiss:
@nielsvanvelzen changed title and moved worker into class property

nielsvanvelzen (approved):

PR #4411, opened 2025-01-18
Fix codec keys in new playback module

dtrexler:
Fixes inconsistencies between the reported/expected codec values in the mime type maps of the new playback module. This was causing unnecessary transcoding when using the experimental playback module

Ex. The server reports ```hevc```. The client expected ```h265```, now expects ```hevc```

**Changes**
Expected codec names now match what is reported by the server


nielsvanvelzen (approved):

Hukuma1:
Is this why my Shield was all-of-a-sudden transcoding stuff for some reason potentially? (all content is x265)

PR #4422, opened 2025-01-26
Fixed RTL languages punctuation mark issue

### LINKED ISSUE #4296: Hebrew subtitles don't show up correctly on the latest version (0.18.2) on TV.
### Describe the bug

Hey.
On my TV, Jellyfin updates automatically, and I've been told that recent updates (0.18) on TV has been changed a lot in regards to subtitles.

My issue is, with the latest version (0.18.2) on my TV (Sony A90K, Google TV OS) is that Hebrew subtitles show up incorrectly.
The period, or any punctuation, for that matter... show up wrong.
So instead of showing up correctly as:
.אוכל

It shows up as:
אוכל.

FYI, on my PC (windows 10), it shows up just fine (double clicking the Jellyfin icon from the task bar, on Chrome. Not via the "Jellyfin Media Player" thingy). But on my TV, it's wrong.

I've downgraded to version 0.17.9 from the f-droid link, the subtitles show up just fine now.

It seems like something in the latest versions was changed and now it's all messed up.

Do be aware though, that the Hebrew letters themselves do show up just fine. It's only the inverted punctuation.

### Logs

_No response_

### Application version

0.18.2

Google Play

### Device information

Sony A90K OLED, XR-42A90K

### Android version

Android version 10, Google TV OS version I'm not sure how to check...

### Jellyfin server version

10.10.3, doesn't matter probably since (I'm guessing) the version of the Jellyfin TV that is messing up with the subtitles, not the server version.

orenskl:
**Changes**
Add the unicode LTRM character for subtitles

**Issues**
Fixed #4296

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
@@ -26,6 +26,8 @@ import androidx.media3.common.TrackGroup;
 import androidx.media3.common.TrackSelectionOverride;
 import androidx.media3.common.TrackSelectionParameters;
 import androidx.media3.common.Tracks;
+import androidx.media3.common.text.Cue;
+import androidx.media3.common.text.CueGroup;
 import androidx.media3.common.util.UnstableApi;
 import androidx.media3.datasource.DefaultDataSource;
 import androidx.media3.datasource.DefaultHttpDataSource;
@@ -111,6 +113,23 @@ public class VideoManager {
                 Timber.e("***** Got error from player");
                 if (mPlaybackControllerNotifiable != null) mPlaybackControllerNotifiable.onError();
                 stopProgressLoop();
+            }
+
+            @Override
+            public void onCues(@NonNull CueGroup cueGroup) {
+                List<Cue> modifiedCues = new ArrayList<>();
+                final String LTRM = "\u200E";
+                for (Cue cue : cueGroup.cues) {
+                    CharSequence modifiedText =  LTRM + cueGroup.cues.get(0).text;
+                    Cue modifiedCue = new Cue.Builder()
+                            .setText(modifiedText)
+                            .setPosition(cue.position)
+                            .setLine(cue.line, cue.lineType)
+                            .build();
+                    modifiedCues.add(modifiedCue);
+                }
+                CueGroup cg = new CueGroup(modifiedCues, cueGroup.presentationTimeUs);
+                mExoPlayerView.getSubtitleView().setCues(cg.cues);
             }
 
             @Override


nielsvanvelzen (changes requested):
While I'm sure this fixes the LTR issues, I'm not sure if this is a proper fix for the underlying issue. Modifying each line to force unicode directional marks into them seems like a hack we should not need to do, and it may even introduce side-effects.

orenskl:
I agree this is a hack/patch for this issue, while I did tested this with both LTR and RTL subtitles and it seems ok.
Subtitle direction should be handled correctly in the `SubtitleView` but I think there is a bug or an issue in the `media3` library so I guess we will have to wait for it to be fixed.

itsrn:
I tried your method, however it only fixed the RTL of the first subtitle line. Moreover, it sometimes mixed together English letters or numbers inside RTL subtitle lines.  For this reason, I decided to modify your code, which works only with Hebrew, but is easily adaptable to other RTL languages.:

```java
@Override
public void onCues(@NonNull CueGroup cueGroup) {
  List < Cue > modifiedCues = new ArrayList < > ();
  final String LTRM = "\u200E"; // Left-to-Right Mark

  for (Cue cue: cueGroup.cues) {
    CharSequence originalText = cue.text;

    if (originalText == null) {
      modifiedCues.add(cue);
      continue;
    }

    boolean containsHebrew = false;
    String textStr = originalText.toString();
    for (int i = 0; i < textStr.length(); i++) {
      char c = textStr.charAt(i);
      if ((c >= 0x0590 && c <= 0x05FF) || // Hebrew range
        (c >= 0xFB1D && c <= 0xFB4F)) { // Hebrew presentation forms
        containsHebrew = true;
        break;
      }
    }

    CharSequence modifiedText;
    if (containsHebrew) {
      String[] lines = textStr.split("\n");
      StringBuilder ltrText = new StringBuilder();

      for (int i = 0; i < lines.length; i++) {
        if (i > 0) ltrText.append("\n");
        ltrText.append(LTRM).append(lines[i]);
      }

      modifiedText = ltrText.toString();
    } else {
      modifiedText = originalText;
    }

    Cue.Builder builder = new Cue.Builder()
      .setText(modifiedText)
      .setPosition(cue.position)
      .setLine(cue.line, cue.lineType);

    if (cue.size != Cue.DIMEN_UNSET) {
      builder.setSize(cue.size);
    }
    if (cue.textAlignment != null) {
      builder.setTextAlignment(cue.textAlignment);
    }

    modifiedCues.add(builder.build());
  }

  CueGroup cg = new CueGroup(modifiedCues, cueGroup.presentationTimeUs);
  mExoPlayerView.getSubtitleView().setCues(cg.cues);
}
```

This is ideal for someone looking for a temporary solution (although it may cause additional complications, as previously indicated, even though I couldn't find any with Hebrew and English subtitles).

nielsvanvelzen:
Closing as stale and unwanted. We should not hack around the subtitle rendering code of an upstream library, it should be fixed in the media3 code instead.

PR #4423, opened 2025-01-26
Subtitles background opacity

OwOchle:
**Changes**
Option to configure subtitles background opacity and function to apply it onto existing colors.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -171,11 +171,6 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		var subtitlesBackgroundColor = longPreference("subtitles_background_color", 0x00FFFFFF)
 
 		/**
-		 * Subtitles background opacity
-		 */
-		var subtitlesBackgroundOpacity = floatPreference("subtitles_background_opacity", 1f)
-
-		/**
 		 * Subtitles foreground color
 		 */
 		var subtitlesTextColor = longPreference("subtitles_text_color", 0xFFFFFFFF)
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -170,6 +170,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 */
 		var subtitlesBackgroundColor = longPreference("subtitles_background_color", 0x00FFFFFF)
 
+		/**
+		 * Subtitles background opacity
+		 */
+		var subtitlesBackgroundOpacity = floatPreference("subtitles_background_opacity", 1f)
```
nielsvanvelzen, line 176: The subtitlesBackgroundColor color already includes transparency so there is no need for a new preference.

OwOchle: Yes but since preset colors have `FF` as alpha, it's easier to have opacity as a separate config. Otherwise, changing color would reset your opacity to 100%. Changing the color selector to include opacity would require a lot of rewrite, which I can do if you think it's a better way to do it. Just tell me

OwOchle: Alright, I'll do that then. I switch this PR to Draft while I work on that

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/preference/screen/PlaybackPreferencesScreen.kt
```diff
@@ -1,5 +1,6 @@
 package org.jellyfin.androidtv.ui.preference.screen
 
+import okhttp3.internal.toHexString
```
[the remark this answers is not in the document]
OwOchle: Oh sorry, I thought it was the kotlin stdlib function

nielsvanvelzen (approved):
I'm not really satisfied with the color conversions needed for the preferences but that's a limitation of the current `colorList()` function so I'll work on that separately. Otherwise the changes here look fine, thanks!

PR #4438, opened 2025-02-04
Decouple external video playback from internal playback code

nielsvanvelzen:
This is an older branch and I'm not entirely sure what I didn't finish 🙈. Needs testing (@BotBlake, @TimGels). I need this for a different change I'm working on, that is needed for yet another change I want to work on 🗿.

**Changes**
- Rewrite ExternalPlayerActivity
  - Detached from other playback code (except for the queue)
  - Cleaner code, removed a bunch of cruft
  - Now uses Kotlin SDK
  - **always** directplays the file

**Issues**


TimGels:
@nielsvanvelzen which external player needs to be tested? I have a few that I can test like vlc and mx player.

nielsvanvelzen:
Just in general to see if the Jellyfin app still behaves like it should, you can test that with whatever external player you want.

BotBlake:
From playing arround a bit, I didnt find direct Playback Issues.
(Tested on `CCwGTV HD` using `MXPlayer`)

All issues I found:
- autoplay of the next Episode does not work.
- VLC support is back? - Using VLC as external player not working on stable but on this branch

- ~On Some files AC3 Audio doesnt play~ (also an issue on stable)
- ~Client Ignores External Player Setting when using Playlist~ (also an issue on stable)
- ~Shuffle all episodes completely breaks Seeking~ (also an issue on stable)
- ~I forgot what I wanted to write because I fell asleep~ (a me issue)

nielsvanvelzen:
I've now re-implemented support for external subtitles & queue advancing again. I believe this should now have feature parity with the previous implementation. Let me know if you find any incorrect behavior compared to the stable version of the app.

theguymadmax:
I'm still encountering a few of the issues I mentioned earlier:  
1. VLC is not saving the resume position (this worked in the stable version).  
2. When there's a saved position, both Just Player and VLC ignore it and start playback from the beginning (this also worked in the stable version).

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
```diff
@@ -0,0 +1,214 @@
...
+			finish()
+			return
+		}
+
+		val (item, mediaSource) = currentItem!!
+		val extras = result?.extras
+
+		val endPosition = if (extras == null) null else resultPositionExtras
+			.firstOrNull { extra -> extras.containsKey(extra) }
+			?.let { extra -> extras.getInt(extra, 0).milliseconds }
```
theguymadmax, line 184: Accounting for VLC's long position should fix the issue of not saving the resume position

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
```diff
@@ -154,7 +158,7 @@ class ExternalPlayerActivity : FragmentActivity() {
 			putExtra(API_VLC_FROM_START, true)
```
theguymadmax, line 158: This is overriding the position and causing VLC to always start from the start. You could conditionally set it if there is a resume position

nielsvanvelzen:
Thanks for the testing @theguymadmax, should now be able to resume and save position with VLC.

theguymadmax:
I got around to testing the queue:
For Just Player, it doesn't work—once the episode ends, it just returns to the details page (works in stable).
For VLC, it works. The episode ends and the next one starts automatically.

VLC - there's an issue with episodes not starting. On two different series, new episodes stop right after starting. I can't reliably replicate this issue.

crobibero (approved):

nielsvanvelzen:
I feel like the state of the PR is now good enough to merge. We have feature parity with the old implementation, if we do encounter bugs please open an issue and it will be looked at.

PR #4459, opened 2025-02-14
Add option to always burn in subtitles when transcoding

### LINKED ISSUE #4399: subtitle out of sync when transcoding

### Describe the bug

1. video name: Squid.Game.S02E01.2021.2160p.NF.WEB-DL.DDP5.1.Atmos.DV.HDR.H.265-HHWEB.mkv
2. subtitle out of sync when transcoding play
3. the issue happens both on androidtv app and web.
4. On the web, if I set "Always burn in subtitle when transcoding", the issue solved. 
5. But androidtv app doesn't have this option，the issue can't be avoided.

### Logs

_No response_

### Application version

0.18.3

None

### Device information

Xiaomi TV，HK1 RBOX X4，X96X10pro

### Android version

Android9，Android 11

### Jellyfin server version

10.10.3

### Other sources

_No response_

GertSallaerts:
**Changes**
Adds support for toggling the `alwaysBurnInSubtitleWhenTranscoding` option that was introduced to Jellyfin in 10.10.

**Issues**
Fixes #4399

**Notes**
This is a first draft, partly to check whether or not you're amenable to adding this feature at all and partly to check if it's going in the right direction.

Things I know may need work:

- I have to double-check the whole `burningSubs` logic (PlaybackController and PlaybackControllerHelper) for different scenarios
- I was wondering if you could tell me whether or not the new playback code is far enough along for me to try adding it there as well of if it's something best left for later (or never).

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -245,6 +245,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		var assDirectPlay = booleanPreference("libass_enabled", false)
 
 		/**
+		 * Always burn in subtitles when transcoding.
+		 */
+		var subtitlesBurnDuringTranscode = booleanPreference("subtitles_burn_during_transcode", false)
+
+		/**
 		 * Enable PGS subtitle direct-play.
 		 */
 		var pgsDirectPlay = booleanPreference("pgs_enabled", true)
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -85,6 +85,23 @@ public class PlaybackController implements PlaybackControllerNotifiable {
     protected VideoOptions mCurrentOptions;
     private int mDefaultAudioIndex = -1;
     protected boolean burningSubs = false;
+
+    // The server does not update the subtitle delivery method when alwaysBurnInSubtitleWhenTranscoding
+    // is set, so we need to assume subs are burned in when transcoding with the option enabled.
+    protected boolean shouldBurnInSubtitles(PlayMethod playMethod) {
+        return userPreferences.getValue().get(UserPreferences.Companion.getSubtitlesBurnDuringTranscode())
+                && playMethod == PlayMethod.TRANSCODE;
+    }
+
+    private void updateBurningSubs(StreamInfo response) {
+        if (response.getSubtitleDeliveryMethod() == SubtitleDeliveryMethod.DROP) {
+            burningSubs = false;
+            return;
+        }
+
+        burningSubs = response.getSubtitleDeliveryMethod() == SubtitleDeliveryMethod.ENCODE
+                || shouldBurnInSubtitles(response.getPlayMethod());
+    }
     private float mRequestedPlaybackSpeed = -1.0f;
 
     private Runnable mReportLoop;
@@ -502,6 +519,7 @@ public class PlaybackController implements PlaybackControllerNotifiable {
         VideoOptions internalOptions = new VideoOptions();
         internalOptions.setItemId(item.getId());
         internalOptions.setMediaSources(item.getMediaSources());
+        internalOptions.setAlwaysBurnInSubtitleWhenTranscoding(userPreferences.getValue().get(UserPreferences.Companion.getSubtitlesBurnDuringTranscode()));
         if (playbackRetries > 0 || (isLiveTv && !directStreamLiveTv)) internalOptions.setEnableDirectPlay(false);
         if (playbackRetries > 1) internalOptions.setEnableDirectStream(false);
         if (mCurrentOptions != null) {
@@ -564,7 +582,7 @@ public class PlaybackController implements PlaybackControllerNotifiable {
                     if (mVideoManager == null)
                         return;
                     mCurrentOptions = internalOptions;
-                    if (internalOptions.getSubtitleStreamIndex() == null) burningSubs = internalResponse.getSubtitleDeliveryMethod() == SubtitleDeliveryMethod.ENCODE;
+                    updateBurningSubs(internalResponse);
                     startItem(item, position, internalResponse);
                 }
 
@@ -995,6 +1013,9 @@ public class PlaybackController implements PlaybackControllerNotifiable {
                 public void onResponse(StreamInfo response) {
                     if (!isActive()) return;
                     mCurrentStreamInfo = response;
+
+                    updateBurningSubs(response);
+
                     if (mVideoManager != null) {
                         mVideoManager.setMediaStreamInfo(api.getValue(), response);
                         mVideoManager.start();
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackManager.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackManager.kt
@@ -111,6 +111,7 @@ class PlaybackManager(
 					allowVideoStreamCopy = true,
 					allowAudioStreamCopy = true,
 					autoOpenLiveStream = true,
+					alwaysBurnInSubtitleWhenTranscoding = options.alwaysBurnInSubtitleWhenTranscoding,
 				)
 			).content
 		}
--- a/app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
@@ -195,6 +195,17 @@ fun SettingsPlaybackAdvancedScreen() {
 			)
 		}
 
+		item {
+			var subtitlesBurnDuringTranscode by rememberPreference(userPreferences, UserPreferences.subtitlesBurnDuringTranscode)
+
+			ListButton(
+				headingContent = { Text(stringResource(R.string.pref_burn_subtitles_when_transcoding)) },
+				captionContent = { Text(stringResource(R.string.pref_burn_subtitles_when_transcoding_description)) },
+				trailingContent = { Checkbox(checked = subtitlesBurnDuringTranscode) },
+				onClick = { subtitlesBurnDuringTranscode = !subtitlesBurnDuringTranscode }
+			)
+		}
+
 		item { ListSection(headingContent = { Text(stringResource(R.string.pref_live_tv_cat)) }) }
 
 		item {
--- a/app/src/main/java/org/jellyfin/androidtv/data/compat/VideoOptions.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/compat/VideoOptions.kt
@@ -3,4 +3,5 @@ package org.jellyfin.androidtv.data.compat
 class VideoOptions : AudioOptions() {
 	var audioStreamIndex: Int? = null
 	var subtitleStreamIndex: Int? = null
+	var alwaysBurnInSubtitleWhenTranscoding: Boolean = false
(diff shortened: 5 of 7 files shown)


nielsvanvelzen:
Personally I dislike this feature because you'll lose all customization on subtitles when they are burned into the video. I'd rather look into what causes the subtitles to be out-of-sync. There's a possibility (although small) that it is already fixed in the master branch after #4455 was merged.
If we have a better understanding of what causes the sync issues and determine it is not (easily) fixable we can always look into the burning option again.

gnattu:
It is indeed a problem that is hard to track down. Neither Emby nor Plex provided full solution to this yet. The transcoding is a very complex process and the time codes can change during that process, and sometimes it would be larger enough to make the original subtitles to look very off with the original time codes. You might even notice that the transcoded video can have a different video duration that is longer/shorter than the original one, and sometimes even the actual fps is slightly modified (like 29.97 being speeded up to 30). There are too many reasons that the time code might being altered and that is why I implemented that brute force hack as a workaround.

Is it bad? Yes. And the way it is implemented is also very hacky. Because we don't have a good API to determine the actual playback method before the playback has started (the one returns before the playback does not differ transcode and remux and both will be reported as remux), the is transcoding is detected after the playback has started using the api to check for playback stats, and then determine should the client load the subtitle or not to prevent double-subtitles.

johnpc:
Allowing burn-in subtitles when transcoding would also address this problem: https://github.com/jellyfin/jellyfin-androidtv/pull/4522

PilotFlying:
The option would be useful even when transcoding is for audio only (TV app: subtitles will go out of sync if you are downmixing audio and you interrupt things by seeking, in either direction). 

Fixing the root issues would be preferrable but seems difficult and, until then, having the option to force burn-in would really help

ppp7032:
Any progress on this being added? It seems discussion on this is now dead despite the fact that this feature would make jellyfin leagues more usable on androidtv imo.

GertSallaerts:
Hey @nielsvanvelzen, this PR has a merge conflict and I'd rather not rebase if the answer is going to be no, so I'm checking in.

I've done some more testing and research since we last spoke:

- **Still reproducible**: I can confirm the subtitle desync is still present on the latest stable (server 10.11.6, Android TV 0.19.7) as well as on the latest unstable server build. External SRT subtitles desync after seeking during transcoding. Burn-in fixes it every time.
- I was hopeful it was fixed on jellyfin's main branch when I saw these, which is why I tested unstable too. (jellyfin/jellyfin#15845 identified `-noaccurate_seek` as a cause and PR jellyfin/jellyfin#15926 removes it during transcodes.) However, even on unstable with that fix applied, the desync still occurs, confirming @gnattu's earlier point that there are likely multiple causes for the timecode drift.
- I saw the Roku client shipped this option too, if that matters (https://github.com/jellyfin/jellyfin-roku/pull/511)
- It doesn't sound like an upstream fix is on the horizon right now (going from previous comments here)

This is an opt-in toggle. Users who don't need it won't be affected, and users who do currently have no way to get working subtitles during transcoding on Android TV.

Are you open to merging this, or should I close it?

nielsvanvelzen:
Ok after reconsidering I'm still not the biggest fan of this as it will prevent the subtitle customization from working. But I also understand that some people would be fine with that trade off when it "solves" the synchronization issue. Therefor I think it's fine to add a setting for this in the advanced playback preferences with at least a heads up about it ignoring the customization options in the description.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -231,6 +231,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 */
 		var assDirectPlay = booleanPreference("libass_enabled", false)
 
+		/**
+		 * Always burn in subtitles when transcoding.
+		 */
+		var alwaysBurnInSubtitleWhenTranscoding = booleanPreference("pref_burn_subtitles_when_transcoding", false)
```
nielsvanvelzen: We don't use the `pref_` prefix anymore for new preferences. I also think it would be best to start this preference name with "subtitles" for consistency. e.g.


```suggestion
		var subtitlesBurnDuringTranscode = booleanPreference("subtitles_burn_during_transcode", false)
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -976,6 +983,9 @@ public void seek(long pos, boolean skipToNext) {
                 public void onResponse(StreamInfo response) {
                     if (!isActive()) return;
                     mCurrentStreamInfo = response;
+
+                    updateBurningSubs(response);
```
nielsvanvelzen, line 1017: Was there a reason to also update the burningSubs variable from this place now?

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -564,7 +571,7 @@ public void onResponse(StreamInfo internalResponse) {
                     if (mVideoManager == null)
                         return;
                     mCurrentOptions = internalOptions;
-                    if (internalOptions.getSubtitleStreamIndex() == null) burningSubs = internalResponse.getSubtitleDeliveryMethod() == SubtitleDeliveryMethod.ENCODE;
+                    updateBurningSubs(internalResponse);
```
nielsvanvelzen, line 585: Was there a reason the previous if statement is removed here?

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackManager.kt
```diff
@@ -111,6 +113,7 @@ class PlaybackManager(
 					allowVideoStreamCopy = true,
 					allowAudioStreamCopy = true,
 					autoOpenLiveStream = true,
+					alwaysBurnInSubtitleWhenTranscoding = userPreferences[UserPreferences.alwaysBurnInSubtitleWhenTranscoding],
```
nielsvanvelzen: The preference should be communicated via VideoOptions here. similar to the other options.

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
```diff
@@ -187,6 +187,17 @@ fun SettingsPlaybackAdvancedScreen() {
 			)
 		}
 
+		item {
+			var burnInSubtitles by rememberPreference(userPreferences, UserPreferences.alwaysBurnInSubtitleWhenTranscoding)
```
nielsvanvelzen: Name of the var should always be equal to the preference name

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
GertSallaerts: Being a bit too safe maybe. If I got it right, this path rebuilds the stream when seeking on a non-seekable transcoded stream. Added `updateBurningSubs` here for consistency with the initial play callback. Both receive a fresh StreamInfo response, so both recalculate. In practice the play method and subtitle delivery are unlikely to change during a seek.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
GertSallaerts: The original if statement only updated `burningSubs` when no subtitle was explicitly selected. This broke when switching bitrate mid-playback:

> 1. User plays with subtitles, `subtitleStreamIndex` is set (e.g. track 2)
> 2. User changes bitrate, `refreshStream()` calls `stop()` which resets `burningSubs = false`
> 3. `play()` rebuilds the stream, reusing `subtitleStreamIndex = 2` from the previous options
> 4. Server responds, possibly now transcoding with burned-in subs
> 5. If statement matches `subtitleStreamIndex != null`, skips the `burningSubs` update
> 6. `burningSubs` stays false, client renders its own subs on top of the burned-in ones, double subtitles

I replaced the if statement with an early return for `SubtitleDeliveryMethod.DROP`, which is what `StreamInfo.getSubtitleDeliveryMethod()` returns when no subtitle is active. This correctly skips the update when there are no subs, without blocking it when a subtitle is selected.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
GertSallaerts: Renamed.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackManager.kt
GertSallaerts: Done.

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
GertSallaerts: Done.

GertSallaerts (commented):
Sorry @nielsvanvelzen I wasn't done with my changes when you reviewed them, committed and pushed my pending code when I finished for the day. Ready now and took care of the comments you already left.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -85,6 +85,23 @@ public class PlaybackController implements PlaybackControllerNotifiable {
     protected VideoOptions mCurrentOptions;
     private int mDefaultAudioIndex = -1;
     protected boolean burningSubs = false;
+
+    // The server does not update the subtitle delivery method when alwaysBurnInSubtitleWhenTranscoding
+    // is set, so we need to assume subs are burned in when transcoding with the option enabled.
+    protected boolean shouldBurnInSubtitles(PlayMethod playMethod) {
```
nielsvanvelzen, line 91: I verified this and to me this feels like a server bug. @gnattu can you confirm if this was intended behavior or not?

GertSallaerts: I think it was intentional, going from [this PR](https://github.com/jellyfin/jellyfin/pull/12676) but I couldn't explain why exactly.

gnattu: Well this is very hacky and I think current implementation here is not "correct" in behavior.

Our streambuilder currently has a limitation that, the actual play method cannot be determined before the client requests the HLS stream (the m3u8 manifest).

This caused a problem that we can only determine if it’s direct playing from the API. However, we can only confirm if it’s actually transcoding or remuxing after the actual request to start ffmpeg by calling the HLS endpoint. Then, the server will perform a stream copy test to decide if transcoding or stream copy should be used.

Which means both playback method and subtitle delivery method returned from the get playback info API will return stale info, because the actual playback decision is made after that API call.

If I understand the code here correctly, it is also using the wrong method to tell if the stream is transcoded because, remuxing will also be reported as transcoding in this api.

Web used a super hack approach: it query playback sessions of current device and see current playback method after the playback has started, as that is (currently) the only api that will update its results after the playback started to reflect the real playback method.

I know this is a terrible design from an API perspective overall and I do agree we need to make a more consistent API. But that requires an even bigger refactor and was out of the scope when I introduced this workaround.

nielsvanvelzen:
@GertSallaerts if you can fix the merge conflicts we can go ahead and merge this.

GertSallaerts:
@nielsvanvelzen rebased on the master branch

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks for working on this!

PR #4483, opened 2025-02-25
Fix progress bar calculation

lukas-fichtner:
Hello there! This is my first Jellyfin PR – Feedback Welcome!

**Changes**
- Corrected condition to properly check if the current time is within the program's duration.
- Fixed progress percentage calculation using double division for accuracy.
- Replaced getPremiereDate() with getStartDate() for correct program start time.
- Refactored getRealTimeProgress() to ensure proper duration calculation.

**Issues**
Both Live TV progress bars are not working properly.
Couldn't find a matching issue, but this likely affects many. If one exists, please link it. Thanks!


nielsvanvelzen (approved):
Thanks! Seems like some of these slipped through in a migration I did some  time ago 😅

PR #4487, opened 2025-02-26
Use External Player Popup

### LINKED ISSUE #4261: Bring back per item External Player

Previous versions of the application has had an option to set each series/movie to a specific playback setting:
- internal player
- internal player with other codec
- external player

However, recently (about a few update cycles ago maybe) the option has been removed. It was always incredibly helpful to have that on the items selection page to quickly toggle. 

As it stands the process is as follows: go back to the jellyfin homepage > settings > playback > advanced playback > use external player then when I'm done with that show or what have you I have to go all the way back to toggle it off to make sure I keep things like playback times sync'd etc.

I'd like to humbly request it to come back to prod.

### LINKED ISSUE #4465: Add option to play via external player directly from movie/episode screen

### Problem description

Currently, multiple playback issues with HDR, DV, etc., on different platforms are reported here. In many cases, the temporary workaround is to use an external player. To do that, you need to go to Settings, find it, enable it, and return to the movie/episode screen again - which seems to be a lot of back and forth especially of only a few videos have that problem in your large media library.

I understand the technical complexity of fixing playback for all platforms or implementing "smart choice" based on capabilities, so that kinda "temporary" option might be a good alternative for a lot of users

### Proposed solution

I propose to add a button/option/whatever on the movie/episode screen to force use the external/system player.
Literally, and button "Play on Extrernal/System player" somewhere near button "Play" or "Continue from hh:ss".
Alternatively, it can be hidden under "three dots" menu or whatever 

### Alternatives considered

As an alternative solution, it might be to move the "Use external player" setting somewhere on the top level.

### Additional information

_No response_

### Other sources

_No response_

haydenkz:
**Changes**

Adds a long-click popup to the play button with the option to use an external player. 

**Issues**

Fixes #4465 
Fixes #4261


marinierb:
This would be awesome to have! Please merge!

nielsvanvelzen (changes requested):
The getPlaybackDestination function in PlaybackLauncher is supposed to choose the destination. Adding a new getExternalPlayer function is against that concept.

I'm also not convinced we need this feature. If you want to use an external player you can configure that in the app settings.

SerhiyRomanov:
@nielsvanvelzen I've described why we would like to have that functionality here https://github.com/jellyfin/jellyfin-androidtv/issues/4465

kingp0dd:
> @nielsvanvelzen I've described why we would like to have that functionality here #4465

I for one agree with this. I want default player to be jellyfin but there are certain video formats that my TV can't play natively

nielsvanvelzen:
Closing as stale (author did not address review feedback or comment)

cch85:
Definitively something I would like to see again (pretty sure it did exist in the past), few another request are based on this one.
IMHO this has few use cases, i,e playing video not well supported by the default player, and external player would definitively help especially if Jellyfin does not have transcoding enabled.

cch85:
> PlaybackLauncher

Just curious what do you have in mind as alternative if adding a new getExternalPlayer function is against the concept of getPlaybackDestination function in PlaybackLauncher supposed to choose the destination?
Would you mind clarifying?

pavichokche:
This would be a super nice feature, because sometimes random outlier movies/episodes might not function with the built-in player and it can save SO MUCH time of going in/out of the advanced video player settings to toggle external player on and then off later. PLEASE consider it

IlIlIIlIlIlIIllIIlIlI:
We definitely **DO** need this feature, anyone who actually uses the TV app on a daily basis with complex media files will tell you how annoying it is to have to dig to the external player toggle over and over. If you put out a survey 80% of TV app users would say the feature is needed. 

I prefer to use the embedded played, but when I can't, I use the specific external player (Kodi) I prefer, but then have to remember to go turn it back off in order to be able to use the built in player. At least if the toggle wasn't 4 menu's deep it wouldn't be as bad, but it is that deep in the options.

BenMcLean:
> We definitely **DO** need this feature, anyone who actually uses the TV app on a daily basis with complex media files will tell you how annoying it is to have to dig to the external player toggle over and over. If you put out a survey 80% of TV app users would say the feature is needed.
> 
> I prefer to use the embedded played, but when I can't, I use the specific external player (Kodi) I prefer, but then have to remember to go turn it back off in order to be able to use the built in player. At least if the toggle wasn't 4 menu's deep it wouldn't be as bad, but it is that deep in the options.

He's obviously right. Screwing up all your other videos across the entire app just to get an alternative player one time for one specific video by navigating into a four-menus-deep obscure setting not just once, but ***twice*** in order to get all your other videos back to normal afterwards is not a solution. Choosing which player is fundamentally not a universal system-wide choice: it is an individual video choice.

PR #4488, opened 2025-02-27
Add preference to show date and time

### LINKED ISSUE #3254: Display date next to time

I think it would be helpful to see the date next to the time in the upper right hand corner of the screen. Maybe an option to toggle it on/off and date options? Not sure if anyone would be opposed to this, and it seems like it should be simple enough to implement. Although, I suppose it would complicate things for different countries with different date formats. Just a thought. Okay, thanks.

haydenzeller:
**Changes**
Added a show date option under show clock in preferences. I added a second TextClock to clock_user_bug and view_toolbar XML files using the date format. updated display logic accordingly.

**Issues**
fixes #3254


nielsvanvelzen:
What would be the use-case of showing the date in a media player? I don't really see how this feature makes sense.

PR #4492, opened 2025-02-28
Fix filtering not working in collections view

### LINKED ISSUE #4475: Filter "unplayed" in collections doesn't work

### Describe the bug

1. Go to the media library "collections" 
2. Enable the filter with the tick mark to show only (partially) unplayed collections
3. Bug: still all elements are shown. Even those with a tick mark
4. FYI: this happens since I use the app (several months). In the web app this filter works in collections. This kind of filter works in other libraries like movies and tv series works, though (also in Android TV).

### Logs

_No response_

### Application version

0.18.4

Sideloaded APK

### Device information

Zidoo Z9X 8K

### Android version

Android 11

### Jellyfin server version

10.10.6

### Other sources

_No response_

nielsvanvelzen:
Bit of old code that was migrated to the SDK with exact same behavior. We shouldn't need to apply additional constrictions when requesting collections, API request now matches more similarly with jellyfin-web.

**Changes**
- Fix filtering not working in collections view

**Issues**
Fixes #4475

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowsingUtils.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowsingUtils.kt
@@ -288,7 +288,7 @@ object BrowsingUtils {
 			parentId = parent.id,
 		)
 
-		if (parent.type == BaseItemKind.USER_VIEW || parent.type == BaseItemKind.COLLECTION_FOLDER) {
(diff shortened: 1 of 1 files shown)


thornbill (approved):

nielsvanvelzen:
This fix introduced some regression (reported by @BotBlake) so reverting now and opening another fix.

PR #4502, opened 2025-03-05
Clientside Support for Online Subtitles featuring OpenSubtitles.com and SUBDL, and also subtitle delay option for these online subtitles

ahmethascelik:
Hi, all this is my first contribution to an opensource project, so I need your patient and support, thanks.

**Changes**

Seamless Online External Subtitle support for SubDL and Opensubtitles.com:
- When user hits play button, rest calls are made automatically with configured subtitle sources, and just in second, those results which are fetched from Subtitle Apis are added to CC list. So users can select subtitles from the UI that they are familiar with. 

- SubDL does not requires authentication, but in any case I added to change API KEY option to Settings. Users can also manage to change their preferred languages from settings.
- Opensubtitles requires authentication, so users can login by visiting Settings. Until then subtitle search won't be available for opensubtitles. Users can also change their preferred languages, also in a case of deprecation of api token, they can enter an optional one. 

- When an user selects and online subtitle, It will automatically download and loads to Exoplayer by design.
- If selected subtitle is an online subtitle user also can align timings by 2 custom actions, show subtitle earlier and show subtitle later buttons. 

- It's easy to add more online subtitle source because of interface design. 
- I didn't add a custom http lib, I used default one : Okhttp . 
- I also added default api keys for this 2 subtitle sources (Opensubtitles and SubDL), I don't pay, they are free but I don't know if they have limits. 

cc
onlinesubsprefs
onlinesubs
cctiming


ahmethascelik:
Hi,

I totally understand your concerns, It would be hard to maintain for every client, I see. But first, let me explain my scenario of my problem. 

I don't have access to server which is managed by someone, and he shares his Jellyfin media with users from around the world. English is not my native language and I love watching media in original language with Turkish Subtitles. This concept was very popular among Plex users, which called PlexShare and Plex has support for users searching subtitles even Media server have not that subtitles via OpenSubtitles. https://support.plex.tv/articles/subtitle-search/ So As a user, I don't care if subtitle search is on client or on server, I just want that feature. Commercial client Infuse also have that feature. https://support.firecore.com/hc/en-us/articles/215090967-Using-Subtitles. Also Maybe in future, there would be an external subtitle selection from external drive and I can prepare my subtitles before movie night, like the have this feature in Infuse.

So first, I searched a solution by changing external media player, I tried all players including Nova Player, VLC, MX Player but I had no success. I don't know how media info is shared with external players, but when I use Nova Player, It search subtitles for name "Stream" and it was impossible to change that string, (It failed to show video by the way). Also there are some cons about using external player, so I think it would be a better solution to implement this as an integrated feature.

I developed it very loosely coupled, everything is around OnlineSubtitleHelper class. If something is wrong it is easy to remove.  So I say, let's try with analytics, If the feature is used heavily then we can implement it for other clients if we have demand. As you post that link,  Web client is already have similar pull request .

nielsvanvelzen:
Closing as unwanted, as mentioned by @thornbill this is something that should be implemented server-side. If the server then exposes a more abstract "subtitle search" API we can use that in the client. But we will **never** use third-party API's directly from the client.

PR #4508, opened 2025-03-07
Add codec features, color formats and profile levels to device profile report

nielsvanvelzen:
Extends the tool from #4482 with even more data.

**Changes**

- Add color formats
- Add profile levels
- Add device codec features
- Use collapsed detail blocks
- Updated formatting of current values to be easier to read
  - Ranges are now `45-70` instead of `[45,70]`
  - null values now hidden

Updated the gist with new example: https://gist.github.com/nielsvanvelzen/cba33b29c1dfc82350735fd5f0190abd

**Issues**


crobibero (approved):
That's pretty cool

Issue #1327, opened 2022-01-02
"Continue watching?" dialog similar to Netflix


hypervtechnics:
### Describe the feature you'd like

As an end user I want to have an option which activates a mechanism which pauses the playback of a show (!) after a configurable amount of time and/or episodes and asks me "Do you want to continue?" giving three options "I want to continue", "I want to continue and don't ask again" (until next start of JF App) and "Bring me to the menu" (essentially: No). During display of the dialog any wakelocks or similar mechnisms should be disabled to allow the system to go into stand-by mode (essentially turning off the tv) after a certain amount of time (primarily configured in the system settings).

Pentaphon:
@nielsvanvelzen this should be closed with the release of 0.19 beta 1.

PR #4389, opened 2025-01-07
feat(VideoManager): add "Continue watching" dialog

kinhelm:
First draft, feel free to suggest modifications

**Changes**
Adding a WatchTracker singleton to track watch time and watched episodes. It resets counts on user interaction

**Issues**
Request of #1327

PR #4509, opened 2025-03-09
Add still watching feature

ConnorS1110:
**Changes**
Monitor how many episodes have been watched in a row with no remote interaction. Default behavior is if 3 episodes have been watched, or there was 90 minutes of watch time, with no interaction, a timer starts when watching the 4th episode. If timer runs out with no interaction, "Are you still watching prompt" appears. Another countdown starts and a progress bar fills the "No" button. When countdown ends, and a choice hasn't been selected playback ends. There are a series of user customizable options to configure the conditions the screen will trigger, or you can disable it all together.

**Issues**
Adds a [feature](https://features.jellyfin.org/posts/814/are-you-still-watching-prompt) that has been requested for 5 years now
Also requested in #1327

## Summary by CodeRabbit

- **New Features**
  - Introduced a "Still Watching" screen that prompts users to confirm continued viewing after a configurable period of inactivity or episode count.
  - Added user preference settings to customize the "Still Watching" behavior, including various timeout durations and a disable option.
  - Enhanced playback experience with a new overlay and timer-based confirmation dialog during binge-watching sessions.

- **Improvements**
  - Replaced the previous screensaver logic with an improved interaction tracker for more accurate user activity monitoring.
  - Updated preference screens and localization to support the new inactivity timeout options.

- **Bug Fixes**
  - Improved focus handling and navigation in the new "Still Watching" overlay to ensure smooth TV remote interactions.


ConnorS1110:
I have this like 80% done I would say, but I have hit a roadblock I could use some help with. I have the correct logic in place that will display the are you still watching screen in the correct scenario, but I can't get the actual fragment for this screen to appear. Changing the `replace` flag on the navigate function makes no difference. Currently just a black screen appears when trying to display the fragment, there are no errors that occur either. I largely followed the pattern used by the `NextUpFragment` since this is really just another variant of that screen. The playback code in general is very difficult for me to make sense of so I am hoping someone sees something I don't

Dnkhatri:
This feature might not be accepted as I think it would be preferred for this to be in the jellyfin server. So that it can be implemented by all the different client devices

ConnorS1110:
What part are you envisioning is at the server level? The logic to decide if the still watching screen should be displayed? There still would need to be changes in each of the consuming clients to display the client specific implementation of the screen

Dnkhatri:
> What part are you envisioning is at the server level? The logic to decide if the still watching screen should be displayed? There still would need to be changes in each of the consuming clients to display the client specific implementation of the screen and client specific implementations to notify the server if there has been any any device specific inputs to reset the counter.

Yes the logic should be decided on the server once that is implemented it can be adopted for different devices/apps to give a consistent system. @nielsvanvelzen would probably be the guy to answer this. But as far as I can tell device specific features are unlikely to get implemented.

johnpc:
@Dnkhatri I don't think that's right - this specific feature resets the timer every time the android tv remote is used, which is not an event the server will ever be aware of. This feature definitely should be implemented on the client.

ConnorS1110:
We have discussed it on element/discord and the consensus is that there is no reason for this to be server side at all, except for user config settings to customize the conditions to trigger the screen. So this will largely stay as is

johnpc:
Have you seen https://github.com/jellyfin/jellyfin-androidtv/pull/4389? It seems to be seeking to solve the same problem.

I tried pulling both your solution and the one in https://github.com/jellyfin/jellyfin-androidtv/pull/4389 and neither seemed to work right though, and also seem to introduce other buggy behavior.

ConnorS1110:
@johnpc I hadn't seen that PR thanks for sharing. It does look like it hasn't been worked on in 2 months though, idk if the contributor is still working on it and just hasn't pushed their changes or what. It does look like it is doing what I was hoping to accomplish. I am still not done with this one, I am having issues with fragment management to actually get the screen to appear. From my testing, it does still properly track the condition to display the still watching screen, but displaying the fragment itself is proving to be more troublesome than I thought. If you experienced other issues I would be interested to know what those issues were

ConnorS1110:
A lot of work was done here. This was based off the work @kinhelm did in this [PR](https://github.com/jellyfin/jellyfin-androidtv/pull/4389). If/when this merges, they should also get credit for it as well. I did add 2 temp settings for this feature to make it easier to test. 1 to test the episode count condition, and 1 to test the min watch time condition.

johnpc:
Screenshot 2025-03-18 at 10 00 28 PM

It's beautiful!

app/src/main/java/org/jellyfin/androidtv/ui/playback/WatchTrackerViewModel.kt
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+
+	private fun checkPrompt() {
+		if (!::videoManager.isInitialized) return
+
+		val stillWatchingSetting = StillWatchingStates.getSetting(userPreferences[UserPreferences.stillWatchingEnabled].toString())
+
+		val currentEpisodeProgress = videoManager.currentPosition.toFloat() / videoManager.duration.toFloat()
+		val minMinutesInMs = stillWatchingSetting.minMinutes * MILLIS_PER_MIN
```
johnpc, line 77:
```suggestion
		val minMinutesInMs = stillWatchingSetting.minMinutes.toLong() * MILLIS_PER_MIN
```

app/src/main/java/org/jellyfin/androidtv/ui/playback/stillwatching/StillWatchingButtonsView.kt
```diff
@@ -0,0 +1,73 @@
...
+import android.util.AttributeSet
+import android.view.LayoutInflater
+import android.widget.FrameLayout
+import org.jellyfin.androidtv.databinding.FragmentStillWatchingButtonsBinding
+
+class StillWatchingButtonsView @JvmOverloads constructor(
+	context: Context,
+	attrs: AttributeSet? = null
+) : FrameLayout(context, attrs) {
+	private var countdownTimer: CountDownTimer? = null
```
johnpc, line 14: How long does the countdown timer last by default in other in other platforms like Netflix? It was jarring when it returned me to the Home Screen - I would have rather let it just sit there forever on "Are you still watching?". I could imagine myself falling asleep watching TV, then waking up and clicking "yep continue actually" even though it would have been paused for a long time by the time I wake up.

In that case I suppose I would be suggesting to remove the count down timer altogether. This is just my own design preference though so if the standards and expectations in other apps prefer a countdown, it's probably better to stick with whatever conforms best to user expectations.

johnpc (commented):
This feature is awesome - I hope it gets included in the next release.

app/src/main/java/org/jellyfin/androidtv/ui/playback/WatchTrackerViewModel.kt
johnpc: I had to make this change in order for the build to succeed on my machine.

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -161,6 +164,8 @@ public void onCreate(Bundle savedInstanceState) {
 
         playbackControllerContainer.getValue().setPlaybackController(new PlaybackController(mItemsToPlay, this, mediaPosition));
 
+        playbackController = playbackControllerContainer.getValue().getPlaybackController();
```
johnpc, line 167: This is well intentioned but it makes the diff a lot harder to read. Is there any change you intended to make to this file?

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -135,9 +135,12 @@ public class CustomPlaybackOverlayFragment extends Fragment implements LiveTvGui
     private final Lazy<NavigationRepository> navigationRepository = inject(NavigationRepository.class);
     private final Lazy<BackgroundService> backgroundService = inject(BackgroundService.class);
     private final Lazy<ImageHelper> imageHelper = inject(ImageHelper.class);
+    private final Lazy<WatchTrackerViewModel> watchTracker = inject(WatchTrackerViewModel.class);
```
johnpc, line 138: It doesn't look like this is used, is it?

app/src/main/java/org/jellyfin/androidtv/ui/browsing/MainActivity.kt
```diff
@@ -168,6 +170,10 @@ class MainActivity : FragmentActivity() {
 		super.onUserInteraction()
 
 		screensaverViewModel.notifyInteraction(false)
+
+		if (watchTracker.getItemIsEpisode()) {
+			watchTracker.onUserInteraction()
```
johnpc, line 175: I think I'd rather see `watchTracker.notifyInteraction()` and then the implementation within watchTracker would check if it's an episode or not.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -481,6 +487,8 @@ public void onClick(DialogInterface dialog, int which) {
                 mFragment.setPlayPauseActionState(0);
                 mFragment.setCurrentTime(position);
 
+                if (watchTracker.getItemIsEpisode() && !watchTracker.getItemWasInterrupted()) watchTracker.updateLastUpdateTime();
```
johnpc, line 490: same comment as above, I think I prefer `watchTracker.notifyPlay()` and let the watchTracker be in charge of isEpisode/interupted logic.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -1179,6 +1190,8 @@ public void onProgress() {
                     stopSpinner();
                 }
             }
+
+            if (watchTracker.getItemIsEpisode() && !watchTracker.getItemWasInterrupted()) watchTracker.updateWatchTime();
```
johnpc, line 1194: same comment here with like `watchTracker.notifyProgress()`

app/src/main/java/org/jellyfin/androidtv/ui/playback/WatchTrackerViewModel.kt
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	fun updateLastUpdateTime() {
+		lastUpdateTime = System.currentTimeMillis()
+	}
+
+	fun onEpisodeWatched() {
+		Timber.i("Watcher onEpisodeWatched")
+		if (!itemWasInterrupted) episodeCount++
+		itemWasInterrupted = false
```
johnpc, line 42: I don't think `itemWasInterrupted` is needed at all, and is just leading to bugs and complexity. I think whenever user interaction occurs, you can just reset the lastUpdateTime and watchedEpisodeCount. Tracking "interruption" separately is confusing unless I'm missing something.

app/src/main/java/org/jellyfin/androidtv/ui/playback/WatchTrackerViewModel.kt
```diff
@@ -0,0 +1,109 @@
...
+
+	private fun checkPrompt() {
+		if (!::videoManager.isInitialized) return
+
+		val stillWatchingSetting = StillWatchingStates.getSetting(userPreferences[UserPreferences.stillWatchingEnabled].toString())
+
+		val currentEpisodeProgress = videoManager.currentPosition.toFloat() / videoManager.duration.toFloat()
+		val minMinutesInMs = stillWatchingSetting.minMinutes * MILLIS_PER_MIN
+
+		// At episode count, your watch time is above min minute threshold, and you are at least 10% of the way through the next episode
```
johnpc, line 79: This 10% requirement feels out of the blue.

app/src/main/java/org/jellyfin/androidtv/ui/playback/WatchTrackerViewModel.kt
```diff
@@ -0,0 +1,109 @@
...
+		val episodeRequirementMet = episodeCount == stillWatchingSetting.episodeCount && watchTime >= minMinutesInMs && currentEpisodeProgress >= 0.1
```
johnpc, line 80: why does `episodeRequirementMet` include watch time? Isn't watch time supposed to be considered in the next line?

app/src/main/java/org/jellyfin/androidtv/ui/playback/WatchTrackerViewModel.kt
```diff
@@ -0,0 +1,109 @@
...
+
+		// Above min minute threshold and you are not on the episode to meet episode requirement
```
johnpc, line 82: Can you explain `and you are not on the episode to meet episode requirement`? This feels superfluous

app/src/main/java/org/jellyfin/androidtv/ui/playback/stillwatching/StillWatchingButtonsView.kt
```diff
@@ -0,0 +1,73 @@
...
+			override fun onTick(millisUntilFinished: Long) {
+				view.fragmentStillWatchingButtonsNoProgress.apply {
+					max = duration
+					progress = (duration - millisUntilFinished).toInt()
+				}
+			}
+
+			override fun onFinish() {
+				// Perform a click so the event handler will activate
+				view.fragmentStillWatchingButtonsNo.performClick()
```
johnpc, line 50: This reminds me - do other parts of the codebase call `.performClick()` in a fashion that would cause a user action event? For example, on the NextUpButtonsView.kt calls `.performClick()` and I'm concerned that might be treated as a user interaction that resets the Still Watching timer/episode count.

app/src/main/java/org/jellyfin/androidtv/ui/playback/stillwatching/StillWatchingFragment.kt
```diff
@@ -0,0 +1,105 @@
...
+					else -> Unit
+				}
+			}.launchIn(lifecycleScope)
+	}
+
+	override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+		_binding = FragmentStillWatchingBinding.inflate(inflater, container, false)
+
+		binding.fragmentStillWatchingButtons.apply {
+			// duration = userPreferences[UserPreferences.nextUpTimeout]
```
johnpc, line 67: If we must have a countdown timer for this (I really don't like the countdown timer the more I think about it), then we should definitely have a user preference instead of hardcoding 30 seconds (which is actually super duper fast unless your remote is taped to your hands).

app/src/main/java/org/jellyfin/androidtv/ui/playback/stillwatching/StillWatchingStates.kt
```diff
@@ -0,0 +1,19 @@
+package org.jellyfin.androidtv.ui.playback.stillwatching
+
+// minMinutes needs to be changed back to Int after testing and 2 test conditions need to be removed
+data class StillWatchingStates(val enabled: Boolean, val episodeCount: Int, val minMinutes: Number) {
+	companion object {
+		fun getSetting(type: String): StillWatchingStates {
+			return when (type.lowercase()) {
+				"test_episode_count" -> StillWatchingStates(enabled = true, episodeCount = 2, minMinutes = 1)
+				"test_min_minutes" -> StillWatchingStates(enabled = true, episodeCount = 5, minMinutes = 0.5)
+				"short" -> StillWatchingStates(enabled = true, episodeCount = 2, minMinutes = 60)
```
johnpc, line 10: How does this affect long shows with >60m episodes?

app/src/main/java/org/jellyfin/androidtv/ui/playback/WatchTrackerViewModel.kt
ConnorS1110: Huh I see android studio is giving me an error on that line as well. Idk how it built before but it will be fixed in my next push

app/src/main/java/org/jellyfin/androidtv/ui/playback/stillwatching/StillWatchingButtonsView.kt
ConnorS1110: I will just remove it and if there is a feature request for it later, I can add it back. Should be straightforward to re-add if needed.

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
ConnorS1110: As I was working on this it looked like I was going to need to make changes to this file, and as I was working in it, I realized there was A LOT of duplicate code to get the playbackController, so I just refactored it. It turns out I didn't need to touch this file at all though. I think I would still advocate for keeping the refactor, even though the changes are no longer relevant to the feature, since it changes no logic and just makes the file more readable. For review purposes it could be largely ignored. It would be easy to revert if there was an issue. I'll leave it up to you if these changes should stay or not.

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
ConnorS1110: You're right it will be removed in my next push

app/src/main/java/org/jellyfin/androidtv/ui/playback/stillwatching/StillWatchingButtonsView.kt
ConnorS1110: So to address this, it is going to run counter to what you suggested earlier, re: Having the viewmodel be completely in charge of deciding whether there was an interaction/it should start/etc. I need to ignore inputs when the still watching fragment is open, so I am not resetting things when I shouldn't be. I need the fragmentManager to do that, and thats a big ViewModel no-no since the ViewModel shouldn't be handling UI logic. In the activity, I am going to still need to keep the interaction function wrapped in an if statement so I can properly determine if the fragment is visible or not

app/src/main/java/org/jellyfin/androidtv/ui/playback/stillwatching/StillWatchingFragment.kt
ConnorS1110: I am getting rid of the timer

app/src/main/java/org/jellyfin/androidtv/ui/playback/WatchTrackerViewModel.kt
ConnorS1110: I will need this flag for notifyProgress, because we no longer need to track progress for an episode that experienced an interruption. We will start over on the next episode

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
ConnorS1110: Will be changed in next push

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
ConnorS1110: Will be changed in next push

app/src/main/java/org/jellyfin/androidtv/ui/browsing/MainActivity.kt
ConnorS1110: Will be changed in next push

ConnorS1110:
@johnpc Pushed again and addressed everything to the best of my ability. Now I will explain the logic behind the 2 requirements.

Episode Requirement: We need to ensure we are above the min minute threshold in order to say we have met the episode requirement. Imagine we are watching episodes that are 10 minutes long, Within 30 minutes, we would hit the condition for the default setting to show the screen, even though the min number of minutes is 90. So if we had normal to longer episodes (45 minutes - 1 hr +), this condition will likely get tripped

Time Requirement: There needs to be another condition that handles if you are watching a lot of short episodes. That is what this is for. The logic has been changed so that watchTime must be greater than or equal to the min minute threshold of your user setting AND your episode count must be greater than the number from your user setting. This ensures we capture short media to still display the screen at an appropriate interval.

Hope this makes it make more sense

app/src/main/java/org/jellyfin/androidtv/ui/playback/WatchTrackerViewModel.kt
johnpc: Interesting. So if I say "yes still watching" you want it to not ask again until the next episode even if my time setting is set to very short for testing?

johnpc: This is another one of those things where I am concerned could cause customer confusion. Resetting the timer/counter is how I assumed it would work. The "interrupted" tracker feels unexpected to me. I would assume that people's timers/counters wouldn't re-fire again in the same episode anyway unless they have it set comically low for testing.

johnpc:
Also just a reminder, my suggestions come from an interest in your feature, a love of jellyfin and a desire to build great software, but I am not a maintainer just a guy so that's all they are is suggestions. :D

app/src/main/java/org/jellyfin/androidtv/ui/playback/stillwatching/StillWatchingButtonsView.kt
johnpc: > I need to ignore inputs when the still watching fragment is open

It might be a bit fragile if other `.performClick()`s get added in the future since you'll have to work around those in order to keep the Still Watching feature behaving as expected.

app/src/main/java/org/jellyfin/androidtv/ui/playback/WatchTrackerViewModel.kt
```diff
@@ -0,0 +1,120 @@
...
+
+	fun notifyProgress() {
+		if (itemIsEpisode() && !itemWasInterrupted) {
+			calculateWatchTime()
+			checkPrompt()
+		}
+	}
+
+	private fun itemIsEpisode(item: BaseItemDto? = null): Boolean {
+		val newItem = item
```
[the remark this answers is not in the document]
johnpc: is ktlintFormat set up for this project? If so you might just have to run that.

ConnorS1110:
At this point, I am going to wait for a maintainer to give thoughts on my implementation before working on this more. Hopefully this can get finalized soon and I can start working on making a matching implementation for web

nielsvanvelzen (changes requested):
We already have interaction tracking for our screensaver so I'd rather add a new "InteractionTracker" that will be used by both the screensaver and this feature.

We will **not** be adding popups during video playback, that's an anti-pattern. Instead, this feature should be implemented in the next-up screen: when a video item ends and we want to know if a user is still watching it will show a different screen that will behave similarly to the popup you currently have.

All new UI **must** be written in Compose.

ConnorS1110:
@nielsvanvelzen is it acceptable for you if I just move the contents of `ScreensaverViewModel` into `WatchTrackerViewModel` and just rename it to `InteractionTrackerViewModel` per your suggestion and obviously update references and such

ConnorS1110:
@nielsvanvelzen almost done with the changes you requested, but a design question. Thoughts on the screen looking nearly identical to the "Next Up" screen, except the words "Are you still watching?" along with the buttons are in the center of the screen? I thought about still having the next up info in the bottom right corner, just minus the buttons. Or did you envision this screen would look literally EXACTLY the same as the Next Up screen, just with the verbiage describing the next episode removed with just the "Are you still watching?" question with action buttons?

ConnorS1110:
Alright I think I have finalized what was asked to be changed and I personally think the design I have looks pretty good. Going to say ready for a formal re-review @nielsvanvelzen 
Still_watching_more_info

app/src/main/java/org/jellyfin/androidtv/ui/playback/stillwatching/StillWatchingViewModel.kt
```diff
@@ -0,0 +1,68 @@
...
+
+	private suspend fun loadItemData(id: UUID) = withContext(Dispatchers.IO) {
+		val item by api.userLibraryApi.getItem(itemId = id)
+
+		val thumbnail = when (userPreferences[UserPreferences.nextUpBehavior]) {
+			NextUpBehavior.EXTENDED -> imageHelper.getPrimaryImageUrl(item)
+			else -> null
+		}
+		val thumbnailBlurhash = item.imageBlurHashes?.get(ImageType.PRIMARY)?.get(item.imageTags?.get(ImageType.PRIMARY))
+		val logo = imageHelper.getLogoImageUrl(item)
```
nielsvanvelzen, line 56: The way we work with images has recently changed, you can simplify this by using the `itemImages` extension properties. Those have objects  that include blurhash, aspect ratio etc.

You can look at t he next up view model to see how it works

nielsvanvelzen (changes requested):
Finally had some time to look at this PR. Started out with testing it out and here are my findings from that:

- When the title of the next playing episode is too long the thumbnail is incorrectly positioned
- The buttons don't have the most intuitive labels, I'd say maybe use "Continue watching" and "Exit" instead of "yes" and "no".
- The preference title should not use word casing and there is no need for the "settings" suffix. Maybe call it "Show inactivity timeout popup" or something like that
- While useful for testing, I don't think the 2eps/1min or 5eps/0.5min options should be kept
- I find the default option should be to disable this behavior

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -87,6 +88,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 */
 		var cinemaModeEnabled = booleanPreference("pref_enable_cinema_mode", true)
 
+		/**
+		 * Enable still watching
+		 */
+		var stillWatchingBehavior = enumPreference("pref_enable_still_watching", StillWatchingBehavior.DEFAULT)
```
nielsvanvelzen, line 94: The `pref_` prefix is only used for old preference so we don't have to migrate their keys. Remove it for new ones.

app/src/main/java/org/jellyfin/androidtv/preference/constant/StillWatchingBehavior.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.jellyfin.androidtv.R
+import org.jellyfin.preference.PreferenceEnum
+
+enum class StillWatchingBehavior(
+	override val nameRes: Int
+) : PreferenceEnum {
+	TEST_EPISODE_COUNT(R.string.lbl_still_watching_test_episode_count),
+	TEST_MIN_MINUTES(R.string.lbl_still_watching_test_min_minutes),
+	/**
+	 * Takes shorter than Netflix implementation to show still watching screen.
```
nielsvanvelzen, line 12: I don't know how Netflix implements this. Just describe how it actually works.

app/src/main/java/org/jellyfin/androidtv/ui/InteractionTrackerViewModel.kt
```diff
@@ -0,0 +1,193 @@
...
+import org.jellyfin.androidtv.ui.playback.stillwatching.StillWatchingStates
+import org.jellyfin.androidtv.util.TimeUtils.MILLIS_PER_MIN
+import org.jellyfin.sdk.model.api.BaseItemDto
+import org.jellyfin.sdk.model.api.BaseItemKind
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.inject
+import kotlin.time.Duration.Companion.milliseconds
+
+class InteractionTrackerViewModel : ViewModel(), KoinComponent {
+	// Screensaver vars
```
nielsvanvelzen, line 25: Group name comments should always have an empty line following them

app/src/main/java/org/jellyfin/androidtv/ui/InteractionTrackerViewModel.kt
```diff
@@ -0,0 +1,193 @@
...
+import org.jellyfin.androidtv.ui.playback.PlaybackControllerContainer
+import org.jellyfin.androidtv.ui.playback.stillwatching.StillWatchingStates
+import org.jellyfin.androidtv.util.TimeUtils.MILLIS_PER_MIN
+import org.jellyfin.sdk.model.api.BaseItemDto
+import org.jellyfin.sdk.model.api.BaseItemKind
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.inject
+import kotlin.time.Duration.Companion.milliseconds
+
+class InteractionTrackerViewModel : ViewModel(), KoinComponent {
```
nielsvanvelzen, line 24: Don't use KoinComponent and inject dependencies via constructor instead

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
```diff
@@ -1199,6 +1201,7 @@ public void onResponse(List<BaseItemDto> response) {
                     return;
                 }
 
+                interactionTracker.getValue().notifyStartSession(item, response);
```
nielsvanvelzen, line 1204: I think it might be better to call this from the actual video player when it starts playback

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
nielsvanvelzen: We explicitly use a holder-class because the actual playback controller can change. So while it's fine to store the container in a variable you should NOT store the playback controller itself.

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -115,6 +120,7 @@ public PlaybackController(List<BaseItemDto> items, CustomPlaybackOverlayFragment
         mFragment = fragment;
         mHandler = new Handler();
 
+        interactionTracker = lazyInteractionTracker.getValue();
```
nielsvanvelzen, line 123: If you're gonna assign it directly on construct anyway you don't need to use a lazy injection. But that goes against the code style of all our existing Java code so I'd prefer not doing that.

app/src/main/java/org/jellyfin/androidtv/ui/playback/stillwatching/StillWatchingFragment.kt
```diff
@@ -0,0 +1,261 @@
...
+import org.jellyfin.androidtv.ui.navigation.NavigationRepository
+import org.jellyfin.androidtv.util.TimeUtils.MILLIS_PER_SEC
+import org.jellyfin.androidtv.util.apiclient.getUrl
+import org.jellyfin.sdk.api.client.ApiClient
+import org.jellyfin.sdk.model.serializer.toUUIDOrNull
+import org.koin.androidx.compose.koinViewModel
+import org.koin.compose.koinInject
+import java.util.UUID
+
+const val TIMEOUT_IN_MS = 10 * MILLIS_PER_SEC
```
nielsvanvelzen, line 62: Use durations instead of millisecond longs

nielsvanvelzen (changes requested):

ConnorS1110:
@nielsvanvelzen had time to go back and address all your comments. Let me know if there's any additional changes you want. ALso tried testing your long title issue and I couldn't repro. If the title is long, it just turns into an ellipsis (the same as the Next Up screen) and the thumbnail looks correct. Maybe you had a weird build or something or maybe its something specific about the title you were seeing?

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -131,14 +131,16 @@ public class CustomPlaybackOverlayFragment extends Fragment implements LiveTvGui
...
-    private final Lazy<PlaybackControllerContainer> playbackControllerContainer = inject(PlaybackControllerContainer.class);
+    private final Lazy<PlaybackControllerContainer> lazyPlaybackControllerContainer = inject(PlaybackControllerContainer.class);
     private final Lazy<CustomMessageRepository> customMessageRepository = inject(CustomMessageRepository.class);
     private final Lazy<NavigationRepository> navigationRepository = inject(NavigationRepository.class);
     private final Lazy<BackgroundService> backgroundService = inject(BackgroundService.class);
     private final Lazy<ImageHelper> imageHelper = inject(ImageHelper.class);
 
     private final PlaybackOverlayFragmentHelper helper = new PlaybackOverlayFragmentHelper(this);
 
+    private PlaybackControllerContainer playbackControllerContainer;
```
nielsvanvelzen: Remove this as requested in previous review

app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
```diff
@@ -61,12 +63,13 @@
...
 import io.github.peerless2012.ass.media.parser.AssSubtitleParserFactory;
 import io.github.peerless2012.ass.media.type.AssRenderType;
+import kotlin.Lazy;
 import timber.log.Timber;
 
 @OptIn(markerClass = UnstableApi.class)
 public class VideoManager {
     private ZoomMode mZoomMode;
-    private Activity mActivity;
+    private FragmentActivity mActivity;
```
nielsvanvelzen: Why is this changed to FragmentActivity?

app/src/main/java/org/jellyfin/androidtv/util/TimeUtils.java
```diff
@@ -10,12 +10,12 @@
...
-    private static final long MILLIS_PER_MIN = TimeUnit.MINUTES.toMillis(1);
-    private static final long MILLIS_PER_HR = TimeUnit.HOURS.toMillis(1);
+    public static final int MILLIS_PER_SEC = 1000;
+    public static final long MILLIS_PER_MIN = TimeUnit.MINUTES.toMillis(1);
+    public static final long MILLIS_PER_HR = TimeUnit.HOURS.toMillis(1);
 
-    private static final int SECS_PER_MIN = 60;
-    private static final long SECS_PER_HR = TimeUnit.HOURS.toSeconds(1);
+    public static final int SECS_PER_MIN = 60;
+    public static final long SECS_PER_HR = TimeUnit.HOURS.toSeconds(1);
```
nielsvanvelzen: These should not be made public. Any conversion between different time units should be done with the Duration class instead.

app/src/main/java/org/jellyfin/androidtv/ui/playback/stillwatching/StillWatchingStates.kt
```diff
@@ -0,0 +1,19 @@
+package org.jellyfin.androidtv.ui.playback.stillwatching
+
+// minMinutes needs to be changed back to Int after testing and 2 test conditions need to be removed
+data class StillWatchingStates(val enabled: Boolean, val episodeCount: Int, val minMinutes: Number) {
+	companion object {
+		fun getSetting(type: String): StillWatchingStates {
```
nielsvanvelzen: You have an enum. Use it.

ConnorS1110:
@nielsvanvelzen all comments have been addressed

nielsvanvelzen:
@coderabbitai full review

nielsvanvelzen (approved):
I could probably find more nitpicks but overall these changes are good and the functionality works as expected. I've made a few final small changes in the last commits.

Thanks for your work on this!

PR #4513, opened 2025-03-10
Add custom connectivity checker for Coil that always claims to be connected

### LINKED ISSUE #4379: The image does not display properly
### Describe the bug

jellyfin-androidtv installed on ThundeaL Projector TD80W Android 13 system does not display pictures, these pictures can be displayed briefly when the cursor is moved, but not when the cursor is stopped movingImage

### Logs

_No response_

### Application version

0.18.3

None

### Device information

ThundeaL Projector TD80W

### Android version

android 13

### Jellyfin server version

10.10.3

nielsvanvelzen:
Coil uses a "connectivity checker" to skip the network if it thinks there is no available connection. In Android this is sometimes detected by pinging a known host (e.g. google.com) which might fail for users that blackhole these hosts or only use the app on their local network.

**Changes**

Add custom connectivity checker that will always return `true` to force a network request to be made.

**Issues**

Fixes #4379

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/di/AppModule.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/di/AppModule.kt
@@ -3,6 +3,7 @@ package org.jellyfin.androidtv.di
 import android.content.Context
 import android.os.Build
 import coil3.ImageLoader
+import coil3.annotation.ExperimentalCoilApi
 import coil3.gif.AnimatedImageDecoder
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
Some more context: when Coil thinks there is no network connection it will force OkHttp to load from cache, when OkHttp cannot load something from cache it will return a 504 response code.

crobibero (approved):

PR #4514, opened 2025-03-11
feat: Add ability to use latest audio and subtitles

carlosjfcasero:
**Changes**
Add ability to remember audio and subtitle languages from last played episode. By enabling correspondent option in settings section, every new episode played will take into account last played episode to select audio and/or subtitles track.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -2,6 +2,8 @@ package org.jellyfin.androidtv.ui.playback;
 
 import static org.koin.java.KoinJavaComponent.inject;
 
+import static java.util.Objects.requireNonNull;
+
 import android.annotation.TargetApi;
 import android.app.AlertDialog;
 import android.content.DialogInterface;
@@ -23,6 +25,7 @@ import org.jellyfin.androidtv.preference.constant.NextUpBehavior;
 import org.jellyfin.androidtv.preference.constant.RefreshRateSwitchingBehavior;
 import org.jellyfin.androidtv.preference.constant.ZoomMode;
 import org.jellyfin.androidtv.ui.livetv.TvManager;
+import org.jellyfin.androidtv.util.PlaybackHelper;
 import org.jellyfin.androidtv.util.TimeUtils;
 import org.jellyfin.androidtv.util.Utils;
 import org.jellyfin.androidtv.util.apiclient.ReportingHelper;
@@ -41,15 +44,15 @@ import org.jellyfin.sdk.model.api.SubtitleDeliveryMethod;
 import org.jellyfin.sdk.model.serializer.UUIDSerializerKt;
 import org.koin.java.KoinJavaComponent;
 
+import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import kotlin.Lazy;
 import timber.log.Timber;
-
-import java.time.Duration;
 
 public class PlaybackController implements PlaybackControllerNotifiable {
     // Frequency to report playback progress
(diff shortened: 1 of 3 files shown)


johnpc:
Looks like an alternative to waiting for https://github.com/jellyfin/jellyfin-androidtv/pull/3754, which appears stale

nielsvanvelzen:
Just like #3754, this shouldn't need preferences in the client but could be default behavior. Although to be fair, I believe this behavior would be better suited for the server. We already ask it for the (subtitle/audio) streams to use and it already knows we played a previous episode of the same series with a specific subtitle/audio stream. If the change was made on the server it would result in a better experience across all clients.

carlosjfcasero:
I added preferences to:
1. Let the user choose
2. This part of the code is a little tricky to me (plus there are no tests), so I'm not completely sure I'm not breaking anything (although I have tested it manually).

If you prefer, I can remove preferences and let it as default behavior.
I agree this should be come from the server. However, a lot of issues has been raised (including [this one](https://github.com/jellyfin/jellyfin/issues/13087) open by myself, which, btw, has been tagged as "Wrong repo") and there is no solution yet (in web it seems to partially work). So, until it's solved server side, I guess client should be able to handle it

What do you think @nielsvanvelzen?

ConnorS1110:
(If a non-server approach is used) this looks a lot better than what I had in #3754 . I struggled getting playback to actually play, but this looks like it doesn't have the issue I did. I do think I could get behind the server approach for this though. Makes more sense to have the server calculate this in one place the consuming clients could use

mschumacher69:
> I added preferences:
> 
> 1. To let the user choose
> 2. Because this part of the code is a little tricky to me (plus there are no tests), so I'm not completely sure I'm not breaking anything (although I have tested it manually).
> 
> If you prefer, I can remove preferences and let it as default behavior. I agree this should be come from the server. However, a lot of issues has been raised (including [this one](https://github.com/jellyfin/jellyfin/issues/13087) open by myself, which, btw, has been tagged as "Wrong repo") and there is no solution yet (in web it seems to partially work). So, until it's solved server side, I guess client should be able to handle it
> 
> What do you think @nielsvanvelzen?

Yeah default behavior would be better, no one would watch an episode in 1 language and another episode in another language, once set, it should stick by default. And I do agree that this is better implemented server side (user setting), but if this is the only solution right now, I'd take it as long as this change is reverted once/if a server side solution is implemented later.

carlosjfcasero:
Set as default options. Also created https://github.com/jellyfin/jellyfin/issues/13794 to use it once it's solved

carlosjfcasero:
Could a maintainer have a look at the PR?

carlosjfcasero:
Any news regarding reviewing?

nielsvanvelzen (changes requested):
This implementation will cause extra API calls for each playback, this should be avoided.

carlosjfcasero:
> This implementation will cause extra API calls for each playback, this should be avoided.

Of course it does. I don't see any point of the code where application gets the whole list of episodes played (if there is, please, tell me). To avoid that, I added a preference in user settings, but you discouraged me to do so and I removed it...

carlosjfcasero:
> This implementation will cause extra API calls for each playback, this should be avoided.

You mean this implementation will cause extra API call when you, for example, play and resume the content? If so, I guess you mean to do API call just when starting the content(?)

carlosjfcasero:
@nielsvanvelzen

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
```diff
@@ -536,25 +543,80 @@ public void onError(Exception exception) {
...
+                        previousMediaStream.getType().equals(mediaStream.getType()) &&
+                                previousMediaStream.getLanguage().equals(mediaStream.getLanguage()))
+                .collect(Collectors.toList());
+
+        if (matchingMediaStreamList.size() == 1) {
+            newMediaStream = item.getMediaStreams().indexOf(matchingMediaStreamList.get(0));
+        } else {
+            newMediaStream = matchingMediaStreamList
+                    .stream()
+                    .filter(mediaStream -> previousMediaStream.getTitle().equals(mediaStream.getTitle()))
```
MonsieurNiark, line 595: Sometimes title can be null. You should check nullity before doing equals

carlosjfcasero:
I'm sorry for those interested in this feature. I give up. Contributing like this is impossible

Issue #4187, opened 2024-11-10
Configure the amount of time the skip button is visible


viggy96:
### Describe the feature you'd like

Right now the skip button is only shown briefly at the start of the intro and outro. 

Allowing users to configure this time to be longer or shorter would be nice. For me it's short, the skip button tends to disappear before I get to hit it.

nielsvanvelzen:
There are currently no plans to add a preference for this. Once the new player design is implemented the skip button will always become visible in the controls-overlay, but for now it will only be visible for 8 seconds.

viggy96:
Always visible in the overlay? Do you mean only during the intro/outro, or during the entire video?

woolmonkey:
So it well become visable when contorls are available good. 8 seconds is sometimes not long enough.

johnpc:
@nielsvanvelzen why limit it to 8 seconds? I feel like I always just miss it, and it's so unobtrusive in the ui. 

I created https://github.com/jellyfin/jellyfin-androidtv/pull/4519 to add the setting. It'd be a huge QoL improvement for many!

Issue #4177, opened 2024-11-06
Add "Skip" button to player overlay during media segments


laurentlbm:
### Describe the feature you'd like

The new "Skip" button that appears at the start of a media segment is automatically hidden after 8 seconds. There is no way to skip after that. I propose to have the Skip button available in the controls overlay for the whole duration of the media segment.

foux:
@laurentlbm you get a skip button? I haven't got any button and the app automatically skips segments without any option to prevent it, it is particularly annoying.

laurentlbm:
@foux The button is new in the beta release. As for automatically skipping, that's an option in the settings for the "Intro Skipper" plugin; maybe it's enabled on your server?

foux:
THanks @laurentlbm . I'm already on the beta build, so I should have the button. And no, the option to auto-skip isn't set on my extension.

laurentlbm:
Did you check the options in the app itself?

foux:
Yep, no options whatsoever

Clemv95:
The button is appearing in the Android client, but not in the current beta for me

johnpc:
I made this pr to make it configurable https://github.com/jellyfin/jellyfin-androidtv/pull/4519

Saschl:
What would also be nice is that the button reappears when you bring up the controls (like on Netflix for example) and the intro segment is still running.

Ciechom:
What is the status of this issue? I noticed someone proposed temporarily allowing user to configure the disappearance of the button timer but it was rejected long time ago and the same problem exists right now on a fresh install of the app from Amazon's app shop on the Fire Stick 4k max 2nd gen.

What is the pr number for the button to be shown for the duration of the segment? If no one is even working on it yet can we please implement #4519 untill the better solution exists?

Fnaargh:
I'm adding my +1 to this instead of creating a new request, hoping that the "new" player advertised a year ago arrives in the near future.

That said, I do appreciate immensely the work done on the jellyfin server and app, you have made my life so much better, thank you!
I'll contribute whenever I can to this project.

johnpc:
Is anyone actually working on this feature? The only reference I can find about it is [this comment from nov 2024](https://github.com/jellyfin/jellyfin-androidtv/issues/4187#issuecomment-2466651744), but that is just a comment on an issue that has since been closed. @nielsvanvelzen can you please share any insight

Kastrel7:
[...]

PR #4519, opened 2025-03-13
Enable configuration of Ask To Skip duration for intro skipper

johnpc:
Changes:

* Added a new setting in the Playback Preferences screen that allows users to customize how long the Skip Intro/Outro button stays on screen. 
* The setting provides a slider to adjust the duration from 3 to 20 seconds, with a default of 8 seconds. This gives users more control over their viewing experience based on their reaction speed and preferences.

Issues:

* https://features.jellyfin.org/posts/3035/intro-outro-button-visibility-time

Screenshot 2025-03-13 at 12 07 07 AM

NOTE: This is my first contribution to jellyfin-androidtv. Please let me know if you have any feedback I am happy to make whatever updates are desired by the maintainers.


nielsvanvelzen (changes requested):
As mentioned in #4187 there is no interest in adding this right now. The new player UI will always show a "skip" button (when available) in its control overlay so if you miss the "popup" you can still skip if you missed the timer.

johnpc:
Understood. 

Out of curiosity what is the perceived harm of offering this preference? It seems to align well with the existing preferences like "Next Up Timer Duration" 

Also, I'm not sure what you mean by "The new player UI will always show a "skip" button (when available) in its control overlay". Are you describing an upcoming feature that will mitigate this problem? The control overlay that I see does not currently include a skip button during intros. When I miss the 8 second window, the opportunity is gone. 

Maybe this would be valuable as a stop-gap if the upcoming feature you're referencing is far off? I see that the proposal to add it to the control overlay (https://github.com/jellyfin/jellyfin-androidtv/issues/4177) was from 6 months ago, so I am concerned it will be a while before the demand for configurable skip duration is met. When the feature you're talking about is released, perhaps then we could consider removing this setting if we determine it's just clutter at that point?

nielsvanvelzen:
Closing as invalid (see my comment above).

patricks1:
Seems like there would have only been benefits and no downside to merging this. Would have been nice.

Issue #3269, opened 2023-12-30
Screensaver option safe for OLED TVs


kingp0dd:
### Describe the feature you'd like

The new screensaver option looks good.

Would it be possible to have an option for OLED TVs that are susceptible to screen burn-in?

A plain black screensaver would be nice. Or maybe have the Jellyfin logo in the current screensaver to move around?

nielsvanvelzen:
Duplicate of #1888

PR #4521, opened 2025-03-14
Add support for empty and logo screensaver modes to reduce (OLED) screen burn-in

### LINKED ISSUE #3256: Simpler dark screensaver

I would like to request a simple dark screensaver, not that screen savers are an option. The issue is that I tend to fall asleep in front of the tv and the changing pictures are quite bright with the movie poster. My proposal is something like a floating around jellyfin logo similar to how plex used to (still does?) do it. Okay, thanks for reading.

basraven:
Add support for empty and logo screensaver modes to reduce (OLED) screen burn-in

- Introduces `ScreenSaverType` enum with `NORMAL` (slideshow) and `EMPTY_SCREEN` (black screen) options.
- Adds `screensaverType` user preference to control the screensaver type.
- Adds `screensaverShowLogo` user preference to show or not the logo on the screensaver.
- Implements the empty screen logic in `DreamView`.
- Update DreamHost logic for the new settings.
- Add the new settings on the Customization screen.

**Issues**
Fixes #3256 #3269 #1888

@nielsvanvelzen For some reasons the default screensaver from the Chromecast with Android doesn't work, hence these screensaver changes. Met vriendelijke groet!

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHost.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHost.kt
@@ -6,6 +6,7 @@ import androidx.compose.runtime.getValue
 import org.jellyfin.androidtv.integration.dream.DreamViewModel
 import org.jellyfin.androidtv.preference.UserPreferences
 import org.jellyfin.androidtv.preference.constant.ClockBehavior
+import org.jellyfin.androidtv.preference.constant.ScreenSaverType
 import org.koin.androidx.compose.koinViewModel
 import org.koin.compose.koinInject
 
@@ -20,6 +21,11 @@ fun DreamHost() {
 		showClock = when (userPreferences[UserPreferences.clockBehavior]) {
 			ClockBehavior.ALWAYS, ClockBehavior.IN_MENUS -> true
 			else -> false
-		}
+		},
+		showLogoSetting = userPreferences[UserPreferences.screensaverShowLogo],
+		showBlank = when (userPreferences[UserPreferences.screensaverType]) {
+			ScreenSaverType.EMPTY_SCREEN -> true
+			else -> false
+		},
 	)
 }
--- a/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamView.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamView.kt
@@ -16,27 +16,33 @@ import org.jellyfin.androidtv.integration.dream.model.DreamContent
 fun DreamView(
 	content: DreamContent,
 	showClock: Boolean,
-) = Box(
-	modifier = Modifier
-		.fillMaxSize()
-) {
-	AnimatedContent(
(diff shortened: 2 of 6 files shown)


nielsvanvelzen (changes requested):
If you don't want the screensaver to show up you can just disable it. I don't see any point in the black screen mode.

basraven:
@nielsvanvelzen Thanks for getting back on this, appreciate it!
From a usability point, if you disable the screensaver, then your screen will burn in because for the Google Chromecast with Android TV the native screensaver doesn't take over (tested on multiple devices). So right now we provide only 2 options:
1. No screensaver as you suggested, with the menu burning in the Jellyfin menu in your (OLED) tv
2. Through the screensaver where the logo burns in the TV.
Neither aren't ideal

Bonus points of having an empty screensaver over one with image content: it's better for the environment because your screen consuming less energy.  It's actually the option I use myself now (I already compiled a version for myself at home).

Overall it's a quality of life feature that I'm very much enjoying that I wanted to share here with the rest of the community.

nielsvanvelzen:
If you disable the screensaver in our app the system screensaver is allowed to run. If it does not show up that means you either have disabled the screensaver of there is some bug in your system. At the very least, not something that should be a concern for our app to deal with.

basraven:
> If you disable the screensaver in our app the system screensaver is allowed to run. 
I've tested this on multiple devices because in theory you are absolutely right, that is the expected behavior. On many apps like Netflix or HBO it actually does, but on the official Jellyfin app (installed from the Google Playstore) it does not. I created this fix exactly because of this odd behavior that is specific to Jellyfin.

ethaldeman:
> > If you disable the screensaver in our app the system screensaver is allowed to run.
> 
> I've tested this on multiple devices because in theory you are absolutely right, that is the expected behavior. On many apps like Netflix or HBO it actually does, but on the official Jellyfin app (installed from the Google Playstore) it does not. I created this fix exactly because of this odd behavior that is specific to Jellyfin.

I have observed this behavior on my onn 4K. Dispite disabling the jellyfin screensaver, the system screensaver does not run. However system screen saver does run on other apps.

nielsvanvelzen:
That would be a bug then, perhaps within our app. But that would need to be tracked in a GitHub issue and not worked around like this.

Hukuma1:
Please allow us to simply remove the logo that can be burned in. I'm not even asking for it to be the default value, just allow us to  choose a preference please.

nielsvanvelzen:
@basraven are you still interested in working on this (and addressing the feedback) or should we close it?

basraven:
> @basraven are you still interested in working on this (and addressing the feedback) or should we close it?

I'm still interested to spend a bit of time on it, but it depends on what your specific requirements are that would make it acceptable for you to accept this MR

nielsvanvelzen:
As mentioned before, there is no need for a purely black screen option as screensaver.

nielsvanvelzen:
Closing as abandoned. If you're still interested in addressing review feedback and finishing the PR please re-open.

PR #4522, opened 2025-03-15
Handle multiple srt cues at the same timestamp

johnpc:
Currently, there is a problem in the android-tv client where if multiple cues in an .srt subtitle file occur at the same timestamp, the android-tv client stacks them on top of each other, rendering them completely illegible.

See example, the line is "Maybe the people on this world are cave dwellers.":

Screenshot 2025-03-14 at 2 52 21 AM

You can see in the screenshot that the text is jumbled. It appears to read "Maybeare cave dwellerss world.".

In the jellyfin-web client, this is handled well enough by stacking the cues on top of each other. You can see that cue is broken into two lines "Maybe the people on this world" and "are cave dwellers."

Screenshot 2025-03-14 at 2 46 52 AM

The jellyfin-web client has these out of order, so it reads "are cave dwellers. Maybe the people on this world". It's not exactly right, but at least the text is legible.

This PR implements a custom subtitle handler that maintains the existing behavior except in the case where multiple srt cues exist at the same timestamp.

Screenshot 2025-03-14 at 11 03 20 PM

The result is a fully correct and legible subtitle render that reads  "Maybe the people on this world are cave dwellers."

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomSubtitleHelper.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomSubtitleHelper.java
@@ -1,0 +1,149 @@
+package org.jellyfin.androidtv.ui.playback;
+
+import android.content.Context;
+import android.graphics.Color;
+import android.util.TypedValue;
+import android.view.Gravity;
+import android.view.ViewGroup;
+import android.widget.LinearLayout;
+import android.widget.TextView;
+
+import androidx.annotation.NonNull;
+import androidx.media3.common.text.Cue;
+import androidx.media3.common.text.CueGroup;
+import androidx.media3.ui.CaptionStyleCompat;
+
+import org.jellyfin.androidtv.preference.UserPreferences;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import timber.log.Timber;
+
+/**
+ * Helper class to manage custom subtitle rendering with proper stacking of multiple cues.
+ */
+public class CustomSubtitleHelper {
+    private final Context context;
+    private final LinearLayout subtitleContainer;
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
We've removed our custom subtitle renderer in 0.18 and now rely on the more feature complete implementation of our video player (media3). There is no intention to add a custom renderer for text-based subtitles in the foreseeable future.
Bugs related to text based subtitle rendering should thus be fixed in the upstream library.

johnpc:
The issue in the upstream library https://github.com/google/ExoPlayer/issues/10980 has been open since feb 2023 with comments saying they don't intend to fix it either. 

Why not take ownership of the problem into our own hands?

Is there a specific regression in this implementation? I think it will actually be the same except in the case of the squashed multiline issue.

nielsvanvelzen:
I understand the issue is not great to have, but as pointed out by the ExoPlayer (now media3) developers it's a difficult issue to fix properly. Our subtitle situation has **greatly** improved with 0.18 because we removed the custom text rendering, not going to re-introduce that now.

I'll be closing this PR as I don't see any chance of this being merged.

gmargutti:
Hey folks, sorry for reopening this, but I'm also facing the same issue. I agree it should be fixed in ExoPlayer/Media3, but I found a workaround that works for my use case — although I’m not sure if it’s the same as the one suggested in the PR. I’ve pushed a commit to a fork — could you take a look and let me know if it makes sense? IMO, it doesn't mess with rendering, just repositions cues while preserving the player's logic and subtitle styles.
The commit in question: https://github.com/gmargutti/jellyfin-androidtv/commit/1b0012f9692aea5cf39a7f209f402d3135a93983

johnpc:
Fwiw I have PR to address in the upstream library we'll see if they accept it https://github.com/androidx/media/pull/3151

PR #4523, opened 2025-03-15
Replace androidx.tv with own theme implementation

nielsvanvelzen:
Jetpack Compose (library used for new UI) is build in layers. The top layer is the "theme" package, of which we currently use `androidx.tv`. Unfortunately this library is lacking and IMO not suited for a lot of our uses (like not supporting touch), in addition to already being somewhat unmaintained by Google similar to how they did with the `androidx.leanback` library.

This PR marks the start of our own "Jellyfin" theme. Eventually it will adhere to our design guidelines made by @erikbucik. Right now most of it is close to the previous (material based) theme, as our Figma does not have foundations like typography or shapes ready yet.

The new design composables will eventually be moved to their own Gradle module and be used in both our mobile and TV apps where possible. Some composables will obviously be platform specific due to the different input methods (dpad remote vs touch). 

The API is mostly similar to what the official compose themes do to make it easy to work with for new contributors.

**Changes**

- Add "org.jellyfin.androidtv.ui.base" (will probably be moved later)
  - Add "Text" composable (yes this is a theme thing for some reason)
  - Add "Icon" composable
  - Add `JellyfinTheme" used to access current typograpy/shapes
  - Add Typography tokens (currently only Default which uses platform defaults)
  - Add shape tokens (currently the same values as `androidx.tv`, rounded corner shapes)
- Remove "androidx.tv" dependency
- Add "ButtonBase" composable which is the basis of a button
- Use the new composables
  - ConnectHelpAlert somewhat refactored
  - NowPlayingComposable now uses ButtonBase

Note: ButtonBase will behave a lot better with the next Compose version, which implements better key (dpad) support.

**Screenshots**

Notable differences are found in the ConnectHelpAlert in which letter spacing is now different. Other UI changes are not/barely noticeable.

**Before**

**After**

**Issues**

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -123,7 +123,6 @@ dependencies {
 	implementation(libs.androidx.cardview)
 	implementation(libs.androidx.startup)
 	implementation(libs.bundles.androidx.compose)
-	implementation(libs.androidx.tv.material)
 
 	// Dependency Injection
 	implementation(libs.bundles.koin)
--- a/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentLibraryShowcase.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamContentLibraryShowcase.kt
@@ -14,8 +14,8 @@ import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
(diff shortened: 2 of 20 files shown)


crobibero (approved):
Looks like some new magic numbers, should they be moved to consts?

nielsvanvelzen:
I did hardcode some numbers that should be moved to our "design system" eventually. E.g. "TextStyle.DialogTitle" instead of hardcoded size of 45sp. So this is intended _for now_.

PR #4531, opened 2025-03-19
Rewrite NextUpFragment with Compose

nielsvanvelzen:
**Changes**
- Add "ProgressButton" which is the same as "Button" but with an additional "progress" value (0.0f-1.0f) to show in its background layer
- Rewrite NextUpFragment in Compose, layout is kept the same with some minor design changes
  - Typography will be updated later once available in the Figma (@erikbucik)
  - Button shapes might change later due to feedback on Matrix, will also affect the ConnectHelpAlert
  - Written in two parts: complete screen & control overlay. this would make it possible to re-use just the overlay when we add the next up info during playback.

**Screenshots**

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/base/button/ProgressButtonBase.kt
```diff
@@ -0,0 +1,44 @@
...
+		enabled = enabled,
+		shape = shape,
+		interactionSource = interactionSource,
+	) {
+		Box(Modifier.matchParentSize()) {
+			Box(
+				Modifier
+					.fillMaxHeight()
+					.fillMaxWidth(progress)
+					.background(Color(0x33000000))
```
nielsvanvelzen, line 38: Hardcoded color 👀

crobibero (approved):

PR #4537, opened 2025-03-22
Respect server-side user preference for live TV default view

### LINKED ISSUE #1113: Have ATV client respect "default view" on a per user basis for LiveTV.
Jellyfin offers the option to set a preferred view for LiveTV in the HOME section of user options.  ATV does not currently respect this selection and instead displays only it's own default view.

image

**Additional context**	
It would be nice to either a) default to guide view for LiveTV (which I presume is far and away the most used) or to respect the user preferences as selected in the Jellyfin interface.

johnpc:
**Changes**

This PR respects the server-side user preference for the "Default view" for the Live TV category. 

This is configurable on the web client in your `User Preferences -> Home -> Live TV -> Default Screen`. See

Screenshot 2025-03-22 at 4 12 53 PM

For example, my preference is the TV Guide, so it skips directly to the TV Guide view:

https://github.com/user-attachments/assets/05da4bc3-3b0d-4417-af94-3e6a04559581

Worth noting not every view has an equivalent on android tv, so if it's an unsupported selection, it will show the same default screen that appears today.

If we like this idea of having the client respect the server preferences, I can do the same for other categories as well (like movies / shows / music / etc) but it's much cleaner to do one thing at a time.

Let me know what else you'd like me to include to get this over the finish line! Tbh for Live TV, none of the views other than Guide are useful for me at all, so this would save a lot of clicks!

**Issues**

* Resolves https://github.com/jellyfin/jellyfin-androidtv/issues/1113
* Resolves https://features.jellyfin.org/posts/1831/straight-to-live-tv-guide

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/di/AppModule.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/di/AppModule.kt
@@ -132,4 +132,7 @@ val appModule = module {
 	single<PlaybackHelper> { SdkPlaybackHelper(get(), get(), get(), get(), get(), get(), get()) }
 
 	factory { (context: Context) -> SearchFragmentDelegate(context, get(), get()) }
+
+	// Add coroutine scope for async operations
+	single { kotlinx.coroutines.MainScope() }
 }
--- a/app/src/main/java/org/jellyfin/androidtv/di/KoinInitializer.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/di/KoinInitializer.kt
@@ -18,9 +18,9 @@ class KoinInitializer : Initializer<KoinApplication> {
 			playbackModule,
 			preferenceModule,
 			utilsModule,
+			liveTvModule,
 		)
 	}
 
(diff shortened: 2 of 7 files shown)


app/src/main/java/org/jellyfin/androidtv/di/AppModule.kt
```diff
@@ -132,4 +132,7 @@ val appModule = module {
 	single<PlaybackHelper> { SdkPlaybackHelper(get(), get(), get(), get(), get(), get(), get()) }
 
 	factory { (context: Context) -> SearchFragmentDelegate(context, get(), get()) }
+
+	// Add coroutine scope for async operations
+	single { kotlinx.coroutines.MainScope() }
```
nielsvanvelzen, line 137: Coroutine scopes should **never** be added to dependency injection

nielsvanvelzen (changes requested):
As mentioned in chat. This cannot be merged unless we have UI to navigate to the different live TV screens (in a separate pull request).

In the current state of the PR if you default to the guide its impossible to open recordings for example.

PR #4550, opened 2025-03-26
Directplay VobSub and DVDSub

nielsvanvelzen:
The the recent media3 1.6 release added support for VobSub. This pull request enables it for Jellyfin. In our case both VobSub and DVDSub are recognized as the "dvdsub" codec. It cannot be extracted by the server so only embedded & encoded is supported for now.

**Changes**
- Directplay VobSub and DVDSub

**Issues**


crobibero (approved):
#yolo

nielsvanvelzen:
During testing a few notable things came up:

1. This change is working as expected with MKV files: embedded dvdsub works, external will transcode
2. This change is **not** working as expected with MP4 files: media3.exoplayer is unable to detect the subtitle tracks, making our app unable to activate the subtitle track

I'll merge this PR for now as I'm confident this same issue already exists for other subtitle formats in mp4, like pgs. I'll do some more testing later and come up with a solution. Ideally this should be fixed in media3 but we should also make sure the app can deal better with subtitle switching issues.

PR #4567, opened 2025-04-06
Adds a profanity filter like Vidangel

jadams777:
This PR adds a Vidangel-style custom content filter.

**Changes**
- Adds a custom blacklist of words and phrases that users don't want to see or hear, e.g. profanity.
- Mutes audio when blacklisted words and phrases are played. 
- Replaces blacklisted content in subtitles with asterisks.


nielsvanvelzen (changes requested):
A feature like this must be implemented server side, media segments feel like the best solution here. There is no change of this being merged like this.

nielsvanvelzen:
Closing as invalid and stale

PR #4593, opened 2025-04-13
Add hearing impaired indicator for subtitles (SDH)

### LINKED ISSUE #3463: Add hearing impaired subtitle stream indicator

https://github.com/jellyfin/jellyfin/pull/7379

This will let the deaf community know when subtitles are for them or not.

This would be for 10.9.0 update.

nielsvanvelzen:
Would it be better to show CC when there is _any_ non-sdh subtitle stream and SDH when there is _any_ SDH stream? Let me know in the comments (don't forget to like and subscribe).

**Changes**
- Add "SDH" instead of "CC" as indicator to the info row when subtitles are marked as hearing impaired

image

**Issues**

Fixes #3463


iPleyGit:
The nomenclatures SDH or CC are often not understood by the average user, so the best option, I suppose, would be to use “for the deaf.”
If this isn’t possible due to abbreviation, the best choice is to stick with SDH, which is the most familiar.

Pentaphon:
>Would it be better to show CC when there is any non-sdh subtitle stream and SDH when there is any SDH stream?

I think it should show both CC and SDH if both are detected and the one that is detected when only 1 type exists. That will cover everybody.

mschumacher69:
I think SDH is enough when there's any SDH subtitle stream.

crobibero (approved):
unsubscribe

nielsvanvelzen:
I've updated the pull request to show the "CC" badge when there is at least 1 subtitle stream that is not hearing impaired and "SDH" if there is. They can also both show.
Unfortunately we don't have a lot of space so using "hearing impaired" as text would be too large. Both CC and SDH are translatable if a language has other abbreviations for them.

PR #4612, opened 2025-04-23
Add rating option to only display ratings on media info screen

zackslash:
Currently ratings can be '_displayed_' or '_hidden_', this new additional option "Only in media info" allows ratings to be hidden on the home screen items while still appearing in the media info.

Selection:
Screenshot 2025-04-23 at 5 19 19 pm


nielsvanvelzen (changes requested):
This preference is used to decide which rating type to show. Not where to show it. I'm also not interested in a new preference to choose where to show ratings. I don't see any use-case for it.

zackslash:
The use case is customisable preference. Not clear if your response means I should re-implement as a seperate preference or the feature is unwelcome.

zackslash:
This is disheartening, I retract my contribution. Marking as closed.

PR #4619, opened 2025-04-24
Change screensaver to cycle through random selection and avoid repeats

claireratigan:
**Changes**
I was watching my Jellyfin screensaver and noticed that there were some media that were displayed repeatedly, and found that this was because the screensaver has no memory of the media it recently cycled through. This is a rough draft of an alternative model; instead of requesting a new random library item to display each time, request many distinct random library items from the outset, and then play them in (the provided random) order. This ensures better coverage of the media library and reduces repetition.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/integration/dream/DreamViewModel.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/dream/DreamViewModel.kt
@@ -61,18 +61,21 @@ class DreamViewModel(
 		emit(null)
 		delay(2.seconds)
 
-		var randomItems = getRandomLibraryItems();
-		while (randomItems == null)
-			delay(3.seconds)
-			randomItems = getRandomLibraryItems();
-
-		val isCompleteList = randomItems.totalRecordCount == randomItems.items.size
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
I think preloading a bunch of items is a good idea, although 1000 is too many. We should limit it to about 100 to avoid hammering the server (loading items is expensive) and client (takes up bandwidth and memory). A good way to implement it is probably to add a queue that we add the items to and take the first item from each time we want the next showcase. We should avoid complicating the while loop as that is purely used for timing.

nielsvanvelzen:
@claireratigan the build is failing and there is a merge conflict

claireratigan:
@nielsvanvelzen Yeah sorry, I accidentally merge from the web interface and didn't get around to fixing it. This is a more careful rebase that hopefully should build 😅

claireratigan:
I see the build is failing, I'm not the most familiar with kotlin but I'll have a go at debugging it

claireratigan:
No worries, this looks much cleaner! Thanks for working on this, I'm glad to see it merged :)

PR #4630, opened 2025-04-29
Add experimental ASS direct play using libass

nielsvanvelzen:
**Changes**

Add libass integration to render ssa & ass subtitle formats using [libass-android](https://github.com/peerless2012/libass-android). It needs to be enabled using a developer preference (also visible in release builds) and is **experimental** for now.

The used library is a bit messy to work with as media3.exoplayer is not the easiest to extend with new formats and the library implements multiple different methods of rendering (with questionable names). We only support a single method, the one called "legacy" (which is by no means a "legacy" method), subtitles will be rendered as bitmap and added to a cue. The other available methods do not work on Nvidia Shield devices making it a no-go until fixed.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -133,6 +133,7 @@ dependencies {
 	implementation(libs.androidx.media3.exoplayer.hls)
 	implementation(libs.androidx.media3.ui)
 	implementation(libs.jellyfin.androidx.media3.ffmpeg.decoder)
+	implementation(libs.libass.media3)
 
 	// Markdown
 	implementation(libs.bundles.markwon)
--- a/app/src/main/java/org/jellyfin/androidtv/di/PlaybackModule.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/di/PlaybackModule.kt
@@ -24,7 +24,6 @@ import org.jellyfin.playback.media3.exoplayer.ExoPlayerOptions
 import org.jellyfin.playback.media3.exoplayer.exoPlayerPlugin
 import org.jellyfin.playback.media3.session.MediaSessionOptions
 import org.jellyfin.playback.media3.session.media3SessionPlugin
-import org.jellyfin.sdk.api.client.ApiClient
 import org.jellyfin.sdk.api.client.HttpClientOptions
 import org.koin.android.ext.koin.androidContext
 import org.koin.core.scope.Scope
@@ -71,9 +70,9 @@ fun Scope.createPlaybackManager() = playbackManager(androidContext()) {
 	}
 
 	val userPreferences = get<UserPreferences>()
-	val api = get<ApiClient>()
 	val exoPlayerOptions = ExoPlayerOptions(
 		preferFfmpeg = userPreferences[UserPreferences.preferExoPlayerFfmpeg],
+		enableLibass = userPreferences[UserPreferences.assDirectPlay],
 		enableDebugLogging = userPreferences[UserPreferences.debuggingEnabled],
 		baseDataSourceFactory = get<HttpDataSource.Factory>(),
 	)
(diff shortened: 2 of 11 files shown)


ethaldeman:
https://github.com/androidx/media/issues/1620 is a link to Nvidia devices issue.

Hello-World-Traveler:
The testing I did found ass and dvdsubs display with all dev settings without transcoding from the server.

No issue with HEVC or AV1 files and external subtitles other then not showing in the set colour [yellow].

However I find an error with audio out of sync unless bitstream DD is unticked. 

Start of the ass subtitle file

```
ScriptType: v4.00+
PlayResX: 384
PlayResY: 288

[V4+ Styles]
Format: Name, Fontname, Fontsize, PrimaryColour, SecondaryColour, OutlineColour, BackColour, Bold, Italic, Underline, StrikeOut, ScaleX, ScaleY, Spacing, Angle, BorderStyle, Outline, Shadow, Alignment, MarginL, MarginR, MarginV, Encoding
Style: Default,Arial,16,&Hffffff,&Hffffff,&H0,&H0,0,0,0,0,100,100,0,0,1,1,0,2,10,10,10,0
```


Downloaded this build https://github.com/jellyfin/jellyfin-androidtv/actions/runs/14800036650/artifacts/3052243080

ec04012:
What's the different between MEDIA_CUES, MEDIA_EFFECT, and MEDIA_EFFECT_GL under the the hood? From the limited testing that I did on my device, the two media_effect options are better at handling subtitles that have motion (media_cues either displays static subtitles or freezes). But the media_effect options also struggled and had choppy playback with a 4k file.

jessejohn01:
> What's the different between MEDIA_CUES, MEDIA_EFFECT, and MEDIA_EFFECT_GL under the the hood? From the limited testing that I did on my device, the two media_effect options are better at handling subtitles that have motion (media_cues either displays static subtitles or freezes). But the media_effect options also struggled and had choppy playback with a 4k file.

@peerless2012 has a readme writeup on the underlying implementation you can find here:
https://github.com/peerless2012/libass-android/blob/master/lib_ass_media/README.md?plain=1

nielsvanvelzen:
Further work will be done in future pull requests. This feature is now good enough to be introduced as experimental feature in the upcoming 0.19 release.

thornbill (approved):

gummykage:
I'm kind of new to this, would this be compatible with FireTV? Where does one get the files to try this build with the experimental support?

nielsvanvelzen:
I have not tested these changes with Fire TV specifically yet but in general, yes it should work. Depending on the performance of the device and the subtitles played there may be stuttering.
To enable the experimental option you need to sideload a debug development build of the app to your device.

MrGcGamer:
FYI: I just pulled the latest artifact from here https://github.com/jellyfin/jellyfin-androidtv/actions/runs/15623249913
and it doesn't seem to work for me.
After enabling it in the developer settings, when I try to play something with ass subtitles
1. no subtitles are being displayed
2. playback freezes after a few seconds (UI is still responsive)
3. switching of audio tracks doesn't work anymore
And for some reason it is also remuxing (which it's also doing on stable) and I've got no clue why, as playing other mkv files is no problem.
Playback of other media is extremely laggy / stuttery, but that might be due to the debug build?

Haven't tried https://github.com/jellyfin/jellyfin-androidtv/pull/4685
as the build artifact expired

nielsvanvelzen:
I've updated #4685 so you can try a fresh build. If it still doesn't work please share client logs (in a new issue) so we can see what is going on.

MrGcGamer:
Subtitles get rendered now!
Though they tend to get out of sync with the playback and the playback in general is kinda unstable, but I suspect that's due to it being a debug and not a release build in combination with my TV being pretty old / slow.

PR #4635, opened 2025-04-30
Add intermediate model for image information

### LINKED ISSUE #4628: Request: Use random background image for media during screensaver

### Problem description

Jellyfin server allows adding multiple background images to TV shows and movies. But when the Android TV app is displaying the screensaver, it always uses the first image.

Possibly related issue: https://github.com/jellyfin/jellyfin-androidtv/issues/4626 (but it only discusses music and fanart)

### Proposed solution

The app should use a random background image instead of always using the first one.

### Alternatives considered

Right now I have to manually reorder background images in Jellyfin server for the media every couple of months. For large libraries, this can be very tiresome.

### Additional information

N/A.

Thank you!

### Other sources

_No response_

nielsvanvelzen:
Add a new "JellyfinImage" model that contains all image information retrievable from API responses. This is accessible via the next ".images" extension getters like `BaseItemDto.itemImages`, `BaseItemDto.parentBackgroundImages` etc.

The goal is to make it much easier to use images from the API by grouping information like the item id, blurhash, tag etc in a single model. That model can then be passed through the application and resolved to a URL easily.

Current implementation does not use it everywhere yet but it does in the background/screensaver/nowplaying code and fixes at least one bug that has been here for a long time and we only recently became aware of 👀.

**Changes**
- Add intermediate model for image information

**Issues**

Fixes #4628

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
@@ -15,13 +15,14 @@ import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.launch
 import org.jellyfin.androidtv.auth.model.Server
 import org.jellyfin.androidtv.preference.UserPreferences
+import org.jellyfin.androidtv.util.apiclient.getUrl
+import org.jellyfin.androidtv.util.apiclient.itemBackdropImages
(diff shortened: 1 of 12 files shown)


nielsvanvelzen:
Note: I've also started work on an proposal to change our API to do something similar as this by default, which in the long-term should replace all existing image information from API responses.

thornbill (approved):

sheharyarn:
Thank you! Looking forward to this in the next release!

Issue #4572, opened 2025-04-09
Fire TV Android App crashing on start


kaledev:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

Regularly, on my Fire TV only, the Jellyfin app is crashing on launch and displays the "No Playable Items" error. Then re-launching it solves the problem.

### Logs

---
client: Jellyfin for Android TV
client_version: 0.18.6
client_repository: https://github.com/jellyfin/jellyfin-androidtv
type: crash_report
format: markdown
---
### Logs
***Stack Trace***:
```log
java.lang.IllegalStateException: FragmentManager is already executing transactions
at androidx.fragment.app.FragmentManager.ensureExecReady(FragmentManager.java:1947)
at androidx.fragment.app.FragmentManager.execSingleAction(FragmentManager.java:1977)
at androidx.fragment.app.BackStackRecord.commitNow(BackStackRecord.java:317)
at org.jellyfin.androidtv.ui.browsing.DestinationFragmentView.activateHistoryEntry(DestinationFragmentView.kt:159)
at org.jellyfin.androidtv.ui.browsing.DestinationFragmentView.navigate(DestinationFragmentView.kt:94)
at org.jellyfin.androidtv.ui.browsing.MainActivity.handleNavigationAction(MainActivity.kt:118)
at org.jellyfin.androidtv.ui.browsing.MainActivity.access$handleNavigationAction(MainActivity.kt:33)
at org.jellyfin.androidtv.ui.browsing.MainActivity$onCreate$2.invokeSuspend(MainActivity.kt:66)
at org.jellyfin.androidtv.ui.browsing.MainActivity$onCreate$2.invoke(Unknown Source:8)
at org.jellyfin.androidtv.ui.browsing.MainActivity$onCreate$2.invoke(Unknown Source:4)
at kotlinx.coroutines.flow.FlowKt__TransformKt$onEach$$inlined$unsafeTransform$1$2.emit(Emitters.kt:50)
[...]

Issue #4441, opened 2025-02-05
Jellyfin client won't start on first try


BuggusMaximus:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

Jellyfin client won't start on first try on my Android TV. When i turn on my TV and then open Jellyfin app it will flash welcome screen and then show errors along these lines:

"No content to play back" or "Something went wrong a report was sent to your jellyfin server"

Immediately when i try to start it again it will launch without any issues. I can then close it again and on next try it will also start without errors. I can even turn my TV off and then retry and it will launch flawlessly. It only happens on fresh start after TV was off for some time, can't tell exactly how long but rather hours than minutes.

### Logs

Crash log
[upload_org.jellyfin.androidtv_0.18.4_20250205174224_a41677efe4a24005b58363dae5838495.log](https://github.com/user-attachments/files/18676551/upload_org.jellyfin.androidtv_0.18.4_20250205174224_a41677efe4a24005b58363dae5838495.log)

Jellyfin log
[log_20250205.log](https://github.com/user-attachments/files/18676561/log_20250205.log)

### Application version

0.18.4

### Where did you install the app from?

Google Play Store

### Device information

Sharp 70gp6760e

### Android version

Android 11

### Jellyfin server version

10.10.5

frankhuurman:
Exact same issue on Android TV(the chromecast one) starting from Jellyfin 10.10.1 and still occurs at 10.10.5.
I don't exactly know where this crash log goes, have tried to look it up but could not find anything specific related to this start-up "no content to play" issue.

I'm assuming it's an Android client issue rather than a jellyfin server issue.

mbrbug:
Same here.
Same logs. Same versions of the server/client. Google streamer.
Client is starting only after second launch of app.

mbrbug:
looks like with latest 10.10.6 server issue was fixed

frankhuurman:
> looks like with latest 10.10.6 server issue was fixed

[...]

PR #4640, opened 2025-05-04
Fix crash on startup

DreierF:
**Changes**

Handles already running transactions. Not an android expert, but judging from the stack trace this could fix the issue.

**Issues**

#4572
#4441


Pentaphon:
I've seen this a few times on my Fire TV 4K. Nice work. Hope this makes 0.18.x before the 0.19.x betas begin.

nielsvanvelzen:
I _think_ we should actually check for FragmentManager.isDestroyed and don't commit at-all (but write a warning message to the logs). Unfortunately I cannot reproduce the crash so I can also not test a potential fix.

DreierF:
Thanks for taking a look! I adjusted the code as suggested.

nielsvanvelzen (changes requested):

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks!

Issue #1150, opened 2021-09-22
Video Backdrop and Theme song playback


Thursty24:
Hi guys

Great job on Jellyfin!

Just one request for now:
Please add support for Video Backdrop and Theme song playback when browsing the library, just like the web interface

Kaschdn:
I would also love to have background theme songs in the library as I have them on the web-client or the normal android client.

Hackintosys:
+1

KruyKaze:
+1

SysCrasher:
In both cases enable the user specific settings in your client and rescan after placing the files.

Theme songs: 
Install the "Theme Songs"-Plugin (https://github.com/danieladov/jellyfin-plugin-themesongs not sure if it's needed to play theme songs by default). If you need your own theme song, place it as theme.mp3 in the root directory of your specific movie.

Theme video:
In the root directory of your movie create folder "backdrops" and move your own video file into it.

Kaschdn:
> In both cases enable the user specific settings in your client and rescan after placing the files.
> 
> Theme songs: Install the "Theme Songs"-Plugin (https://github.com/danieladov/jellyfin-plugin-themesongs not sure if it's needed to play theme songs by default). If you need your own theme song, place it as theme.mp3 in the root directory of your specific movie.
> 
> Theme video: In the root directory of your movie create folder "backdrops" and move your own video file into it.

Are you sure they work for the Android-**TV-Client**?

SysCrasher:
I can't confirm it for Android-TV, but theme songs work with the Android app on my Phone and Tablet.
For TV you can use an additional pc, if it doesn't work native.

Kaschdn:
I know it's working for the other clients but **not** for the Android TV - client which is why @Thursty24 and me were asking for it. Using a PC for the TV is not an option. 

> I can't confirm it for Android-TV, but theme songs work with the Android app on my Phone and Tablet. For TV you can use an additional pc, if it doesn't work native.

darinmanica:
+1 from me.  Theme music for TV series and movies creates a polished user experience when browsing a collection.

nielsvanvelzen:
This is something I'd like to add and should be possible when #1057 is ready.

ReenigneArcher:
+1, I'm the developer of [Themerr](https://github.com/LizardByte/Themerr-jellyfin) and a user just reported that it doesn't work on Android TV.

Starfoxfs:
[...]

PR #4653, opened 2025-05-10
Support theme song playback

mehrmoudi:
**Changes**
Adds theme song playback support using the `RewriteMediaManager` class. Behaviour:
- Upon navigating to an item that has a theme song, it starts playing in the background if there is no other media playing or queued.
- Further navigation to other items with the same theme song doesn't interrupt or restart the playback.
- Once playback finishes, it won't start again even when navigating back and forth between pages. However it will happen if if the user navigates to an item with a different theme song, or goes back to the home screen, and then comes back again. 

**Issues**
https://github.com/jellyfin/jellyfin-androidtv/issues/1150

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -16,6 +16,7 @@
  - [MajMongoose](https://github.com/majmongoose)
  - [Olaren15](https://github.com/Olaren15)
  - [dtrexler](https://github.com/dtrexler)
+ - [mehrmoudi](https://github.com/mehrmoudi)
 
 # Emby Contributors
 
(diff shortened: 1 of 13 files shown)


almightiest:
Volume is a must-have, thanks!  I always set mine to ~30% so when i'm scrolling shows later at night it doesn't blast me.

nielsvanvelzen (changes requested):
You should not use the MediaManager class. Instead manipulate the playback queue directly. We also need some indicator on queue items that they are for theme playback (most likely using a new element type). Those items should be ignored by various services like playback reporting and media sessions.

nielsvanvelzen:
@mehrmoudi still planning to finish this?

nielsvanvelzen:
Closing as stale

PR #4568, opened 2025-04-06
Add Random Sort

mehmetu61:
Add Sort Random

**Changes**

**Issues**

PR #4663, opened 2025-05-16
Add random sort (cleaned up #4568)

josephclaytonhansen:
This is a functional version of https://github.com/jellyfin/jellyfin-androidtv/pull/4568 - it has just one commit and one file changed, which resolves the feedback on #4568 . I'd like to have this feature, so I took the liberty of fixing the issues with the existing PR

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java
@@ -177,6 +177,8 @@ public class BrowseGridFragment extends Fragment implements View.OnKeyListener {
             if (mFolder.getCollectionType() != null && mFolder.getCollectionType() == CollectionType.MOVIES) {
                 sortOptions.put(7, new SortOption(getString(R.string.lbl_runtime), ItemSortBy.RUNTIME, SortOrder.ASCENDING));
             }
+            
+            sortOptions.put(8, new SortOption(getString(R.string.lbl_random), ItemSortBy.RANDOM, SortOrder.DESCENDING));
         }
 
         setDefaultGridRowCols(mPosterSizeSetting, mImageType);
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java
```diff
@@ -177,6 +177,8 @@ public void onCreate(Bundle savedInstanceState) {
             if (mFolder.getCollectionType() != null && mFolder.getCollectionType() == CollectionType.MOVIES) {
                 sortOptions.put(7, new SortOption(getString(R.string.lbl_runtime), ItemSortBy.RUNTIME, SortOrder.ASCENDING));
             }
+            
+            sortOptions.put(8, new SortOption(getString(R.string.lbl_random), ItemSortBy.RANDOM, SortOrder.DESCENDING));
```
nielsvanvelzen, line 181: The lbl_random string does not exist.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java
josephclaytonhansen: whoops, good catch. It does now

nielsvanvelzen (changes requested):
While this does seem to work on the initial page, it somewhat breaks after fetching more items. Because everything is random the subsequent pages can contain duplicate items and the app stops loading more pages when it thinks it has received all items in the library. This causes two issues:

1. Some items will never be shown
2. Some items will be shown multiple times

Especially the second one can cause issues because this means not all grid items are unique, which may break some code that restores focus etc.

I think the easy fix here is to limit display to just a single page when random sorting is used.

josephclaytonhansen:
Yeah, that makes sense, I'll add that tomorrow 👍🏻

nielsvanvelzen:
Closing as stale

PR #4679, opened 2025-05-23
Change/Fix Resume Button Behaviour on Series

### LINKED ISSUE #3363: UI for TV shows is illogical. Suggestions.

When i click on a TV show, i can select "play all" or "play next". If i have not started watching the show, having "play next" seems odd as there would be no next, considering i haven't started watching yet.
If i click "play all" it starts from episode 1, which is fine.

Instead of having "play all" and "play next", it would make more sense having a "play from beginning" and nothing else, having a "play next" on a show that hasn't been started makes no sense.
When a show has started, a new button should appear next to "play from beginning" saying "resume" and it resumes from the last played episode.

If i go watch something else, when i return to the TV show, there is no option to resume, there's "play all" and "play next" but if  i were in the middle of an episode "play next" isn't an option. 
Instead, to resume, i have to.
1. Click the show
2. Then click down and enter the show once again but this time showing individual episodes
3. Choose episode
4. Then click resume.

That's a lot of steps for something that should just be a button on the TV shows main page.

osiris86:
**Changes**
The "Play next up" Button now asks the user via a PopupMenu if the next episode should be resumed or started from the beginning if the next episode was played before:
image

Also if the series wasn't played before, the "Play next up" button disappears. For this to implement I had to include a request to the API to get the next episode and find out if it was already partially played and I'm quite unhappy with that approach. If you think there's a better way of doing things, I'm happy to refactor.

**Issues**
Fixes #3363


app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragmentHelper.kt
```diff
@@ -231,47 +232,72 @@
...
+				sortBy = setOf(
+					ItemSortBy.PARENT_INDEX_NUMBER,
+					ItemSortBy.INDEX_NUMBER,
+					ItemSortBy.SORT_NAME
+				),
+				limit = 1
+			).content
+		}
+		callback(episodes.items.firstOrNull())
+	} catch (err: ApiClientException) {
```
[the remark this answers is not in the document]
osiris86: Done

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragmentHelper.kt
```diff
@@ -231,47 +232,72 @@
...
 	}
 }
 
+fun FullDetailsFragment.showResumeMenu(
+	view: View,
+	nextUpEpisode: BaseItemDto
+) = popupMenu(requireContext(), view) {
+	item(getString(
+		R.string.lbl_resume_from,
+		TimeUtils.formatMillis((nextUpEpisode.userData!!.playbackPositionTicks / 10000) - resumePreroll)
```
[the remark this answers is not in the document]
osiris86: I replaced the magic number and introduced a new static. Also replaced the other occurrences of 10000 in FullDetailsFragment and FullDetailsFragmentHelper

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
```diff
@@ -257,7 +258,7 @@ public void run() {
                                 boolean resumeVisible = (mBaseItem.getType() == BaseItemKind.SERIES && !mBaseItem.getUserData().getPlayed()) || JavaCompat.getCanResume(mBaseItem);
                                 mResumeButton.setVisibility(resumeVisible ? View.VISIBLE : View.GONE);
                                 if (JavaCompat.getCanResume(mBaseItem)) {
-                                    mResumeButton.setLabel(getString(R.string.lbl_resume_from, TimeUtils.formatMillis((mBaseItem.getUserData().getPlaybackPositionTicks() / 10000) - getResumePreroll())));
+                                    mResumeButton.setLabel(getString(R.string.lbl_resume_from, TimeUtils.formatMillis((mBaseItem.getUserData().getPlaybackPositionTicks() / TICKS_IN_MILLISECONDS) - getResumePreroll())));
```
nielsvanvelzen, line 261: If you're going to change these you must use the `.ticks` extension getter, which from Java would be something like `TicksExtensionsKt.getTicks(timestamp)`

osiris86: I'm not sure if I understand ...
In `FullDetailFragmentHelper` I changed my code to calculate the position with the `.ticks` extension. This makes sense.
However, if I try to do the same in java, the `TicksExtensionsKt.getTicks()` returns a `long`. If I'm not mistaken this will be in nanoseconds and then has to be converted back to milliseconds for the `TimeUtils` class. I came up with two alternatives:

``` java
Duration duration = Duration.ofNanos(TicksExtensionsKt.getTicks(mBaseItem.getUserData().getPlaybackPositionTicks());
Duration prerollDuration = Duration.ofMillis(getResumePreroll());
mResumeButton.setLabel(getString(R.string.lbl_resume_from, TimeUtils.formatMillis(duration.minus(prerollDuration).toMillis())));
```
Here I'm using the `Duration` class to mimic something similar to what Kotlin does. To be honest, I think this makes things just much more complicated then it needs to be.

As an alternative I see something like this:
``` java
long position = TicksExtensionsKt.getTicks(mBaseItem.getUserData().getPlaybackPositionTicks()) / 1000;
mResumeButton.setLabel(getString(R.string.lbl_resume_from, TimeUtils.formatMillis(position - getResumePreroll())));
```
I think this is much cleaner than the above, but if I'm completely honest, I don't see the benefit compared to the current solution that just divides the ticks by 10000.

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragmentHelper.kt
```diff
@@ -231,47 +232,72 @@ fun FullDetailsFragment.populatePreviousButton() {
...
+					ItemSortBy.PARENT_INDEX_NUMBER,
+					ItemSortBy.INDEX_NUMBER,
+					ItemSortBy.SORT_NAME
+				),
+				limit = 1
+			).content
+		}
+		callback(episodes.items.firstOrNull())
+	} catch (err: ApiClientException) {
+		Timber.w("Failed to get next up items", err)
```
nielsvanvelzen, line 256: Throwable should be the first argument, not the second

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragmentHelper.kt
```diff
@@ -231,47 +232,72 @@ fun FullDetailsFragment.populatePreviousButton() {
 	}
 }
 
-fun FullDetailsFragment.resumePlayback() {
+fun FullDetailsFragment.getNextUpEpisode(callback: (nextUpEpisode: BaseItemDto?) -> Unit) = lifecycleScope.launch {
```
nielsvanvelzen, line 235: the callback is never called when the request fails. I think it would be better to call the callback with no next episode instead of showing the toast.

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragmentHelper.kt
```diff
@@ -231,47 +232,72 @@ fun FullDetailsFragment.populatePreviousButton() {
...
+fun FullDetailsFragment.resumePlayback(v: View) {
 	if (mBaseItem.type != BaseItemKind.SERIES) {
 		val pos = (mBaseItem.userData?.playbackPositionTicks?.ticks
 			?: Duration.ZERO) - resumePreroll.milliseconds
 		play(mBaseItem, pos.inWholeMilliseconds.toInt(), false)
 		return
 	}
 
-	val api by inject<ApiClient>()
+	getNextUpEpisode {
```
nielsvanvelzen, line 273: I understand why you did this (avoid duplicate code) but we should avoid callbacks in the Kotlin code.

nielsvanvelzen (changes requested):
Works fine during testing but got a few smaller comments left

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragmentHelper.kt
osiris86: what do you think of that? 
converted to a suspend function and (to be able to call it from java) added another function with a callback. pro: no duplication, con: two functions with almost the same name

nielsvanvelzen: I think this works although I'm not the biggest fan of the function names. I believe you can rename `getNextUpEpisodeCb` to just `getNextUpEpisode` because it has a different signature.

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragmentHelper.kt
osiris86: Done

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragmentHelper.kt
osiris86: Done

app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragmentHelper.kt
osiris86: True ... Changed the function name

nielsvanvelzen (approved):

Issue #4067, opened 2024-10-11
Refresh rate switching forces audio to be played in PCM


Kiara1250:
### Describe the bug

1. Enable refresh rate switching
2. Play a movie with audio codec other then PCM
3. The movie is played with PCM audio, it is not transcoded on the server and seems to be completely caused by androidtv (or exoplayer?)

I tried fixing this myself by disabling transcoding and investigating the source code but could not find the issue.

### Media info of the file

```shell
Most movies i play are HEVC and in a mkv container
```

### Application version

v0.17.6

### Device information

Nvidia Shield Pro (2017)

### Android version

Android 9.1.1

### Jellyfin server version

10.9.11

Cyruz143:
Verified this still happens on v18 client and 10.10.3 server.

frankko:
Confirmed as still an issue using Jellyfin server 10.10.3, Android TV client 0.18.2.

This has been driving me nuts all week! I couldn't figure out why two different Shield TV Pros were outputting LPCM audio without any server logs indicating transcoding. Sure enough, disabling refresh rate switching has restored my clients ability to play untouched audio.

labodj:
Same problem here, client 0.18.2 on Shield TV pro 2019

RyuzakiKK:
In the Jellyfin server side apparently nothing gets logged when this issue appears. From the server point of view, the file is simply direct playing.

Is there a way to force the client side to upload its logs, even if it is not crashing?

FTR, I'm also using Jellyfin server 10.10.3, client 0.18.2 and the Nvidia Shield TV pro 2019. I can reproduce the unexpected PCM audio issue when setting the "Refresh rate switching" to either "Scale on TV" or "Scale on device".

RolandRMG:
Same here. Could be a handshake Problem with hdmi. It switches the the Resolution Frame rate and Looses Connection for a Bit. Handshake List too. Sometimes i can get Dolby back After i Jump ahead in the Video, wich forces s New handshake

jjmirks:
I'm having this issue too

Server: 10.10.5
Client: 0.18.4
Using shield TV (2019)

ne0phyte:
[...]

PR #4686, opened 2025-05-27
Allow Exoplayer to invalidate the track selection on capabilities change

RyuzakiKK:
**Changes**
If the capabilities of a device change, Exoplayer can invalidate the current selection to adhere to the new list of capabilities.

This helps the Nvidia Shield and its content matching refresh rate option because, during the refresh rate switching period, the reported capabilities change. Without this flag, the device could reproduce audio as PCM instead of the expected passthrough.

**Issues**
Addresses https://github.com/jellyfin/jellyfin-androidtv/issues/4067, additional context https://github.com/androidx/media/issues/2258

I'm using "addresses" because I don't want to automatically close https://github.com/jellyfin/jellyfin-androidtv/issues/4067. To fully fix it, we also need to bump media3 to the upcoming release (1.6.1+) because we need https://github.com/androidx/media/commit/f825ad7e476c7190f21ff0bddb8e02ddf9a5a856 and https://github.com/androidx/media/commit/93cf82df151b16b4220d0c9f646abd45a3d1aed5. On that matter, we probably just need to wait for them to tag a new release (bumping media3 is not a blocker for this PR).

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
@@ -217,6 +217,7 @@ public class VideoManager {
                         .setAudioOffloadMode(TrackSelectionParameters.AudioOffloadPreferences.AUDIO_OFFLOAD_MODE_ENABLED)
                         .build()
                 )
+                .setAllowInvalidateSelectionsOnRendererCapabilitiesChange(true)
                 .build()
         );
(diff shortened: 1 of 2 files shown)


nielsvanvelzen (changes requested):
This change also needs to be made in the new playback code (ExoPlayerBackend.kt)

nielsvanvelzen:
Thanks for the PR and the effort you put into a good solution from the media3 side!

PR #4696, opened 2025-05-31
Show all Users of all Servers Option

KeksBombe:
**Changes**
Added an Option in the Usersettings, to display all Users of all Servers like username@servername. Makes switching Profiles accross Servers way easyer and it's not a big change :)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/auth/store/AuthenticationPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/auth/store/AuthenticationPreferences.kt
@@ -19,6 +19,7 @@ class AuthenticationPreferences(context: Context) : SharedPreferenceStore(
 
 		val sortBy = enumPreference("sort_by", AuthenticationSortBy.LAST_USE)
 		val alwaysAuthenticate = booleanPreference("always_authenticate", false)
+		val showAllServersUsers = booleanPreference("show_all_servers_users", false)
 
 		// Persistent state
 		val lastServerId = stringPreference("last_server_id", "")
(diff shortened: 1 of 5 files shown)


nielsvanvelzen:
Thanks for the pull request but I don't believe this is something we want. Within the authentication UI the server screen is for just a single server with related metadata (e.g. splash screen image, server disclaimer, server settings etc.). Showing users from different servers there would be confusing.
I also don't see much of a use-case here as normally you don't really use multiple Jellyfin servers.

KeksBombe:
Okay, thanks! My friend and I really missed the easy switching between our servers when we migrated from Plex. Currently, you have to switch the selected server and then select a profile. I thought that maybe more people have the same problem as us, and making this an option on the menu would be a good idea. Thanks for the response, though!

nielsvanvelzen:
I think some of the new user interface parts I'm currently working on might solve that issue for you. Here is a (early prototype) screenshot:

92400005c7

Instead of directly going to the user-select screen, the "profile" button in the toolbar shows this intermediate screen that has a quick switch feature built-in.

KeksBombe:
That is already looking very good! But will Profiles from other Servers I recently used also be shown? 🙃

nielsvanvelzen:
The quick switch will mix previously used users from multiple servers yes

KeksBombe:
Okay so it kinda does what my addition dit, just limiting the users to the recent 5. Than im looking forward to the update containing your changes and i hope i can someday provide an addition to this awesome project 😄

nielsvanvelzen:
My plan is to finish it for the next big app release (0.19). I'll close this pull request for now. For next time I'd recommend discussing your planned changes first via our android development chat (listed at jellyfin.org/contact).

PR #4724, opened 2025-06-14
Add option for bold subtitles (#4559)

### LINKED ISSUE #4559: Add settings option to make subtitles bold

### Problem description

There's no way to change the subtitle weight from normal to bold, and the client does not seem to respect the server subtitle weight setting. 

### Proposed solution

Please add an option to change between normal and bold subtitles. 

### Alternatives considered

Or possibly just make the client respect the server's subtitle weight setting?

### Additional information

_No response_

### Other sources

_No response_

EMohlin7:
Hello! This is my first pull request, so any feedback is welcome.

**Changes**
Adds a checkbox for setting all subtitles to bold text. The setting is located in Preferences -> Playback -> Subtitles

**Issues**
Fix #4559

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/preference/screen/PlaybackPreferencesScreen.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/preference/screen/PlaybackPreferencesScreen.kt
@@ -223,11 +223,11 @@ class PlaybackPreferencesScreen : OptionsFragment() {
 
 			checkbox {
 				setTitle(R.string.pref_subtitles_bold)
-				bind{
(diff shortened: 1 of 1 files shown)


EMohlin7:
I updated the implementation to use text weights instead of the default bold typeface. I set the regular weight as 400 and the bold weight as 700, as recommended in the [api reference](https://developer.android.com/reference/android/graphics/Typeface.html#create(android.graphics.Typeface,%20int,%20boolean)). 

Would it be of interest to implement a list that allow the user to select from the weights in the reference, instead of using a single checkbox for on or off?

Issue #2766, opened 2023-05-15
ExoPlayer PGSSUB implementation doesn't correctly render multiple on-screen captions


silentTeee:
### Describe the bug

Cross-post of https://github.com/jellyfin/jellyfin-android/issues/1045, as this issue affects all Android apps that uitlize ExoPlayer and enable PGSSUB native rendering. Issue was introduced with https://github.com/jellyfin/jellyfin-androidtv/pull/2690, and seems to be caused by https://github.com/google/ExoPlayer/issues/7458

### Media info of the file

```shell
Mob Psycho 100 S01E09
Container: mkv
Size: 4530 MB

Subtitle
Title: Forced Only - English - PGSSUB
Language: eng
Codec: PGSSUB
Default: No
Forced: Yes
External: No

Audio
Title: Surround 5.1 - English - TRUEHD - Default
Language: eng
Codec: TRUEHD
Layout: 5.1
Channels: 6 ch
Bitrate: 2623 kbps
Sample rate: 48000 Hz
Bit depth: 24 bit
Default: Yes
Forced: No
External: No

Video
Title: 1080p H264 SDR
Codec: H264
AVC: Yes
Profile: High
Level: 41
Resolution: 1920x1080
Aspect ratio: 16:9
Interlaced: No
Framerate: 23.976025
Bitrate: 26170 kbps
Bit depth: 8 bit
Video range: SDR
Video range type: SDR
Pixel format: yuv420p
Ref frames: 1
NAL: 4

Audio
Title: Surround 5.1 - English - Dolby Digital
Language: eng
Codec: AC3
Layout: 5.1
Channels: 6 ch
Bitrate: 448 kbps
Sample rate: 48000 Hz
Default: No
Forced: No
External: No

Audio
Title: Stereo - Japanese - TRUEHD
Language: jpn
Codec: TRUEHD
Layout: stereo
Channels: 2 ch
Bitrate: 839 kbps
Sample rate: 48000 Hz
Bit depth: 24 bit
Default: No
Forced: No
External: No

Audio
Title: Stereo - Japanese - Dolby Digital
Language: jpn
Codec: AC3
Layout: stereo
Channels: 2 ch
Bitrate: 192 kbps
Sample rate: 48000 Hz
Default: No
Forced: No
External: No

Subtitle
Title: English - PGSSUB
Language: eng
Codec: PGSSUB
Default: No
Forced: No
External: No
```

### Application version

0.15.7

### Device information

[...]

PR #4729, opened 2025-06-17
Add option to disable PGS direct play

linkandzelda91:
**Changes**
Put a checkbox in Preferences (cog icon on home screen) > Playback > Advanced playback preferences that controls whether the client will report to the server it supports embedded PGS subtitles. Default behavior is to not advertise support to the server, instead requiring PGS subs to be burned in.

This changes the current behavior which advertises PGS subtitle support with no option to remove.

**Issues**
Not sure I'd call it a "fix", more "workaround/bandaid" for https://github.com/google/ExoPlayer/issues/7458
https://github.com/androidx/media/issues/2534
#2766


linkandzelda91:
Initially I was just going to remove support entirely, but the awkward thing here being that if you only have media with one subtitle on the screen at a time, this is never an issue. So for people who have been happily enjoying their media with PGS subtitles without having to burn in, this is a negative change for them. I am hoping making it opt in with a note about the lack of support for multiple onscreen subtitles is a good compromise.

The original issue on the exoplayer repo is from 2020 with no progress. That repo is now deprecated so I reposted it on the new one and it ended up being marked low priority enhancement, so I don't think they plan on fixing it anytime soon.

linkandzelda91:
Also sort of now realizing I am not sure if this change only effects exoplayer, or if it'll impact external players too.

nielsvanvelzen (changes requested):
I understand there's some problems with the PGS rendering that basically requires this (advanced) option to be available until those issues are fixed upstream. However, we should not disable it by default, as there's been almost no reports of users encountering this issue. Implying this issue isn't that widespread.

linkandzelda91:
Thanks for the feedback.
I changed the default to be checked/enabled. testing seems good.

I also grabbed an external player to check my earlier concern about it having an impact on using external players. with VLC I dont even get an ffmpeg log or notification in the server dashboard that it's direct play. I assume it's bypassing everything anyway, as no server ffmpeg log = no way it can be burning in.

I updated the button title to reflect that this only effects the internal player.

...also search is broken when I build but Im pretty sure that has nothing to do with my changes it happens on the current master as well, not sure if it's me using the android tv x86 emulator in android studio or something else.

nielsvanvelzen (approved):
Thanks for the PR! I've made some final changes in the translations and pushed it in a final commit.

silentTeee:
Thank you for working on this! I am one of the rare weirdos who experienced this issue, but I could never find the time to learn enough to submit a PR.

Now I can fully enjoy my anime Blu-Rays again. 😊

PR #4731, opened 2025-06-18
Reduce dream mode memory usage

cameocoder:
**Changes**
Reduce memory usage by not preloading images and converting them to bitmaps.

When dream mode starts, the code was prefetching the images and converting them to bitmaps using Coil.  However, we can load the images on demand in the view using Coil AsyncImage instead.   That way the code doesn't need to create additional copies of the images and store the bitmaps in the LibraryShowcase data class.  The changes also address the StaticFieldLeak lint warning due to injecting Context into the ViewModel.

In my testing the behaviour of the dream mode seems to behave the same, but further testing could be performed to confirm.  Clearing app cache would be necessary to make sure there are no cached images and test the fade between scenes still works correctly.

**Issues**
I believe these changes fix the following stack trace that I see using latest master:
```
2025-06-18 19:22:23.765   868-1037  CoilTimber...mageLoader org.jellyfin.androidtv.debug         E  🚨 Failed - NullRequestData (Ask Gemini)
                                                                                                    coil3.request.NullRequestDataException: The request's data is null.
                                                                                                    	at coil3.RealImageLoader.execute(RealImageLoader.kt:109)
                                                                                                    	at coil3.RealImageLoader.execute(RealImageLoader.kt:77)
                                                                                                    	at org.jellyfin.androidtv.integration.dream.DreamViewModel$asLibraryShowcase$2$logoDeferred$1.invokeSuspend(DreamViewModel.kt:141)
                                                                                                    	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                                                                                                    	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
                                                                                                    	at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run(LimitedDispatcher.kt:124)
                                                                                                    	at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:89)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:586)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:820)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:717)
                                                                                                    	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:704)
```
The logoUrl is nullable, so the image request should not be attempted.  The AsyncImage has better handling for that.


nielsvanvelzen:
As mentioned already in chat, this is not a desired change. The screensaver explicitly preloads images to avoid flickering and this PR does not result in any memory saving as the imaged still needs to be loaded.

PR #4763, opened 2025-06-30
Moving logo randomly in the screensaver to prevent burn-ins

### LINKED ISSUE #1888: Move logo and clock randomly in the screensaver to prevent burn-ins on OLEDs

patricks1:
Adds a feature that randomly moves the Jellyfin logo in the screensaver so it does not burn in on plasma and OLED TVs.
- Modifies `DreamHeader` so that it takes a `contentKey` and randomly moves the Jellyfin logo's horizontal position when that `contentKey` changes. 
- Where `DreamView` calls `DreamHeader`, it provides the `content` of the title poster being displayed as the argument for `contentKey`. This makes it so when the app shows a new poster, the logo position changes. 
- `DreamHeader` takes a `fadeMillis` argument and is now an `AnimatedContent` so it can fade out and in smoothly at the same rate as the title poster.

**Relevant Issue**
Fixes #1888. In the interest of making the simplest possible change, I did not modify the clock. The user can turn off the clock to prevent it from burning in.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -16,6 +16,7 @@
  - [MajMongoose](https://github.com/majmongoose)
  - [Olaren15](https://github.com/Olaren15)
  - [dtrexler](https://github.com/dtrexler)
+ - [patricks1](https://github.com/patricks1)
 
 # Emby Contributors
 
--- a/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHeader.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/integration/dream/composable/DreamHeader.kt
@@ -1,8 +1,11 @@
 package org.jellyfin.androidtv.integration.dream.composable
 
+import androidx.compose.animation.AnimatedContent
 import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.animation.core.tween
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
+import androidx.compose.animation.togetherWith
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Row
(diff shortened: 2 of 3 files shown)


nielsvanvelzen:
In my testing this positioning feels a little too random.

patricks1:
@nielsvanvelzen can you clarify what you’d like me to change? I designed it to be fully random. I thought that was the goal.

Pentaphon:
> I did not modify the clock. The user can turn off the clock to prevent it from burning in.

Considering that the Google TV screensaver has the clock on the bottom right, I think it would be worthwhile to have the clock also shift between top right and bottom right by default. Most people with an OLED aren't going to look for an option to disable the clock, so it would be fine to randomly shift the clock as well. Just my 2 cents.

patricks1:
@Pentaphon I'd possibly be willing to work on that in a separate pull request. Once we get this one squared away, that'll inform me as to how I should go about coding the clock.

As an aside, though, if someone with an OLED is unconcerned with disabling the clock then they they're probably outside the target demographic of people who care about this feature.

Pentaphon:
> As an aside, though, if someone with an OLED is unconcerned with disabling the clock then they they're probably outside the target demographic of people who care about this feature.

I would argue that a slight majority of people who own OLEDs are blissfully unaware of the possibility of burn-in so I think it is always prudent for TV app developers to take every possible measure to prevent burn-in for the user so the clock switching between top right and bottom right by default is more than welcome. I can't tell you what to do with your PRs, of course. I am just saying that I think it is always the right thing to do, regardless of the ignorance of end-users.

nielsvanvelzen:
> @nielsvanvelzen can you clarify what you’d like me to change? I designed it to be fully random. I thought that was the goal.

Currently the logo is all-over the place and that just doesn't look right. The issue you linked is about adding better support for OLED displays to the screensaver. Moving the logo is one of the ways you could fix this, although it should still be somewhat in the top-left corner.

patricks1:
Okay, thanks. Currently, it's restricted to a single vertical position and allowed to randomly move horizontally. I'll make the change you suggested by selecting the horizontal starting position randomly within the left 30% of the screen and the vertical starting position from the top 30% of the screen.

nielsvanvelzen:
Closing as an alternative fix was merged via #4837

Issue #4114, opened 2024-10-29
Tune to live TV channel by clicking the channel name in the live TV guide


lewmiere:
### Describe the feature you'd like

When clicking the channel name in the live TV guide list, nothing happens. It would make sense when you click the channel name that it begins streaming the selected TV channel.

PR #4777, opened 2025-07-11
Start Live TV player when activating channel header

kylep09:
Enhance LiveTVGuide Issue #4114

**Changes**
Clicking on the channel name now tunes to the LiveTV channel.

**Issues**
Enhancement #4114

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideFragment.java
@@ -321,7 +321,7 @@ public class LiveTvGuideFragment extends Fragment implements LiveTvGuide, View.O
             case KeyEvent.KEYCODE_ENTER:
             case KeyEvent.KEYCODE_DPAD_CENTER:
                 if ((event.getFlags() & KeyEvent.FLAG_CANCELED_LONG_PRESS) == 0) {
-                    if (mSelectedProgramView instanceof ProgramGridCell) {
+                    if ((mSelectedProgramView instanceof ProgramGridCell) || (mSelectedProgramView instanceof GuideChannelHeader)) {
                         if (mSelectedProgram.getStartDate().isBefore(LocalDateTime.now()))
                             playbackHelper.getValue().retrieveAndPlay(mSelectedProgram.getChannelId(), false, requireContext());
                         else
--- a/app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideFragment.java
@@ -40,7 +40,6 @@ import org.jellyfin.androidtv.ui.ObservableScrollView;
 import org.jellyfin.androidtv.ui.ProgramGridCell;
 import org.jellyfin.androidtv.ui.ScrollViewListener;
 import org.jellyfin.androidtv.ui.navigation.ActivityDestinations;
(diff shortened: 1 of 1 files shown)


kylep09:
> While this works when there is EPG data, it won't when there isn't. It would make more sense to just play the channel in the video player instead of trying to play the current program.

I can implement this using the "Tune to Channel" feature. Though tracking it all down seems to implement the same code:

`playbackHelper.getValue().retrieveAndPlay(mSelectedProgram.getChannelId(), false, requireContext());`

I did test this feature as it is currently implemented with no program data. I deleted all my EPG data and the code gets down into the same conditional, with or without the EPG data. `if (mSelectedProgram.getStartDate().isBefore(LocalDateTime.now()))`

Would you still like to see it implemented via the 'Tune to Channel" feature?

nielsvanvelzen:
It would work similarly to activating a channel card in the initial Live TV screen which goes through ItemLauncher and then calls this snippet with the channel base item:

```java
playbackHelper.getValue().getItemsToPlay(context, response, false, false, new Response<List<BaseItemDto>>() {
  @Override
  public void onResponse(List<BaseItemDto> response) {
    playbackLauncher.getValue().launch(context, response);
  }
});
```

If you don't have the channel base item you should be able to do it the way you said.

kylep09:
I took a slightly different approach and tuned by the channels ID. Works well with or without EPG data as it is not dependent on program data.

nielsvanvelzen (changes requested):
Looking at the code you do have access to the channel's baseItem so I'd rather use that instead of calling retrieveAndPlay with the id. Because now it needs to refetch the channel from the server when you already have it, which is inefficient.

kylep09:
Ok, i've updated to use the itemLauncher directly by passing it the baseItem. Tested with and without EPG data and works on both.

app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideFragment.java
```diff
@@ -328,6 +330,13 @@ private boolean onKeyUp(int keyCode, KeyEvent event) {
                             showProgramOptions();
                         return true;
                     }
+                    else if (mSelectedProgramView instanceof GuideChannelHeader) {
```
nielsvanvelzen:
```suggestion
                    } else if (mSelectedProgramView instanceof GuideChannelHeader) {
```

is how our code style normally looks

app/src/main/java/org/jellyfin/androidtv/ui/livetv/LiveTvGuideFragment.java
```diff
@@ -328,6 +330,13 @@ private boolean onKeyUp(int keyCode, KeyEvent event) {
...
+                        // Tuning directly to a channel
+                        GuideChannelHeader channelHeader = (GuideChannelHeader) mSelectedProgramView;
+                        BaseItemDto pChannel = channelHeader.getChannel();
+                        Lazy<ItemLauncher> itemLauncher = inject(ItemLauncher.class);
+                        itemLauncher.getValue().launch(new BaseItemDtoBaseRowItem(pChannel),null,requireContext());
```
nielsvanvelzen: This has the same problem as before, the item launcher will refetch the item which is not necessary. You can just call the getItemsToPlay on playbackHelper function directly just like what happens within the item launcher.

kylep09: ah, i didnt notice itemLauncher doing the same fetching. I think we're on the same page now.

PR #4780, opened 2025-07-12
Support removing items from playback queue

### LINKED ISSUE #4516: Music - Remove from queue not working correctly

### Describe the bug

1. Launch JellyFin and select an album to play.
2. Add to queue or just play the album.
3. Goto the queue and select a queued up song and click on it and Remove from queue.
4. The first song that you remove might be removed from the queue display list.
5. Select a second song in the queue and remove from the queue.
6. The 2nd song might or might not be removed. The behavior is inconsistent, but whatever it is doing is not what is expected.

### Media info of the file

```shell

```

### Client logs

_No response_

### FFmpeg logs

```shell

```

### Application version

0.18.5

Google Play Store

### Device information

Google TV Streamer

### Android version

Android TV OS version 14

### Jellyfin server version

10.10.6

### Other sources

_No response_

nielsvanvelzen:
To integrate the new fancy playback code with old UI I moved the old MediaManager to an interface and implemented it on a best-effort basis. We still don't use the queue system in its intended way and have a dirty supplier implementation that is just a list of baseitems. However, the queue system is only made for those suppliers to grow in size while fetching and never shrink or grow after they stopped supplying. This did happen with the queue removal function causing very weird states in the app.

Those issues are now fixed, just took me a few hours 🫠.

**Changes**

- Add new AudioQueueBaseRowAdapter
  - Replacement for the ItemAdapter that was maintained by MediaManager
  - Works stand-alone (e.g. it manages its own items)
- Use queue entries instead of just base items in AudioQueueBaseRowItem
- Support _any_ adapter type in ItemLauncher
- Do not reset adapter when items change in AudioNowPlayingFragment
  - This results in cards animating on removal/addition/move/update instead of the row flashing
- Require queue entry for removeFromAudioQueue and playFrom functions in MediaManager
- Allow reading all queue suppliers in a queue
- Allow reading all (non-removed) fetched entries in a queue
  - This is used to update the AudioQueueBaseRowAdapter when entries are removed
- Add indexOf function to a queue
  - This is used to skip ahead in a queue (MediaManager.playFrom)
- Add support for removing entries from a queue
  - This will update internal caches and modify indices accordingly
  - Will automatically skip to next item if the removed item is currently playing
  - This is the primary change of this PR which required a few additional changes
- Update QueueService to always use "entry" instead of the old "item" term
- Update KeyProcessor for all other mentioned changes
- Change RewriteMediaManager.BaseItemQueueSupplier to be immutable
  - Queue additions just add new suppliers
  - Queue removals use the new removal functionality

**Issues**

Fixes #4516
Part of #1057

### DIFF
--- a/playback/core/src/main/kotlin/queue/QueueService.kt
+++ b/playback/core/src/main/kotlin/queue/QueueService.kt
@@ -23,20 +23,24 @@ import kotlin.math.max
 class QueueService internal constructor() : PlayerService(), Queue {
 	private val suppliers = mutableListOf<QueueSupplier>()
 	private var currentSupplierIndex = 0
-	private var currentSupplierItemIndex = 0
-	private val fetchedItems: MutableList<QueueEntry> = mutableListOf()
+	private var currentSupplierEntryIndex = 0
+	private val fetchedEntries: MutableList<QueueEntry> = mutableListOf()
+	private var removedEntries = 0
 
 	private var defaultOrderIndexProvider = DefaultOrderIndexProvider()
 	private var orderIndexProvider: OrderIndexProvider = defaultOrderIndexProvider
 	private var currentQueueIndicesPlayed = mutableListOf<Int>()
 
-	override val estimatedSize get() = max(fetchedItems.size, suppliers.sumOf { it.size })
+	override val estimatedSize get() = max(fetchedEntries.size, suppliers.sumOf { it.size } - removedEntries)
 
 	private val _entryIndex = MutableStateFlow(Queue.INDEX_NONE)
 	override val entryIndex: StateFlow<Int> get() = _entryIndex.asStateFlow()
 
 	private val _entry = MutableStateFlow<QueueEntry?>(null)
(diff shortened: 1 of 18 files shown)


nielsvanvelzen:
@coderabbitai full review (don't mess with with the PR description, you can just reply with your dumb summary)

playback/core/src/main/kotlin/queue/QueueService.kt
```diff
@@ -71,36 +75,71 @@ class QueueService internal constructor() : PlayerService(), Queue {
...
 			val supplier = suppliers[currentSupplierIndex]
-			val nextItem = supplier.getItem(currentSupplierItemIndex)
+			val nextEntry = supplier.getItem(currentSupplierEntryIndex)
 
-			if (nextItem != null) {
-				// Add item to cache and icnrease item index
-				fetchedItems.add(nextItem)
-				currentSupplierItemIndex++
+			if (nextEntry != null) {
+				// Add entry to cache and icnrease entry index
```
crobibero:
```suggestion
				// Add entry to cache and increase entry index
```

PR #4783, opened 2025-07-13
Refetch server information on authenticate when cached server version is unsupported

nielsvanvelzen:
This is mostly useful for users upgrading their server and then re-opening the app when it was active on their previous server version less than 10 minutes ago. Now, the app will recheck the server version before showing the "server version not supported" warning which should result in a smoother upgrade.

**Changes**
- Refetch server information on authenticate when cached server version is unsupported

**Issues**


nielsvanvelzen:
@coderabbitai review

Issue #3276, opened 2024-01-01
Adjustable skip time


schumi4:
### Describe the feature you'd like

Being able to adjust the skip time for left and right clicking would be very beneficial in my opinion. Something like 5, 10, 20 seconds.
The current system with the fixed times does not really make sense.

nielsvanvelzen:
You can already change the skip duration. It is not currently possible to do so within the app, but you can open Jellyfin in the browser and change "skip forward length" and "skip back length" in the playback settings.

schumi4:
That changes the behaviour of the skip buttons, I'm talking about simply pressing the left or right button on the remote during playback, sorry I was a bit unclear there.

Doing this it always skips to the nearest :00 or :30 seconds, which is not very practical.

nielsvanvelzen:
Thanks for clarifying. The user interface of the video player is largely managed by a third-party library right now and we have little control over its behavior. This is similar to #3154 (same cause).

schumi4:
I see, very unlucky. Thanks for clarifying and the work you are putting into this project!

Issue #3154, opened 2023-10-31
On seek video get paused on Fire TV


0x6d61726b:
### Describe the bug

When we seek in a video forward/backward with the Fire TV remote, the seek bar appears and the video playback pauses until the play button is pressed again using both, ExoPlayer and LibVLC. This is very annoying and leads to massive family complaints.
This issue is directly related to the closed issue #2399 (which was auto-closed and I am unable to re-open) where I provided additional information.

Is this an issue related to a misconfiguration on our clients I did during setup/configuration?
If not, is this issue possibly fixed with #2826 and is there a test version available?

In general Jellyfin and Jellyfin-AndroidTV is a great software I would like to continue using it for our family. I would also like to provide some support. Unfortunately, I have no experience with Java programming language, nor Android programming.

### Application version

0.15.12

### Device information

Amazon AFTR (FireTV Cube 2nd Gen)

### Android version

Fire OS 7.6.6.4

### Jellyfin server version

10.8.10.0

nielsvanvelzen:
This is intended behavior (it was programmed to work this way) but we should definitely change it. Unfortunately this behavior is not defined by us but a third-party library we use so we cannot change it until we rewrite the player user interface.

hmbqzidixiwc:
I find this to be an annoyance as well but I do understand it is not under the control of jellyfin.

As a workaround I do have a suggestion, would it be possible to detect seeking then re-instate the last playback.

If a seek occurs, check previous playback state, on seek completed restore playback state.

So if it were seeking when paused then pause, if it were playing then resume playing.

Or jellyfin is completely out of the loop during playback and none of that is available?

0x6d61726b:
keep open: issue still persists

0x6d61726b:
keep open: issue still persists (checked with v0.16.11)

sfyfq:
[...]

Issue #2592, opened 2023-03-14
Seeking does not respect preferences and is locked to 30 seconds


brettp:
### Describe the bug

Opening an issue as requested by @mueslimak3r for the problem as reported here: https://reddit.com/r/jellyfin/comments/xzyhv9/seek_is_stuck_to_30_seconds_when_using_google_tv/

1. Set seeking preferences for profile
2. Open a video
3. Try to seek
4. Observe seeking is locked at 30s intervals

### Application version

0.15.4

### Device information

CCWGTV

### Android version

Android 12, STTE.220920.015.H1.9408258

### Jellyfin server version

10.8.9

nielsvanvelzen:
Duplicate of #2529

Issue #2529, opened 2023-02-22
Iconsistent seek when using dpad buttons on remote


d8f6:
### Describe the bug

So, the onscreen buttons to rewind and fast-foward go ten seconds back or thirty seconds forward respectively, but when using the KEYCODE_DPAD_LEFT or KEYCODE_DPAD_RIGHT both go backwards or fowards thirty seconds only. When using fast-forward and rewind buttons on remote, things work as expected.

### Application version

0.15.3

### Where did you install the app from?

Sideloaded APK

### Device information

Nvidia Shield 2015

### Android version

LineageOS for microG [foster]

### Jellyfin server version

10.8.9

francislavoie:
Some of this is mentioned in https://github.com/jellyfin/jellyfin-androidtv/issues/460 as well.

francislavoie:
This is still a problem.

glat:
Same here on Shield Pro 2019 + App for Android TV from Google Play.

nielsvanvelzen:
Should be fixed in the next 0.16 beta

0-MegaMind-0:
@nielsvanvelzen  i'm on 0.16.1 (updated to the latest version available on google play store). The changelog indicates that this should be fixed in 0.16.0.

But trying this on my TV the timer is still fixed to default 30 seconds. The ui Settings->playback doesn't list any options to change the seek timings either. 
I even tried to change the timings for the same logged in user on web browser and change the playback seek timings there and clear data on Android TV and re login to see if the new timings will be synced but no changes.

PR #4803, opened 2025-07-19
Use user defined seek times, relative to current position, with optional confirmation

itayavra:
**Changes**

- Seeking now uses configurable forward/backward durations (5-30s) and operates relative to current position instead of snapping to predefined positions. 
- Add optional seek confirmation mode that previews position with thumbnails before applying changes. 
- Implement custom thumbnail preview system to support the new seeking behavior since it no longer uses the native player seek mechanism. 
- Add settings for backward seek time to complement the forward seek time, and seek confirmation toggle in playback preferences.

**Issues**
https://github.com/jellyfin/jellyfin-androidtv/issues/460
https://github.com/jellyfin/jellyfin-androidtv/issues/2529
https://github.com/jellyfin/jellyfin-androidtv/issues/2592
https://github.com/jellyfin/jellyfin-androidtv/issues/3276
https://github.com/jellyfin/jellyfin-androidtv/issues/3154
https://github.com/jellyfin/jellyfin-androidtv/issues/2399


itayavra:
@nielsvanvelzen I created a new PR with better understanding of the codebase and hopefully a much cleaner solution
image

itayavra:
@nielsvanvelzen just clarifying this is ready for review on my part

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -252,6 +252,12 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
   		 * Enable PGS subtitle direct-play.
 		 */
 		var pgsDirectPlay = booleanPreference("pgs_enabled", true)
+
+		/**
+		 * Require confirmation before seeking to allows users to preview seek positions before confirming them.
+		 * When disabled, D-pad left/right will seek immediately.
+		 */
+		var seekConfirmationRequired = booleanPreference("seek_confirmation", true)
```
nielsvanvelzen, line 260: We should choose one behavior for the seek control and stick with that. No preferences to complicate it necessary.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/ThumbnailPreviewHandler.java
```diff
@@ -0,0 +1,75 @@
...
+import android.view.View;
+import android.view.animation.Animation;
+import android.view.animation.AnimationUtils;
+import android.widget.ImageView;
+
+import androidx.leanback.widget.PlaybackSeekDataProvider;
+
+import org.jellyfin.androidtv.R;
+
+public class ThumbnailPreviewHandler {
```
nielsvanvelzen: New classes must be written in Kotlin

app/src/main/res/layout/vlc_player_interface.xml
```diff
@@ -116,6 +116,18 @@
             app:layout_constraintTop_toTopOf="parent" />
     </androidx.constraintlayout.widget.ConstraintLayout>
 
+    <ImageView
```
nielsvanvelzen: New UI must be written in compose unless there is a good reason not to.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/ThumbnailPreviewHandler.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public ThumbnailPreviewHandler(PlaybackSeekDataProvider seekProvider,
+                                   ImageView thumbnailView,
+                                   long skipForwardLength) {
+        this.seekProvider = seekProvider;
+        this.thumbnailView = thumbnailView;
+        this.skipForwardLength = skipForwardLength;
+    }
+
+    public void updateThumbnailPreview(long previewSeekPosition) {
```
nielsvanvelzen: This doesn't properly cancel already loading images causing race conditions.

nielsvanvelzen (changes requested):
Initial review based on a quick look over the code and testing in emulator. I'm currently unsure if this is the correct approach or if we should replace the entire seekbar UI to avoid hacking into leanback code.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
itayavra: To me, the intuitive behavior would be to pause the video when seeking, especially when thumbnail previews can be utilized, this is what I'm used to from netflix, youtube (well, smart tube), etc.
However, I saw that there were some issues open specifically about that by users that prefer the other behavior (https://github.com/jellyfin/jellyfin-androidtv/issues/2399, https://github.com/jellyfin/jellyfin-androidtv/issues/3154, ...). 
I can leave only one option, but I would consider giving the option to choose, since it seems like it would allow more users to feel at home/enjoy the behavior they're used to (especially given supporting both behaviors in the code was pretty easy) 🙏 .
Anyway, let me know what you decide.

itayavra:
> Initial review based on a quick look over the code and testing in emulator. I'm currently unsure if this is the correct approach or if we should replace the entire seekbar UI to avoid hacking into leanback code.

Not sure what the implication/complexity of that would be since I'm too new here, but if this is not something that is planned anytime soon then perhaps one does not negate the other (meaning this PR as the short term solution)?

I must say I am pretty shocked this isn't the native leanback behavior and every client has to implement this separately though.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/ThumbnailPreviewHandler.java
itayavra: @nielsvanvelzen I thought about this issue and figured there are 2 possible race conditions here:
1. Multiple requests for thumbnails/tiles from the same sheet (e.g. tiles 0-99 all fetching the same tile sheet simultaneously)
2. Callbacks returning in wrong order, showing outdated thumbnails during fast scrubbing

For the first issue - I think the `ImageLoader` in `CustomSeekProvider` handles caching/deduplication network requests? Although even if not then the tile sheets are relatively small, so this shouldn't be too heavy on the network. 
As an optimization I added a prefetch for the sheet that includes the tiles near the starting position of the media.

For the second issue (which I assume you were referring to) - I tested simply canceling all previous requests, but this creates terrible UX during fast scrubbing since very few thumbnails actually get shown while the user is moving.

In the end I chose the following, which improved performance a lot in my tests:
- Ignore callbacks from older requests (meaning, we already showed a thumbnail from a newer request)
- Ignore stale callbacks that are 20+ requests behind (this threshold gave the best feel in my opinion)

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/ThumbnailPreviewHandler.java
itayavra: Done

app/src/main/res/layout/vlc_player_interface.xml
itayavra: Done

itayavra:
@nielsvanvelzen how would you like to proceed with this?

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/ThumbnailPreviewHandler.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    fun updateThumbnailPreview(previewSeekPosition: Long) {
+        val currentRequestNumber = requestNumber.incrementAndGet()
+        
+        val thumbnailCallback = object : PlaybackSeekDataProvider.ResultCallback() {
+            override fun onThumbnailLoaded(bitmap: Bitmap?, index: Int) {
+                if (bitmap == null) return
+                
+                // Ignore requests that are older than what we last showed
+                if (currentRequestNumber <= lastShownRequestNumber.get()) return
```
lgruen, line 38: This guard is the seek-preview "black box" bug — the real thumbnail never replaces the placeholder.

`CustomSeekProvider.getThumbnail()` fires the `ResultCallback` **twice** for a single request: first from coil's `onStart` with the placeholder bitmap (the `black_transparent_light` one added in #4946), then from `onSuccess` with the real frame — both carrying the same `currentRequestNumber`.

The `onStart` placeholder arrives first, passes this guard, and sets `lastShownRequestNumber = currentRequestNumber` below. When `onSuccess` then delivers the real frame for the **same** request, `currentRequestNumber <= lastShownRequestNumber` is now true, so it returns early and the real thumbnail is discarded — leaving a permanent translucent-black box (deterministic on any cache-miss, since `onStart` always precedes `onSuccess`).

The comment above says *"older than what we last showed"*, but `<=` also rejects the *equal* (same-request) `onSuccess`. Strict `<` matches that intent — a request can still replace its own placeholder, while genuinely older requests are still dropped during fast scrubbing:

```suggestion
                if (currentRequestNumber < lastShownRequestNumber.get()) return
```

(The native leanback scrub path is unaffected, since it uses leanback's own `ResultCallback`, which renders whatever bitmap it receives.) Confirmed on an Nvidia Shield build cherry-picking this PR onto current `master` — this one-line change restores the previews.

lgruen:
For anyone landing here who just wants **VLC-style D-pad seeking** on Android TV today rather than waiting on this PR — I ended up going a smaller, different route that's been working well on an Nvidia Shield, in case it's useful to others.

Rather than a custom seek overlay, it's a small patch on current `master` (no new UI, no settings), in `CustomPlaybackOverlayFragment`'s key interceptor:

- **Short press** left/right → exact ±skip via `PlaybackController.fastForward()/rewind()`: instant, keeps playing, no UI; rapid taps stack.
- **Hold** left/right → reveal leanback's **native** seekbar/timeline (with its trickplay thumbnail, full-res frame behind, keeps playing). On **key release**, read the scrubbed position from `getControlsRow().getCurrentPosition()` and `seek()` there directly, then hide the controls — so holding scrubs with trickplay and letting go lands you there, no separate confirm, and nothing pauses.

Reusing the built-in seek UI sidestepped the fullscreen-upscaling and pause behaviours, and it's only a few dozen lines.

Branch (off clean upstream `master`): https://github.com/lgruen/jellyfin-androidtv/tree/release-vlc-seek — logic is in [`CustomPlaybackOverlayFragment.java`](https://github.com/lgruen/jellyfin-androidtv/blob/release-vlc-seek/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java). It's a debug/personal build, not PR-ready, but happy to turn it into a proper PR if there's interest.

Thanks @itayavra for kicking this off.

nielsvanvelzen:
Closing as stale (unresolved merge conflicts for months)

PR #4804, opened 2025-07-20
Remove WebSocket subscriptions when process is inactive

nielsvanvelzen:
We have two classes ("services") that listen for WebSocket events in the background; the SocketHandler (app) and PlaySessionSocketService (new playback module). Until now those would subscribe when initialized and simply never unsubscribe.
This is not the desired behavior. Instead, we want them to unsubscribe when the app is inactive. With this PR we now use the ProcessLifecycle to gracefully unsubscribe when the app is closed (no more visible activities).

**Changes**
- Remove WebSocket subscriptions when process is inactive

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/eventhandling/SocketHandler.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/eventhandling/SocketHandler.kt
@@ -4,10 +4,14 @@ import android.content.Context
 import android.media.AudioManager
 import android.os.Build
 import android.widget.Toast
+import androidx.lifecycle.Lifecycle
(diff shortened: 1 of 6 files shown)


nielsvanvelzen:
@coderabbitai review please. Also, what do you think about changing the PlayerService class to use lifecycles by default, and then some kind of way to indicate if an implementation is foreground (bound to process lifecycle) or background only? If you think it's a good idea please open a pull request implementing it!

nielsvanvelzen:
@coderabbitai Yup open the PR, I'm curious what kind of AI slop you'll create! Also approve this one please 😉

crobibero (approved):

PR #4814, opened 2025-07-23
Run Detekt from root module only

nielsvanvelzen:
Unfortunately GitHub made a breaking change in how they process SARIF files without properly communicating it (I follow the release blog RSS feed and I didn't know about this until they made the actual change). This PR updates the way we lint using Detekt so all results are in a single run so GitHub is happy again.

This change must be backported if we do another 0.18 release.

**Changes**

- Run Detekt from root module only

**Issues**

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -16,20 +16,25 @@ java {
 	}
 }
 
-subprojects {
-	// Configure linting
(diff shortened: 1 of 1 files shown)


crobibero (changes requested):
Make the linter happy

PR #4819, opened 2025-07-27
Fix wrong Trickplay thumbnail displayed while seeking

tal-sarid:
**Changes**
Clear Trickplay thumbnail in current index while starting the request for a new one.

**Issues**
[I didn't find any issue regarding this one and I also didn't open one myself because it is an unsupported feature]
Instead of seeing the same thumbnail over and over again while seeking, now the old thumbnails are being cleared until the right image is loaded.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomSeekProvider.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomSeekProvider.kt
@@ -42,6 +42,8 @@ class CustomSeekProvider(
 		val currentRequest = imageRequests[index]
 		if (currentRequest?.isDisposed == false) currentRequest.dispose()
 
+		callback.onThumbnailLoaded(null, index)
+
 		val item = videoPlayerAdapter.currentlyPlayingItem
 		val mediaSource = videoPlayerAdapter.currentMediaSource
 		val mediaSourceId = mediaSource?.id?.toUUIDOrNull()


tal-sarid:
> The caller of the getThumbnail function (leanback) already provides caching and invalidation so there is no need to manually clear the thumbnail. In fact, the callback should never be invoked twice.

Maybe I did it wrong, but it does fixes the said issue on my end. Can you try reproducing the bug in your end with and without this fix?

The steps to reproduce the bug on my setups are:
1. Seek once and stop until all thumbnails in view are loaded and visible.
2. Press right without releasing for maximum seeking speed.
3. Thumbnails on the right side (depending on loading speed and streamer performance) start to display duplicated older images which is quiet annoying. This happens to me on NVIDIA Shield TV, Google TV Streamer and Chromecast with Google TV 4K.

Clearing the image in the newer indexes before loading them seem to resolve this, now only the images to the left are loading correctly while the right ones are transparent until you stop and wait for them to load, which is what I think should happen in the first place.

If I did it wrong, even if it is fixing the issue, I'll be more than happy to do it "the right way" if there is one.

tal-sarid:
OK, I did some digging and I found the root cause of the issue I'm trying to fix and why the added call fixes it. I'll also try to explain why my fix is correct...

From the documentation of `getThumbnail`:
> Called to get thumbnail bitmap. This method is called on UI thread. When provider finds cache bitmap, it may invoke PlaybackSeekDataProvider.ResultCallback.onThumbnailLoaded(Bitmap, int) immediately. Provider may start background thread and invoke PlaybackSeekDataProvider.ResultCallback.onThumbnailLoaded(Bitmap, int) later in UI thread. The method might be called multiple times for the same position, PlaybackSeekDataProvider must guarantee to replace pending PlaybackSeekDataProvider.ResultCallback with the new one.

It doesn't explicitly prohibit calling the callback twice but only to never use an older callback for the same index, which is not what I'm doing here.

Secondly, I believe that this snippet is what triggering the `getThumbnail` from Leanback when seeking back and forward:
```java
    void updateThumbsInSeek(int thumbHeroIndex, boolean forward) {
            if (mThumbHeroIndex == thumbHeroIndex) {
                return;
            }

            final int totalNum = mThumbsBar.getChildCount();
            if (totalNum < 0 || (totalNum & 1) == 0) {
                throw new RuntimeException();
            }
            final int heroChildIndex = totalNum / 2;
            final int start = Math.max(thumbHeroIndex - (totalNum / 2), 0);
            final int end = Math.min(thumbHeroIndex + (totalNum / 2), mPositionsLength - 1);
            final int newRequestStart;
            final int newRequestEnd;

            if (mThumbHeroIndex < 0) {
                // first time
                newRequestStart = start;
                newRequestEnd = end;
            } else {
                forward = thumbHeroIndex > mThumbHeroIndex;
                final int oldStart = Math.max(mThumbHeroIndex - (totalNum / 2), 0);
                final int oldEnd = Math.min(mThumbHeroIndex + (totalNum / 2),
                        mPositionsLength - 1);
                if (forward) {
                    newRequestStart = Math.max(oldEnd + 1, start);
                    newRequestEnd = end;
                    // overlapping area directly assign bitmap from previous result
                    for (int i = start; i <= newRequestStart - 1; i++) {
                        mThumbsBar.setThumbBitmap(heroChildIndex + (i - thumbHeroIndex),
                                mThumbsBar.getThumbBitmap(heroChildIndex + (i - mThumbHeroIndex)));
                    }
                } else {
                    newRequestEnd = Math.min(oldStart - 1, end);
                    newRequestStart = start;
                    // overlapping area directly assign bitmap from previous result in backward
                    for (int i = end; i >= newRequestEnd + 1; i--) {
                        mThumbsBar.setThumbBitmap(heroChildIndex + (i - thumbHeroIndex),
                                mThumbsBar.getThumbBitmap(heroChildIndex + (i - mThumbHeroIndex)));
                    }
                }
            }
            // processing new requests with mThumbHeroIndex updated
            mThumbHeroIndex = thumbHeroIndex;
            if (forward) {
                for (int i = newRequestStart; i <= newRequestEnd; i++) {
                    mSeekDataProvider.getThumbnail(i, mThumbResult);
                }
            } else {
                for (int i = newRequestEnd; i >= newRequestStart; i--) {
                    mSeekDataProvider.getThumbnail(i, mThumbResult);
                }
            }
            // set thumb bitmaps outside (start , end) to null
            for (int childIndex = 0; childIndex < heroChildIndex - mThumbHeroIndex + start;
                    childIndex++) {
                mThumbsBar.setThumbBitmap(childIndex, null);
            }
            for (int childIndex = heroChildIndex + end - mThumbHeroIndex + 1;
                    childIndex < totalNum; childIndex++) {
                mThumbsBar.setThumbBitmap(childIndex, null);
            }
        }
```

```java
    PlaybackSeekDataProvider.ResultCallback mThumbResult =
                new PlaybackSeekDataProvider.ResultCallback() {
                    @Override
                    public void onThumbnailLoaded(Bitmap bitmap, int index) {
                        int childIndex = index - (mThumbHeroIndex - mThumbsBar.getChildCount() / 2);
                        if (childIndex < 0 || childIndex >= mThumbsBar.getChildCount()) {
                            return;
                        }
                        mThumbsBar.setThumbBitmap(childIndex, bitmap);
                    }
        };
```

From this code, it seems like bitmaps outside of the view are the only ones being explicitly cleared. But the left-most or right-most bitmap that is in view are not being assigned but only being requested with a call to `getThumbnail`. That's what causing a brief duplication of the previous slot image being displayed until `getThumbnail` are replacing it.

My fix clears those image slots explicitly from the `getThumbnail` calls synchronously (immediately) before the `ImageRequest`'s asynchronous handling there, which takes time because of the network fetching and/or cropping needed (when there is a cache miss).

From looking at the source and documentation, I don't see anything wrong that I did but maybe I can improve the current fix by doing the clearing in the `onStart` (and `onError` when the image is still in view but failed to be fetched) callback instead:
```kotlin
target(
    onStart = { _ -> callback.onThumbnailLoaded(null, index) },
    onError = { _ -> callback.onThumbnailLoaded(null, index) },
    onSuccess = { image ->
        val bitmap = image.toBitmap()
	callback.onThumbnailLoaded(bitmap, index)
    },
)
```

What do you think?

nielsvanvelzen (approved):

incon:
I still have this issue on v0.18.11

tal-sarid:
> I still have this issue on v0.18.11

That's because it was merged after that version was released.

PR #4822, opened 2025-07-29
Fixed playback overlay to display the currently playing items name and episode for Live TV

### LINKED ISSUE #4781: Add a "Now Playing" Button to Live TV View

### Problem description

Currently, when watching a Live TV channel, it's difficult to determine which media item is actively playing. Selecting the channel selector or guide often displays information from a previous media item rather than the current one. This leads to confusion and disrupts the viewing experience. It would be especially helpful when the TV channel is playing short media items like music videos.

### Proposed solution

Introduce a "Now Playing" button or persistent indicator in the Live TV interface. This could show the exact media item currently playing on the active channel with minimal metadata. Enough to identify what is playing. The display could also disappear after a few seconds.

### Alternatives considered

- Updating the channel selector to more accurately reflect the current media item.

- Manually checking logs or schedules, which is cumbersome and not user-friendly during playback.

### Additional information

Just bought you a bug. Thanks for your hard work!

### Other sources

_No response_

kylep09:
**Changes**
Updated the Playback overlay to display the name of the show and episode when watching live tv. This also updates the progress bar and end time when the current show ends and a new one starts on the same channel.

**Issues**
Fixes #4781


app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -1248,6 +1279,13 @@ public void updateDisplay() {
             if (current.getType() == BaseItemKind.EPISODE) {
                 binding.itemTitle.setText(current.getSeriesName());
                 binding.itemSubtitle.setText(BaseItemExtensionsKt.getDisplayName(current, requireContext()));
+            } else if (current.getType() == BaseItemKind.TV_CHANNEL) {
+                binding.itemTitle.setText(current.getName());
+                if (current.getCurrentProgram() != null) {
+                    String nowPlaying = current.getCurrentProgram().getName() + " - " + current.getCurrentProgram().getEpisodeTitle();
```
nielsvanvelzen: This string seems to be unused?

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -1248,6 +1279,13 @@ public void updateDisplay() {
...
+                    binding.itemSubtitle.setText(current.getCurrentProgram().getName() + " - " + current.getCurrentProgram().getEpisodeTitle());
```
nielsvanvelzen: This would display "wrong' when there is no episode title.

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
```diff
@@ -612,6 +617,31 @@ public boolean onKey(View v, int keyCode, KeyEvent event) {
                     //and then manage our fade timer
                     if (mFadeEnabled) startFadeTimer();
                 }
+
+                if ( (playbackControllerContainer.getValue().getPlaybackController().isLiveTv()) && !mGuideVisible && keyCode != KeyEvent.KEYCODE_BACK ) {
+                    //do some stuff here
```
nielsvanvelzen: This comment is not really descriptive of what is happening here. What is the purpose of this check on EVERY key press??

app/src/main/java/org/jellyfin/androidtv/ui/livetv/TvManager.java
```diff
@@ -239,4 +239,8 @@ private static void addRow(Context context, List<BaseItemDto> timers, Presenter
         ListRow scheduleRow = new ListRow(new HeaderItem(TimeUtils.getFriendlyDate(context, timers.get(0).getStartDate(), true)), scheduledAdapter);
         rowAdapter.add(scheduleRow);
     }
+    public static BaseItemDto getChannelByID(UUID channelID) {
```
nielsvanvelzen, line 243: Missing an empty line before this

nielsvanvelzen (changes requested):
Initial review

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
kylep09: Removed unused string.

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
kylep09: updated to check if episode title is null first.

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
kylep09: Updated comment. I chose to implement this function here because i could not find any other mechanism for making this check. This function fetches the currently playing item, and if its changed since we first started playing this channel, update the currently playing item. 

At the moment the currently playing item does not update as the channel continues playing. So if you're tuned to it for 2 hours, the currently playing item is still stuck on the original item with no good way to update it. I think updating the item only really matters when you want to view the overlay information. So i chose this action to update it.

Any suggestion for a better trigger mechanism to update the currently playing item?

app/src/main/java/org/jellyfin/androidtv/ui/livetv/TvManager.java
kylep09: Added empty line

app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
nielsvanvelzen: Ok I understand now. We have a reference to the base item of the channel but we want to show the information of the current program, which can/will change during playback.

To be perfectly honest, I think we should avoid introducing logic for that refreshing in the current player implementation as it's extremely messy right now. So just showing the channel title without program info would be the best we can do for now.

kylep09: Bummer. I have an idea for an EPG event listener type of approach. I'll see if i can get i working and see what you think about that sort of approach instead. Though it might make the player implementation slightly more messy :grimacing:

PR #4835, opened 2025-08-03
Target Android 16 (API 36)

nielsvanvelzen:
Starting in August 2025 Google enforced API 35 or newer on GPlay, they'll bump it to 36 in the future so better make sure we're ahead so we don't need to change this during 0.19 patch releases.

**Changes**
- Target Android 16 (API 36)
- Remove deprecated screenOrientation attributes
  - This allows the app to be used in portrait mode. Don't do this.

**Issues**


crobibero (approved):

Dnkhatri:
I hope you are able to keep compatibility with older android versions as long as possible

nielsvanvelzen:
> I hope you are able to keep compatibility with older android versions as long as possible

#4836 🙈

Dnkhatri:
> > I hope you are able to keep compatibility with older android versions as long as possible
> 
> #4836 🙈

Well ouch I just checked my device is android 6 and api 23 which will loose support in next year 2026 so the next android tv will be the last version I can use I guess. A lot of old media boxes with non google android are on Android 6 as that was the version where google came out with android tv UI if recall correctly.

PR #4865, opened 2025-08-16
Add new video player UI

nielsvanvelzen:
This PR adds the first (of many) changes for the new (video-)player user interface. It's built in two layers so everything is reusable for different player types. This will eventually be used for music, audiobooks, photos, live tv etc. as well. With this base it will be easier to add features as the UI can be developed at the same time.

This is not the final design. Things will change.

Currently the UI implements the following features:

- Display video with correct aspect ratio
- Display subtitles with correct aspect ratio
  - They are currently rendered on top of the controls, this may change in the future
- Display controls on any key press
  - Auto-hide after 5 seconds or on back press
- Play/pause control
- Rewind and fast forward controls
- Seekbar (already backported in #4844)

**Changes**

- Add composables for video and subtitle surfaces
- Add compose based video player

**Screenshots**

Big Buck Bunny

**Issues**

Part of #1057

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/player/video/VideoPlayerControls.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/player/video/VideoPlayerControls.kt
@@ -1,0 +1,127 @@
+package org.jellyfin.androidtv.ui.player.video
+
+import androidx.compose.animation.AnimatedContent
+import androidx.compose.foundation.focusGroup
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.focus.focusRequester
+import androidx.compose.ui.focus.focusRestorer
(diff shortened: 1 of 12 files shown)


crobibero (approved):

app/src/main/java/org/jellyfin/androidtv/ui/player/video/VideoPlayerControls.kt
```diff
@@ -0,0 +1,127 @@
...
+					)
+				}
+			}
+		}
+	}
+}
+
+@Composable
+private fun RewindButton(
+	playbackManager: PlaybackManager,
```
cameocoder, line 107: For all of your controls you should hoist the playbackManager actions to be handled by the caller.

```
@Composable
fun RewindButton(
    onRewind: () -> Unit,
    modifier: Modifier = Modifier,
    enabled: Boolean = true
) {
    IconButton(
        onClick = onRewind,
```

Then in the caller, handle it as:
```
RewindButton(onRewind = {playbackManager.state.rewind()})
or
RewindButton() { playbackManager.state.rewind() }
```

I would suggest removing the playbackManager and add the onAction arguments all the way up to the VideoPlayerScreen.  

The problem here is that playbackManager is unlikely to be stable which will result in unnecessary recompositions.  You should be able to observe that in the Android Studio Layout Inspector.

It also makes these composables "dumb ui components" but also previewable, testable and reusable.

nielsvanvelzen: The playbackManager instance will never change, it's constructed once on app start and re-used everywhere. This shouldn't make any difference in performance. Although admittedly, I have not verified this as performance was not a goal of these changes right now. This is just to have some initial UI so I can more easily work on other features in the playback code itself. The backing composables for this UI are all separated (IconButton, PlayerOverlayLayout, PlayerHeader etc.) already.

PR #4866, opened 2025-08-17
SkipOverlayView: adaptive auto-hide (80 % of remaining segment)

queukat:
### What’s changed
* The skip overlay now hides automatically after 80 % of the time remaining to the target segment, providing a smoother UX for both short and long segments.
* Fallback to the existing `AskToSkipAutoHideDuration` when the remaining duration is zero or negative.
* Added `kotlin.time.times` import to support `remaining * 0.8` expression.

### Why
Static timeout occasionally hid the overlay too early (short segments) or too late (very long segments).  
Adapting the delay to the segment’s actual length keeps the overlay visible long enough to act without overstaying.

### Implementation notes
* Logic is isolated inside the existing `LaunchedEffect`; no public API changes.
* UI, layout, and property names remain untouched to keep the diff minimal.
* Verified on device & emulator with segments from 5 s to 10 min — overlay hides as expected, no regressions observed.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/SkipOverlayView.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/SkipOverlayView.kt
@@ -35,6 +35,7 @@ import org.jellyfin.androidtv.ui.base.Text
 import org.jellyfin.androidtv.ui.playback.segment.MediaSegmentRepository
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.milliseconds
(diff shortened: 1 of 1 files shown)


nielsvanvelzen:
Was this made with AI?

queukat:
yep

nielsvanvelzen:
Ok thanks for confirming

queukat:
did I miss any policy? the changes here are tested, not just copied code from the internet

nielsvanvelzen:
We do not accept AI generated code. It's a waste of time for everyone involved.

PR #4872, opened 2025-08-19
add calculatable duration to skip button

queukat:
**Changes**
replace 8 seconds overlay with actual duration

**Issues**
N/A


nielsvanvelzen:
As mentioned previously, we do not accept AI code and this isn't even a wanted change.

queukat:
em, these changes I made myself. the problem exists, here is the solution. You, kind of, accept requests. let it seem to you like a small thing, and I solved it. it's not a small thing for me. people created products, like the creators of the "intro-skipper" plugin, which I use every day. it perfectly complements jellyfin for me. why not provide them with better integration? if I'm now with a "black mark", well, then remove all those changes I made to the project

PR #4876, opened 2025-08-24
Remove libass preference in non-development builds

nielsvanvelzen:
Thanks to @JPVenson reminding me about this. We've recently added the libass-android library to allow direct-play of the ssa and ass formats for subtitles. Unfortunately the library has been plagued with memory issues causing the app to crash and has a non-transparent release pipeline. So far it doesn't appear that work is being done on these issues.
This PR removes the support from the UI in releases (e.g. upcoming 0.19 version) to avoid users creating issues even though we already tell them no support is provided.

**Changes**
- Move the "assDirectPlay" preference in the developer preference UI to require a development build (e.g. debug build without version information)

**Issues**


JPVenson:
IMG-20170803-WA0002

PR #4885, opened 2025-08-26
Trim whitespace from search query

mkst:
**Changes**
Trim any leading/trailing whitespace from the user's input before performing the search - my phone adds a trailing space when using autocomplete, e.g. `"movie "` which does not match anything in my library.

NOTE: I didnt bother `.trim()`-ing the `SearchVoiceInput` query text as I assume that wouldn't have any leading/trailling whitespace, but I can update the PR if it makes sense.

**Issues**
[This](https://github.com/jellyfin/jellyfin/pull/11896) is an unmerged/unfinished PR for performing the trim server-side where it was suggested that the trim be performed client-side.


nielsvanvelzen (changes requested):
I think it's good to do this stripping client side but it would need to happen in the view model itself so the behavior is consistent and we don't need to maintain multiple places to modify the string.

nielsvanvelzen:
Thanks for the PR!

PR #4886, opened 2025-08-26
Calculate playback progress using derived state

cameocoder:
**Changes**
Calculate playback progress using derived state in order to prevent recompositions

The PlayerSeekbar was being recomposed on every frame due to the calculation of the progress being "unstable" from the perspective of the Seekbar Composable.

Layout inspector screenshots were taken after ~20s with a change to increase the player overlay visibility timeout.

Layout inspection before:
Screenshot 2025-08-26 at 3 31 38 PM

The solution is to remember the progress calculation and pass that to the Seekbar.

Layout inspector after:
Screenshot 2025-08-26 at 3 34 47 PM
Notice the number of recompositions have been reduced and the recompositions of the Seekbar have been moved to the "skipped" column.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/player/base/PlayerSeekbar.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/player/base/PlayerSeekbar.kt
@@ -2,6 +2,7 @@ package org.jellyfin.androidtv.ui.player.base
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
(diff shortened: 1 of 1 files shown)


cameocoder:
> Seems like this optimizes the seekbar a bit too much. It no longer shows any progress now 🙃

Of course, there is nothing better to improve performance than to do nothing at all. 🙃

The fix for making the progress appear again is to change it to:
```
val seekbarProgress by remember(progress, positionInfo) {
		derivedStateOf { determinePlayerProgress(progress, positionInfo) }
	}
```

However that does not help with recompositions.  I think the problem is from the precision of the progress value.  The progress value really is changing even if that value may not result in a visual change of the seekbar.

PR #4889, opened 2025-08-30
Rewrite picture viewer in compose

nielsvanvelzen:
**Changes**
- Rename "picture viewer" to "photo player"
- Rewrite the UI with compose based on our new player UI so it has the same look & feel as the new video player
  - Now shows the photo name, album name and time in the overlay
- Add ScreensaverLock composable
- Fix player overlay showing by default

**Screenshots**
image

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/player/photo/PhotoPlayerScreen.kt
```diff
@@ -0,0 +1,45 @@
...
+		enabled = presentationActive,
+	)
+
+	Box(
+		modifier = Modifier
+			.background(Color.Black)
+			.fillMaxSize()
+	) {
+		PhotoPlayerContent(
+			item = item,
```
cameocoder, line 38: For all of your composables, consider passing in only the parts of the BaseItemDto that are actually needed as arguments, or map the BaseitemDto to a UI data class as a state holder.  For example:
```
@Composable
fun PhotoPlayerContent(
    imageUrl: String?,
    blurHash: String?,
    aspectRatio: Float,
    modifier: Modifier = Modifier,
)
```

Consider mapping the viewModel.currentItem to a UI state object so BaseItemDto can be garbage collected (assuming there are no children with direct requirements on BaseItemDto).  It will result in a UI that is more memory efficient, testable, previewable, reusable as well as decoupling your UI from your SDK network layer.

app/src/main/java/org/jellyfin/androidtv/ui/player/photo/PhotoPlayerHeader.kt
```diff
@@ -0,0 +1,43 @@
...
+import androidx.compose.ui.unit.sp
+import org.jellyfin.androidtv.ui.base.LocalTextStyle
+import org.jellyfin.androidtv.ui.base.Text
+import org.jellyfin.androidtv.ui.player.base.PlayerHeader
+import org.jellyfin.sdk.model.api.BaseItemDto
+
+@Composable
+@Stable
+fun PhotoPlayerHeader(
+	item: BaseItemDto?,
```
cameocoder, line 16: It looks like all you need to pass here is the name and album.  It simplifies the logic by removing the nullability check.

app/src/main/java/org/jellyfin/androidtv/ui/player/photo/PhotoPlayerViewModel.kt
```diff
@@ -21,7 +21,7 @@ import org.jellyfin.sdk.model.api.SortOrder
 import java.util.UUID
 import kotlin.time.Duration.Companion.seconds
 
-class PictureViewerViewModel(private val api: ApiClient) : ViewModel() {
+class PhotoPlayerViewModel(private val api: ApiClient) : ViewModel() {
 	private var album: List<BaseItemDto> = emptyList()
```
cameocoder, line 25: Consider mapping BaseItemDto to an PhotoPlayerUIState or Album data class.
```
data class Album(
    val id: String,
    val title: String,
    val albumName: String,
    val coverImageUrl: String?
    val aspectRatio: Float,
    ...
)

fun BaseItemDto.toAlbum(): Album {
    return Album(
        id = this.id,
        title = this.name,
        coverImageUrl = ...
    )
}
```

Having a PhotoPlayerUIState allows you to add Loading, Success and Error states to the UI.
sealed class PhotoPlayerUiState {
    object Loading : PhotoPlayerUiState()
    data class Success(val album: Album) : PhotoPlayerUiState()
    data class Error(val message: String) : PhotoPlayerUiState()
}

PR #4907, opened 2025-09-05
Fix some languages using wrong code

nielsvanvelzen:
Android uses the wrong language codes for some languages. Weblate did not use those and as such the languages changed in this PR never worked.

**Changes**
- Fix some languages using wrong code

**Issues**


crobibero (approved):
Good luck weblate

nielsvanvelzen:
I paused Weblate before making these changes, and the mobile app does already use the correct code for Hebrew so I expect Weblate to just pick it up and work 🤞

PR #4937, opened 2025-09-13
Change seekbar seeking to be relative to the current position

### LINKED ISSUE #4427: Relative rewind and seek (+/- 30s instead of fixed marks at 0:00, 0:30, ...) and customize rewind/seek independently

### Problem description

When I could not understand a dialogue or want to watch a scene again, I don't want to rewind to a fixed interval, but instead replay the last 30 (or 10, 20) seconds again.
As of now, any rewind / seek always jumps to the next full 30 seconds. I.e. I am at 5:10 and rewind. I'd expect the first rewind to jump to 04:40, but it jumps to 05:00.

Rewinding is also usually desired to be fine-grained, i.e. 5 or 10 seconds (option added in https://github.com/jellyfin/jellyfin-androidtv/pull/3794), but seeking forward should skip fast (e.g. due to playback not starting at the correct point in time). Hence, going backward and forward should each have their own slider for step size.

This was already requested here (https://github.com/jellyfin/jellyfin-androidtv/issues/4257) but I believe it was closed due to misunderstanding - using jellyfin-androidtv 0.18.4 release, the described enhancement is not implemented.

### Proposed solution

1. Rewind and seek don't jump to absolute points in time, but relatively to the current point
2. Step size of rewind and seek can be configured independently, proposal: [5-30s rewind] [5-60s seek]

### Alternatives considered

Setting the seek time to 10s works better, but introduces the problem of seeking forward for very long when needing to manually seek from 0:00 to (e.g.) 42:50.

### Additional information

_No response_

### Other sources

_No response_

Syquel:
**Changes**
Currently the available seek positions are fixed at multiples of the configured seek amount.  
Now seek positions are calculated relatively to the current position in increments of the configured seek amount.

**Issues**
Fixes #4427

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomSeekProvider.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomSeekProvider.kt
@@ -20,8 +20,6 @@ import org.jellyfin.sdk.api.client.ApiClient
 import org.jellyfin.sdk.api.client.extensions.trickplayApi
 import org.jellyfin.sdk.api.client.util.AuthorizationHeaderBuilder
 import org.jellyfin.sdk.model.serializer.toUUIDOrNull
-import kotlin.math.ceil
-import kotlin.math.min
 
 class CustomSeekProvider(
 	private val videoPlayerAdapter: VideoPlayerAdapter,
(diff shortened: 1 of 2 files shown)


Syquel:
@nielsvanvelzen could you take a look, please?  
Would be great if it were included in v0.19

nielsvanvelzen:
Don't ping for review requests, I look at stuff when I have time for it. Pushing to a PR will only move it lower on my backlog.
The feature-set for 0.19 is basically set in stone now and the focus is on fixing (newly introduced) bugs.

Syquel:
Didn't intend to offend, but just fixed issues I encountered on my dev build and added some tests.  
While I understand your displeasure about being pinged your comment was kind of rude TBH. I simply tried to contribute something to this project.

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomSeekProvider.kt
```diff
@@ -29,17 +27,44 @@ class CustomSeekProvider(
 	private val forwardTime: Long
 ) : PlaybackSeekDataProvider() {
 	private val imageRequests = mutableMapOf<Int, Disposable>()
+	private var currentSeekPositions = LongArray(0)
```
nielsvanvelzen, line 33: We shouldn't cache the results of `getSeekPositions` as it's unspecified when `getSeekPositions()` is called so this can cause issues if the leanback library (that controls this UI) is updated with behavior changes.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/overlay/CustomSeekProvider.kt
Syquel: This is intentional to satisfy the contract with `getThumbnail` which explicitly describes its `index` parameter as follows in its javadoc:
> @param index Index of position in {@link #getSeekPositions()}.

https://android.googlesource.com/platform/frameworks/support/+/300475fc42431cefb0d8f1bfc70de1a7a052a501/leanback/src/main/java/android/support/v17/leanback/widget/PlaybackSeekDataProvider.java#64

If I read the javadoc of [PlaybackSeekDataProvider](https://android.googlesource.com/platform/frameworks/support/+/300475fc42431cefb0d8f1bfc70de1a7a052a501/leanback/src/main/java/android/support/v17/leanback/widget/PlaybackSeekDataProvider.java) and its methods correctly it is explicitly intended for this provider to be stateful while seeking and to calculate the seek positions only once at the start of seeking and to reset them when seeking stops.

Side note: because it has been some time since I opened this PR I don't remember if `videoPlayerAdapter.currentPosition` changes while seeking is in progress or not. So it might be at least required to save the `videoPlayerAdapter.currentPosition` when seeking starts.

nielsvanvelzen: Apologies for the long wait times, with the big backlog of PR's I had this one kind of fell of my radar until jut now.

Good point, so we can likely assume the index to be correct for the latest returned value of getSeekPositions. meaning the current approach of caching should work there.

If you can rebase the PR we can continue with the reviews. I'd say if seeking works as expected the current implementation is probably safe to use.

Syquel: I rebased it onto master now.

nielsvanvelzen (approved):
Seems to be working perfectly fine. Thanks for the PR!

Issue #4936, opened 2025-09-13
Truncated text on multiline labels in Hebrew language


tal-sarid:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

1. Switch to Hebrew language
2. Go to any media page
3. Make sure to stop the playback in the middle so the "Resume from mm:hh" will be visible.
4. Any multiline labels under the action buttons will be truncated.

In English the second line is also a bit truncated but it is much less obvious (letters like p, q, g are a bit cut off for example when they are in the second line)

<img width="1074" height="845" alt="Image" src="https://github.com/user-attachments/assets/7db7d79d-e987-400a-9375-da504418bddf" />

### Application version

0.19.0

### Where did you install the app from?

Sideloaded APK

### Device information

Homatics Box R 4K Plus

### Android version

Android 14

### Jellyfin server version

10.10.7

854562:
The issue affects languages with Latin script as well, albeit less obvious. Letters like p and g are cut off when they are on the second line.

tal-sarid:
> The issue affects languages with Latin script as well, albeit less obvious. Letters like p and g are cut off when they are on the second line.

I stand corrected, it does happen in English as well but, as you said, it is less obvious than in Hebrew. Maybe these labels' size should be dynamic based on their text.

PR #4941, opened 2025-09-15
Fix truncated labels under buttons in media view UI

tal-sarid:
**Changes**
Changed button row layout's height to be adaptive to its content instead of a fixed `70dp` value.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-androidtv/issues/4936


nielsvanvelzen (approved):
Thanks for the PR!

PR #4946, opened 2025-09-18
Show placeholder thumbnail while Trickplay thumbnail loads

tal-sarid:
**Changes**
Introducing a placeholder thumbnail that is displayed while the actual Trickplay thumbnail is being loaded. This provides visual feedback to the user and improves the perceived performance of Trickplay seeking.

The placeholder is a simple black transparent bitmap matching the dimensions of the expected thumbnail and it is cached to avoid redundant creations and memory allocations.


tal-sarid:
I also explained [here](https://github.com/jellyfin/jellyfin-androidtv/pull/4819#issuecomment-3135445778) why it is OK to call `onThumbnailLoaded` twice for the same index.

PR #4956, opened 2025-09-23
Fix low quality card images

### LINKED ISSUE #4955: Low Resolution/Pixelated Cover Art Images

### Describe the bug

On the homepage, the cover art looks low res and pixelated compared to every other Jellyfin client like the mobile android and web client. The other clients look perfectly fine, I have tried clearing the image cache multiple times as well, same issue. I have also tried multiple versions of Jellyfin, the latest stable version and 0.19.0 beta 3 all have this same low resolution and pixelated look for the cover art.

### Logs

_No response_

### Application version

0.19.0 beta 3

Google Play Store

### Device information

Nvidia Shield TV Pro (2019)

### Android version

Android 11

### Jellyfin server version

10.10.7

### Other sources

_No response_

Syquel:
**Changes**
In some cases card images are loaded from the server without limiting the image dimensions.  
This results in loading images in their original size and requires the client to scale them which results in low quality images.
To avoid low quality card images the scaling is moved to the server-side by requesting images with appropriate dimensions in all cases.

**Issues**
Fixes #4955

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/BaseItemDtoBaseRowItem.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/BaseItemDtoBaseRowItem.kt
@@ -130,7 +130,7 @@ open class BaseItemDtoBaseRowItem @JvmOverloads constructor(
 		val seriesPrimaryImage = baseItem?.seriesPrimaryImage
 
 		return when {
-			preferSeriesPoster && seriesPrimaryImage != null -> imageHelper.getImageUrl(seriesPrimaryImage)
+			preferSeriesPoster && seriesPrimaryImage != null -> imageHelper.getImageUrl(seriesPrimaryImage, fillWidth, fillHeight)
 
 			imageType == ImageType.BANNER -> imageHelper.getBannerImageUrl(
 				requireNotNull(
--- a/app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java
@@ -391,10 +391,17 @@ public class CardPresenter extends Presenter {
         int fillWidth = Math.round(holder.getCardWidth() * holder.mCardView.getResources().getDisplayMetrics().density);
         int fillHeight = Math.round(holder.getCardHeight() * holder.mCardView.getResources().getDisplayMetrics().density);
 
-        holder.updateCardViewImage(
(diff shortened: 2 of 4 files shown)


nielsvanvelzen:
We deliberately no longer set width/height for the item images so we can load them from the local app cache. If we request them in various sizes the cache would just grow and we can't re-use the images.

Resizing the images client-side should not be a problem and I can't reproduce the weird scaling results from the linked issue.

Syquel:
I agree that it would be better to just load the "full quality" posters but
1. Not everything is loaded without width / height although I did not check in which cases: https://github.com/jellyfin/jellyfin-androidtv/blob/4d532f806066ab447d8fcfac6c7e62a355a40ab3/app/src/main/java/org/jellyfin/androidtv/ui/presentation/CardPresenter.java#L395
2. It's reliably reproducible on my Google Streamer with a 4K TV for Posters as well as the user cards (shows, movies etc cards). I tried to experiment with scaling options but the quality of the downscaled posters was always bad. The only viable solution I found was loading the downscaled image from the backend.
3. Do we ever need the full quality version of those images? Doesn't it take up more space than downloading the downscaled version from the backend in 2-3 sizes?

What would be a better solution? Downloading the full quality version and downscaling it adhoc in-memory before setting it as poster image to avoid the android downscaling logic which produces those bad results?

// EDIT
I also noticed slightly slower loading of posters (not much but quite noticable) even within my intranet and using the full quality images might be the cause for this too.  
If people use Jellfin over the Internet it might be really noticable.  

I increasingly think loading the (possibly) multiple downscaled versions of those images is way more beneficial than downloading the full quality version only once.

clams4shoes:
I was able to reproduce this issue on my Nvidia Shield TV (2017) as well.

From what I've seen, the typical approach is for the server to generate and store/cache multiple image sizes, with clients requesting the most appropriate size for their context. This approach has several advantages:

- **Reduced storage:** For an individual image stored in a small number of sizes, keeping a few scaled versions cached often uses less space than caching the full-size original.

- **Faster loading:** Smaller payloads improve network performance and first-time load times.

- **Consistent and efficient rendering:** It avoids inconsistencies and performance issues with client-side scaling. Scaling large posters down to thumbnail sizes can produce suboptimal results and be resource-intensive on less powerful devices, especially when processing many images at once.

The main downside is potential cache bloat, and some images may bypass the cache.

A practical compromise would be to maintain a small, predefined set of sizes (e.g., thumbnail, medium, poster). When the client needs a size not in that set, it can request the next larger size and scale it down. This generally minimizes pixelation or aliasing, keeps only a small set of image sizes cached, and helps limit cache bloat.

An additional benefit is that images are cached lazily, or on demand. For example, if the largest size in the set is never used, it won’t be stored in the cache.

PR #4985, opened 2025-10-02
Changed subtitle positioning to move up when the player controls are visible

### LINKED ISSUE #4815: Move subtitles up when player controls are visible

### Problem description

Currently, in the Jellyfin Android app, the on-screen display (OSD) controls, which include the timeline, play/pause buttons, and other settings, overlay the subtitles when activated. This makes it impossible to read the subtitles for the several seconds that the controls are visible on the screen. This is particularly disruptive when pausing, seeking, or adjusting playback, as crucial dialogue can be missed.

### Proposed solution

I propose that the subtitles be dynamically repositioned upwards whenever the player controls are visible. Once the controls auto-hide, the subtitles should return to their original position.

This is the standard behavior implemented in other popular video playback apps like Plex and YouTube. It's an intuitive solution that ensures an uninterrupted viewing experience by keeping the subtitles readable at all times.

### Alternatives considered

Add an option to remove the auto-hide delay for controls: An option could be added to the settings to make the player controls hide instantly without any delay.

### Additional information

_No response_

### Other sources

_No response_

jimlei:
**Changes**
Implemented methods in PlaybackController and VideoManager to adjust subtitle position based on control visibility and pixel offset.
Added functionality to restore subtitle position with a delay in LeanbackOverlayFragment when hiding controls.

**Issues**
Fixes #4815 

@yslpn, would be great if you could have a look and see if this is what you had in mind :rocket:


yslpn:
@jimlei Hello, thank you for your work. Currently, I don't have enough time to test. Probably, I will try this week.

Do you know an easy way how I can test it? Maybe the project has some guide.

nielsvanvelzen (commented):
Depending on the subtitle formats used this can cause some issues with the subtitle rendering. For text subtitles (srt/webvtt etc) this would work fine, for others like PGS that are image based it will cause subtitles to show at the wrong positions and even be warped.

nielsvanvelzen (changes requested):

nielsvanvelzen:
Closing as stale

PR #4986, opened 2025-10-02
Changed audio preferences to include and respect default language from server

### LINKED ISSUE #2289: Configurable default language for audio/subtitle

Hi,

I think would be great if the default language for audio/subtitle is configurable.
Now always the default audio/subtitle track selected is the first or with the forced flag enabled.

Thanks in advance.

jimlei:
Implement preferred audio language selection, sync of property from server, and update audio stream switching methods

I needed this myself to use Jellyfin because my 4yos Bluey sessions were a constant "select audio language" annoyance that kept me using a different media server.

Haven't done anything with subtitles but if this PR looks acceptable audio wise I can probably look into the subtitles side as well, either in this or a separate PR.

**Changes**
- Added a new preference for selecting the preferred audio language in UserSettingPreferences.
- Updated PlaybackController to handle audio stream switching based on user preference.
- Refactored switchAudioStream methods for clarity and added automatic selection logic.
- Introduced OptionsLanguageList for managing audio language options in the preferences screen.
- Updated UI components to reflect changes in audio stream selection behavior.

**Issues**
Fixes #2289  (partly)

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/eventhandling/SocketHandler.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/eventhandling/SocketHandler.kt
@@ -126,7 +126,7 @@ class SocketHandler(
 				val index = message["index"]?.toIntOrNull() ?: return@onEach
 
 				withContext(Dispatchers.Main) {
-					playbackControllerContainer.playbackController?.switchAudioStream(index)
+					playbackControllerContainer.playbackController?.switchAudioStreamByUser(index)
 				}
 			}
 			.launchIn(coroutineScope)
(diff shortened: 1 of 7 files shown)


hcharbonnier:
There seems to be an issue. Opening the playback menu triggers an infinite loop that repeatedly calls the server on /Localization/Cultures, and clicking on ‘Preferred Languages’ appears to crash the application most of the time.

jimlei:
@hcharbonnier closing this because #5150 is much further along. 

I didn't follow this closely enough after it went stale with no feedback for a while. Happy you picked it up though :smile:

hcharbonnier:
@jimlei For my part, the matter has been dropped since my merge request was rejected.

jimlei:
@hcharbonnier Thanks for the update. My PR went stale without any review/feedback just like yours initially did. Sucks because this is the one feature that stops jellyfin from being a viable alternative for me.

PR #4990, opened 2025-10-03
Add item play button LongClickListener to use external player

### LINKED ISSUE #4261: Bring back per item External Player

### LINKED ISSUE #4465: Add option to play via external player directly from movie/episode screen

krakj:
**Changes**
Change the play button for an item to use the external player on a long click. 

**Issues**
Fixes #4465 
Fixes #4261 
Addresses feedback given on #4487


nielsvanvelzen (changes requested):
As mentioned previously, there is currently no intention to add this feature. If you want to use an external player you can configure it in the app settings.

BenMcLean:
> As mentioned previously, there is currently no intention to add this feature. If you want to use an external player you can configure it in the app settings.

Do we really need to fork the entire program in order to get a player that works?

On discovering that the internal player is broken on some video you've got, you have to go all the way back out, find some really obscure setting to change the player permanently and globally for all your media, then navigate back in to watch the one odd video that doesn't work right with the internal player, find that the external player doesn't sync the video progress with jellyfin, then when you're done watching that video, you're either stuck with that alternative player that doesn't sync right forever or else you have to remember to go all the way back out, navigate into the setting menu, navigate into the playback menu, find where the alternate players option is under "advanced" or wherever the hell it is and change it back every single time. Four menus deep, ***twice***.

What in the hell? That's broken. This isn't a feature request: this is a bug fix.

The concept of which player to use fundamentally isn't universal: it's very very individual to each resource you're trying to play back.

IlIlIIlIlIlIIllIIlIlI:
Why? Simply saying there is no intention to add the feature doesn't minimize the need. To be blunt, external players play HDR video way more consistently, but the problem is they don't track watch history. So, when you go from watching movies, to watching lets say non HDR content (Anime), you have to CONSTANTLY switch back and forth between the options, and its buried **deep** within the Jellyfin settings to change external player. This all seems extremely purposeful to stop people from using external players.

BenMcLean:
I think it must be important, from a design perspective, that the settings real users need the fastest and most often be buried several levels nested in menu dives deeper than the Marianas trench.

nielsvanvelzen:
Closing as unwanted

BenMcLean:
Is this important enough for a fork of the whole program just to get it done? Because this decision seems incredibly bad.

PR #4991, opened 2025-10-03
Add device-specific profile override support

### LINKED ISSUE #4892: Dolby Vision HDR10+, Fire TV 4K Max

### Describe the bug

I did research and i found this issue https://github.com/jellyfin/jellyfin-androidtv/issues/4860. As i am understanding this happens if there is no HDR support at all. So I'm having a different Issue?

My TV Supports HDR10+ and DV. If I play a File, that has DV and HDR10+ it either generates a splitscreen (happens if i put the screen setting in Android to "always hdr". If i put the setting to "adaptive" (so the menu etc. is in SD and the Stick changes if compatible media is played), i get a purple screen and only audio is playing. 

Pictures are attached. 

If u need logs, tell me which ones and i will provide them. 

My Server ist on RC5 and the Android App is also on the newest version. I was very hopefull, that the new version would fix my problem, because nearly all my Media is combined HEVC HDR10+ /DV Content. 

I do not use an LG TV. As soon as i force transcoding, the video switches to DV and i can watch the file. 

Image
Image
Image
Image
Image

### Media info of the file

```shell

```

### Client logs

_No response_

### FFmpeg logs

```shell

```

### Application version

0.18.11

Amazon Appstore

### Device information

Fire TV Stick 4K Max 

### Android version

Fire OS 8.1.4.5

### Jellyfin server version

10.11.0 - rc5

### Other sources

_No response_

clams4shoes:
**Changes**
This allows storing and applying device-specific overrides to the device profile.

This is useful for handling devices that incorrectly report capabilities or have known issues (e.g., a faulty hardware decoder).

Additionally, this fixes Dolby Vision playback with an HDR10+ base layer on Fire TVs.

**Issues**
Fixes #4892

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/DeviceProfileBuilderOverrideManager.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/DeviceProfileBuilderOverrideManager.kt
@@ -1,0 +1,114 @@
+package org.jellyfin.androidtv.util.profile
+
+import org.jellyfin.sdk.model.api.CodecProfile
+import org.jellyfin.sdk.model.api.ContainerProfile
+import org.jellyfin.sdk.model.api.DirectPlayProfile
+import org.jellyfin.sdk.model.api.ProfileCondition
+import org.jellyfin.sdk.model.api.TranscodingProfile
+import org.jellyfin.sdk.model.deviceprofile.DeviceProfileBuilder
+import timber.log.Timber
+
+/// Rudimentary override handling for device profile
+// Consider moving this override logic into the SDK -- DeviceProfileBuilder should handle this
+// TODO Handle partially overlapping profiles by separating into their own distinct profile
+object DeviceProfileBuilderOverrideManager {
+    fun applyOverrideProfiles(builder: DeviceProfileBuilder, vararg overrides: CodecProfile) =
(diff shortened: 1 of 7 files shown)


clams4shoes:
> Adding a special condition for a known faulty device is fine, but the state of this PR is overkill.

I understand that the current PR might feel a bit overkill. I chose this approach because it seemed to give us the most flexibility -- simply updating the JSON allows us to address device-specific issues quickly and easily in the future.

I'm happy to hear any suggestions for a lighter approach, or any specific concerns you might have.

nielsvanvelzen:
Adding special conditions for specific devices is something we won't do often, in fact: this is the first time I'm agreeing with one. We should **always** try feature detection first and only if that is not possible and the device in question is popular enough will we add a special rule.

My suggestion would be to just add a profile condition in the code for the device we're currently trying to have working. No fancy JSON files with custom parsing. Especially if we want to get this into a 0.19.x release.

clams4shoes:
Alright, I'll set this aside for now and just add a special case for the relevant Fire TVs in the device profile.

If more similar situations come up in the future, we can revisit this.

I'll go ahead and close this PR and reopen a new one when I implement the lighter Fire TV specific approach.

Thanks again for the review and feedback -- it's much appreciated! 🙂

PR #5023, opened 2025-10-17
Update translation files

zhao-hw:
**Changes**
Change the Chinese translation of playback related strings to "播放".

**Issues**
"回放" means replay,  therefore "播放" is more proper.


crobibero:
Translations are managed through https://translate.jellyfin.org/projects/jellyfin/

PR #3110, opened 2023-10-09
Add preferences to configure audio codecs and 4k transcoding/direct play

t0mas:
Allow users to select which audio codecs to direct play, and in case of transcoding select the preferred codec to transcode into. Main use-case is for systems where the connected audio equipment fails to play some codecs. For example transcode Dolby TrueHD and DTS into AC3 or EAC3 if those are supported by the audio receiver or select AAC if more advanced formats are not supported.

Allow users to select whether 4k video streams can be accepted or not. Main example is the "Chromecast with Google TV" device, which in the old method will request 4k streams but is unable to play those (there exists a 4K version of the device). The new preference allows users to disable 4k if their player is not capable of playing it. The setting defaults to the same as the original behaviour (device detection).

I saw PR #3109 by @MichaelRUSF overlaps with this but only does AAC to DolbyDigital and not the other options.

**Changes**
- Add user preferences to enable audio codecs
- Add user preference to select transcode audio target
- Add user preference to enable/disable 4k video
- Changed ExoPlayerProfile to be based on the selected preferences

**Issues**
Fixes #2867, #2602, #2991 for audio and #2971, #2516 for video

PR #3109, opened 2023-10-09
Add option to transcode AAC to AC3

MichaelRUSF:
**Changes**

Add an option in the playback preferences menu to transcode AAC to AC3 audio. 

**Issues**
#2867  #2602 #2991

Issue #2602, opened 2023-03-19
AAC Plays as Stereo 


BJET67:
### Describe the bug

When watching AAC 5.1 or 7.1 Content it remuxes to stereo I'm all use cases, I've tried 2 different audio receivers as well as 2 different media devices (Chromecast With Google TV and a TiVO Box) And the behavior is the same, Only formats I've gotten to work flawlessly are AC3 and EAC3, all other codecs I've tried have had some sort of issue.

This is a problem because a lot of the content I have on my server is AV1/opus 5.1/7.1 which will transcode to HEVC/AAC and will come through as Stereo.

I also tried LibVLC which has its own issues and all codecs come through as Stereo and the video playback 
Is choppy.

I have an open issue for the opus support and it looks like it is potentially getting merged but AAC should still play as surround

### Application version

0.15.5

### Where did you install the app from?

Google Play

### Device information

TiVO model unknown

### Android version

Android 10

### Jellyfin server version

10.8.9

branches78:
I know the answer to your problem, the device you are using has Android 9, turns out, Google screwed up that version because every single codec different than Dolby and DTS (AAC, FLAC, WAV, etc), gets downmixed to stereo by that version, it's a bug they will never fix, the Jellyfin app is absent of this problem, however, Android 10 and beyond versions are playing multichannel PCM correctly so I suggest to get another device with a newer Android version if the one you have won't see an update anytime soon.

Note: AAC and the other codecs I mentioned always get decoded as PCM internally by the device before it's output.

BJET67:
The original issue is wrong, the TiVo device is on Android 10. @branches78

branches78:
I see, in that case, please describe your environment since on mine it's able to play multichannel AAC without problems.

BJET67:
I have a Yamaha Rx-V363 receiver connected via optical to my android tv box

branches78:
> connected via optical

Optical doesn't support multichannel PCM, it only supports Dolby Digital, DTS and PCM 2.0.

I think the Tivo transcodes everything to Dolby Digital as far as I heard, I guess that's what you were getting?.

Why don't you use the HDMI inputs in your AV receiver instead?, I see it has two inputs.

BJET67:
> Why don't you use the HDMI inputs in your AV receiver instead?, I see it has two inputs.


The inputs on the rx-v363 are video only
[...]

PR #5025, opened 2025-10-17
Allow prioritization of AC3 over AAC

### LINKED ISSUE #2867: Add a fallback to AC3/EAC3 option

Some audio codecs such as TrueHD and DTS dont always play correctly so I think the easiest way to get bitstream surround sound is to request a AC3 or EAC3 transcode from the server instead of an AAC or any other codec.

From what I believe bitstreaming AC3 or EAC3 is the only way to get surround over optical so this would be a great addition

andreasgerner:
This PR allows users to prioritize transcoding into AC3 over AAC. This behavior can be configured in the advanced settings menu if bit streaming of Dolby Digital audio is enabled.

Further details:
- If the corresponding setting is enabled in the advanced playback menu and a file is transcoded (no matter if transcoding happens only for audio or for audio & video), Jellyfin now prioritizes audio transcoding into AC3 instead of AAC.
- Because AC3 is one of Dolby Digitals audio codecs, this setting only can be modified if bit streaming of Dolby Digital audio is enabled.
- As ffmpeg currently cannot transcode any audio into EAC3 (https://github.com/jellyfin/jellyfin/issues/9045), I focused on AC3. Prioritizing EAC3 currently wouldn’t have any impact on the transcoding of audio streams.
- If the source audio is encoded in EAC3, it will be kept and only video will be transcoded. This behavior does not change in comparison to the current version and is exactly the same in other clients.
- Direct Play does not have a prioritization of audio codecs, so changing the order of the codecs in that part of the device profile does not have an impact on streaming.

I tested all of these aspects on a FireTV stick and several emulated Android TVs.

**Changes**
The order of the audio codecs in the device profile changes their priorities, thus making AC3 the first element tells the Jellyfin server to prefer it over AAC. `jellyfin-web` handles this the same way if you configure a preferred audio codec ([jellyfin-web:apphost.js](https://github.com/jellyfin/jellyfin-web/blob/3215be4cd8d2c0db5d487f239dac728dd017f4cf/src/components/apphost.js#L108)). 

Considerations while implementing:
- I tried to make the new checkbox automatically turn off if Dolby Digital bit streaming is disabled, but changing the `ac3Preferred` user preference in addition to the `ac3Enabled` user preference does not trigger a refresh for the new checkbox.
- I didn't see any use cases right now for a selection of a specific preferred codec (as implemented in the web client).
- Same goes for selecting a specific video codec, I think the current implementation (using HVEC whenever possible) is the best way to handle this.

**Issues**
Fixes #2867, #2602, #2991, #2432 and possibly a lot of similar issues.
Rework for #3109.
Rework for some parts of #3110.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -125,6 +125,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 */
 		var ac3Enabled = booleanPreference("pref_bitstream_ac3", true)
 
+		/**
+		 * Prefer AC3 over AAC
+		 */
+		var ac3Preferred = booleanPreference("prefer_ac3", false)
+
 		/* Live TV */
 		/**
 		 * Use direct play
--- a/app/src/main/java/org/jellyfin/androidtv/ui/preference/screen/PlaybackAdvancedPreferencesScreen.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/preference/screen/PlaybackAdvancedPreferencesScreen.kt
@@ -133,6 +133,13 @@ class PlaybackAdvancedPreferencesScreen : OptionsFragment() {
 				setContent(R.string.desc_bitstream_ac3)
 				bind(userPreferences, UserPreferences.ac3Enabled)
 			}
+
+			checkbox {
+				setTitle(R.string.lbl_prefer_ac3)
+				setContent(R.string.desc_prefer_ac3)
+				bind(userPreferences, UserPreferences.ac3Preferred)
+				depends { userPreferences[UserPreferences.ac3Enabled] }
+			}
 		}
 
 		category {
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
@@ -62,23 +62,40 @@ fun createDeviceProfile(
 	mediaTest = MediaCodecCapabilitiesTest(context),
 	maxBitrate = userPreferences.getMaxBitrate(),
 	isAC3Enabled = userPreferences[UserPreferences.ac3Enabled],
+	isAC3Preferred = userPreferences[UserPreferences.ac3Preferred],
 	downMixAudio = userPreferences[UserPreferences.audioBehaviour] == AudioBehavior.DOWNMIX_TO_STEREO,
 	assDirectPlay = false,
 	pgsDirectPlay = userPreferences[UserPreferences.pgsDirectPlay],
 )
 
+private fun createSupportedAudioCodecs(
+	isAC3Enabled: Boolean,
+	isAC3Preferred: Boolean
+): Array<String> {
+	if (!isAC3Enabled) return supportedAudioCodecs
+		.filterNot { it == Codec.Audio.AC3 || it == Codec.Audio.EAC3 }
+		.toTypedArray()
+
+	if (isAC3Preferred) return supportedAudioCodecs
+		.filterNot { it == Codec.Audio.AC3 }
+		.toMutableList()
+		.apply { add(0, Codec.Audio.AC3) }
+		.toTypedArray()
+
+	return supportedAudioCodecs
+}
+
 fun createDeviceProfile(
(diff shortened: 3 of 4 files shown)


Pentaphon:
I hope this becomes a high priority for 0.19 stable. There might be a dozen issues that this fixes. Maybe more!

n8llcaster:
Can confirm that this fix works on older Sony's Android TV, used latest Beta to merge this to.

xxgmoneyxx:
> Can confirm that this fix works on older Sony's Android TV, used latest Beta to merge this to.

Anyway for you to share the apk you merge the fix to?

sudo-kraken:
> Can confirm that this fix works on older Sony's Android TV, used latest Beta to merge this to.

Can you confirm this worked for 7.1 and 5.1?

w-marco:
As an owner of a LG OLED G5 which famously lacks DTS pass through support, I’d love to see this merged. 
Currently, DTS gets transcoded into stereo AAC it seems. This PR would hopefully make it be 5.1 AC3 instead so I can use the full potential of my sound system even with DTS audio files. 

Update: I merged this into the current master state (a fork in my profile) and installed that version of the app. I enabled the feature and played a video file with DTS-HD 7.1 on my Xiaomi TV Box S 3rd gen, which is connected to my LG G5. 
Sadly, my sound system does not report Surround Sound received (which it would if the audio were properly converted to at least 5.1 ac3). 
I tested this with ffmpeg audio in the developer options of jellyfin enabled and disabled.

I also have Bitstream DD in jellyfin active and also have the TV Box as well as the TV itself all set to passthrough, so nothing except jellyfin should convert audio.

andreasgerner:
> As an owner of a LG OLED G5 which famously lacks DTS pass through support, I’d love to see this merged. Currently, DTS gets transcoded into stereo AAC it seems. This PR would hopefully make it be 5.1 AC3 instead so I can use the full potential of my sound system even with DTS audio files.
> 
> Update: I merged this into the current master state (a fork in my profile) and installed that version of the app. I enabled the feature and played a video file with DTS-HD 7.1 on my Xiaomi TV Box S 3rd gen, which is connected to my LG G5. Sadly, my sound system does not report Surround Sound received (which it would if the audio were properly converted to at least 5.1 ac3). I tested this with ffmpeg audio in the developer options of jellyfin enabled and disabled.
> 
> I also have Bitstream DD in jellyfin active and also have the TV Box as well as the TV itself all set to passthrough, so nothing except jellyfin should convert audio.

Did you sent the Jellyfin profile report to your Jellyfin instance? If "AC3" is moved to the front of your audio capabilities array, than everything of this PR works as expected. Jellyfin Core handles the conversion, I'm just following their specifications. If so, there is another error somewhere in your setup.

w-marco:
> Did you sent the Jellyfin profile report to your Jellyfin instance? If "AC3" is moved to the front of your audio capabilities array, than everything of this PR works as expected. Jellyfin Core handles the conversion, I'm just following their specifications. If so, there is another error somewhere in your setup.

Hi, yes indeed. However I since switched to plugging my Android TV Box directly into the Soundbar so DTS and other formats work natively. Hence I did not investigate further, though for now I would just assume there was some other issue with my particular setup.

Ranft65:
I was inspired by your PR and created another PR that does what yours does and also allows disabling AAC support completely.   See https://github.com/jellyfin/jellyfin-androidtv/pull/5419

Thank you for your (and others) efforts.

nielsvanvelzen (changes requested):
We can add a new section in the settings where it is possible to configure the audio codecs (which are enabled and in what order). But it should not be a single checkbox.

andreasgerner:
@nielsvanvelzen will close this PR for now in flavor of https://github.com/jellyfin/jellyfin-androidtv/pull/5419?

I didn't have time to implement your changes, so @Ranft65 did a PR inspired by this.

thinhlegolas:
Both PRs submitted by Ranft65 were rejected. So we're back to square one.

Ranft65:
If you want to try it out it is available at https://github.com/Ranft65/jellyfin-androidtv/releases/tag/Test-Prioritize-Audio-Codecs-v1.0.0.  I've moved away from fixing the issue on the client side and are doing so now with the Jellyfin Windows server now.   The Jellyfin server approach is available at https://github.com/Ranft65/jellyfin/releases/tag/v1.2.0.  Please do not discuss these further on this forum.  Any further discussion should be on the forks themselves.  Such discussion is deemed to be off-topic here.

Issue #4978, opened 2025-10-01
Hybrid Dolby Vision and HDR10+ files should display multiple badges in info row


hwneh:
### This request respects the following points:

- [x] This request is **not** already on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/labels/enhancement) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Problem description

Media with Dolby Vision with HDR10+ or HDR10/SDR fallback only correctly display the HDR badge in the item info row. Badges for other VideoRangeType are only displayed if Dolby Vision is absent.

<img src="https://github.com/user-attachments/assets/31b29297-3701-44c2-aab1-273589d07352" />

### Proposed solution

Media with Hybrid Dolby Vision and other HDR formats should display multiple separate badges.

<img  src="https://github.com/user-attachments/assets/686131c6-fc54-4683-893d-c2279999c643" />

### Alternatives considered

Alternatively, you could display a single badge that displays both.

<img src="https://github.com/user-attachments/assets/20013ad2-0a7f-4614-9a51-4a6e4886d3b1" />

### Additional information

I've created a working version by modifying BaseItemInfoRow.kt but it requires using the snapshot jellyfin-sdk because of the extended VideoRangeType.

PR #5038, opened 2025-10-21
Add support for multiple badges for media with hybrid HDR or HDR with fallback

hwneh:
**Changes**

- Added support for displaying multiple info row badges for media that has multiple Dynamic HDR or Dynamic HDR with SDR fallbacks. 
- Slightly reduced the InfoRowItem badge size to avoid cutting the last badges off when there are many.

**Issues**
#4978

**Notes**
- Requires changes from #5037

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/composable/inforow/BaseItemInfoRow.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/composable/inforow/BaseItemInfoRow.kt
@@ -189,23 +189,25 @@ fun InfoRowMediaDetails(mediaSource: MediaSourceInfo) {
 		}
 	}
 
-	// Video stream
-	val videoCodecName = when {
-		!videoStream?.videoDoViTitle.isNullOrBlank() -> stringResource(R.string.dolby_vision)
-		videoStream?.videoRangeType != null && videoStream.videoRangeType != VideoRangeType.SDR && videoStream.videoRangeType != VideoRangeType.UNKNOWN ->
-			videoStream.videoRangeType.serialName.uppercase()
-
-		else -> videoStream?.codec?.uppercase()
+	// Video range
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/browsing/composable/inforow/BaseItemInfoRow.kt
```diff
@@ -189,21 +189,24 @@ fun InfoRowMediaDetails(mediaSource: MediaSourceInfo) {
...
-	// Video stream
-	val videoCodecName = when {
-		!videoStream?.videoDoViTitle.isNullOrBlank() -> stringResource(R.string.dolby_vision)
-		videoStream?.videoRangeType != null && videoStream.videoRangeType != VideoRangeType.SDR && videoStream.videoRangeType != VideoRangeType.UNKNOWN ->
-			videoStream.videoRangeType.serialName.uppercase()
-
-		else -> videoStream?.codec?.uppercase()
+	// Video range
+	val videoRangeNames: Set<String> = when (videoStream?.videoRangeType) {
+		VideoRangeType.SDR -> setOf(VideoRangeType.SDR.serialName)
```
nielsvanvelzen: The serialName is the internal name used to communicate with the API. Names that are shown in the UI should always use a string resource.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/browsing/composable/inforow/BaseItemInfoRow.kt
hwneh: I've made the requested changes.

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks!

PR #4995, opened 2025-10-04
Fix DoVi/HDR10+ playback on certain Fire TV devices

clams4shoes:
**Changes**
Mitigates a known hardware decoding issue on certain Fire TV devices where the Dolby Vision decoder fails to properly decode and play streams with an HDR10+ base layer.

**Issues**
Fixes #4892

**Notes**
- Requires Jellyfin Server 10.11.* or later to take effect.
- Adds additional device identifiers to the profile report to help identify hardware commonalities behind cross-device issues.

PR #5049, opened 2025-10-25
Add older FireTV 4K Device and fix the Issues with DOVIWithHDR10 Videos

ddsc01:
The previous fix did still not work for DOVIWithHDR10 Videos, also a older Device was mentioned here https://github.com/jellyfin/jellyfin-androidtv/issues/4892#issuecomment-3445945775

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
@@ -430,6 +430,7 @@ fun createDeviceProfile(
 		}
 
 		if (jellyfinTenEleven && KnownDefects.hevcDoviHdr10PlusBug) {
+			add("DOVIWithHDR10")
 			add("DOVIWithHDR10Plus")
 			add("DOVIWithELHDR10Plus")
 		}
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
@@ -66,6 +66,7 @@ fun createDeviceProfile(
 	assDirectPlay = false,
 	pgsDirectPlay = userPreferences[UserPreferences.pgsDirectPlay],
 	jellyfinTenEleven = serverVersion >= ServerVersion(10, 11, 0),
+	dolbyVisionHDR10Disabled = userPreferences[UserPreferences.disableHDR10]
 )
 
 fun createDeviceProfile(
@@ -76,6 +77,7 @@ fun createDeviceProfile(
 	assDirectPlay: Boolean,
 	pgsDirectPlay: Boolean,
 	jellyfinTenEleven: Boolean,
+	dolbyVisionHDR10Disabled: Boolean
 ) = buildDeviceProfile {
 	val allowedAudioCodecs = when {
 		downMixAudio -> downmixSupportedAudioCodecs
@@ -430,9 +432,16 @@ fun createDeviceProfile(
 		}
 
 		if (jellyfinTenEleven && KnownDefects.hevcDoviHdr10PlusBug) {
-			add("DOVIWithHDR10")
 			add("DOVIWithHDR10Plus")
 			add("DOVIWithELHDR10Plus")
+		}
+
+		if (jellyfinTenEleven && dolbyVisionHDR10Disabled) {
+			add("DOVIWithHDR10")
+			if (!KnownDefects.hevcDoviHdr10PlusBug) {
+				add("DOVIWithHDR10Plus")
+				add("DOVIWithELHDR10Plus")
+			}
 		}
 	}
 
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -100,6 +100,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		var useExternalPlayer = booleanPreference("external_player", false)
 
 		/**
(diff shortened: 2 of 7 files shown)


nielsvanvelzen (approved):

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
```diff
@@ -430,6 +430,7 @@ fun createDeviceProfile(
 		}
 
 		if (jellyfinTenEleven && KnownDefects.hevcDoviHdr10PlusBug) {
+			add("DOVIWithHDR10")
```
nielsvanvelzen: Why add this one? The known defect list is only for devices with the 10 plus bug.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
ddsc01: The same problem with the Black screen/only audio appears with Dolby Vision and HDR10 Videos, not only HDR10+ Videos, if the TV does not Support HDR10 at all. Adding this helped view them on my old TV.

nielsvanvelzen: So you're saying that the first generation Fire TV 4k has the black screen issue with HDR10 in addition to the other ones, but it does show up as supported in the reported capabilities?

ddsc01: The first Generation FireTV 4K was mentioned having the same issues, by another User here: https://github.com/jellyfin/jellyfin-androidtv/issues/4892#issuecomment-3445945775

The HDR10 Issue is with newer Devices also. Tested with AFTKM and AFTKRT on my side, the issue it resolves by adding that line

theguymadmax: I have the Fire TV 4K 2nd gen (non-Max version) AFTKM and can play Dolby Vision with HDR10 without issue.

Leonashub: Same with the Firestick 4k max 2nd gen. Dv with hdr works fine (coincidentally they're all 16:9 aspect ratio).
Dv with hdr10+ gave issues but is still not fully resolved since it doesnt choose hdr10 but transcodes to sdr.
I wonder if that line would break anything for people where dv is now working on dv+hdr files

ddsc01:
Some people do not have issues with HDR10, so instead give the user the option to disable DOVIWithHDR10/DOVIWithHDR10+ manually in the Advanced Playback Preferences

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
```diff
@@ -433,6 +435,14 @@ fun createDeviceProfile(
...
 			add("DOVIWithELHDR10Plus")
 		}
+
+		if (jellyfinTenEleven && dolbyVisionHDR10Disabled) {
+			add("DOVIWithHDR10")
+			if (!KnownDefects.hevcDoviHdr10PlusBug) {
+				add("DOVIWithHDR10Plus")
+				add("DOVIWithELHDR10Plus")
+			}
+		}
```
clams4shoes, line 445: This new setting would affect **HEVC Dolby Vision content** but leave **AV1 Dolby Vision with an HDR10 base layer** unaffected.

**Reference:** [Dolby Vision Profiles](https://professionalsupport.dolby.com/s/article/What-is-Dolby-Vision-Profile?language=en_US)

It also disables **Profile 7.6 with HDR10+ metadata** (`DOVIWithELHDR10Plus`), while leaving the standard HDR10 metadata variant (`DOVIWithEL`) unaffected.

---

It seems this PR may be conflating separate issues. It appears to try to address both the [Fire TV HEVC Dolby Vision/HDR10+ bug](https://github.com/jellyfin/jellyfin-androidtv/pull/4995) for an additional device and a new Dolby Vision HDR10 issue. The approach for the latter feels like a quick workaround for a problem that isn’t yet fully understood.

It would be helpful to open a new issue so we can track and investigate it properly.

valkum:
I also ran into this issue, as in I have a file with DoVi 8.1 with HDR fallback which played back with sound only and a black screen on a Fire TV Stick 4k (2018).
```
Video range: HDR
Video range type: DOVIWithHDR10
DV title: Dolby Vision Profile 8.1 (HDR10)
DV version major: 1
DV version minor: 0
DV profile: 8
DV level: 6
DV rpu preset flag: 1
DV el preset flag: 0
DV bl preset flag: 1
DV bl signal compatibility id: 1
```

Turns out I had 'Bitstream Dolby Digital audio' enabled, which prevented the fallback to HDR for some reason.
Once disabled, the player switched to HDR.

nielsvanvelzen:
Closing as stale (unresolved merge conflicts for months)

DeNohvo:
> I also ran into this issue, as in I have a file with DoVi 8.1 with HDR fallback which played back with sound only and a black screen on a Fire TV Stick 4k (2018).
> 
> ```
> Video range: HDR
> Video range type: DOVIWithHDR10
> DV title: Dolby Vision Profile 8.1 (HDR10)
> DV version major: 1
> DV version minor: 0
> DV profile: 8
> DV level: 6
> DV rpu preset flag: 1
> DV el preset flag: 0
> DV bl preset flag: 1
> DV bl signal compatibility id: 1
> ```
> 
> Turns out I had 'Bitstream Dolby Digital audio' enabled, which prevented the fallback to HDR for some reason. Once disabled, the player switched to HDR.

This worked for jellyfin and wholphin, but now the audio is transcoding.

PR #5052, opened 2025-10-26
Reset still watching state on confirmation

### LINKED ISSUE #5051: [0.19.0-beta7] - Binge-watching watchdog counts is not reseted

### Describe the bug

1. Watch multiple episodes.
2. At the end, wait to have the next one (or select you want to watch the next one).
3. After sometime or several watched episodes, a menu ask you if you are still watching.
4. Answer you want to see the next episode.
5. After that, you are asked each time at the end of all episode.

I think the count should be reset each time the user interact with the TV (ie: volume changes, manual segment skipping..) or at least, when your answer you want to watch the next episode.

### Logs

_No response_

### Application version

0.19.0-beta7

Sideloaded APK

### Device information

Nvidia Shield Pro (2019)

### Android version

Android 11

### Jellyfin server version

10.11.0

### Other sources

_No response_

nielsvanvelzen:
**Changes**

- Reset the "still watching" state when the user confirms they're still watching

**Issues**

Fixes #5051

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/di/AppModule.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/di/AppModule.kt
@@ -144,7 +144,7 @@ val appModule = module {
 	viewModel { UserLoginViewModel(get(), get(), get(), get(defaultDeviceInfo)) }
 	viewModel { ServerAddViewModel(get()) }
 	viewModel { NextUpViewModel(get(), get(), get()) }
-	viewModel { StillWatchingViewModel(get(), get(), get()) }
+	viewModel { StillWatchingViewModel(get(), get(), get(), get()) }
 	viewModel { PhotoPlayerViewModel(get()) }
 	viewModel { SearchViewModel(get()) }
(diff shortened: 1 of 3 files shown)


BlackstormFirst:
Hello, thanks for the fix.
Unfortunatly it doesn't fix the state where you already have triggered the "showStillWatching" state.

I don't known if this is the best way, but adding "showStillWatching = false" into the new "private fun resetSession()" seems to fix the issue.

nielsvanvelzen:
Good catch, I only checked if the reset function was called in my testing and didn't verify the UI wouldn't appear again.. should've done that 😅.

BlackstormFirst:
> Good catch, I only checked if the reset function was called in my testing and didn't verify the UI wouldn't appear again.. should've done that 😅.

Hello, I probably miss your update ;) do i need to create a new issue about this new fix ? or i am just to much of a hurry, and you need some more time to try it on your side ?

nielsvanvelzen:
No need for a new issue, I already fixed the remaining issue with #5056 which also refactors the code to be easier to maintain.

PR #5063, opened 2025-10-29
Support theme song playback (v2)

AurumTitanos:
**Changes**
Adds an opt-in "series theme song preview" on the series detail screen.

When a series (FullDetailsFragment) is opened AND the feature is enabled:

* We request its theme audio from the server (`/Items/{id}/ThemeMedia?inheritFromParent=true`).
* A dedicated `ThemeSongService` builds (or reuses) a lightweight preview playback pipeline.
* The theme track is queued as a single item and starts playing automatically.

When leaving the series screen, the preview queue is cleared and playback stops (without tearing down the player).

**User-facing toggle (off by default)**

* A new `Series theme song preview` checkbox was added under **Preferences → Developer options**.
* This is backed by `UserPreferences.themeSongPreviewEnabled` and persists like other app settings.
* Theme preview only runs when this toggle is ON. By default it is OFF.

**Playback isolation / crash safety**

* The preview player does **not** use the main app player / MediaManager.
* It uses its own `PlaybackManager`, its own queue supplier, and its own Media3 `MediaSession` (separate notification/channel).
* We now reuse a single preview `PlaybackManager`/`MediaSession` across series instead of creating a new one every time.

  * On exit we just clear the queue instead of destroying the session.
  * This resolves the `IllegalStateException: Session ID must be unique` crash when switching between shows.

**New / changed code**

* `ThemeSongService.kt`: fetches theme media, owns the preview player, handles enter/leave lifecycle.
* `ThemePreviewQueueSupplier.kt`: supplies a one-item queue for the theme track.
* `DeveloperPreferencesScreen.kt`: adds the new opt-in toggle and binds it to `UserPreferences.themeSongPreviewEnabled`.
* `UserPreferences.kt`: new `themeSongPreviewEnabled` boolean preference (default `false`).
* `FullDetailsFragment`: now calls `onEnterItem(...)` / `onLeaveItem()` on the service.
* Removed the previous temporary `ThemeSongSettings` helper.

**Open items / draft notes**

* Audio currently restarts when navigating inside the same series (clicking seasons / episodes). Intended behavior is: only restart when switching to a different series.
* The preview `MediaSession` currently surfaces as its own notification; long term UX might want that suppressed or merged.
* Some comments / naming still need cleanup before final.

Tested on NVIDIA Shield (Android TV):

* With the toggle ON, opening a series reliably plays that show's theme.
* With the toggle OFF, nothing plays.

Marking this as draft for early review/feedback, especially around lifecycle and preference wiring.

**Issues**
[#1150](https://github.com/jellyfin/jellyfin-androidtv/issues/1150)


AurumTitanos:
I'll work on this more this weekend to fix remaining issues

PR #4713, opened 2025-06-09
Keep same audio language while binge watching a series

bin101:
**TODO**

**Changes**
- persist last played audio track language inside the lifecycle of a `PlaybackController` and reuse it for following items
- try to keep audio language if next episode is played via next up card

**Issues**
Fixes #4675

## Summary by CodeRabbit

* **New Features**
  * The app now remembers your last selected audio language and automatically selects it for future playback sessions.
* **Improvements**
  * Audio stream selection is enhanced to prioritize your preferred audio language across different videos.
  * Playback options reset appropriately when switching between different media items to ensure consistent audio language preferences.

PR #5071, opened 2025-10-30
fix: audio track selection on subsequent video media items

bin101:
It seems that I introduced some audio track selection issues with [this](https://github.com/jellyfin/jellyfin-androidtv/pull/4713) change.

I wasn't aware that the `VideoQueueManager` state is not cleared when navigating back to the episode detail screen.

**Changes**
Reset `VideoOptions` in `PlaybackController` during start of a media item.

**Issues**
Didn't created one


nielsvanvelzen (approved):

nielsvanvelzen:
During release testing I found a critical issue so the changes have been reverted with #5083. Please resubmit the pull request with a fix.

Transcoded videos appear to get stuck in a loop where, once the transcoded video starts, the app will restart.

PR #5083, opened 2025-11-01
Revert "fix: audio track selection on subsequent video media items"

nielsvanvelzen:
Reverts jellyfin/jellyfin-androidtv#5071


crobibero (approved):
What is the reason for revert?

nielsvanvelzen:
See comment in the original PR

Issue #4772, opened 2025-07-06
Adhere to “Max Days in Next Up” Established in User Settings


bobobado:
### This request respects the following points:

- [x] This request is **not** already on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/labels/enhancement) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Problem description

Max Days in Next Up set up in user settings is not followed on Android TV app

### Proposed solution

Please consider adding the limit set in the web user interface

### Alternatives considered

A duplicate of this setting in the app would be great if it’s easier to implement

Issue #2406, opened 2023-01-11
Hide shows from next up if they haven't been watched in X days


lunarlily:
### Describe the feature you'd like

In Jellyfin Web, shows that haven't been watched for 1 year (configurable in settings>display) are automatically hidden from the next up section. This feature is really useful, as I've started watching some shows that I'm not really interested in continuing, and so they just clutter things up (I can't even manually remove them from the TV home screen, as they get re-added immediately after opening the app)

It'd be cool if this could be added to the Android tv client as well, preferably with the option being synced with other clients (although I'm pretty sure that'd need some server-side changes, as it doesn't appear to sync between two web clients either)

reitenth:
this would be great. any plans?

nielsvanvelzen:
This setting is not saved on the server so we cannot use it in the app. Someone will need to update jellyfin-web first to save the value using display preferences, otherwise we cannot add it to the app.

sjorge:
I was looking for the same feature. In the meantime if you are using a reverse proxy in front of your jellyfin I found a workaround, although it's not ideal as it applies to **all** users!

`/usr/local/bin/infuse_nextup_hack.sh`:
```
#!/bin/bash
while read key; do
	date -u -d "$key days ago" +"%Y-%m-%dT%H:%M:%S.000Z"
done
```

Some where in your apache config
```apache
## NextUp limit to X days (hack for Infuse and ATV)
RewriteEngine On
RewriteMap date_cutoff prg:/usr/local/bin/infuse_nextup_hack.sh
RewriteCond %{HTTP_USER_AGENT} ^(Infuse-Direct|Ktor\ client) [NC]
RewriteCond %{REQUEST_URI} ^/Shows/NextUp$ [NC]
RewriteCond %{QUERY_STRING} !NextUpDateCutoff= [NC]
RewriteRule ^(.*)$ $1?NextUpDateCutoff=${date_cutoff:180} [QSA,R]
```

Would this time I wasted on this been better spend trying to get whatever is missing into jellyfin-server, probably, but I know apache pretty well while I have no idea where to even begin touching jellyfin-server.

tl;dr: This is a hack, use at your own risk. It enforced a 180 day NextUp cutoff for *ALL* users that use Infuse OR the AndroidTV client (Possible the Android client too, I just target "Ktor client" user-agent.

PR #5103, opened 2025-11-06
Add ability to set next up retention policy

Na3blis:
Add ability to set next up retention policy

**Changes**
Allows users to set a the number of days a show will stay in the Next Up row.

**Issues**
Fixes: https://github.com/jellyfin/jellyfin-androidtv/issues/4772, https://github.com/jellyfin/jellyfin-androidtv/issues/2406

next-up-customization
next-up-retention


crobibero:
This should be using the `nextUpDateCutoff` parameter on the api call and not filtering client-side

https://api.jellyfin.org/#tag/TvShows/operation/GetNextUp

nielsvanvelzen:
Closing as abandoned

Issue #4648, opened 2025-05-08
"Play All" or "Shuffle" for an artist or albums plays a list of random songs


vmpyr:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

1. Open an album/artist's page.
2. Click on "Play All" (for artist page) or "Shuffle" (for artist page as well as album page).
3. Random assortment of songs is played.

The random assortment for artist page "Play All" is seemingly always the same (maybe a set seed somewhere in the code). I have to always go to the first song and then click play from here instead. Because of this, I'm unable to play a single artist without creating a playlist.

### Application version

0.18.8

### Where did you install the app from?

Amazon Appstore

### Device information

Fire TV Stick (3rd Gen)

### Android version

Fire OS 7.6.9.0

### Jellyfin server version

10.10.3

mal20k:
Also seeing this issue. In the page for an album, "Shuffle All" will create a shuffled queue consisting of music from the entire library. The first song is always the same one not from the selected album. From the album page, "Play All" works as expected and adds the full album to the queue with no extras.

From the artist page, both "Play" and "Shuffle All" will create a shuffled queue of music from the entire library. The first song is always the same like the "Shuffle All" from the album page, but is a different song.

Application version
0.18.9

Where did you install the app from?
Google Play Store

Device information
Nvidia Shield (2019)

Android version
9.0.1(33.1.0.283)

Jellyfin server version
10.10.7

BloodyIron:
I am having this issue for both Jellyfin Android TV and Jellyfin on Android sending via Chromecast. It seems all shuffle buttons/instructions I can find for both app versions results in what is being described above:

1. It doesn't shuffle based on current context, it shuffles ALL audio files in ALL libraries
2. It plays the same song first in this shuffle list, except it's _never_ a song I select

I can't find a workaround for this and I don't know when this started as I don't (yet) use Jellyfin in these regards frequently enough to have said back-history.

[...]

PR #5135, opened 2025-11-12
Fix play all and shuffle album/artist playback

balaios:
SDK queries for `MUSIC_ALBUM` and `MUSIC_ARTIST` were missing filtering parameters

**Changes**
- Add  filtering  IS_NOT_FOLDER

**Code assistance**
None

**Issues**
#4648

### DIFF
--- a/.github/workflows/app-lint.yaml
+++ b/.github/workflows/app-lint.yaml
@@ -28,7 +28,7 @@ jobs:
       - name: Run detekt and lint tasks
         run: ./gradlew detekt lint
       - name: Upload SARIF files
-        uses: github/codeql-action/upload-sarif@0499de31b99561a6d14a36a5f662c2a54f91beee # v4.31.2
+        uses: github/codeql-action/upload-sarif@014f16e7ab1402f30e7c3329d33797e7948572db # v4.31.3
         if: ${{ always() }}
         with:
           sarif_file: .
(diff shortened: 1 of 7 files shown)


nielsvanvelzen (changes requested):
The media type filter should already filter out folders, did this change actually make a difference for you?

balaios:
Yes, it makes a critical difference. In my testing, the media type filter alone was not sufficient, and other items were still being added to the queue. I am attaching screenshots and a video showing the behavior before and after this patch.

### Before
mpv-shot0001

https://github.com/user-attachments/assets/606dbdcc-c8ee-457f-95e9-b186f74bcfa3

### After
mpv-shot0002

https://github.com/user-attachments/assets/01c46c6f-a83c-4516-a9b4-307e49cf07d1

This change aligns with the jellyfin-web logic, which also explicitly adds an `IsNotFolder` filter when building the playback queue, as seen in `playbackmanager.js`:
- [Line 1848](https://github.com/jellyfin/jellyfin-web/blob/e1724f5ec11269f78ae458ad9585877ab20ec2c8/src/components/playback/playbackmanager.js#L1848)
- [Line 171](https://github.com/jellyfin/jellyfin-web/blob/e1724f5ec11269f78ae458ad9585877ab20ec2c8/src/components/playback/playbackmanager.js#L171)

nielsvanvelzen:
Ok weird, didn't expect that. Thanks a lot for the explanation though! Can you rebase the PR (as there are conflicts because of your other PR)?

nielsvanvelzen (approved):

PR #5136, opened 2025-11-12
Fix music playback sort order to match web client

balaios:
Tracks were sorted only by SortName, causing incorrect playback order across albums.
Now uses: Album, ParentIndexNumber, IndexNumber, SortName to ensure tracks play
in correct album and track sequence.

**Changes**
- Update MUSIC_ARTIST and MUSIC_ALBUM sort order
- Album → disc number → track number → name
- Aligns with jellyfin-web implementation

**Code assistance**
None

**Issues**


nielsvanvelzen (commented):
Do you have a reference to where jellyfin-web uses these sorting values?

balaios:
> Do you have a reference to where jellyfin-web uses these sorting values?

Yes, here is the reference:

https://github.com/jellyfin/jellyfin-web/blob/026349babbf65270672bc13eba603af4bfb3b63d/src/components/playback/playbackmanager.js#L1848C1-L1858C39

nielsvanvelzen:
Awesome, thanks!

Issue #654, opened 2020-12-14
show titles in poster view "all media"


SonicPotions:
**Describe the feature you'd like**	
Currently when I view all media, only the posters/covers are shown, but no text below it with the title of the album/movie.
The title is only shown for the selected media. This is okayish if you have all covers, but especially for music where I'm missing covers there is no way to see what album it is when the album is not selected.

Sveninndh:
Same issue for fotos or private home video libs.

acordtag:
This request helps a lot visually when navigating on the "All items" view.

oneseventhree:
Switching back to Emby because the posters view on AndriodTV app is much better due to having Title, Year, Resolution of video. Please implement! :)

scottrobertson:
This would be really good. It's so hard to navigate movies/tv shows without titles.

rpm5099:
Lots of requests for this in #1778, I think this issue is not being found because of language barrier. Highly desired feature, very difficult to identify shows in the android tv app.

kiwichick888:
I also want this. Especially as the poster titles are displayed in the web app and Android mobile - it makes no sense to no have them displayed for Android TV (and any other platforms that don't have them). Better spacing between each title for Android TV is also needed.

Web:
<img width="480" alt="Image" src="https://github.com/user-attachments/assets/6cd87180-da9a-4fa8-84c6-2fa27500cbbd" />

Android mobile:
![Image](https://github.com/user-attachments/assets/0ebf0b48-aaed-4e1a-9b83-c7b70d7a5c35)

Android TV:
![Image](https://github.com/user-attachments/assets/6b0114d6-894c-401b-abdb-706e7e261f64)

hard2carry:
I also think its hard to navigate as of now. The users eyes always have to bounce from the selection to the top for the title and its really tiring when browsing a library.
Showing the title underneath the poster / cover would improve the user experience by a lot. Its pretty much standard for most tv apps and also the android and web version of jellyfin.

scottrobertson:
If I spent some time learning how to code for Android, would you accept a PR for this?

nielsvanvelzen:
Yes pull requests properly implementing this would be accepted. Unfortunately the current code for the item grid is extremely messy and has a lot of hardcoded calculations that all need to be adapted and tested so it might not be trivial.

GenericMaleUser:
@nielsvanvelzen Is this where the item grid code you are referring to is located?

[...]

PR #5146, opened 2025-11-14
Adding a Compose-based BrowseGrid

razor67rus:
**Changes**
A grid based on Compose. 

Screenshot_20251114_142439

The current functionality is implemented by 80%. 
Focus, pagination, grid settings, filters, sorting are working.  
Reuse of the ItemRowAdapter - a wrapper has been made, which, according to the idea, should simulate the work of the Repository pattern. (This is, of course, a controversial point).
The card is made as "stupid" as possible. The filling logic is handled by the Mapper. (this was previously done by CardPresenter).
The label of viewed, favorites, and progress indicator are working.

To switch to a new grid, replace the link "BrowseGridFragment" with "BrowseGridComposeFragment" in the Destinations file in the "libraryBrowser" method. I did not include this in PR, because the functionality of the component has not yet been 100% implemented.

**Not implemented:**
- Context menu.
- In a horizontal grid, the picture stretches too much, shifting the caption beyond the boundaries of the cell. (everything is fine in the vertical grid). I don't understand how to fix this yet.
- All sorts of little things, like getting card changes without going back to the grid

**Issues**
[Issues #654](https://github.com/jellyfin/jellyfin-androidtv/issues/654)
[discussion #3390](https://github.com/jellyfin/jellyfin-androidtv/discussions/3390)


razor67rus:
Screenshot_20251119_192224

Added an accent and letter filter.
I had 2 weeks of free time. I did my best. So far, I'm sharing what I have at the moment. I will be glad of any help.

Pentaphon:
> letter filter.

I think that would look better on the right-hand side?

razor67rus:
> I think that would look better on the right-hand side?

I agree, it's better on the right. But when switching to the horizontal grid, it will be difficult to shift the focus to this panel, you will have to scroll through the entire list to the end.

Pentaphon:
> But when switching to the horizontal grid, it will be difficult to shift the focus to this panel, you will have to scroll through the entire list to the end.

You could put the letters somewhere on the top in that configuration? 

Also, could you add the year next to the title?

sjau:
Can this be merged to the android tv app? Have 100s of shows just with thumbnails it's so annoying to navigate. Name of the show below makes it a lot easier.

Pentaphon:
> Can this be merged to the android tv app?

I don't think this will be merged. Otherwise it would have been acknowledged by now.

sjau:
not having TV show name shown below the poster is a real turn off on the android app.

razor67rus:
I doubt that my PR is of any value. Otherwise, they would have already left me at least some kind of comment.

PR #5156, opened 2025-11-16
Add setting to toggle Android TV home recommendations

pricci1:
**Changes**
Adds a new user preference to disable content recommendations in Android TV's home.
I'm actually not sure if the recommendations in home are just a thing of Google TV's version of Android TV.

I wasn't sure where to put this new setting, so I created a new category in the customization preferences called "Android TV".

Recommendations are shown by default, which is the current behavior. After disabling, the changes are actually applied the next time the app is opened (it was simpler than applying immediately).

### DIFF
--- a/app/src/main/res/values-es/strings.xml
+++ b/app/src/main/res/values-es/strings.xml
@@ -404,6 +404,7 @@
     <string name="pref_theme">Tema</string>
     <string name="pref_browsing">Navegación</string>
     <string name="pref_behavior">Comportamiento</string>
+    <string name="pref_android_tv">Android TV</string>
     <string name="pref_accounts">Cuentas</string>
     <string name="sort_accounts_by">Ordenar cuentas por</string>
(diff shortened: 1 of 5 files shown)


nielsvanvelzen:
I don't understand why we need an extra option in the app to disable this. You already can disable this from the Android launcher itself.

afbeelding

mschumacher69:
Why do home recommendations not work on Google TV?

pricci1:
> You already can disable this from the Android launcher itself.

Then I guess this really is a limitation of the launcher in "Chromecast with Google TV". You cannot even set an alternative launcher as default. The only option is to [enable the](https://www.androidcentral.com/how-get-rid-personalized-recommendations-chromecast-google-tv) "Apps only mode".

> Why do home recommendations not work on Google TV?

They do work, but you can't disable them for a specific app.

---

I guess this is a niche feature, I'm closing my PR.

PR #5215, opened 2025-12-03
Migrate telemetry settings to compose

nielsvanvelzen:
**Changes**

Rewrite the existing `CrashReportingPreferencesScreen` using compose within the new settings UI. This is mostly a proof of concept to see what parts of the new UI need improvement, although it's perfectly usable as-is.

image

**Code assistance**

**Issues**


nielsvanvelzen:
Gonna make additional changes later, have some ideas how to improve the UX and I also think we'll have to make some tweaks to the design like using colors for the checkboxes and making them square (because circles usually indicate radio buttons). But those will be in separate PR's :smile:

Issue #4716, opened 2025-06-10
[Feature Request] Add Toggle for Backdrop Blur on Home and Media Pages


nothing2obvi:
### This request respects the following points:

- [x] This request is **not** already on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/labels/enhancement) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Problem Description
Backdrop blurring was recently added to the Jellyfin Android TV client, and while it may enhance readability or visual consistency for some users, it significantly diminishes one of the things I enjoyed most about the platform: appreciating my hand-selected backdrops in full clarity. I’ve invested considerable time and care curating those images, and the new blur effect makes that effort feel wasted. I imagine other users who take pride in customizing their libraries may feel similarly.

### Proposed Solution
It would be great to have user-accessible settings that allow us to toggle backdrop blurring on or off—separately for the home screen and individual media item pages. This would allow those who prefer the visual clarity of unblurred artwork to preserve that experience, while still accommodating users who like the new blurred aesthetic.

### Alternatives considered

I don't think there's an alternative.

tal-sarid:
I agree that this should have been an option when the feature introduced. I can suggest an alternative, that might be even better - a slider for blur intensity to be used for the backdrops, which can be from 0 (without the blur, like it used to be) to some maximum (which can be like it is now, or even higher). The default can be the same intensity as it is now so it won't be noticed by someone who got used to it and like it that way.

nothing2obvi:
@tal-sarid That would be great. As long as there are options to blur on the home screen, in the libraries, and when viewing anything deeper.

For example I would prefer the home screen and the libraries (when browsing several titles) to have a blurred background for the sake of clarity but be able to have it unblurred when I'm viewing an individual movie, show, or episode.

nielsvanvelzen:
> I would prefer the home screen and the libraries (when browsing several titles) to have a blurred background for the sake of clarity but be able to have it unblurred when I'm viewing an individual movie, show, or episode.

This is the planned default behavior, use a blurred backdrop during browsing and show the full graphic when visiting the item itself.

tal-sarid:
Personally, I would prefer two toggles for both scenarios, or two sliders as I suggested earlier. I don't want the blur at all.

shahribms:
[...]

PR #5216, opened 2025-12-04
Add ability to unblur backdrops

akhilmulpurii:
**Changes**
1. Added a User preference option for toggling between backdrop blurring
2. Respect the user preference setting in `BackgroundService.kt`

**Code assistance**
No

**Issues**
Resolves https://github.com/jellyfin/jellyfin-androidtv/issues/4716


akhilmulpurii:
@nielsvanvelzen I have updated the pull request, now using enum BackdropBehavior with DISABLED, BACKDROP_WITH_BLUR, BACKDROP_WITHOUT_BLUR. 

the same is reflected in the customization screen with the ability to select the options from the list instead of checkboxes.

Let me know if any changes are requested from my side.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -42,7 +43,7 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		/**
 		 * Enable background images while browsing
 		 */
-		var backdropEnabled = booleanPreference("pref_show_backdrop", true)
+		var backdropEnabled = enumPreference("pref_show_backdrop", BackdropBehavior.BACKDROP_WITH_BLUR)
```
nielsvanvelzen: Preference should be renamed as the type changed and a migration should be added to convert the old pref to the new one (in the toVersion=9 block below).


```suggestion
		var backdropBehavior = enumPreference("backdrop_behavior", BackdropBehavior.BACKDROP_WITH_BLUR)
```

Without this the app just crashes on start (as it just did for me while testing :p)

nielsvanvelzen (changes requested):

akhilmulpurii:
@nielsvanvelzen I have added the changes to reflect the new variable name and run the migration.

Issue #5090, opened 2025-11-03
Live TV playback skipping and rewinding.


kingezk:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

1. Updated the app to 0.19.2 from Amazon
2. Launched app 
3. Selected live channel to view
4. Live channel plays
5. Play back restarts a few seconds later (sometimes playback freezes as audio still plays)

Side loaded  0.18.11. No issues with streaming.

### Media info of the file

```shell

```

### Client logs

[2025-11-03 03:35:20.528 +00:00] [INF] Opening channel stream from "Xtream Live", external channel Id: "3e4c775d-0000-8bfa-0000-000000000000"
[2025-11-03 03:35:20.545 +00:00] [INF] Starting restream for channel "3d4c775d-0000-0000-0000-fa8b00000000".
[2025-11-03 03:35:21.742 +00:00] [INF] Live tv media info probe took "0.017679" seconds
[...]

PR #5229, opened 2025-12-08
Fix Live TV direct play stability

tobbi007:
### Changes

- Switch stream URL construction in `PlaybackManager.kt` from `getVideoStreamUrl()` to `getVideoStreamByContainerUrl()`, ensuring the `/Videos/{itemId}/stream.{container}` format like in the 0.18.11 client and passing `container`, `liveStreamId`, and `playSessionId` for Live TV sessions.
- Configure `OkHttpDataSource.Factory` in `PlaybackModule.kt` with explicit timeouts (`connectTimeout = 6s`, `readTimeout = 30s`), matching the legacy `DefaultHttpDataSource` behavior for stable long-lived connections.

### Chronological debugging steps

1. **Added `container`, `liveStreamId`, and `playSessionId` to API calls.**
   - Result: Live TV no longer remuxed/transcoded unnecessarily, but streams stopped after some time.
2. **Observed in Dispatcharr:** Streams were stopped and restarted repeatedly until they stopped completely.
3. **Reverted to `DefaultHttpDataSource` with legacy timeouts.**
   - Result: Stream crashes were fixed.
4. **Switched back to OkHttp with explicit timeouts.**
   - Result: Playback has been stable since.

### Code assistance

> GitHub Copilot in VS Code (Claude Opus 4.5 & GPT-4.1) was used to analyze differences between the v0.18.11 and v0.19.x client codebases and their respective SDK versions. This analysis helped identify changes in Live TV playback handling, especially the HTTP client configuration difference that caused stream instability.

And it was used to generate this PR text 👍 

### Related issues

- Related to #5090
- Related to #4962
- [Forum thread: LiveTV errors and playback stuttering since 0.19.x](https://forum.jellyfin.org/t-livetv-errors-and-playback-stuttering-since-0-19-x)

### DIFF
--- a/playback/jellyfin/src/main/kotlin/mediastream/JellyfinMediaStreamResolver.kt
+++ b/playback/jellyfin/src/main/kotlin/mediastream/JellyfinMediaStreamResolver.kt
@@ -37,9 +37,11 @@ class JellyfinMediaStreamResolver(
 				conversionMethod = MediaConversionMethod.None,
 				url = api.videosApi.getVideoStreamUrl(
 					itemId = baseItem.id,
+					container = mediaInfo.mediaSource.container,
 					mediaSourceId = mediaInfo.mediaSource.id,
 					static = true,
 					tag = mediaInfo.mediaSource.eTag,
+					liveStreamId = mediaInfo.mediaSource.liveStreamId,
 				)
 			)
 
@@ -49,9 +51,11 @@ class JellyfinMediaStreamResolver(
 				conversionMethod = MediaConversionMethod.None,
 				url = api.audioApi.getAudioStreamUrl(
 					itemId = baseItem.id,
+					container = mediaInfo.mediaSource.container,
 					mediaSourceId = mediaInfo.mediaSource.id,
 					static = true,
 					tag = mediaInfo.mediaSource.eTag,
+					liveStreamId = mediaInfo.mediaSource.liveStreamId,
 				)
 			)
 


app/src/main/java/org/jellyfin/androidtv/di/PlaybackModule.kt
```diff
@@ -39,12 +39,14 @@ val playbackModule = module {
...
-		val httpClientOptions = get<HttpClientOptions>()
-
-		OkHttpDataSource.Factory(okHttpFactory.createClient(httpClientOptions))
+		val okHttpClient = OkHttpClient.Builder()
+			.connectTimeout(6, TimeUnit.SECONDS)
+			.readTimeout(30, TimeUnit.SECONDS)
+			.followRedirects(true)
+			.followSslRedirects(true)
+			.build()
+		OkHttpDataSource.Factory(okHttpClient)
```
nielsvanvelzen: the okhttp client send to media3 must be built using our own factory. it also should not change options that don't need changing to fix the bug

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackManager.kt
```diff
@@ -38,11 +38,14 @@ private fun createStreamInfo(
...
 			source.isRemote && source.path != null -> source.path
-			else -> api.videosApi.getVideoStreamUrl(
+			else -> api.videosApi.getVideoStreamByContainerUrl(
 				itemId = itemId,
+				container = source.container ?: "ts",
 				mediaSourceId = source.id,
 				static = true,
 				tag = source.eTag,
+				liveStreamId = source.liveStreamId,
+				playSessionId = response.playSessionId,
```
nielsvanvelzen: Did adding the play session id truly make a difference in behavior or was it added just because?

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackManager.kt
```diff
@@ -38,11 +38,14 @@ private fun createStreamInfo(
 		container = source.container
 		mediaUrl = when {
 			source.isRemote && source.path != null -> source.path
-			else -> api.videosApi.getVideoStreamUrl(
+			else -> api.videosApi.getVideoStreamByContainerUrl(
 				itemId = itemId,
+				container = source.container ?: "ts",
```
nielsvanvelzen: The container fallback should be null

app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackManager.kt
tobbi007: I don't think it is needed. I will remove it.

app/src/main/java/org/jellyfin/androidtv/di/PlaybackModule.kt
tobbi007: I see that in OkHttpFactory.kt most of the timeouts I had in the default OkHttpClient are already set. 
I noticed that in OkHttpFactory.kt additionally the callTimeout was set (to 30s).

So I tried adding that to the solution I provided earlier to confirm that this is what is causing the streams to stop.
--> With callTimeout set to 30s the stream stops playing. So this is the root cause.

My question is, what do we do here. I don't know the project enough to know why the callTimeout is set in OkHttpFactory.kt and what side effects it can have if we remove it.

nielsvanvelzen: Does this work?

```
	single<HttpDataSource.Factory> {
		val okHttpFactory = get<OkHttpFactory>()
		val httpClientOptions = get<HttpClientOptions>().copy(
			// Disable socket timeout for media playback as this causes issues with Live TV
			socketTimeout = Duration.ZERO
		)

		OkHttpDataSource.Factory(okHttpFactory.createClient(httpClientOptions))
	}
```

tobbi007: Yes that works, but it is the requestTimeout that we need to change, not the socketTimeout. 

```
public fun createClient(httpClientOptions: HttpClientOptions): okhttp3.OkHttpClient = base.newBuilder()
		.followRedirects(httpClientOptions.followRedirects)
		.connectTimeout(httpClientOptions.connectTimeout.toJavaDuration())
		.callTimeout(httpClientOptions.requestTimeout.toJavaDuration())
		.readTimeout(httpClientOptions.socketTimeout.toJavaDuration())
		.writeTimeout(httpClientOptions.socketTimeout.toJavaDuration())
		.build()
```

Code changes commited.

nielsvanvelzen:
Thanks!

martenumberto:
I'm facing problems with this Code change. In my case it throws an Live-TV/Player Error after 10 seconds and a remux playback starts. Can provide logs later.

brianjmurrell:
I was soooo hopeful that this was going to correct the issues I am having with Live TV DirectPlay, but sadly it is not.  I added this [comment](https://github.com/jellyfin/jellyfin/issues/15459#issuecomment-3708819659) about it with details from the Jellyfin server side.  TL;DR: stream was still transcoded due to a _DirectPlay_ error.

From the jellyfin-androidtv side, this is what is in the [`adb log`](https://github.com/user-attachments/files/24426570/logcat.txt).

I suppose I really should open a new issue.

PR #5238, opened 2025-12-11
Fix French label for goto series

louseee:
**Changes**
This PR fixes a minor French translation issue.
The label lbl_goto_series previously displayed "Aller vers les séries", which is grammatically incorrect in context.
It is now updated to the correct singular form "Aller vers la série".

**Code assistance**
No automated code generation was used.
Only a manual translation correction was performed.

**Issues**
No issue was opened for this change.


nielsvanvelzen:
Translations need to be changed via translate.jellyfin.org. We do not accept string changes except for the source file on GitHub.

PR #5260, opened 2025-12-20
Use fragmented MP4 as HLS container

nielsvanvelzen:
Needs testing

**Changes**

Use fragmented MP4 instead of TS when transcoding video as this allows for more audio codecs and is more modern and better supported by media3.

**Code assistance**

**Issues**


Leonashub:
When using this build on a firestick 4k max 2nd gen:

Files with dts: no audio (also no audio when forcing transcoding)
Files with truehd: only audio when direct play is supported. Transcoding or remuxing also causes no audio. (This is because the video codec wasn't supported on those files and fell back to hdr or sdr)

Files with dd+ atmos work fine regardless of direct play or transcoding.
Haven't tested other codecs thoroughly but I suspect they work fine

PR #5261, opened 2025-12-21
Fix SDR fallback selection over HDR for HLS streams

### LINKED ISSUE #5245: ExoPlayer selects SDR over HDR remux on first playback for unsupported HDR formats

### Describe the bug

When playing content with an unsupported HDR format (e.g., Dolby Vision Profile 7 or HDR10+) on a device that supports HDR10, the server responds with an HLS manifest containing both:

- an HDR remux variant (PQ / HDR10)
- an SDR transcode variant

On first playback after opening the app, ExoPlayer selects the SDR transcode instead of the HDR remux. This occurs before any playback heuristics (e.g., bandwidth estimation) have run. Since ExoPlayer does not currently parse or use the draft `VIDEO-RANGE` attribute and both variants report the same bitrate, resolution, etc., initial selection favors the SDR stream because it has lower codec requirements.

After ~1 minute, ExoPlayer attempts to switch to the HDR variant, at which point playback freezes, possibly because the server no longer retains the HDR track after the SDR variant has been selected (could be mistaken; haven't fully investigated).

#### Example HLS Manifest

```
#EXTM3U
#EXT-X-STREAM-INF:BANDWIDTH=60367918,AVERAGE-BANDWIDTH=60367918,VIDEO-RANGE=PQ,CODECS="hvc1.2.4.L153.B0,mp4a.40.2",RESOLUTION=3840x2160,FRAME-RATE=23.976
main.m3u8?...
#EXT-X-STREAM-INF:BANDWIDTH=60367918,AVERAGE-BANDWIDTH=60367918,VIDEO-RANGE=SDR,CODECS="hvc1.2.4.L150.B0,mp4a.40.2",RESOLUTION=3840x2160,FRAME-RATE=23.976
main.m3u8?...
#EXT-X-IMAGE-STREAM-INF:BANDWIDTH=8241,RESOLUTION=320x180,CODECS="jpeg",URI="Trickplay/320/tiles.m3u8?MediaSourceId=***&ApiKey=***"
```

#### Possible Fixes

1. **Workaround:** Setting `setForceHighestSupportedBitrate(true)` on the `DefaultTrackSelector` causes ExoPlayer to select the HDR variant on first playback. This can be scoped to only non-live and transcoded HLS content.
2. **Wrap HLS parser / metadata injection**: Inject `VIDEO-RANGE` metadata into each variant’s `Format` object and override default track selection to prefer HDR when all else is equal. Again, could scope this behavior.

### Media info of the file

```shell

```

### Media capabilities report

_No response_

### Client logs

_No response_

### FFmpeg logs

```shell

```

### Application version

0.19.5

Google Play Store

### Device information

Nvidia Shield TV (2017), Chromecast with Google TV (2022)

### Android version

Android 11, Android 14

### Jellyfin server version

10.11.5

### Other sources

_No response_

clams4shoes:
This PR fixes an issue where ExoPlayer selects an SDR fallback variant (typically a transcode) instead of an available HDR variant (typically a remux) when playing HLS streams.

**Changes**

- Wrap the HLS playlist parser to map the `VIDEO-RANGE` attribute to variant streams and add it to their `Format` metadata (`VideoRangeEntry`)
- Wrap the adaptive track selection to prefer HDR variants over SDR when tracks are otherwise equivalent (resolution, bitrate, and frame rate)

**Code assistance**

- Code review

**Issues**

Fixes #5245

**Additional**
- Has not been tested with Live TV; behavior could be scoped to transcodes and remuxes.
- Feedback, suggestions, and testing are greatly appreciated.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/AdaptiveHdrTrackSelection.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/AdaptiveHdrTrackSelection.kt
@@ -1,0 +1,120 @@
+package org.jellyfin.androidtv.ui.playback
+
+import androidx.annotation.OptIn
+import androidx.media3.common.Format
+import androidx.media3.common.MimeTypes
+import androidx.media3.common.Timeline
+import androidx.media3.common.TrackGroup
+import androidx.media3.common.util.UnstableApi
+import androidx.media3.exoplayer.source.MediaSource
+import androidx.media3.exoplayer.source.chunk.MediaChunk
+import androidx.media3.exoplayer.source.chunk.MediaChunkIterator
+import androidx.media3.exoplayer.trackselection.AdaptiveTrackSelection
+import androidx.media3.exoplayer.trackselection.BaseTrackSelection
+import androidx.media3.exoplayer.trackselection.ExoTrackSelection
+import androidx.media3.exoplayer.upstream.BandwidthMeter
+
+@OptIn(UnstableApi::class)
+class AdaptiveHdrTrackSelection(
+	private val delegate: AdaptiveTrackSelection
+) : BaseTrackSelection(
+	delegate.trackGroup,
+	IntArray(delegate.length()) { it },
+	TYPE_UNSET
+) {
+
+	companion object {
+		const val SELECTION_REASON_ADAPTIVE_HDR_UPGRADE = 1001
+	}
+
+	private var selectedIndex = 0
+
+	override fun getSelectedIndex(): Int = selectedIndex
+
+	override fun updateSelectedTrack(
+		playbackPositionUs: Long,
+		bufferedDurationUs: Long,
+		availableDurationUs: Long,
+		queue: List<MediaChunk>,
+		mediaChunkIterators: Array<out MediaChunkIterator>
+	) {
+		delegate.updateSelectedTrack(
+			playbackPositionUs,
+			bufferedDurationUs,
+			availableDurationUs,
+			queue,
+			mediaChunkIterators
+		)
+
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
While looking into the issue I kinda came to the same conclusion; media3 doesn't save the dynamic range to the format and doesn't use it in the selection so we need to do that ourselves.
However, maintaining large customizations on top of media3 is prone to issues. Especially when using the UnstableApi marker is needed.

I'd rather see this changed on the server (e.g. add a new option to device profiles that defaults to current behavior. Shouldn't be too hard and makes sure transcoded media always uses correct streams) and/or media3 upstream. I believe other users of the media3 could also benefit from better behavior regarding hdr/non-hdr entries so I hope the media3 developers are open for this.

clams4shoes:
I agree -- I'd strongly prefer to avoid overriding core Media3 functionality, since that can be more fragile. This solution could likely avoid some of the use of the `@UnstableApi` if the track selection logic were moved into an override on the track selector. While digging into this, I was surprised that the track selector builder doesn't provide an easy way to preference HDR tracks.

Another, more hacky, option would be to remove the SDR variant before it reaches the parser, which would implicitly force selection of the HDR variant.

Like you said though, it would be nice if the device profile could signal the server to exclude the SDR fallback variant entirely, at least until media3 improves its track selection capabilities.

What do you think is the best way to loop the server team into this discussion? And would you prefer to keep this PR open until we have a definitive solution, or close it and move the discussion back to the issue tracker?

nielsvanvelzen:
Closing as invalid, issue should be fixed either in media3 or on the server to produce different HLS playlists (potentially with a device profile option? idk).

> What do you think is the best way to loop the server team into this discussion?

Open an issue in the server repository / join the development chat.

PR #5262, opened 2025-12-21
Use dedicated screen for about section in settings

nielsvanvelzen:
**Changes**
- Add an "about" screen to settings instead of the previous section
- Add support for copying app name/device model
- Add "copyAction" utility

**Code assistance**

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/SettingsMainScreen.kt
```diff
@@ -69,8 +69,12 @@ fun SettingsMainScreen() {
 			)
 		}
 
-		settingsAboutItems(
-			openLicenses = { router.push(Routes.LICENSES) }
-		)
+		item {
+			ListButton(
+				leadingContent = { Icon(painterResource(R.drawable.ic_jellyfin), contentDescription = null) },
```
nielsvanvelzen, line 74: Wrong string

Issue #4838, opened 2025-08-03
CCwGTV 4K direct plays AV1 since Android 14 upgrade


solidsnake1298:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

Since my CCwGTV 4K was upgraded to Android 14 it now direct plays AV1 via software decoding, resulting in extremely poor frame rates.

Setting audio output to "downmix to stereo" is a workaround to trigger transcoding to HEVC.

### Media info of the file

```shell
KPop Demon Hunters (2025) [tmdbid-803796]
Container: mkv
Path: /media/library/Movies/KPop Demon Hunters (2025) [tmdbid-803796]/KPop Demon Hunters (2025) [tmdbid-803796].mkv
Size: 2292 MB

Audio
Title: English - OPUS - 5.1 - Default
Language: eng
Codec: OPUS
AVC: No
Layout: 5.1
Channels: 6 ch
Bitrate: 768 kbps
Sample rate: 48000 Hz
Default: Yes
Forced: No
External: No

Video
Title: 1080p AV1 SDR
Codec: AV1
AVC: No
Profile: Main
Level: 8
Resolution: 1920x800
Aspect ratio: 2.40:1
Anamorphic: No
Interlaced: No
Framerate: 23.976025
Bitrate: 3216 kbps
Bit depth: 10 bit
Video range: SDR
Video range type: SDR
Pixel format: yuv420p10le
Ref frames: 1

Subtitle
Title: English (forced) - SUBRIP
Language: eng
Codec: SUBRIP
AVC: No
Default: No
Forced: No
External: No

Subtitle
Title: English - SUBRIP
Language: eng
Codec: SUBRIP
AVC: No
Default: No
Forced: No
External: No

Subtitle
Title: English (SDH) - SUBRIP
Language: eng
Codec: SUBRIP
AVC: No
Default: No
Forced: No
External: No
```

### Client logs

n/a

### FFmpeg logs

```shell
n/a
```

### Application version

0.18.11

### Where did you install the app from?

Google Play Store

### Device information

[...]

PR #5275, opened 2025-12-26
Add option to not advertise software codecs as supported

luca-schlecker:
Sometimes, software-codecs can cause performance-related issues like seen in #4838. The added option allows disabling software codecs, forcing a transcode to a non-software (hardware-accelerated) codec.

The latest version did not fix the issue described in #4838 for me. The software decoder still struggled with my particular file, even though the resolution was within the advertised range.

**Changes**
- A new checkbox in the _Advanced Playback Preferences_, toggling whether or not software codecs are used.
- An accompanying new user preference (`use_software_codecs`)
- Updated logic in `createDeviceProfile` and `MediaCodecCapabilitiesTest` to allow excluding software-based codecs.
- A new translatable string `pref_use_software_codecs`.

**Code assistance**
No AI-based assistance tools were used.

**Issues**
I don't know of an open issue for this.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -259,6 +259,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Enable PGS subtitle direct-play.
 		 */
 		var pgsDirectPlay = booleanPreference("pgs_enabled", true)
+
+		/**
+		 * Enable the use of software-based codecs.
+		 */
+		var useSoftwareCodecs = booleanPreference("use_software_codecs", true)
 	}
 
 	init {
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -254,6 +254,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Enable PGS subtitle direct-play.
 		 */
 		var pgsDirectPlay = booleanPreference("pgs_enabled", true)
+
+		/**
+		 * Enable the use of software-based codecs.
+		 */
+		var softwareCodecsEnabled = booleanPreference("software_codecs_enabled", true)
 	}
 
 	init {
--- a/app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
@@ -217,6 +217,16 @@ fun SettingsPlaybackAdvancedScreen() {
 			)
 		}
 
+		item {
+			var useSoftwareCodecs by rememberPreference(userPreferences, UserPreferences.useSoftwareCodecs)
+
+			ListButton(
+				headingContent = { Text(stringResource(R.string.pref_use_software_codecs)) },
+				trailingContent = { Checkbox(checked = useSoftwareCodecs) },
+				onClick = { useSoftwareCodecs = !useSoftwareCodecs }
+			)
+		}
+
 		item { ListSection(headingContent = { Text(stringResource(R.string.pref_live_tv_cat)) }) }
 
 		item {
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
@@ -1,6 +1,7 @@
 package org.jellyfin.androidtv.util.profile
 
 import android.content.Context
+import android.media.MediaCodecInfo
 import android.media.MediaCodecInfo.CodecProfileLevel
 import android.media.MediaCodecList
 import android.media.MediaFormat
@@ -15,6 +16,7 @@ class MediaCodecCapabilitiesTest(
 ) {
 	private val display by lazy { ContextCompat.getDisplayOrDefault(context) }
 	private val mediaCodecList by lazy { MediaCodecList(MediaCodecList.REGULAR_CODECS) }
+	var useSoftwareCodecs = true
 
 	// Map common Dolby Vision Profiles to their corresponding CodecProfileLevel constant
 	private object DolbyVisionProfiles {
@@ -207,6 +209,7 @@ class MediaCodecCapabilitiesTest(
 
 		for (info in mediaCodecList.codecInfos) {
 			if (info.isEncoder) continue
+			if (!useSoftwareCodecs && info.isSoftwareCodec()) continue
 
 			try {
 				val capabilities = info.getCapabilitiesForType(mime)
@@ -226,6 +229,7 @@ class MediaCodecCapabilitiesTest(
 	private fun hasDecoder(mime: String, profile: Int, level: Int): Boolean {
 		for (info in mediaCodecList.codecInfos) {
 			if (info.isEncoder) continue
+			if (!useSoftwareCodecs && info.isSoftwareCodec()) continue
 
 			try {
 				val capabilities = info.getCapabilitiesForType(mime)
@@ -253,6 +257,7 @@ class MediaCodecCapabilitiesTest(
 	private fun hasCodecForMime(mime: String): Boolean {
 		for (info in mediaCodecList.codecInfos) {
 			if (info.isEncoder) continue
+			if (!useSoftwareCodecs && info.isSoftwareCodec()) continue
 
 			if (info.supportedTypes.any { it.equals(mime, ignoreCase = true) }) {
 				Timber.i("found codec %s for mime %s", info.name, mime)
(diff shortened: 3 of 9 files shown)


app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
```diff
@@ -308,4 +315,16 @@ class MediaCodecCapabilitiesTest(
...
+
+	private fun MediaCodecInfo.isSoftwareCodec(): Boolean {
+		return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+			this.isSoftwareOnly
+		} else {
+			val name = this.name
+
+			// detect software-based codecs by name; approach taken from here:
+			// https://android.googlesource.com/platform/frameworks/av/+/d5dbc2b31c492be23fc4d4a96619330b63499b06/media/libstagefright/MediaCodecList.cpp#364
+			name.startsWith("OMX.google.", ignoreCase = true) || name.startsWith("c2.android.", ignoreCase = true)
```
nielsvanvelzen: The codecs available on a device are highly vendor specific. This line assumes a lot of things and will definitely be wrong on a lot of devices. We should ONLY trust the `isSoftwareOnly` property, which means this check can't work on Android 9 and older.

luca-schlecker: Ok, I've removed the problematic check and instead always return false there. This results in the default behavior prior to this change if `isSoftwareOnly` is not supported. Should there be an extra description in the settings option to communicate this limitation to the user?

nielsvanvelzen: The setting should be hidden on those devices as it won't function.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -259,6 +259,11 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 * Enable PGS subtitle direct-play.
 		 */
 		var pgsDirectPlay = booleanPreference("pgs_enabled", true)
+
+		/**
+		 * Enable the use of software-based codecs.
+		 */
+		var useSoftwareCodecs = booleanPreference("use_software_codecs", true)
```
nielsvanvelzen:
```suggestion
		var softwareCodecsEnabled = booleanPreference("software_codecs_enabled", true)
```

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
```diff
@@ -308,4 +315,12 @@ class MediaCodecCapabilitiesTest(
 
 		return Size(maxWidth, maxHeight)
 	}
+
+	private fun MediaCodecInfo.isSoftwareCodec(): Boolean {
```
nielsvanvelzen: function could be simplified to a getter

```kt
private val MediaCodecInfo.isSoftwareCodec: Boolean
	get() = Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q && isSoftwareOnly
```

app/src/main/res/values-de/strings.xml
```diff
@@ -646,4 +646,5 @@
     <string name="playback_buffer_extra_large">Sehr groß</string>
     <string name="pref_burn_subtitles_when_transcoding">Untertitel beim Transkodieren einbrennen</string>
     <string name="pref_burn_subtitles_when_transcoding_description">Untertitel beim Transkodieren immer einbrennen. Dies kann Fehler bei der Untertitel-Synchronisation beheben, aber verhindert individualisierte Untertitel.</string>
+    <string name="pref_use_software_codecs">Software-Codecs verwenden</string>
```
nielsvanvelzen: Please do not push translations via pull requests. Only the `values/strings.xml` should be changed, other files need to be managed via translate.jellyfin.org after merging.

app/src/main/res/values-en-rGB/strings.xml
```diff
@@ -642,4 +642,5 @@
     <string name="sdr">SDR</string>
     <string name="hdr10">HDR10</string>
     <string name="hlg">HLG</string>
+    <string name="pref_use_software_codecs">Use software codecs</string>
```
nielsvanvelzen: Please do not push translations via pull requests. Only the `values/strings.xml` should be changed, other files need to be managed via translate.jellyfin.org after merging.

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
```diff
@@ -98,7 +99,10 @@ fun createDeviceProfile(
 	pgsDirectPlay: Boolean,
 	userAVCLevel: Int?,
 	userHEVCLevel: Int?,
+	useSoftwareCodecs: Boolean,
 ) = buildDeviceProfile {
+	mediaTest.useSoftwareCodecs = useSoftwareCodecs
```
nielsvanvelzen: No static properties, pass variable via called functions

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
luca-schlecker: The settings entry is now hidden if it is not supported by the device. 👍

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
luca-schlecker: I've opted for making it a parameter of the primary constructor of the `MediaCodecCapabilitiesTest` class as it is used in multiple functions across it. 👍

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
```diff
@@ -216,6 +217,18 @@ fun SettingsPlaybackAdvancedScreen() {
...
 
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+			item {
+				var softwareCodecsEnabled by rememberPreference(userPreferences, UserPreferences.softwareCodecsEnabled)
+
+				ListButton(
+					headingContent = { Text(stringResource(R.string.pref_software_codecs_enabled)) },
+					trailingContent = { Checkbox(checked = softwareCodecsEnabled) },
+					onClick = { softwareCodecsEnabled = !softwareCodecsEnabled }
+				)
```
nielsvanvelzen: Should add a caption here with some more info about what this option does, I'd say something like how it is more likely to transcode with this disabled.

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
```diff
@@ -216,6 +217,18 @@ fun SettingsPlaybackAdvancedScreen() {
 			)
 		}
 
+		if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+			item {
```
nielsvanvelzen: I'd say it makes more sense to put this setting on the "media codecs" page

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
luca-schlecker: How do you like _"Some software-based implementations may degrade performance. Disabling this option makes transcoding more likely."_ as a description?

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
luca-schlecker: Will move it there 👍

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
nielsvanvelzen: Sounds good, just need to make sure it fits in the UI but I think it will.

luca-schlecker: Will test it this evening.

nielsvanvelzen: Yeah seems to fit fine

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks for the PR!

luca-schlecker:
Just as a follow-up, the option description does seem to fit quite nicely:
image

PR #5318, opened 2026-01-05
fix: stuttering 4k HEVC streaming on Raspberry Pi 4b

flxzt:
**Changes**
I am running Android TV on a Raspberry Pi 4 and have a lot of 4K HEVC encoded movies that I want to stream using Jellyfin.
Even though the Raspberry Pi seems to be able to decode 4k HEVC with hardware acceleration,
the practical implementation in ffpmeg apparently is so inefficient that I am still experiencing only ~5 fps. (see dev comment: http://disq.us/p/34flfsd)

As a workaround, I am proposing to limit reported HEVC decoding support to 1080p, which then results in server side transcoding, fixing the stuttering playback.

I am aware that this is a bit of a "quick fix", but I am so far unfamiliar how all parts fit together.
Is there another preferred way this could be implemented?

**Code assistance**
No AI was used.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/KnownDefects.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/KnownDefects.kt
@@ -3,7 +3,7 @@ package org.jellyfin.androidtv.util.profile
 import android.os.Build
 
 /**
- * List of devie models with known HEVC DoVi/HDR10+ playback issues.
+ * List of device models with known HEVC DoVi/HDR10+ playback issues.
  */
 private val modelsWithDoViHdr10PlusBug = listOf(
 	"AFTKRT", // Amazon Fire TV 4K Max (2nd Gen)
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:
Thanks for the pull request, unfortunately we only add known defects for devices with high usage rates with their official Android distribution. This does not include one very specific raspberry pi model where no official distribution exists for.

flxzt:
I see. Would adding an option to toggle this behaviour be something that will get accepted?

nielsvanvelzen:
There are plans for more customization of the device profile but that won't be as specific as changing the maximum supported resolution per codec, so no.

flxzt:
Is there something else that can be added to make 4k HEVC work smoothly for the Raspberry Pi 4b Android TV build? (as in forcing server-side transcoding, which probably is the only solution in this case)

I don't know the exact numbers, but I would assume it is popular as custom hardware for Android TV, and skimming through old issues there seem to be other weakly powered TV boxes that report support for 4k HEVC, but are actually not powerful enough.

Do you have a suggestion how it can be solved?

nielsvanvelzen:
You're using some unofficial build of lineageos for a single-board computer that isn't even made for running Android that wrongfully exposes support for HEVC. We cannot add workarounds for such niches. Ask the maintainer of that lineageos build to fix the reporting instead.

Issue #5093, opened 2025-11-03
[0.19.2] Regression: Jellyfin Android TV 0.19.2 triggers unnecessary transcoding for HEVC Dolby Vision video (was Direct Play in previous versions)


DenisNikolas:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

**Environment**
Server: Jellyfin 10.11.1
Client: Jellyfin Android TV 0.19.2
Device: Google TV (Android 12, built-in player)
Connection: Local network
**Media file:**
Container: MKV
Video: HEVC (Main 10), Dolby Vision Profile 8.6 (HDR10 compatible), 3840×2160
Audio: DTS-HD MA 5.1 (Russian, default track)
**Behavior in** 0.18.x: Direct Play
**Behavior in** 0.19.2: Full video+audio transcoding
**Expected:** Direct Play (device supports HEVC, DV, DTS natively)
**Problem description**
Since updating to Android TV client 0.19.2, Jellyfin server forces transcoding of files that are natively supported by the TV.
In all previous versions of the client, the same file played via Direct Play without transcoding.
The TV supports:
HEVC Main10 / Dolby Vision Profile 5–8.x
DTS and EAC3 passthrough
HDR10 and Dolby Vision playback in MKV container
However, the new client incorrectly reports these codecs as unsupported, triggering full software transcoding on the server.
**Expected behavior**
The file should Direct Play without any transcoding.
The server should correctly detect the client’s HEVC and DTS capability.
**Additional context**
The same file plays perfectly on this TV via system player (HEVC + Dolby Vision + DTS).
Previous Jellyfin Android TV versions (0.18.x) played this file without transcoding.
Regression appeared after 0.19.2 updates.
**Possible cause**
Regression in the Android TV client capability reporting —
HEVC and DTS are incorrectly marked as unsupported in the device profile,
causing the server to force transcoding even when not required.

In version 0.19.2, both the built-in player and external player support (e.g. ViMu https://github.com/jellyfin/jellyfin-androidtv/issues/5092) are broken.
[...]

Issue #5073, opened 2025-10-30
Dolby Vision profile 7 with HDR 10 does not fallback to HDR


Kranzes:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

1. Update Jellyfin server from 10.10.7 to 10.11.1
2. Update Jellyfin Android TV app from 0.18.11 to 0.19.0
3. Re-scan all the media library as advised
4. Movies with Dolby Vision profile 7 no longer direct play as they previously did

These media files have HDR10 or HDR10+ fallback, and some of them I direct played with HDR less than 24 hours ago.

Forgive me for creating another one of these issues, but the other ones seem to be rather confusing or Amazon Firestick related (which I do not have). Feel free if you think this issue is a duplicate of another one already open.
 

### Media info of the file

```shell
Video
Title: 4K HEVC Dolby Vision Profile 7.6 (HDR10)
Codec: HEVC
AVC: No
Profile: Main 10
Level: 153
Resolution: 3840x2160
Aspect ratio: 16:9
Anamorphic: No
Interlaced: No
Framerate: 23.976025
Bitrate: 76176 kbps
Bit depth: 10 bit
Video range: HDR
Video range type: DOVIWithEL
DV title: Dolby Vision Profile 7.6 (HDR10)
DV version major: 1
DV version minor: 0
DV profile: 7
DV level: 6
DV rpu preset flag: 1
DV el preset flag: 1
DV bl preset flag: 1
DV bl signal compatibility id: 6
Color space: bt2020nc
Color transfer: smpte2084
Color primaries: bt2020
Pixel format: yuv420p10le
Ref frames: 1
```

### Client logs

Device playback compatibility report:  https://gist.github.com/Kranzes/da7da3b22d5c9afe689c75c6a818ca85

### Application version

0.19.0

### Where did you install the app from?

Sideloaded APK

### Device information

TCL C8K

### Android version

Android 12

### Jellyfin server version

10.11.1

clams4shoes:
[...]

PR #5324, opened 2026-01-10
Add option to open single items in an external player

dhartung:
**Motivation**
The internal player works for me 95% of the time, however it fails for some specific media types. I want to have an option to open just those entries in an external player without changing my global settings.

**Changes**
- Add general context menu option to open in external player
- Add context menu option in FullDetailsFragment
- Add new string "Open in external player", translation needed

**Code assistance**
LLM (Claude) was used to identify relevant files, code changes were mainly done by hand.

**Issues**
none that I know of

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
@@ -1227,6 +1227,10 @@ public class FullDetailsFragment extends Fragment implements RecordingIndicatorV
         });
     }
 
+    void playExternal(final BaseItemDto item) {
+        playbackHelper.getValue().retrieveAndPlayExternal(item.getId(), false, getContext());
+    }
+
     void play(final List<BaseItemDto> items, final int pos, final boolean shuffle) {
         if (items.isEmpty()) return;
         if (shuffle) Collections.shuffle(items);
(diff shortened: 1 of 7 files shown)


nielsvanvelzen (changes requested):
Just copying my comment from another PR for almost the exact same thing:

> 

As mentioned previously, there is currently no intention to add this feature. If you want to use an external player you can configure it in the app settings.

dhartung:
I haven't found any other PRs mentioning this feature. Could you briefly explain why this feature is not needed (or link to the other PR if it has already been discussed there)?

As mentioned, the current all-or-nothing approach is not ideal for my use case.

eyJhb:
@nielsvanvelzen there are numerous bugs in internal video player, that makes playing some HDR/DV profiles impossible, but works well in the external video player. 99% of the time one wants to use the internal player, but for some files it's quite useful to use the external player. However, it's buried far way in the settings.

This PR would make it easier to do it one off, at least until it's fixed. As far as I understand, in previous versions the external player was available on the details page.

It could be nice to add this back, at least until the playback issues has been fixed. This PR seems to be fairly easy to revert if it's not needed anymore. 

Are there any changes that can be made to this PR, so it might be merged? Should it be put behind a settings flag? Or under the `...`?

Related issues:
- https://github.com/jellyfin/jellyfin-androidtv/issues/49
- https://github.com/jellyfin/jellyfin-androidtv/issues/5073
- https://github.com/jellyfin/jellyfin-androidtv/issues/5093

nielsvanvelzen:
Closing as unwanted

PR #5326, opened 2026-01-12
Fix audio track not persisting across episode transitions

skalthoff:
**Changes**
Removed an unnecessary reset of audioStreamIndex in startItem() that was overriding the language preference already set by buildExoPlayerOptions(). The existing mechanism that stores the user's audio language was working correctly—it was just being immediately undone.

**Code assistance**
Debugging guidance to trace the audio preference flow and identify where it was being reset.

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -607,7 +607,8 @@ public class PlaybackController implements PlaybackControllerNotifiable {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (changes requested):
This essentially reverts #5185, please elaborate why this change is needed

skalthoff:
i think i triaged this wrong, the probleem i was running into was, watching a series having selected a different audio track, the next episode plays, goes back to the original audio track, but the track selector still visually shows the other audio track despite it not being played. I will reapproach because it does directly revert the changes from the otehr PR  so likely i missed the root cause

(SO SORRY that I made this pr without having done all of the diligence i needed to, thank you for your time spent reviewing it :)

PR #5334, opened 2026-01-14
Fix: Enable audio passthrough by setting correct AudioAttributes

ne0x86:
This PR addresses an issue where audio passthrough (bitstream) to receivers (5.1/7.1/Atmos) was not working on some Android TV devices, while other players (like VLC or Prime Video) worked correctly.

The Fix: ExoPlayer was missing specific AudioAttributes configuration. Android requires USAGE_MEDIA and CONTENT_TYPE_MOVIE to properly route raw audio bitstreams for passthrough.

I have set these attributes in both:

VideoManager.java
 (Legacy/Current Player)

ExoPlayerBackend.kt
 (Media3 Backend)
Verification: Tested on a physical Android TV device connected to a 5.1 receiver.

Before: Audio was decoded to PCM stereo by the system.
After: Receiver correctly detects and decodes the 5.1 bitstream signal.
Changes:

Added setAudioAttributes with USAGE_MEDIA and CONTENT_TYPE_MOVIE to ExoPlayer builders.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
@@ -81,7 +81,8 @@ public class VideoManager {
     public boolean isContracted = false;
 
     private final UserPreferences userPreferences = KoinJavaComponent.get(UserPreferences.class);
-    private final HttpDataSource.Factory exoPlayerHttpDataSourceFactory = KoinJavaComponent.get(HttpDataSource.Factory.class);
+    private final HttpDataSource.Factory exoPlayerHttpDataSourceFactory = KoinJavaComponent
+            .get(HttpDataSource.Factory.class);
 
     public VideoManager(@NonNull Activity activity, @NonNull View view, @NonNull PlaybackOverlayFragmentHelper helper) {
         mActivity = activity;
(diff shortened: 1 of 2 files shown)


gradle/libs.versions.toml
```diff
@@ -27,7 +27,7 @@ androidx-window = "1.5.1"
 androidx-work = "2.11.0"
 coil = "3.3.0"
 detekt = "1.23.8"
-java-jdk = "21"
```
nielsvanvelzen, line 30: Why did you change the JDK we use? This seems extremely unrelated

playback/media3/exoplayer/src/main/kotlin/ExoPlayerBackend.kt
```diff
@@ -96,6 +96,7 @@ class ExoPlayerBackend(
 			.setMediaSourceFactory(mediaSourceFactory)
 			.setAudioAttributes(AudioAttributes.Builder().apply {
 				setUsage(C.USAGE_MEDIA)
+				setContentType(C.AUDIO_CONTENT_TYPE_MOVIE)
```
nielsvanvelzen: We support both video and audio playback with our new playback backend. You'll have to update the audio attributes based on the currently playing media.

gradle/libs.versions.toml
ne0x86: My apologies! I changed the JDK locally to debug a build issue on my machine and accidentally included it in the commit. I have force-pushed a fix, so the PR should now only contain the relevant audio passthrough changes. Thanks for catching that.

playback/media3/exoplayer/src/main/kotlin/ExoPlayerBackend.kt
ne0x86: Thanks for the feedback! You are absolutely right.

I have updated 

ExoPlayerBackend.kt
 to dynamically inspect the current media stream's tracks in 

playItem
. It now checks if a video track exists (

MediaStreamVideoTrack
) and sets the AudioAttributes to CONTENT_TYPE_MOVIE only in that case; otherwise, it defaults to CONTENT_TYPE_MUSIC. This ensures passthrough is enabled for video content while preserving correct behavior for music playback.

ne0x86:
Thanks for the feedback! You're absolutely right. I've updated both ExoPlayerBackend.kt and VideoManager.java to set the AudioAttributes dynamically.

Now, it inspects the media before playing: it uses CONTENT_TYPE_MOVIE when a video track is present (enabling passthrough for movies/shows) and defaults to CONTENT_TYPE_MUSIC for audio-only streams.

I've also merged the latest changes from the master branch to ensure everything is up-to-date.

nielsvanvelzen:
Unfortunately I cannot really review this anymore as you made a ton of unrelated formatting changes now.

ecdye:
@ne0x86 I think this may fix some of the issues I have been having, is there any chance you will push it through to get it reviewed?

nielsvanvelzen:
I've created a new PR from scratch with #5384. Closing this PR as stale and superseded.

Issue #608, opened 2020-10-26
List view


Jon12345:
I'd like a list view similar to "Media info" view in Kodi. I think this makes the best use of screen real estate.
(Someone's requested this in reddit: https://www.reddit.com/r/jellyfin/comments/fi99ox/does_jellyfin_for_android_tv_have_list_view/ )

The Jellyfin web client also has a pretty good list view.

Voodoopupp:
I think it would be great if the already existing "list view" of the web interface could be also done as well on the Android TV version.

here you can see the mentioned view in the background:
![list_view](https://user-images.githubusercontent.com/9829603/197151059-f1727875-6b06-448f-8fbf-006d111d908e.jpg)

brcarls:
I also would really like this feature... It is so much easier to browse my libraries in list view on the web client than trying to find media using posters on the FireTV client.

lionking102:
Thats the one big thing mission for me in the android-tv app. Especially if you have tv-recordings without any pictures it is a bit painful to browse through 100 items all looking the same :)

vgnx:
Hi,
I'm a new Jellyfin user, and this feature would be of great benefit

UserX404:
Hi,
anyone got a respose from devs. Is this something on the to-do list or is it not planed?
Had a peek at the ATV-app a few weeks ago and was kind of pissed of the poster wall. No idea what poster belongs to what series until it's active selected. No idea how far i have to scroll through to find Series with 'U'.
This probably is an issue because i'm used to Kodi's list view but searching through a big fat list is definitly more convenient in Kodi.
Would love to see something comparable here too.
Thx

nielsvanvelzen:
<img width="295" height="81" alt="Image" src="https://github.com/user-attachments/assets/1a44563d-6a9c-44c8-bb49-6de34aa14218" />

jeroenst:
I also mis listviewnon android tv, grid view with many items is very uncomfortable for me

PR #5369, opened 2026-01-27
Listview #608

bjr86148:
Hi,
based on  #608 I'm trying to implement with help of claude ai, it's a few years since i've wrote some code with android.

I'm pleased to get some feedback or/and I would may have some time to change my PR if it doesn't fit or may requires some more changes.

**Changes**
- ListView for movies
- toggleView button added to toolbar

**Code assistance**
claude ai utilized.

**Issues**
#608 

This is how it looks like at the moment, I've marked the new button which switches from the current to the list view.
image


nielsvanvelzen (changes requested):
- New code should be written in Kotlin
- New UI should be written in compose
- Legacy UI like BrowseGridFragment is in maintenance mode and should not get significant features added

bjr86148:
thank you much for review. I changed and commited the VerticalGridPresenter to Kotlin.

BrowseGridFragment references the Presenters so I don't figured out yet where/what should be changed? Should BrowseGridFragment  translated to Kotlin first?

nielsvanvelzen:
I don't think you really understood, the browse grid fragment is old UI and we're not adding significant changes to it, like a list view.

bjr86148:
as far as i understand it should be migrated to android's new(?) jetpack compose ui architecture like the new settings ui discussion #5317 or #5146 and would be more work to get rid of *fragments.

Is there an approximate estimate when #608 could start?

nielsvanvelzen:

PR #5370, opened 2026-01-29
Remove drop shadows from BrowseGridFragment

TheMejmun:
### Removed drop shadows from BrowseGridFragment, which had become visible due to recent changes in corner rounding radii:

Screenshot_20260128-221505

Screenshot_20260129-010753

**Changes**
Disabled shadows and the focus dimmer of VerticalGridPresenter and HorizontalGridPresenter. These shadows recently became visible when the corner radius for the item cards changed.

Since I assume that this part of the UI will be replaced with Compose in the near future, I tried to keep changes to legacy code minimal, choosing not to add a new constructor to HorizontalGridPresenter. I understand that new features are unwanted in the old java code, but I consider this more of a bugfix, than a new addition.

**Code assistance**
None used.

**Issues**
None.


TheMejmun:
The lack of dimming on cards surrounding the selection might actually reduce readability. Probably not worth merging.

nielsvanvelzen:
I was actually planning to remove those drop shadows as I don't think all UI in 0.20 will be rewritten yet, so we'll have to deal with the old leanback based code for a bit.

TheMejmun:
@nielsvanvelzen I'll reopen this pull request then. You can take a look for yourself. I just think that it can be hard to tell what is selected, when there are more than 2 or 3 rows on screen.

PR #5375, opened 2026-01-30
Restore dimming using Composable Modifier in CardViewHolderContent

TheMejmun:
Restored card dimming as a Composable Modifier.

**Changes**
Added a reusable Compose Modifier for dimming and applied it to any cards that are not focused.

**Code assistance**
None.

**Issues**
This was mentioned in a comment in PR [#5370](https://github.com/jellyfin/jellyfin-androidtv/pull/5370#pullrequestreview-3723356763)


nielsvanvelzen:
In our new design we do not plan to add the dimming back to the cards within the grids. Instead adding more spacing and titles to the cards will make it more clear what card is focused. Together with the scaling this should be sufficient.

PR #5376, opened 2026-01-30
Correct video start delay setting not showing decimal values in settings

JagerSprinkles:
**Changes**
Video start delay setting was not showing decimal values and was truncating the display to a whole number in the settings. For example a value of 1.75 would show as 1. This change sets the text value to a double for displaying in the settings menu to ensure the decimal part of the number is not lost.

**Code assistance**
None, this is a small one line change.

**Issues**
None that I could find.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
@@ -39,6 +39,7 @@ import org.jellyfin.sdk.api.client.ApiClient
 import org.jellyfin.sdk.api.client.extensions.clientLogApi
 import org.jellyfin.sdk.model.ServerVersion
 import org.koin.compose.koinInject
+import java.text.DecimalFormat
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
```diff
@@ -159,7 +159,7 @@ fun SettingsPlaybackAdvancedScreen() {
 						modifier = Modifier.sizeIn(minWidth = 32.dp),
 						contentAlignment = Alignment.CenterEnd
 					) {
-						Text("${videoStartDelay / 1000}s")
+						Text("${videoStartDelay.toDouble() / 1000}s")
```
nielsvanvelzen: Good find, I think we shouldn't show decimals for whole numbers though. Currently it shows as "1.0s" instead of just "1s".

JagerSprinkles: I have made this adjustment. Let me know if there are any other changes needed.

PR #5378, opened 2026-01-31
Remove unused defaultRatingType setting

nielsvanvelzen:
**Changes**

Remove the defaultRatingType setting as it is no longer used by our new item cards (as they don't show ratings at all). The only behavior change here is that now we'll always show ratings during browsing.

**Code assistance**

**Issues**


nielsvanvelzen:
For rating on the item cards: it will likely come back (if the design mockups are to be believed) and will probably configurable
For ratings shown during browsing: I have literally no idea how the new browsing UI will end up looking, so perhaps there will be some customization options again for that.

PR #5379, opened 2026-01-31
Fix play next episode for MX and MPV external players

### LINKED ISSUE #5293: External player next episode and queuing broken

### Describe the bug

- In Jellyfin's Android TV app configure to use an external player (I use MPV-android).
- Play an episode of a show.
- When the episode ends it goes back to the series page instead of playing the next episode. The episode gets marked as watched though so it clearly knows that it ended and played the whole thing. The next episode doesn't get added to the next queue either, possibly related, making it hard to keep track of what I was watching (currently using the favourite feature to keep track of the series I'm watching but were I already using that for its intended purpose then I would have no way to keep track).

This functionality used to work in previous versions of the app. Most likely broken in the external video playback rewrite here: https://github.com/jellyfin/jellyfin-androidtv/pull/4438#issuecomment-2652557871 That pull request indicates that this is a known problem that wasn't actually dealt with before merging (not clear why the reported bug was ignored instead of fixed before merging). This functionality should be fixed so that it functions as it did in previous versions so that the next episode plays and gets added to the next queue for better tracking of what should be watched next.

### Logs

_No response_

### Application version

0.19.5

Google Play Store

### Device information

TCL R646

### Android version

Android (Google TV) 11

### Jellyfin server version

10.11.5

### Other sources

https://github.com/jellyfin/jellyfin-androidtv/pull/4438#issuecomment-2652557871

RealGreenDragon:
**Changes**

Fixes a regression where the "play next episode" feature failed for external players (specifically MX Player and MPV) because they do not return a standard playback position upon completion.

- **MX Player**: Re-implemented specific intent handling from v0.18 (checking the `end_by` flag).
- **MPV**: Added specific handling for MPV's result intent. Since MPV does not return a position extra when playback finishes, the code now correctly interprets this as completion based on the official MPV intent spec.

This restores functionality that was previously covered by the "time elapsed" fallback check, which was removed in v0.19.

**Code assistance**

Used LLM to verify Kotlin syntax and validate the Intent handling logic for specific players.

**Issues**

Fixes #5293

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
@@ -54,6 +54,12 @@ class ExternalPlayerActivity : FragmentActivity() {
 		private const val API_MX_SUBS = "subs"
 		private const val API_MX_SUBS_NAME = "subs.name"
 		private const val API_MX_SUBS_FILENAME = "subs.filename"
+		private const val API_MX_RESULT_ID = "com.mxtech.intent.result.VIEW"
+    	private const val API_MX_RESULT_END_BY = "end_by"
+    	private const val API_MX_RESULT_END_BY_PLAYBACK_COMPLETION = "playback_completion"
+
+		// https://mpv-android.github.io/mpv-android/intent.html
+		private const val API_MPV_RESULT_ID = "is.xyz.mpv.MPVActivity.result"
 
 		// https://wiki.videolan.org/Android_Player_Intents/
 		private const val API_VLC_SUBTITLES = "subtitles_location"
@@ -65,6 +71,7 @@ class ExternalPlayerActivity : FragmentActivity() {
 		private const val API_VIMU_RESUME = "forceresume"
 		private const val API_VIMU_RESULT_ID = "net.gtvbox.videoplayer.result"
 		private const val API_VIMU_RESULT_ERROR = 4
+		private const val API_VIMU_RESULT_PLAYBACK_COMPLETED = 1;
 
 		// The extra keys used by various video players to read the end position
 		private val resultPositionExtras = arrayOf(API_MX_RESULT_POSITION, API_VLC_RESULT_POSITION)
@@ -199,8 +206,16 @@ class ExternalPlayerActivity : FragmentActivity() {
 			else null
 		}
 
+		val playbackCompleted = when (result?.action) {
+			API_MX_RESULT_ID -> extras.getString(API_MX_RESULT_END_BY) == API_MX_RESULT_END_BY_PLAYBACK_COMPLETION
+			API_MPV_RESULT_ID -> endPosition == null // in MPV playback is completed when 'position' extra is absent
+			API_VIMU_RESULT_ID -> result?.resultCode == API_VIMU_RESULT_PLAYBACK_COMPLETED
+			// in VLC is not possible to understand if playback is completed from resulting Intent
+			else -> null
+		}
+
 		val runtime = (mediaSource.runTimeTicks ?: item.runTimeTicks)?.ticks
-		val shouldPlayNext = runtime != null && endPosition != null && endPosition >= (runtime * 0.9)
+		val shouldPlayNext = (playbackCompleted) || (runtime != null && endPosition != null && endPosition >= (runtime * 0.9))
 
 		lifecycleScope.launch {
 			runCatching {
@@ -209,7 +224,7 @@ class ExternalPlayerActivity : FragmentActivity() {
 						PlaybackStopInfo(
 							itemId = item.id,
 							mediaSourceId = mediaSource.id,
-							positionTicks = endPosition?.inWholeTicks,
+							positionTicks = if (playbackCompleted) runtime?.inWholeTicks else endPosition?.inWholeTicks,
 							failed = false,
 						)
 					)
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
@@ -89,7 +89,7 @@ class ExternalPlayerActivity : FragmentActivity() {
 			Toast.makeText(this, R.string.video_error_unknown_error, Toast.LENGTH_LONG).show()
 			finish()
 		} else {
-			onItemFinished(result.data)
+			onItemFinished(result)
 		}
 	}
 
@@ -190,7 +190,7 @@ class ExternalPlayerActivity : FragmentActivity() {
 	}
 
 
-	private fun onItemFinished(result: Intent?) {
+	private fun onItemFinished(activityResult: ActivityResult) {
 		if (currentItem == null) {
 			Toast.makeText(this@ExternalPlayerActivity, R.string.video_error_unknown_error, Toast.LENGTH_LONG).show()
 			finish()
@@ -198,6 +198,7 @@ class ExternalPlayerActivity : FragmentActivity() {
 		}
 
 		val (item, mediaSource) = currentItem!!
+		val result = activityResult.data
 		val extras = result?.extras ?: Bundle.EMPTY
 
 		val endPosition = resultPositionExtras.firstNotNullOfOrNull { key ->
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
```diff
@@ -193,13 +200,31 @@ class ExternalPlayerActivity : FragmentActivity() {
...
+                        Timber.i("MX Player indicated playback completion")
+                        endPosition = runtime
+                    }
+                }
+                API_MPV_RESULT_ID -> {
+                    if (!extras.containsKey(API_MPV_RESULT_POSITION)) {
+                        Timber.i("MPV returned result without position so assuming playback completion")
+                        endPosition = runtime
+                    }
+                }
```
nielsvanvelzen: The "position" extra is already resolved through API_MX_RESULT_POSITION which is in the resultPositionExtras array.

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
```diff
@@ -193,13 +200,31 @@ class ExternalPlayerActivity : FragmentActivity() {
...
 		}
 
 		val runtime = (mediaSource.runTimeTicks ?: item.runTimeTicks)?.ticks
+
+        if (endPosition == null && runtime != null) {
+            when (result?.action) {
+                API_MX_RESULT_ID -> {
+                    if (extras.getString(API_MX_RESULT_END_BY) == API_MX_RESULT_END_BY_PLAYBACK_COMPLETION) {
+                        Timber.i("MX Player indicated playback completion")
+                        endPosition = runtime
```
nielsvanvelzen: We should never assume the end position. We only allow it to be set when the used player explicitly tells us.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
```diff
@@ -193,13 +200,31 @@
...
-		val endPosition = resultPositionExtras.firstNotNullOfOrNull { key ->
+		var endPosition = resultPositionExtras.firstNotNullOfOrNull { key ->
 			@Suppress("DEPRECATION") val value = extras.get(key)
 			if (value is Number) value.toLong().milliseconds
 			else null
 		}
 
 		val runtime = (mediaSource.runTimeTicks ?: item.runTimeTicks)?.ticks
+
+        if (endPosition == null && runtime != null) {
```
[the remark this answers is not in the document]
RealGreenDragon: @nielsvanvelzen should I remove the blank line here?

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
RealGreenDragon: I add API_MPV_RESULT_POSITION for clarity, as use API_MX_RESULT_POSITION in an MPV when case can result strange. That bad? I should replace it?

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
RealGreenDragon: In this case the player tell us explicitly that playback is finished with "end_by" extra (https://mx.j2inter.com/api), nothing is assumed here.
Unfortunately the generic "endPosition" parsing does not work always with MPV (playback end is comunicated "position" extra absence - https://mpv-android.github.io/mpv-android/intent.html) and MX (playback end is comunicated with "end_by" extra set at "playback_completion" - https://mx.j2inter.com/api).
As detailed in PR comment, I simply restored v0.18 correct behavior for both MX and MPV.

nielsvanvelzen: Ok, adding a more abstract approach to determining if "playback successfully completed to the end" is fine. Not as a second fallback for one specific player though.

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
nielsvanvelzen: You're introducing duplicate code in a branch that will never activate.

RealGreenDragon: ok, you are right sorry, I'll refactor it.

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
RealGreenDragon: Yes, you are right, I'll work on it in next days (I think to add another flag for "playback successfully completed to the end", in order to not touch endPosition).

However, I have a doubt about endPosition value in PlaybackStopInfo api call: if set to null in server log I see the message `Stopped at "unknown" ms`. I think it is a little strange if we know the total runtime and player says us the playback completed. Also for PlaybackStopInfo api call is better to not assume endPosition, leaving `Stopped at "unknown" ms` in the server?

nielsvanvelzen: If completed you must specify the position for the server to calculate various things. It can be omitted but this will make the server skip those calculations.
So if the external player doesn't give back this information we'll have to use the item duration.

RealGreenDragon:
@nielsvanvelzen I rewrite PR by adding a playbackCompleted flag that is true only if natively the external player communicate the playback completed at runtime end, this for all external player supported (not only MX and MPV).

Following the cases when we are sure from documentation that playback is completed for each external player:

- MX: if `end_by` extra is present and equal at `playback_completion`
- MPV: if `position` extra is absent (as you suggested I check only if endPosition is null)
- VIMU: if resultCode is 1
- VLC: check is not possible according to documentation (I leaved a comment for clarity)

Finally, as discussed before at the server we send runtime instead of endPosition if playback is completed.

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
RealGreenDragon: I this is correct now, check [new comment](https://github.com/jellyfin/jellyfin-androidtv/pull/5379#issuecomment-3856293404).

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
RealGreenDragon: I this is correct now, check [new comment](https://github.com/jellyfin/jellyfin-androidtv/pull/5379#issuecomment-3856293404).

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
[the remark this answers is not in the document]
RealGreenDragon: I this is correct now, check [new comment](https://github.com/jellyfin/jellyfin-androidtv/pull/5379#issuecomment-3856293404).

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
```diff
@@ -199,8 +206,16 @@ class ExternalPlayerActivity : FragmentActivity() {
...
+			API_MX_RESULT_ID -> extras.getString(API_MX_RESULT_END_BY) == API_MX_RESULT_END_BY_PLAYBACK_COMPLETION
+			API_MPV_RESULT_ID -> endPosition == null // in MPV playback is completed when 'position' extra is absent
+			API_VIMU_RESULT_ID -> result?.resultCode == API_VIMU_RESULT_PLAYBACK_COMPLETED
+			// in VLC is not possible to understand if playback is completed from resulting Intent
+			else -> null
+		}
+
 		val runtime = (mediaSource.runTimeTicks ?: item.runTimeTicks)?.ticks
-		val shouldPlayNext = runtime != null && endPosition != null && endPosition >= (runtime * 0.9)
+		val shouldPlayNext = (playbackCompleted) || (runtime != null && endPosition != null && endPosition >= (runtime * 0.9))
```
nielsvanvelzen:
```suggestion
		val shouldPlayNext = playbackCompleted || (runtime != null && endPosition != null && endPosition >= (runtime * 0.9))
```

nielsvanvelzen (changes requested):

nielsvanvelzen:
Did you even test these changes? Because the code doesn't even compile right now.

app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
RealGreenDragon: Fixed, thanks

RealGreenDragon:
> Did you even test these changes? Because the code doesn't even compile right now.

Sorry no, but now I tested the build and with my last commit it finish without errors.

The build error was caused by resultCode not present in result. I fixed editing onItemFinished sign to pass ActivityResult instead of data Intent.
Also I fixed playbackCompleted when else branch to set false and avoid Boolean conversion warnings.
Finally, I merged also your suggestion.

nielsvanvelzen:
Closing as invalid (changes do not make much sense and are untested)

RealGreenDragon:
@nielsvanvelzen unfortunately a fix is needed as still now using some external player automatic play next of next episode does not work. Sorry for confusing details and explanation (it my first android PR). I will open a more detailed and tested PR asap.

Langes01x:
> Closing as invalid (changes do not make much sense and are untested)

Would really appreciate if you could spare some time to fix it the way you think is appropriate then given that it was your PR that broke the functionality.

Issue #4371, opened 2025-01-03
Android TV App - Hit record on live TV - "Unable to create recording"


ToddCrimson:
### Describe the bug

1. Go to Live TV
2. Watch any show
3. While live watching, hit record
4. Error comes up "Unable to create recording"
5. Nothing in the logs which helps understand.

I tried the exact same steps with same user on (1) Browser or (2) Android Mobile App, it works just fine. No issues at all.
Only issue is on the Android TV App.
What's more the error appears instantaneously. 

### Logs

Really, nothing shows up regarding recording.

### Application version

0.18.3

### Where did you install the app from?

Sideloaded APK

### Device information

Nvidia Shield Pro

### Android version

Android 9

### Jellyfin server version

10.10.3

ToddCrimson:
Seems like it is discussed here in this forum also without resolution
https://forum.jellyfin.org/t-unable-to-create-recording

nielsvanvelzen:
Please share server logs while attempting to start live tv recording.

PR #5396, opened 2026-02-06
Fix recording from playback overlay, passing programId to default gettimer

WizardOfYendor1:
The playback overlay's recordProgram() calls getDefaultTimer() without passing the programId, causing the server to return DateTime.MinValue for StartDate/EndDate. The Kotlin SDK then mangled these into year-0000 dates that the server rejected with HTTP 400 (separate issue).

Pass program.id to getDefaultTimer() so the server returns real dates from the EPG program data. This matches how LiveProgramDetailPopupHelper already does it correctly [https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/ui/LiveProgramDetailPopupHelper.kt#L95-L100
](url)

**Changes**
Added the programid to the getDefaultTImer call. Remove the unnecessary copy in the subsequent isSeries block. 

**Code assistance**
Opus 4.6 was used to aide in researching/debugging and making some of these changes.  Extra scrutiny suggested.

**Issues**
https://github.com/jellyfin/jellyfin/issues/16190
https://github.com/jellyfin/jellyfin-androidtv/issues/4371

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragmentHelper.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragmentHelper.kt
@@ -159,12 +159,12 @@ fun CustomPlaybackOverlayFragment.recordProgram(program: BaseItemDto, isSeries: 
 	lifecycleScope.launch {
 		runCatching {
 			withContext(Dispatchers.IO) {
-				val defaultTimer by api.liveTvApi.getDefaultTimer()
+				val defaultTimer by api.liveTvApi.getDefaultTimer(program.id.toString())
 
 				if (isSeries) {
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (approved):
I'm unable to test this change but it makes sense

nielsvanvelzen:
As a sidenote: the server should NEVER send a min/max value for dates and must always return `null` instead. So this is is also a server issue.

WizardOfYendor1:
> As a sidenote: the server should NEVER send a min/max value for dates and must always return `null` instead. So this is is also a server issue.

@nielsvanvelzen  I agree :-). In fact read this PR comment I have on the server side. There may also need to be a fix put into the SDK.... https://github.com/jellyfin/jellyfin/pull/16191#issuecomment-3865206735

Issue #4893, opened 2025-08-30
Multi-part episodes not working as expected


Fedec96:
### This issue respects the following points:

- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

## Current behavior

- Have a TV show with at least one episode that is split in multiple parts, i.e.:

```
- /shows/Lost (2004)/S01/Lost - S01E24 - Part 1.mkv
- /shows/Lost (2004)/S01/Lost - S01E24 - Part 2.mkv
```

- From the TV client, go to the season 1 of Lost;
- Open episode 24;
- Notice that:
  - Part 2 is not presented/visible in the episode page;
  - When part 1 ends, the client wants to play S02E01 instead of Part 2.

## Expected behavior

- Episode page should present part 2 as well, like it already happens in the web UI;
- When part 1 ends, part 2 should be queued/played.

### Application version

0.18.2

### Where did you install the app from?

Google Play Store

### Device information

Nvidia Shield Pro (2019)

### Android version

Android 11

### Jellyfin server version

10.10.7

nielsvanvelzen:
Please retest in the latest version of the app.

Fedec96:
> Please retest in the latest version of the app.

Hi, apologies for the wrong information - I got confused by the issue template's placeholder.

I am indeed using the latest version of the app, which at the time is 0.18.11.

BotBlake:
Reproduced on current master.

ecdye:
I'm also observing this issue, is there any update on this?

PR #5408, opened 2026-02-11
Show additional parts for episodes

galaterro:
Changes
Added “Additional parts” row for episodes when the item has PartCount > 0, matching the existing movie behavior.

Code assistance
N/A.

Issues
[#4893]


galaterro:
This is a small UI improvement: episodes that are split into multiple parts now show the “Additional parts” row, matching the behavior already present for movies.  
I’ve attached a screenshot of the resulting layout below.

Screenshot_20260211_173422
Screenshot_20260211_173438

Issue #5421, opened 2026-02-14
Playing back recorded live TV (.ts files)  client causes rapid screen flickering — the player starts and stops in a continuous loop.


WizardOfYendor1:
### This issue respects the following points:

- [x] I have discussed this issue in the [Jellyfin forum](https://forum.jellyfin.org/) and/or [troubleshooting chat](https://jellyfin.org/contact) before opening this issue.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

On recorded live tv content, when you attempt to playback the show/content the playback is constantly starting/stopping in fast loop. The effect the user sees is endless flickering of the recording. 

The recorded show plays back fine on and android client. 

**Initial Analysis**

 MPEG-TS containers use PID-based stream identifiers (e.g., video PID 0x91=145, audio PID 0x94=148) rather than sequential indices. The VideoManager track matching methods (getExoPlayerTrack and setExoPlayerTrack) rely on offsetStreamIndex to translate between
  Jellyfin stream indices (sequential: 0, 1) and ExoPlayer track group IDs. This translation assumes group IDs are sequential, which fails for MPEG-TS PIDs:

  - getExoPlayerTrack: Parses the selected audio group ID → 148, calls offsetStreamIndex(148, ...) → fails bounds check (148 > 2 streams) → returns -1
  - setExoPlayerTrack: Converts Jellyfin index 1 → exoTrackID 1, tries to match against group ID 148 → no match → returns false

  This triggers the following loop in PlaybackController:

  1. onPrepared() calls switchAudioStream(1) (the server's default audio index)
  2. getAudioStreamIndex() returns -1 (failed PID lookup), so -1 != 1, the early-return is skipped
  3. setExoPlayerTrack(1, AUDIO, ...) returns false (PID mismatch)
  4. The else branch calls stop() then playInternal() to rebuild the stream
  5. stop() calls clearPlaybackSessionOptions() which resets mCurrentStreamInfo and mDefaultAudioIndex
  6. playInternal() POSTs to /Sessions/Playing, restarts the player → back to step 1

[...]

PR #5422, opened 2026-02-14
FIX - Playback of recorded Live TV (MPEG-TS) causing player to stop/start in a loop.

WizardOfYendor1:
MPEG-TS files (recordings) use PID-based track IDs (e.g. 0x41, 0x44) rather than sequential integers. The internal player's ID-based matching logic assumes   ExoPlayer track IDs map cleanly to Jellyfin stream indices — for TS files this always fails, triggering the restart fallback                                                                                                                                                                              

**Common/Severe case (what created this):  playback never starts (no user interaction required)**
Every time playback begins, switchAudioStream is called automatically to apply the default audio track. On an MPEG-TS recording, with even a single audio track, immediately enters a loop:

  1. Playback starts → switchAudioStream(defaultIndex) fires automatically
  2. getAudioStreamIndex() → ID match fails → returns -1
  3. -1 ≠ defaultIndex → tries setExoPlayerTrack → ID match fails → restarts
  4. Restart triggers onPrepared → back to step 1

The recording never plays. Every MPEG-TS recording was affected  - as least ones with the ATSC PID track scheme. 

**Rare multiple audio tracks case:  user switches audio tracks**
On recordings with multiple audio tracks (e.g. SAP dual-language), successfully getting past the above still left audio switching broken: the in-player switch would always fail and restart playback, and on return the read-back of the current track would again return -1, causing the loop to repeat on every switch attempt. I'm not entirely clear if this would happen in real world usage.

Tested in emulator, FireTV and NVIDIA Shield. Resolved this issue for me.  More regression testing recommended.

**Changes**
Added ordinal-based fallback matching to both getExoPlayerTrack and setExoPlayerTrack in VideoManager.java. When the offset/ID-based matching fails, the fallback matches by ordinal position among tracks of the same type — 
e.g., "the 1st audio stream in Jellyfin's list = the 1st audio track group in ExoPlayer." This works regardless of whether track IDs are sequential (MP4, MKV) or PID-based (MPEG-TS).

The existing ID-based matching is preserved as the primary path; the ordinal fallback only activates when it fails, so non-TS containers are unaffected.

**Code assistance**
Used Opus 4.6 to track down and assist in research of issue. Used agent mode to make initial code changes.  
Need more through review. 

**Issues**
[5421](https://github.com/jellyfin/jellyfin-androidtv/issues/5421)

**Related**
https://github.com/jellyfin/jellyfin/issues/11060

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/compat/StreamInfo.java
+++ b/app/src/main/java/org/jellyfin/androidtv/data/compat/StreamInfo.java
@@ -73,7 +73,12 @@ public class StreamInfo {
     public final org.jellyfin.sdk.model.api.SubtitleDeliveryMethod getSubtitleDeliveryMethod() {
         Integer subtitleStreamIndex = MediaSource.getDefaultSubtitleStreamIndex();
         if (subtitleStreamIndex == null || subtitleStreamIndex == -1) return SubtitleDeliveryMethod.DROP;
-        return MediaSource.getMediaStreams().get(subtitleStreamIndex).getDeliveryMethod();
+        for (MediaStream stream : MediaSource.getMediaStreams()) {
+            if (stream.getType() == MediaStreamType.SUBTITLE && stream.getIndex() == subtitleStreamIndex) {
+                return stream.getDeliveryMethod();
+            }
+        }
+        return SubtitleDeliveryMethod.DROP;
     }
 
     private String PlaySessionId;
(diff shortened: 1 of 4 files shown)


nielsvanvelzen:

dabl2:
A workaround for me is using the wholpin client which doesn't seem to have what I think is this same issue with the jellyfin client on my Nvidia Shield TV

YMMV

Issue #3195, opened 2023-12-01
Next Up Thumbnail Always Episode Even When Prefer Series Checked


linkRace:
### Describe the bug

Go to Settings => Customization => Prefer Series Thumbnails
Next Up List uses Episode Thumbnail regardless
Not sure if this a bug or I'm just missing a setting somewhere.
I do see Show Next Up Without Thumbnail in Playback, but that seems to be for when you are finishing one episode going into another. I'd like it for the homescreen like it is on web/Windows
For what it's worth, I was on the latest stable before trying the beta and it was the same there as well.
**Android TV**
![20231130_155807](https://github.com/jellyfin/jellyfin-androidtv/assets/1369963/3b3d6999-3396-4cac-b11e-3e5615a6b9a1)

**Windows**
![image](https://github.com/jellyfin/jellyfin-androidtv/assets/1369963/8ef8a6d8-ad39-49a9-8ed1-931f2670b93c)

### Application version

0.16.0-beta5

### Where did you install the app from?

Sideloaded APK

### Device information

Nvidia Shield Pro (2019)

### Android version

Android 11

### Jellyfin server version

10.8.5

Aepx:
Just to add, seeing the same behavior on 0.16.0 from the app store and also on the web version.

nielsvanvelzen:
The preference does not affect the "next up" row so it is working as intended.

Aepx:
Maybe on the app, but the option on the web literally says "Use episode images in 'Next Up" and 'Continue Watching' sections.

mattague:
As of Web Version `10.10.7 (f4b8aa0ed)` and Android TV Version `0.18.8` this works on the web but not on Android TV.

PR #5427, opened 2026-02-17
Respect Series Thumbnail Option

linkRace:
**Changes**
UserPreferences.Companion.getSeriesThumbnailsEnabled() was not being passed into the ItemRowAdapter, just a hardcoded false, which meant that the episode thumb was being displayed instead of the series thumb. I changed the preferParentThumb check to look for parentBackdropImages first then parentImages then seriesThumbImage, based on what I saw on the web requests.

**Code assistance:**
Asked Claude to show me where the settings were getting loaded.

**Issues:**
https://github.com/jellyfin/jellyfin-androidtv/issues/3195

**Images:**
**_Current State:_**
image
image

**_New State:_**

image

image


nielsvanvelzen (changes requested):
Using the preference in more places were it makes sense is fine while we still have this setting. We shouldn't use the backdrops as thumbnails though, they're not meant for that purpose.

linkRace:
> Using the preference in more places were it makes sense is fine while we still have this setting. We shouldn't use the backdrops as thumbnails though, they're not meant for that purpose.

I was just using what the web app uses for them when you have the setting enabled there to maintain consistency. If they're not meant for that purpose, why is that what the web is using?

nielsvanvelzen:
> If they're not meant for that purpose, why is that what the web is using?

I'm (fortunately) not responsible for how the web client behaves. But a backdrop is, like the name implies, meant as a backdrop (e.g. background/wallpaper/screensaver etc.) and not a thumbnail. The appropriately named thumb(nail) image type exists for that use-case.

linkRace:
The trouble is that many shows just don't have a series thumbnail, which results in spoiler episode thumbnails being displayed if there isn't one whereas a backdrop solves that issue.
A separate option to say `Use backdrop when no Series Thumb exists` seems like clutter to me when I would think most users like myself checking the `Prefer series thumbnail` option would expect the behavior I've written here to be what happens.
How would you recommend I proceed given that just using the series thumbnail doesn't solve the intent behind this PR? I have changed it currently to use the series thumbnail where it exists before falling back to the backdrop, but again that only solves a few shows.

nielsvanvelzen:
Closing as invalid (contains a variety of unrelated changes)

PR #5430, opened 2026-02-18
replace default UA with device's UA

knighthat:
**Changes**

Current implementation uses default user agent from both JellyfinJDK and OkHttp. This inconsistence makes managing multiple devices a pain in the ass. On some stricter networks, presence of "okhttp" or default user agent will instantly get 403.

image

This PR aims to bring device's user agent as the default, so that devices can be easily ID'd on the network

image

**Code assistance**

No AI model was used to generate code (I don't like them looking at what I'm working on). Though, I did use Gemini to pinpoint the problem. I asked it why I could access Jellyfin on Amazon Silk (FireTV's web browser) but not in the app, coupling with some Nginx configuration on my end, it was able to pinpoint exact problem. I blocked `okhttp` in UA, removing it did solve the issue, but I didn't want to compromise security for a streaming service.

**Issues**

None that I can find

---

If there's another component that also makes network request that isn't "patched", please let me know.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/JellyfinApplication.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/JellyfinApplication.kt
@@ -2,24 +2,32 @@ package org.jellyfin.androidtv
 
 import android.app.Application
 import android.content.Context
+import android.webkit.WebSettings
 import androidx.work.BackoffPolicy
 import androidx.work.ExistingPeriodicWorkPolicy
 import androidx.work.PeriodicWorkRequestBuilder
 import androidx.work.WorkManager
 import androidx.work.await
+import coil3.ImageLoader
+import coil3.PlatformContext
+import coil3.SingletonImageLoader
+import coil3.network.okhttp.OkHttpNetworkFetcherFactory
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import okhttp3.OkHttp
+import okhttp3.OkHttpClient
 import org.acra.ACRA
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
The source of truth for the user agent used in network requests is in our SDK. If we'd want to expand the usage of that user agent to other network requests (media3/coil) we'd have to expose the user agent string from the SDK or, likely the better solution, set it from the ApiClientFactory directly.
Overriding it from the app is NOT something we're going to do, and we're also not going to put some webview user agent into it.

knighthat:
> The source of truth for the user agent used in network requests is in our SDK

That's part of the problem too. Because of that, requests make by the sdk present a generic UA, while requests make by coil3 present a different one. Coil3 uses its default UA because the request never goes through sdk

image

image

> If we'd want to expand the usage of that user agent to other network requests (media3/coil) we'd have to expose the user agent string from the SDK

Then it doesn't solve the problem I'm trying to fix. If 3 TVs connect to the same server, they all present the same UA. 

> we're also not going to put some webview user agent into it.

It could be anything, a custom string set by user is also ok too. I chose webview UA because it was easy to obtain and unique enough to tell if the request came from more than 1 device

nielsvanvelzen:
> I'm trying to fix. If 3 TVs connect to the same server, they all present the same UA.

If all of those devices use the same app version that is intended behavior. The user agent is to meant to identify the app, not the device.

knighthat:
I think the conversation is derailed from what I intended. 

My main goal is to make all requests submit the same UA. So it can be either sdk's UA, or user's custom string.

btw, this is "the same" implementation on Android app, why don't we do the same?

image

nielsvanvelzen:
Closing as invalid (see previous comments)

PR #5432, opened 2026-02-18
Feature -  add program info toggle button to playback overlay

WizardOfYendor1:
Adds an "i" button in the "secondary" actions on the overlay (aka bottom half) on the right, that toggles overlay display of the current in progress program info on the screen. 

**Changes**
Added new program info supporting classes, action handling and top bar sizing change to accommodate program info data. 

**Code assistance**
Used sonnet 4.6 for code review, made a few changes/optimisations based upon that review.


nielsvanvelzen (changes requested):
There's currently no plans to extend the old video player UI. This functionality does make sense to add to our new video player once it's further along though.

PR #5434, opened 2026-02-21
Update: Fix for HDMI Standby / Audio Capability Reset

ne0x86:
Testing revealed a common Android TV OS bug where the system caches incorrect AudioCapabilities (reporting only PCM support) after waking from HDMI standby or when switching from apps that only play stereo.

To ensure stability, I've implemented the following in the latest commit:

Forced AudioCapabilities: When the user has passthrough enabled in settings, the app now bypasses the OS-reported capabilities and provides a manual AudioCapabilities set (AC3, EAC3, DTS, TrueHD) to DefaultAudioSink.
Dynamic Attributes: This is handled within a custom DefaultRenderersFactory override in ExoPlayerBackend.

Compatibility: Standard Stereo content continues to play flawlessly as PCM, while high-definition bitstreams are no longer lost after a system sleep cycle.

### DIFF
--- a/.agent/workflows/debug.md
+++ b/.agent/workflows/debug.md
@@ -1,0 +1,300 @@
+---
+description: debug
+---
+
+---
+name: systematic-debugging
+description: Use when encountering any bug, test failure, or unexpected behavior, before proposing fixes
+---
+
+# Systematic Debugging
+
+## Overview
+
+Random fixes waste time and create new bugs. Quick patches mask underlying issues.
+
+**Core principle:** ALWAYS find root cause before attempting fixes. Symptom fixes are failure.
+
+**Violating the letter of this process is violating the spirit of debugging.**
+
+## The Iron Law
+
+```
+NO FIXES WITHOUT ROOT CAUSE INVESTIGATION FIRST
+```
+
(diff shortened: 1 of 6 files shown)


ne0x86:
I apologize for the noise and the previous duplicate PRs. My local build environment was misconfigured, which led to unrelated log files and debug tools being included by mistake.

I have now performed a complete cleanup of the branch. This PR currently only contains the 3 necessary files for the audio fix.

Summary of the fix:

HDMI Standby Bug: Addressed a known Android TV OS issue where AudioCapabilities are lost after waking from standby or switching from stereo-only apps. We now bypass this by injecting forced capabilities (AC3, DTS, etc.) into DefaultAudioSink when Passthrough is enabled in settings.
Tunneling & Offload: Disabled video tunneling and audio offload, which significantly improves stability for bitstream output on most hardware (Nvidia Shield, Sony Bravia, etc.).
Spatialization: Explicitly set to NEVER to prevent Android 13+ from interfering with the raw bitstream.
Verified on hardware. I appreciate your patience and I'm happy to address any further feedback.

nielsvanvelzen:
Looks like I forgot to submit my comment on the previous PR.

----

PR #5438, opened 2026-02-24
Fix: stable audio passthrough on Android TV (standby fix)

ne0x86:
### Description
This PR addresses intermittent audio passthrough failures and the "PCM fallback" issue on Android TV after the device wakes from HDMI standby.

### Changes
1. **Forced AudioCapabilities**: When passthrough is enabled in settings, we now bypass the Android OS's inconsistent HDMI capability reports (which often reset to PCM-only after standby) by injecting a manual `AudioCapabilities` set (AC3, EAC3, DTS, TrueHD) into ExoPlayer's `DefaultAudioSink`.
2. **Disabled Tunneling**: Explicitly disabled video tunneling as it's a known source of conflicts with bitstream audio on many Android TV devices.
3. **Spatialization Control**: Set `SPATIALIZATION_BEHAVIOR_NEVER` to prevent Android 13+ from attempting to decode the bitstream for spatial audio effects.

These changes ensure a reliable and stable bitstream output directly to the AV receiver.


nielsvanvelzen:
Duplicate of #5434

PR #5460, opened 2026-03-06
Increase to default page size in Live TV guide

jordandalley:
**Changes**
In lieu of a configurable item for the Live TV guide page size, this PR simply doubles the page size from 75 to 150 by modifying the hard-coded 'PAGE_SIZE' variable in both the Playback Overlay (CustomPlaybackOverlayFragment.java) and the Live TV UI (LiveTvGuideFragment.java).

This PR shouldn't have a performance impact for modern Android TV devices, and provides for a better user experience for people with larger channel lists.

**Code assistance**
No code assistance used.

**Issues**
n/a.


nielsvanvelzen (changes requested):
Eventually we'll automatically load the pages, but until then we want to keep the page size low as with 10.11 adding even a single additional item to a page can already have performance impact. So I don't think we can accept this change currently.

jordandalley:
Thanks @nielsvanvelzen - I realise it's a lazy PR. A configurable limit would have been better. I'll await whatever plans you have to fix this in the future :)

PR #5475, opened 2026-03-18
Fix 'continue watching' and 'Next up' sections to re-retrieve and show from 0 items (removed state)

### LINKED ISSUE #2706: Continue Watching doesn't appear after any video playback for the first time
### Describe the bug

If I watch a video that was already marked for "continue watching", it doesn't show app in the app's home screen unless I exit and return.

Steps to reproduce the problem:

1) Make sure there is nothing already added in continue watching, it needs to be empty, because when there is something already there, continue watching works as expected.

2) Play a video not being played previously and fast forward it or select any chapter to mark it for continue watching.

3) Stop playback.

4) go back to Jellyfin's home screen.

5) The video thumbnail should appear immediately but it doesn't.

6) Exit the app to go back to the Android TV or Google TV home screen.

7) Go back to Jellyfin

8) The video thumbnail appears in continue watching.

### Logs

_No response_

### Application version

0.15.7

Google Play

### Device information

Realme Stick, Homatics Box R 4K Plus

### Android version

Both on Android 11

### Jellyfin server version

10.8.10

aivelon:
**Changes**
Hide the continue watching/listening and next up items row instead of removing it. This will keep the object alive and in a hidden items map which can be used to re-retrieve the items and show them back if any items are available to show.

**Scenario tested**

1. Have no item in the continue watching section to begin with, or if you have an item, then mark it watched so that the continue watching section goes away.
2. Watch any episode/movie partially.
3. Go back to home screen.
4. The continue watching section does not show up as it is removed and does not load back.

**Code assistance**

Github copilot was used as a coding assistant for small snippets (not for codebase explanation or generation of large code blocks). Test added by LLM.

**Issues**

Fixes #2706


aivelon:
Can someone review this one please.

nielsvanvelzen (changes requested):
We do not plan to expand the functionality of legacy UI right now. Re-enabling rows on data refresh is something that will be part of the new compose based UI.

nielsvanvelzen:
Closing as stale

PR #5476, opened 2026-03-19
Render items in a row correctly when the focused row is removed

aivelon:
**Changes**

Fixes the issue where a row of items is removed and the focus changes to another row where items are being updated. These updated items did not get rendered until focus is changed again to another row.

**Scenario for testing:**

1. There should be only one episode of a series in 'continue watching'.
2. Mark the video as watched
3. Optionally, mark the same video as unwatched again. The goal is to get the "next up" section to show the upcoming episode.
4. Go back to main screen.
5. The 'Continue watching' section should go away and the next up section should get another item added showing the next episode from the series.

**Before:**

https://github.com/user-attachments/assets/bcc09d29-e25d-4c06-8b9c-c656176c4660

**After**

https://github.com/user-attachments/assets/9b228778-6d45-4567-869d-e9af4a0506c1

**Code assistance**

Minimal usage for snippet genearation.

**Issues**


aivelon:
Can anyone review this?

nielsvanvelzen (commented):
I do not understand what the issue is and what this change is trying to accomplish. When I mark an item in the "continue watching" row as watched from the details page and go back all rows update as expected.

aivelon:
> I do not understand what the issue is and what this change is trying to accomplish. When I mark an item in the "continue watching" row as watched from the details page and go back all rows update as expected.

If you take a look at the videos in the description carefully, I have marked the item in the "continue watching" as watched and then unwatched. This leads that item being shown in the "Next up" section. If you only mark it as watched, then the next item will show in the Next up section. Either way, before, that item would not show in the Next up section right away. You have to move the focus to another row and then it appears. This fix will show the new item right away when you reach the main screen and the focus is on the "Next up" section. PS: you need to have the next up section already with some items in it at the beginning and the "continue watching" section should have only one item in it so that the whole section goes away when that item is marked watched.

app/src/main/java/org/jellyfin/androidtv/ui/presentation/MutableObjectAdapter.kt
```diff
@@ -83,11 +83,15 @@ open class MutableObjectAdapter<T : Any> : ObjectAdapter, Iterable<T> {
...
 	fun removeAt(index: Int, length: Int = 1): Boolean {
-		if (index < 0 || index >= data.size) return false
+		if (index < 0 || index >= data.size || length <= 0 || index + length > data.size) return false
 
 		data.subList(index, index + length).clear()
 		notifyItemRangeRemoved(index, length)
 
+		// Force rebind of shifted items so focused rows redraw immediately after removals.
+		val shiftedCount = data.size - index
+		if (shiftedCount > 0) notifyItemRangeChanged(index, shiftedCount)
```
nielsvanvelzen, line 93: Notifying about removal will already tell the user of the adapter that all items after it will have new positions. There should be no reason to explicitly tell it that.

nielsvanvelzen (changes requested):

nielsvanvelzen:
Closing as stale

PR #5479, opened 2026-03-22
Fix - Season cards always showing series poster instead of season image

WizardOfYendor1:
Season cards unconditionally used the series primary image, ignoring the season's own primary image (if present). Now prefer the season's own image and fall back to the series poster only when no season image exists.

**Changes**
Use the season image if available, otherwise fallback the series image like the default/prior behavior.  Currently the web client behaves this way so this makes this consistent. 
This is a pretty straightforward change with addition of an ternary operator to decide.

**Code assistance**
Open 4.6 was used to identify where season poster was assigned and code reviewed the changes to ensure there were no other systemic impacts. 

**Issues**
I searched, but found none in this repo (which surprised me - but I may not of searched thoroughly enough). I found a similar issue in the web client repo that was resolved in a similar manner. 

Tested in emulator and fire cube/tv clients.


nielsvanvelzen (approved):

nielsvanvelzen:
Thanks!

PR #5490, opened 2026-03-27
Codex/genre filter

Reevoy24:
**Changes**
This adds a genre filter to library grid views in the Android TV app, making genre-based browsing much easier without forcing users into a separate navigation flow. Users can now open a genre menu directly from the browse toolbar, apply a single genre filter to item grids, and clear it again with a localized "All genres" option. The selected genre is persisted with the existing library filter preferences and applied server-side to item queries, so filtered results reload consistently and behave like the other browse filters. This improves usability for larger libraries in particular, where quickly narrowing a grid to one genre is much faster and more practical on TV than browsing long unfiltered lists.

In addition, the new "All genres" label was added to all existing app string resource locales so the filter integrates properly with the current app language.

**Screenshot**
Example of the new genre filter menu in a library grid view:

Screenshot 2026-03-26 230120

**Code assistance**
Code assistance was used for codebase navigation, implementation scaffolding, and localization boilerplate. The final changes were manually reviewed, adjusted, tested on an Android TV emulator, and verified with a local `:app:assembleDebug` build.


nielsvanvelzen:

PR #5511, opened 2026-04-05
Implement D-Pad Playback Skip UX Option

Blackney:
**Changes**
- Add a new optional toggle in Advanced Playback Preferences that when enabled, will make the right/left D-Pad buttons trigger a one off skip on press, or repeatedly skip if held down.
- Add a new "Skip back length" setting to match the existing "Skip forward length" setting, both of which are used as the skip amounts
- Make a new playback overlay to show when the skip occurs, show the direction with an icon and the number of seconds skipped

**Code assistance**
Claude Code used to better understand existing code base, align with existing patterns where possible, then generate code with heavy manual review/testing

**Issues**
Implements https://github.com/jellyfin/jellyfin-androidtv/issues/5510

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -16,6 +16,7 @@
  - [MajMongoose](https://github.com/majmongoose)
  - [Olaren15](https://github.com/Olaren15)
  - [dtrexler](https://github.com/dtrexler)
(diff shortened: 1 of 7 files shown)


nielsvanvelzen:

PR #5513, opened 2026-04-05
Fix missing focus in new player UI

nielsvanvelzen:
crashycrash

**Changes**
- Fix missing focus in new player UI

**Code assistance**

**Issues**

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/base/BaseScreen.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/base/BaseScreen.kt
@@ -15,8 +15,8 @@ import androidx.compose.ui.focus.focusRequester
 fun BaseScreen(
 	content: @Composable () -> Unit,
 ) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/base/BaseScreen.kt
```diff
@@ -0,0 +1,29 @@
...
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.focus.focusRequester
+
+@Composable
+@Stable
+fun BaseScreen(
+	content: @Composable () -> Unit,
+) {
+	// Work around a focus issue with our current fragment based navigation approuch
+	// by always focussing the screen contents whenever the screen is created
```
thornbill:
```suggestion
	// Work around a focus issue with our current fragment based navigation approach
	// by always focusing the screen contents whenever the screen is created
```
🙃

thornbill (approved):

Issue #4364, opened 2025-01-01
Change buffer size to avoid buffering


WaspEma:
### Describe the feature you'd like

Hi,

When i try to play some high quality file like 20GB i have buffering issue on my Android TV, when i switch to lan from wifi this not appen, but speedtest in wifi on tv is gigabit like usual.
A cache setting would be great, someone can implement this?

Or if some pro can give me an hint to change the stock parameter i make a fork for me.

JustSxm:
Hi, it will.
I have 40mbps download, but my ping is 800ms because I'm currently far away

cvanelteren:
@nielsvanvelzen I can see that on my LAN the buffering halts and does not increase over time. This causes an issue for my android tv box; is there a way to merely configure the buffering time to make this work? Note that this box works fine for other streaming services.

PR #5515, opened 2026-04-06
Add configurable playback buffer size setting

### LINKED ISSUE #2855: Set buffer limit and read rate

GeiserX:
**Changes**

Add a user-facing preference under Playback > Advanced > Customization to control ExoPlayer buffer durations. Three presets are available:

- **Auto** — Media3 defaults (50s min/max buffer, 1s playback start, 2s after rebuffer)
- **Large** — 120s max buffer, 2.5s playback start, 5s after rebuffer
- **Extra large** — 240s max buffer, 5s playback start, 10s after rebuffer

Only `setBufferDurationsMs()` is modified per preset. No byte-limit overrides or `setPrioritizeTimeOverSizeThresholds` changes are made, preserving all of Media3's built-in memory safety (track-aware byte target sizing and local/remote distinction).

The setting is placed in the Customization section (not Video) because it applies to both player paths:
- **Legacy `VideoManager`** (default video playback) — reads the preference each time a new player is created, so the setting takes effect on the next video session.
- **Rewrite `ExoPlayerBackend`** (audio playback and experimental new video player) — reads the preference when the singleton is initialized, so a change here requires an app restart.

This directly addresses [#2855](https://github.com/jellyfin/jellyfin-androidtv/issues/2855) and [#4364](https://github.com/jellyfin/jellyfin-androidtv/issues/4364). A [previous attempt](https://github.com/jellyfin/jellyfin-androidtv/pull/3821) was closed because it exposed four granular preferences — this implementation follows the feedback from @nielsvanvelzen and @viown in that PR to use a single preference with pre-defined presets instead.

A companion PR for the Android mobile client is at [jellyfin/jellyfin-android#1946](https://github.com/jellyfin/jellyfin-android/pull/1946).

**Code assistance**

The implementation approach and buffer duration values were informed by research into Media3 1.10.0 `DefaultLoadControl` defaults and prior ExoPlayer issues (google/ExoPlayer#7354 regarding OOM risks with `setPrioritizeTimeOverSizeThresholds`). Code assistance was used for initial scaffolding and iterative review of the buffer parameter choices.

**Issues**

Fixes #2855
Ref #4364


GeiserX:
Same as I said here: https://github.com/jellyfin/jellyfin-android/pull/1946#issuecomment-4190820871
I will be updating soon the screenshots on an android tv

GeiserX:
### Testing on real hardware

Tested the APK on an Android TV device. The buffer size setting works as expected — the setting is accessible, options are displayed correctly, and the buffer behavior changes take effect on the next video session.

**Settings screen** — Playback buffer size selector (Auto / Large / Extra large):

Buffer size settings

**Playback with extra large buffer** — the progress bar shows the larger buffer filling ahead of the playback position:

Playback buffer 1

Playback buffer 2

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackBufferLengthScreen.kt
```diff
@@ -0,0 +1,45 @@
...
+@Composable
+fun SettingsPlaybackBufferLengthScreen() {
+	val router = LocalRouter.current
+	val userPreferences = koinInject<UserPreferences>()
+	var bufferLength by rememberPreference(userPreferences, UserPreferences.bufferLength)
+
+	SettingsColumn {
+		item {
+			ListSection(
+				overlineContent = { Text(stringResource(R.string.pref_customization).uppercase()) },
```
nielsvanvelzen: I don't think this is the right overline

app/src/main/java/org/jellyfin/androidtv/preference/constant/BufferLength.kt
```diff
@@ -0,0 +1,23 @@
+package org.jellyfin.androidtv.preference.constant
+
+import org.jellyfin.androidtv.R
+import org.jellyfin.preference.PreferenceEnum
+
+enum class BufferLength(
+	override val nameRes: Int,
+) : PreferenceEnum {
+	/**
+	 * Use Media3 default buffer durations.
```
nielsvanvelzen: We shouldn't refer to media3 here as the playback backend that is used may not always be media3.

playback/media3/exoplayer/src/main/kotlin/ExoPlayerBackend.kt
```diff
@@ -109,7 +110,19 @@ class ExoPlayerBackend(
...
+		@Suppress("MagicNumber")
+		val loadControl = when (exoPlayerOptions.bufferSize) {
+			ExoPlayerOptions.BufferSize.LARGE -> DefaultLoadControl.Builder()
+				.setBufferDurationsMs(50_000, 120_000, 2_500, 5_000)
+				.build()
+			ExoPlayerOptions.BufferSize.EXTRA_LARGE -> DefaultLoadControl.Builder()
+				.setBufferDurationsMs(80_000, 240_000, 5_000, 10_000)
+				.build()
+			else -> DefaultLoadControl()
+		}
```
nielsvanvelzen: In our code style this would be written more like this

```suggestion
	val loadControl = DefaultLoadControl.Builder().apply {
		when (exoPlayerOptions.bufferSize) {
			ExoPlayerOptions.BufferSize.LARGE -> setBufferDurationsMs(50_000, 120_000, 2_500, 5_000)
			ExoPlayerOptions.BufferSize.EXTRA_LARGE -> setBufferDurationsMs(80_000, 240_000, 5_000, 10_000)
		}
	}.build()
```

playback/media3/exoplayer/src/main/kotlin/ExoPlayerOptions.kt
```diff
@@ -8,4 +8,11 @@ data class ExoPlayerOptions(
 	val enableDebugLogging: Boolean = false,
 	val enableLibass: Boolean = false,
 	val baseDataSourceFactory: DataSource.Factory = DefaultHttpDataSource.Factory(),
-)
+	val bufferSize: BufferSize = BufferSize.AUTO,
+) {
+	enum class BufferSize {
```
nielsvanvelzen: Shouldn't be an inner class, extract to own file instead and rename to `ExoPlayerBufferSize`.

We may also want to just expose the 4 durations here, and convert the user preference to durations on the usage side.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackBufferLengthScreen.kt
GeiserX: Changed to `pref_playback_advanced`.

playback/media3/exoplayer/src/main/kotlin/ExoPlayerOptions.kt
GeiserX: Extracted to its own file as `ExoPlayerBufferSize` with the 4 duration properties (`minBufferMs`, `maxBufferMs`, `bufferForPlaybackMs`, `bufferForPlaybackAfterRebufferMs`). `AUTO` uses `DefaultLoadControl` defaults.

playback/media3/exoplayer/src/main/kotlin/ExoPlayerBackend.kt
GeiserX: Applied your suggestion, thanks! Then went a step further — the hardcoded values are now pulled from `ExoPlayerBufferSize` properties.

app/src/main/java/org/jellyfin/androidtv/preference/constant/BufferLength.kt
GeiserX: Removed the Media3 reference, now just says "Use default buffer durations."

GeiserX:
Thanks for reviewing @nielsvanvelzen , tried to address your comments. LMK if there are more changes you would see

playback/media3/exoplayer/src/main/kotlin/ExoPlayerOptions.kt
nielsvanvelzen: For my second point:

> We may also want to just expose the 4 durations here, and convert the user preference to durations on the usage side.

I meant just adding "minBufferDuration", "maxBufferduration" etc. to the ExoPlayerOptions. No more enum. For the default options you can make it nullable. Also make sure to use Durations instead of millisecond integers.

GeiserX: Refactored — removed the `ExoPlayerBufferSize` enum entirely. `ExoPlayerOptions` now exposes 4 nullable `Duration` properties (`minBufferDuration`, `maxBufferDuration`, `bufferForPlaybackDuration`, `bufferForPlaybackAfterRebufferDuration`). Null = use DefaultLoadControl defaults. The preset→duration mapping lives in `PlaybackModule`. Did I get it right this time?

playback/media3/exoplayer/src/main/kotlin/ExoPlayerBackend.kt
```diff
@@ -109,7 +110,23 @@ class ExoPlayerBackend(
...
+			val play = exoPlayerOptions.bufferForPlaybackDuration
+			val rebuffer = exoPlayerOptions.bufferForPlaybackAfterRebufferDuration
+			if (min != null && max != null && play != null && rebuffer != null) {
+				setBufferDurationsMs(
+					min.inWholeMilliseconds.toInt(),
+					max.inWholeMilliseconds.toInt(),
+					play.inWholeMilliseconds.toInt(),
+					rebuffer.inWholeMilliseconds.toInt(),
+				)
+			}
```
nielsvanvelzen:
```suggestion
				setBufferDurationsMs(
					exoPlayerOptions.minBufferDuration?.inWholeMilliseconds?.toInt() ?: TheDefaultValue,
					...etc
				)
```

This way the options can be used to set just one or two properties and keep the others as default.

app/src/main/java/org/jellyfin/androidtv/di/PlaybackModule.kt
```diff
@@ -75,6 +76,26 @@ fun Scope.createPlaybackManager() = playbackManager(androidContext()) {
...
+		bufferForPlaybackDuration = when (userPreferences[UserPreferences.bufferLength]) {
+			BufferLength.AUTO -> null
+			BufferLength.LARGE -> 2500.milliseconds
+			BufferLength.EXTRA_LARGE -> 5.seconds
+		},
+		bufferForPlaybackAfterRebufferDuration = when (userPreferences[UserPreferences.bufferLength]) {
+			BufferLength.AUTO -> null
+			BufferLength.LARGE -> 5.seconds
+			BufferLength.EXTRA_LARGE -> 10.seconds
+		},
```
nielsvanvelzen: I do think we could add the durations to the BufferLength enum, like you did in the ExoPlayerBufferSize enum. So this wouldn't be as cluttered. And then just forward those values here

nielsvanvelzen (changes requested):
Ok hopefully the final review pass, can you also fix the merge conflicts? 😄

playback/media3/exoplayer/src/main/kotlin/ExoPlayerBackend.kt
GeiserX: Done — ExoPlayerOptions now takes nullable Duration properties (minBufferDuration, maxBufferDuration, etc). Only non-null values get forwarded to setBufferDurationsMs, so callers can override just one or two. Also rebased on master to fix the conflicts.

app/src/main/java/org/jellyfin/androidtv/di/PlaybackModule.kt
GeiserX: Good call — moved the durations into the BufferLength enum itself, so PlaybackModule just forwards `bufferLength.minBufferDuration` etc. Much cleaner now.

playback/media3/exoplayer/src/main/kotlin/ExoPlayerBackend.kt
nielsvanvelzen: I think you forgot to commit this change because it's still the same as before

GeiserX: You're right, missed that. Now uses `||` instead of `&&` so any single non-null value triggers the call, with `DefaultLoadControl` constants as fallback for the rest.

GeiserX: Oops sorry, my bad. should be now

nielsvanvelzen: The entire if statement is not needed, if none of the props are set we still set all the defaults either way.

GeiserX:
@nielsvanvelzen detekt is flagging the 4-part condition as too complex. Would something like this work?

```kotlin
val hasCustomBuffer = listOfNotNull(min, max, play, rebuffer).isNotEmpty()
if (hasCustomBuffer) { ... }
```

Or would you prefer a `@Suppress("ComplexCondition")` instead?

playback/media3/exoplayer/src/main/kotlin/ExoPlayerBackend.kt
GeiserX: True, removed the if entirely — just calling setBufferDurationsMs unconditionally with elvis defaults now.

nielsvanvelzen (approved):

nielsvanvelzen:
Thank you!

PR #5518, opened 2026-04-07
feat: add pin protection

### LINKED ISSUE #3509: PIN protection for users

When having multiple users logged into a TV, one with access to all the libraries and the second one with content restricted, it'd be useful to be able to protect accounts with a PIN so the second user can't access restricted content using the first account just by switching accounts. The PIN doesn't have to be stored server side, it can be a local configuration so it simplifies implementation. If it's forgotten, it'd be a matter of logging back in. Also, entering the PIN number with remote's numbers would increase security since it would be easier to hide while typing.

The main driver behind this request is to be able to restrict access to kids and not allowing them to watch restricted content just by switching to the next account with full access.

A workaround would be to log the full-access user out everytime it's used and log back in when using the app, but it seems quite impractical.

javalsai:
Most of this PR originates from #5352, however, that PR was rejected due to being AI generated. I've revised the code, changed a few things and essentially understood it; no longer being an AI dump and able to explain any detail of this. (I've co-authored the original commit author, which I think is proper attribution)

I don't really develop for Android, so the code is not of high quality, and there's potentially some bad design decisions which I'm willing to improve on. I've done only minor modifications over the original AI code because it seems simple and structured enough to not need any further complexity.

**Changes**

Add PIN protection for users. All PINs are hashed in new repository. 

The proper user settings have been added too (add, change and remove PIN), it has tests, lock SVGs and a visual indicator on the locked accounts in its selection screen.

There's new resource strings that will need translations from other languages, I've taken care of the default english ones and spanish.

**Code assistance**

The only use of AI was from the code already present in #5352 as well as questions to ChatGPT (GPT-5 mini) about Android design concepts and explaining small code snippets. No code was generated from it.

I've asked in https://github.com/jellyfin/jellyfin-androidtv/pull/5352#issuecomment-4034413508 about doing this and I understand this is still at the boundary of what's permitted. But I want to highlight I've put the effort of reviewing it and will change any low quality aspect of the PR as requested.

**Issues**

- closes #3509

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -114,6 +114,7 @@ dependencies {
 	implementation(libs.androidx.fragment)
 	implementation(libs.androidx.fragment.compose)
 	implementation(libs.androidx.leanback.core)
+	implementation(libs.androidx.gridlayout)
 	implementation(libs.androidx.leanback.preference)
 	implementation(libs.androidx.navigation3.ui)
 	implementation(libs.androidx.preference)
--- a/app/src/main/java/org/jellyfin/androidtv/di/PreferenceModule.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/di/PreferenceModule.kt
@@ -6,10 +6,12 @@ import org.jellyfin.androidtv.preference.SystemPreferences
 import org.jellyfin.androidtv.preference.TelemetryPreferences
 import org.jellyfin.androidtv.preference.UserPreferences
 import org.jellyfin.androidtv.preference.UserSettingPreferences
+import org.jellyfin.androidtv.preference.UserPinRepository
 import org.koin.dsl.module
 
 val preferenceModule = module {
 	single { PreferencesRepository(get(), get(), get()) }
+	single { UserPinRepository(get()) }
 
 	single { LiveTvPreferences(get()) }
 	single { UserSettingPreferences(get()) }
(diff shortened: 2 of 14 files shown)


javalsai:
Oh, all this compiles properly, debug builds at least. Tested the app on real hardware, Android TV version 14.

nielsvanvelzen:
There are currently no plans for this functionality from the client perspective. If we were to implement this the pins would need to be configurable server-side. In addition the code appears to be in violation of our LLM policy.

javalsai:
> There are currently no plans for this functionality from the client perspective. If we were to implement this the pins would need to be configurable server-side.

So then is #3509 not planned or do you think that should be server side? IMO that would just be a weaker version of a password, and client-side pins are already what all other apps in the android TV environment do (youtube, netflix, account launcher, etc).

> In addition the code appears to be in violation of our LLM policy.

Only what's already stated in the PR descritpion. If the concern is code quality, I can refactor any code; but I've done few android development and I dont have a keen eye distinguishing good and bad quality code, I'd appreciate some pointers.

cpetry:
@nielsvanvelzen Can you please reconsider this?
I can’t use Jellyfin if my children have access to content that is not for their age. 
A client side pin asked each time the app is opened (webos in my case) seems to be the way to go.

cpetry:
There seems to be a plugin about this from an external repository trying to go a little different route: https://github.com/orgs/jellyfin/discussions/17527#discussion-10545373

A pin code for each library. Would this be a server side solution that is preferred?

PR #3985, opened 2024-09-11
Add codec panel and support for user-defined AVC/HEVC levels

MichaelRUSF:
This pull request introduces a new panel for advanced codec settings and updates the ExoPlayerProfile, PlaybackController, and UserPreferences to support user-defined AVC and HEVC codec levels. This enhancement allows users to bypass automatic codec profile levels and customize settings to their preference.

**Changes**

1. **Added `PlaybackAdvancedCodecPreferencesScreen`:**
   - Introduced a new preferences screen for configuring AVC and HEVC codec levels and audio settings.
     - **AVC Codec Levels:** Users can select from predefined levels, with a default auto option.
     - **HEVC Codec Levels:** Users can select from predefined levels, with a default auto option.
     - **Audio Settings:** Moved the option to enable AC3 bitstream audio from `PlaybackAdvancedPreferencesScreen` to improve organization and clarity of codec-related settings.

2. **Updated `ExoPlayerProfile`:**
   - Enhanced to incorporate user-defined AVC and HEVC codec levels.
     - **New Parameters:**
       - `userAVCLevel: String? = "auto"`
       - `userHEVCLevel: String? = "auto"`
     - **Codec Profiles Adjustment:**
       - **AVC Codec Profiles:** Uses default profiles or user-defined level based on `userAVCLevel`.
       - **HEVC Codec Profiles:** Uses default profiles or user-defined level based on `userHEVCLevel`.
       - Added conditions to handle and apply user-defined AVC and HEVC levels in the `ExoPlayerProfile`.

3. **Modified `PlaybackController`:**
   - Updated to pass user-defined AVC and HEVC codec levels to the `ExoPlayerProfile`.

**Issues**

- Some devices may have codec profiles and levels beyond their advertised capabilities. This update allows users to bypass automatic detection, which can reduce unnecessary transcoding. 
- #3380
- #3401

PR #5520, opened 2026-04-08
Add user-defined AVC/HEVC codec level override

### LINKED ISSUE #3380: HEVC level 156 always transcoding on supported devices
### Describe the bug

I knew something was going on when the media was slow to play or the subtitles skipped ahead.

Reason for transcoding: The video codec's level is not supported

I think this https://github.com/jellyfin/jellyfin-androidtv/pull/3281 has effected HEVC support on devices that once supported it.

Codec: HEVC
Profile: Main 10
Level: 156
Resolution: 1920x1080
Aspect ratio: 16:9

@lawadr @nielsvanvelzen pinging both of you because the pull request has effected playback.

### Logs

Stream mapping:
  Stream #0:0 -> #0:0 (hevc (native) -> h264 (libx264))
  Stream #0:1 -> #0:1 (copy)
Press [q] to stop, [?] for help
[libx264 @ 0x563eef26b080] using SAR=1/1
[libx264 @ 0x563eef26b080] using cpu capabilities: MMX2 SSE2Fast SSSE3 SSE4.2 AVX FMA3 BMI2 AVX2
[libx264 @ 0x563eef26b080] profile Main, level 4.0, 4:2:0, 8-bit

"SupportsTranscoding":true,"SupportsDirectStream":true,"SupportsDirectPlay":true

### Application version

v0.16.5

Sideloaded APK

### Device information

Shield

### Android version

Android 11

### Jellyfin server version

10.8.13

johnpc:
**Changes**

This PR adds user-configurable AVC (H.264) and HEVC (H.265) codec level overrides to address the long-standing issue where devices like Nvidia Shield and Fire TV under-report their actual decoding capabilities via MediaCodec, causing unnecessary transcoding.

The settings are accessible under Playback → Advanced and offer levels from 4.0 to 6.2, with "Auto" as the default which preserves the current device-reported behavior. When a user selects a specific level, the app tells the server it supports content up to that level regardless of what Android reports, allowing direct play of content that would otherwise be transcoded.

This is a resurrection of the approach from PR #3985 which was closed due to author inactivity, updated to work with the current Compose-based settings UI.

  **Code assistance**
  

Initial code was generated by LLM and refined, reviewed, and made to actually work by me.

**Issues**

Fixes #3380


app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -103,6 +103,16 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 */
 		var preferExoPlayerFfmpeg = booleanPreference("exoplayer_prefer_ffmpeg", defaultValue = false)
 
+		/**
+		 * User defined AVC level override. "auto" uses device-reported capabilities.
+		 */
+		var userAVCLevel = stringPreference("user_avc_level", "auto")
```
nielsvanvelzen: Use either an enum or int (-1 is auto) instead of string. Same for the other pref.

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
```diff
@@ -187,6 +187,28 @@ fun SettingsPlaybackAdvancedScreen() {
 			)
 		}
 
+		item {
```
nielsvanvelzen, line 190: I think it would be better to create a new subpage for codec tweaking, as I expect more options to be added.

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
```diff
@@ -260,3 +282,28 @@ fun SettingsPlaybackAdvancedScreen() {
 		}
 	}
 }
+
+fun getAVCLevelOptions(): Map<String, String> = mapOf(
```
nielsvanvelzen: These funcs should be in the strings.kt file

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAVCLevelScreen.kt
```diff
@@ -0,0 +1,47 @@
...
+	SettingsColumn {
+		item {
+			ListSection(
+				overlineContent = { Text(stringResource(R.string.pref_playback_advanced).uppercase()) },
+				headingContent = { Text(stringResource(R.string.user_avc_level)) },
+				captionContent = { Text(stringResource(R.string.codec_level_warning)) },
+			)
+		}
+
+		items(options) { (value, label) ->
```
nielsvanvelzen: We should show the auto option on top followed by a separator before showing the manual options. Similar to how we separate external apps from our own player in the "video player" option

nielsvanvelzen: Also applies to the other screen ofc

nielsvanvelzen (changes requested):
Initial review, overall looks good.

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
johnpc: codec-tweaking-available
codec-tweaking-subpage

Great idea, I created a subpage for this

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAVCLevelScreen.kt
johnpc: h264-override

Makes sense - I have a different area now under "manual override" with auto pinned to the top above that separator.

nielsvanvelzen:
Also please fill in the PR description template

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
johnpc: I removed these functions in favor string resources (similar to pattern for `RefreshRateSwitchingBehavior`)

johnpc:
Added - I think the only part missing was the code assistance which I didn't think mattered very much. But I've added a disclosure now. 

I am reading your llm policy now (https://jellyfin.org/docs/general/contributing/llm-policies) and I believe I'm within the golden rule on this front. I'm a professional, not a slop-cannon.

nielsvanvelzen (approved):
Thanks for the PR! I've made some final tweaks:

- Revert the use of CodecProfileLevel constants, the enums are descriptive enough to the linter issue can be ignored. In addition not all CodecProfileLevel properties are available on older Android versions
- Use `null` instead of `-1` for the auto option
- Split the enums to their own files
- Update translations, move new strings down
- Misc formatting changes

PR #5349, opened 2026-01-23
Add stats for nerds overlay to video player

infomaniac777:
## Add stats for nerds overlay to video player

Fixes #1637 

### Functional Changes (UI/UX)
- Added info button in player controls (next to existing zoom/video size button) to toggle stats overlay
- Stats overlay shows real-time playback statistics (codecs, bitrate, resolution, etc.)

### Code Changes
- Created `PlaybackStatsOverlayView` - custom View for stats display
- Created `StatsInfoAction` - toggle button action
- Created `PlaybackStatsData` - shared data model
- Modified `CustomPlaybackOverlayFragment` - added toggle method and initialization
- Modified `CustomPlaybackTransportControlGlue` - integrated stats action
- Added string resources and `ic_info.xml` drawable

This is a standard feature expectation in modern video players. Very useful for analysis and debugging. The official Jellyfin Android phone app already has the functional equivalent to this.

Tested manually on my Android TV.

PR #5521, opened 2026-04-08
Add playback info overlay to display codec, bitrate, and play method

### LINKED ISSUE #1637: Display playback info

It would be very useful for troubleshooting and for general transparency if we could get a toggle button that would show a small overlay with playback data (showing bitrate, transcode info, whether the file direct plays etc.). Same feature already exists for the android app, as well as in Jellyfin web and Jellyfin media player.

johnpc:
**Changes**

This adds a playback info button to the video player overlay that displays technical details about the currently playing media. When pressed, a popup appears showing the play method (Direct Play, Direct Stream, or Transcoding), container format, video stream details (codec, resolution, bitrate, HDR type), and audio stream details (codec, channels, bitrate, language). This is useful for troubleshooting playback issues and understanding how media is being delivered to the client.

The feature mirrors functionality already present in Jellyfin Web, Jellyfin Media Player, and the Jellyfin Android mobile app. The info button appears in the secondary actions row (bottom) of the playback controls, next to the existing Zoom button.

The implementation includes support for both the current video player and the new Compose-based video player.

### Old Player

Playback info overlay - Old Player

### New Player (in progress rewrite)

Playback info overlay - New Compose Player

**Issues**

Fixes #1637

Related: #5349 (previous attempt, closed by author)


nielsvanvelzen (changes requested):
A debug overlay is something I've been working on for a bit in one of my local branches, although still incomplete. Looking at this PR is somewhat achieves the same but there's some problems with it:

- We don't add any new functionality to the old video player
- New player UI should avoid using Jellyfin specific models (e.g. baseitemdto) as we will re-use it for media without baseitems in the future. So only the information exposed from the playback module can be used. Additional information we want to show should be added to those models, with the baseitem only being used for extra/optional information (e.g. the item id)
- The media info should be below the header so the title but still be visible if enabled and controls are hidden

johnpc:
emulator-screenshot

is this screenshot what you're going for? I've made your recommended changes as well

Pentaphon:
This is long overdue. Not sure why this hasn't been in the player for a while now.

nielsvanvelzen (approved):
Thanks for the PR. I've pushed some small additional changes to cleanup a bit. There are some tweaks to be made in the design of the overlay but I already have those on my own WIP branch so I'll just rebase that on top of this.

PR #5527, opened 2026-04-13
Add audio and subtitle track selection to new video player

johnpc:
**Changes**

Adds audio and subtitle track selection to the new Compose-based video player, bringing it closer to feature parity with the old player.

subtitle-track-selector
audio-track-selector


Pentaphon:
@johnpc please keep an eye on contrast issues when it comes to UI elements. The light checkmark on the light selector background is hard for some people to see. The checkmark should be dark when selected.

johnpc:
Thanks for reviewing! I added a commit to update the style for more contrast, lmk what you think. 

subtitle-track-contrast-fixed
audio-track-contrast-fixed

Issue #5190, opened 2025-11-26
v0.19.4 very slow compared to v0.18.11


DejaVu:
### This issue respects the following points:

- [x] I have discussed this issue in the [Jellyfin forum](https://forum.jellyfin.org/) and/or [troubleshooting chat](https://jellyfin.org/contact) before opening this issue.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

I was reminded to upgrade to v0.19.4 following the recent upgrade for JellyFin Server 10.11.* 

Since, just loading media images freezes or takes an age. 

I've force sideloaded v0.18.11 back to my FireTV Cube and is far better. Loads like it should, practically instant.

Seems something in v0.19.4 is locking up and not allowing anything to load that is not happening in v0.18.11.

Sorry, I don't know how to debug, but this was my experience. 

Will be staying put here til Amazon Store no longer allows this to work and hopefully is resolved by then.

### Media info of the file

```shell

```

### FFmpeg logs

```shell

```

### Application version

0.18.11

### Where did you install the app from?

None

### Device information

FireTV Cube

### Android version

FireTV OS (latest)

### Jellyfin server version

10.11.3

nielsvanvelzen:
Hi where did you discuss this in chat/forum and can you link the source?

nindim:
I am experiencing the same issue. The search is taking about 20 seconds in 0.19.4 whereas in 0.18.11 it's almost instant.

The home screen and continue watching etc also seems to load much more quickly in 0.18.11.

Bendoon:
Exact same issue as above, I downgraded to 0.18.11 and it's much more snappy, search was incredibly slow, nearly unusable. Mainly related to the UI, playback was fine if you got something playing.

Pentaphon:
I do have to admit, even on my modest collection, 19.4 is slower than 18.11 when navigating around the menus on the exact same Onn 4K Plus device.

nielsvanvelzen:
[...]

PR #5543, opened 2026-04-19
Use lighter field set for home screen API requests

johnpc:
**Changes**

This change excludes heavy fields from media items when rendering the Home Screen (eg MediaSources, MediaStreams, Chapters, Trickplay). **THIS IS SAFE!** Full item data is fetched when user selects an item detail page or initiates playback.

This makes the app launch/readiness feel noticeably faster.

I benchmarked on my own library / app startup. Without this change, it's 5+ seconds between launch and all the home rows being ready - the home rows sit empty for a considerable amount of time before the rows populate. With this change, it's 3s, which is noticeably faster. Most of that time is during the launch before the home rows appear at all. 

Combined with https://github.com/jellyfin/jellyfin-androidtv/pull/5544 the performance improvement on app launch is very significant for my library. 

**Code assistance**

I used Claude code to help me add logging and generate the benchmarks to confirm the magnitude of the improvement (~25-50%).

Measurement methodology: Time from MainActivity START to last 'Creating items' log entry (debug timing added by me), averaged over 3 runs per configuration.

**Issues**

None

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/repository/ItemRepository.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/repository/ItemRepository.kt
@@ -22,4 +22,16 @@ object ItemRepository {
 		ItemFields.TAGLINES,
 		ItemFields.TRICKPLAY,
 	)
+
+	// Lighter field set for home screen rows - excludes heavy fields like
+	// MediaSources, MediaStreams, Chapters, Trickplay that aren't needed for display.
+	// Full item data is fetched when user selects an item.
+	val browseFields = setOf(
+		ItemFields.CAN_DELETE,
+		ItemFields.CHILD_COUNT,
+		ItemFields.DATE_CREATED,
+		ItemFields.GENRES,
+		ItemFields.OVERVIEW,
+		ItemFields.PRIMARY_IMAGE_ASPECT_RATIO,
+	)
 }
--- a/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentHelper.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentHelper.kt
@@ -25,7 +25,7 @@ class HomeFragmentHelper(
 	fun loadResume(title: String, includeMediaTypes: Collection<MediaType>): HomeFragmentRow {
 		val query = GetResumeItemsRequest(
 			limit = ITEM_LIMIT_RESUME,
-			fields = ItemRepository.itemFields,
+			fields = ItemRepository.browseFields,
 			imageTypeLimit = 1,
 			enableTotalRecordCount = false,
 			mediaTypes = includeMediaTypes,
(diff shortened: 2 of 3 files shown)


johnpc:
If this is approved for the Home Screen, I'm also interested applying `browseFields` to other screens as well, eg

  - BrowsingUtils.kt - Grid/list views when browsing libraries
  - ByGenreFragment.kt, ByLetterFragment.kt - Browsing by genre/letter
  - GenericFolderFragment.kt - Folder browsing
  - CollectionFragment.kt - Collection views
  - SearchRepository.kt - Search results
  - SuggestedMoviesFragment.kt - Suggestions

since these also load slow for my library. But I will wait to confirm it makes sense for the Home Screen first so that the PR isn't larger than it need be.

nielsvanvelzen (changes requested):
Problem with this change is that, unfortunately, the base items are sent literally everywhere within the app and there are lots of cases where a specific field is used in a place you don't forsee. This caused a wide variety of issues in the past. Because of that we centralized the requested fields in the 0.19 release as a workaround. At the time this barely had any performance impact... until 10.11 released.

We should first fix the passing around of baseitems before we can reduce the requested fields again, which is mostly happening during rewrites.

johnpc:
I've done some profiling

For example, for my server I query "Latest Movies" like this

```
# Light fields, 128ms, 91KB
/Users/$USER_ID/Items/Latest?parentId=f137a2dd-21bb-c1b9-9aa5-c0f6bf02a805&Fields=$LIGHT
# Heavy fields, 420ms, 970KB
/Users/$USER_ID/Items/Latest?parentId=f137a2dd-21bb-c1b9-9aa5-c0f6bf02a805&Fields=$HEAVY
```

I think the 400ms for 970KB is reasonable from the server side. However I am convinced the client is slow at parsing the large jsons. The timing I provided above is for Nvidia shield, which is optimal conditions. But it's even more severe on slower hardware. 

I think this is where https://github.com/jellyfin/jellyfin-androidtv/issues/5190 is coming from, centralizing the requests for these slow-to-load fields 

It could be that it's not safe, as you say, because of how the items are passed around between components. However I have _not_ encountered any problems yet today as I've been using this. Item detail pages and playback refetches as needed.

johnpc:
I'm going to take some time to think this through. I don't know what the answer is yet but I'm not convinced yet that it's just that the server got slow, considering this PR dramatically improves it with (seemingly) no downside.

johnpc:
I added a comment to https://github.com/jellyfin/jellyfin-androidtv/issues/5190#issuecomment-4322920648. 

I cannot reproduce your assertion about

> At the time this barely had any performance impact... until 10.11 released

For my libraries on my Nvidia shield, the current apk is significantly impacted regardless of server version.

johnpc:
> This (light fields) caused a wide variety of issues in the past

I get that this adds complexity to the code, since screens don't know whether they have the full item response or only the light fields and need to refetch. But so far, for the Home Screen I have not encountered any issues. Everywhere refetching is required it's already handled in the existing code. 

I think it's safe to do it. I'm not sure what specific issues we're talking about, but they couldn't have been that bad since v0.18 was reasonably solid.

nielsvanvelzen (approved):
Ok let's try this and we'll see if any issues arise during the 0.20 beta cycle, in which case we can either fix those or if needed revert this change.

nielsvanvelzen:
I feel somewhat confident about backporting this so I've included these changes in 0.19.9 already. It will use a slower rollout (beta users first) so if this does end up causing issue we can halt and revert/fix the change.

PR #5544, opened 2026-04-19
perf: Defer LiveTV check until after main rows are displayed

johnpc:
**Changes**

Previously, the LiveTV API check blocked all home screen rows from appearing. Now the main rows are displayed immediately, and LiveTV rows are added asynchronously after the check completes.

This improves perceived performance - users see content faster even though total load time is similar. The LiveTV rows appear ~1-2 seconds after the main rows when LiveTV is available.

Combined with https://github.com/jellyfin/jellyfin-androidtv/pull/5543, the performance improvement on app launch is very significant for my library. Night and day.

Benchmark: Time to first content visible (Nvidia Shield)
- Baseline: ~3200ms (blocked on LiveTV check)
- With this change: ~500ms (UI shell with loading rows)
- LiveTV rows appear: ~1900ms after UI shell

**Code assistance**

I used Claude code to help me add logging and generate the benchmarks to confirm the magnitude of the improvement (~25-50%).

Measurement: Time from MainActivity START to first row adapter populated, with LiveTV enabled in home screen settings.

**Issues**

None


nielsvanvelzen (changes requested):
This does not work, with this change the live tv rows will always show up at the end, not the user set position.

johnpc:
Good catch. Hmmm... It seems the slow part is 

```
// This is kind of ugly, but it mirrors how web handles the live TV rows on the home screen
// If we can retrieve one live TV recommendation, then we should display the rows
val recommendedPrograms by api.liveTvApi.getRecommendedPrograms(
  enableTotalRecordCount = false,
  imageTypeLimit = 1,
  isAiring = true,
  limit = 1,
)
```

this request is pretty slow, and I don't understand why it's needed at all.

I found the web implementation here: https://github.com/jellyfin/jellyfin-web/blob/master/src/components/homesections/sections/liveTv.ts#L168C12-L179. I don't really understand it, but I think web gets away with it because it's all non-blocking async calls, so it doesn't slow down the whole page.

IMHO, if live tv is configured as a home section, and live tv is enabled for the user, we should just try to add it and let it clean itself up if it's actually empty. This feels like unnecessary latency.

nielsvanvelzen:
> I don't really understand it, but I think web gets away with it because it's all non-blocking async calls, so it doesn't slow down the whole page.

Well that's the thing, I also had no idea why that check existed when I added the home sections support to the app (one of my first contributions btw, I'm not proud of this code 😅). I'm open to changing it but you'll have to find out why this check exists and if removing it is safe before I'll press the merge button.

I _think_ it is to avoid the Live TV section from showing if Live TV isn't configured (so 0 channels available). But that's my best guess looking at this code for 10 seconds.

app/src/main/java/org/jellyfin/androidtv/ui/home/HomeRowsFragment.kt
```diff
@@ -117,6 +103,7 @@ class HomeRowsFragment : RowsSupportFragment(), AudioEventListener, View.OnKeyLi
 			if (!isActive) return@launch
 
 			// Actually add the sections
+			val hasLiveTvAccess = currentUser.policy?.enableLiveTvAccess == true
```
nielsvanvelzen: No real need to assign the variable, can easily inline this in the if-statement.

johnpc:
> I think it is to avoid the Live TV section from showing if Live TV isn't configured (so 0 channels available).

yes that's exactly what it's for. My argument is that it's not needed in reality. Because

1. The user explicitly added the Live TV Home Section, and therefore should expect it to appear
2. The user explicitly configured with `currentUser.policy?.enableLiveTvAccess` so should expect to see live tv
3. The app already has logic to hide the section if its totally empty (see code [here](https://github.com/jellyfin/jellyfin-androidtv/blob/65707c105/app/src/main/java/org/jellyfin/androidtv/ui/itemhandling/ItemRowAdapterHelper.kt#L361))

Right now we have a problem where the app is slow to launch. We're basically blocking the entire app on determining whether this section is empty or not. It's not worth it to keep this cryptic code alive.

nielsvanvelzen:
Unfortunately I do think we need it then because:

1. The default home sections for a user include the Live TV row (see UserSettingPreferences.kt)
2. Creating a new user in Jellyfin will, by default, enable Live TV access
3. The Live TV section adds 2 rows, only one of those is hidden by the retrieveLiveTvRecommendedPrograms function

With this information I'd say the only possible way to avoid the blocking call is to have the retrieveLiveTvRecommendedPrograms function hide both rows. No idea if we can do that in a way that isn't too hacky though.

johnpc:
I see what you mean. 

I updated the PR again with a new approach. It's more like the first approach where I async fetched then appended to the end, but it tracks the position and injects the rows where they're supposed to be. 

It's a little bit of extra state, but it keeps everything non-blocking which in my opinion it is still worth it as I'm laser focused on this launch startup performance issue.

nielsvanvelzen (changes requested):
The current approach does not work properly if multiple live tv sections are used. Although dumb, it's possible for a user to do. I also feel like it's too hacky.
I believe that unless we rewrite a big portion of how the sections are build this isn't really going to work.

johnpc:
That's a great point. 

Took one more stab it, more in line with your recommendation "to have the retrieveLiveTvRecommendedPrograms function hide both rows"

Everything feels a little bit hacky, but at least we can measure it against the hackiness of what's on master.

nielsvanvelzen (approved):
Not the biggest fan of how the siblings are set but this will do for now.

PR #5571, opened 2026-05-03
[codex] Improve ASS subtitle direct play

abhirham:
## Summary

- add a dedicated `playback:libass` module and controller for legacy-player ASS/SSA rendering
- register Jellyfin fallback subtitle fonts for libass and clear the libass track when subtitles are disabled
- advertise ASS/SSA direct subtitle delivery only when the user preference is enabled, while keeping `ENCODE` as fallback
- expose the ASS/libass toggle in advanced playback settings and add routing/profile tests

## Why

Current `master` has basic libass wiring, but ASS/SSA subtitle handling still misses some behavior from the direct-play fix: fallback fonts are not loaded for legacy playback, disabling subtitles can leave the libass renderer with a selected track, and the preference/profile behavior was not covered by tests.

This ports the relevant fix from `69f3b00aff123705100674d46413eafabbeaf351` onto current `master` while preserving newer master changes such as configurable playback buffer length and subtitle burn-in preferences.

## Validation

- `./gradlew :playback:libass:assembleDebug :app:compileDebugKotlin :app:compileDebugJavaWithJavac`
- `./gradlew testDebug`
- `./gradlew detekt` (completes successfully; existing findings are still reported because the project uses `ignoreFailures = true`)
- `./gradlew assembleDebug`


abhirham:
@nielsvanvelzen any reason this was closed?

nielsvanvelzen:
Ah forgot to add my AI slop comment:

PR #5573, opened 2026-05-04
Queue home videos from the same folder

Scorpeg:
**Changes**

When playing a standalone home video item, load sibling videos from the same parent folder and start playback at the selected item. This lets home videos in the same folder play as a queue instead of only playing the selected video.

**Code assistance**

No code assistance, all natural.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/sdk/SdkPlaybackHelper.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/sdk/SdkPlaybackHelper.kt
@@ -235,7 +235,7 @@ class SdkPlaybackHelper(
 				if (parentId == null) {
 					getParts(mainItem)
 				} else {
-					val siblingVideos = getVideosInFolder(parentId)
+					val siblingVideos = getVideosInFolder(parentId, mainItem)
 					if (siblingVideos.any { it.id == mainItem.id } && siblingVideos.size > 1) {
 						siblingVideos
 					} else {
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:

app/src/main/java/org/jellyfin/androidtv/util/sdk/SdkPlaybackHelper.kt
```diff
@@ -248,6 +262,29 @@ class SdkPlaybackHelper(
...
+				sortBy = listOf(ItemSortBy.SORT_NAME),
+				startIndex = videos.size,
+				limit = ITEM_QUERY_LIMIT,
+				fields = ItemRepository.itemFields,
+			)
+
+			totalRecordCount = response.totalRecordCount
+			if (response.items.isEmpty()) break
+			videos += response.items
+		} while (videos.size < totalRecordCount)
```
nielsvanvelzen: Our current playback code does not allow for pagination so I can understand why this loop was added, however this will be problematic for large folders (lets say hundreds to thousands of videos in one folder) as it will cause a lot of fetching here.

We should only request a single page here and (for now) live with the fact that we have no paginated queues yet. I think the best way to do that is to use "nameStartsWithOrGreater" with the sortname of the requested video, as there doesn't seem to be any better filters for this in our API.

app/src/main/java/org/jellyfin/androidtv/util/sdk/SdkPlaybackHelper.kt
```diff
@@ -230,6 +230,20 @@ class SdkPlaybackHelper(
 				}
 			}
 
+			BaseItemKind.VIDEO -> {
+				val parentId = mainItem.parentId
+				if (parentId == null) {
+					getParts(mainItem)
+				} else {
+					val siblingVideos = getVideosInFolder(parentId)
+					if (siblingVideos.any { it.id == mainItem.id } && siblingVideos.size > 1) {
```
nielsvanvelzen: We should assume here that queuing the folder contains the requested item

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/sdk/SdkPlaybackHelper.kt
Scorpeg: Yes, the requested item should normally be included in that single page. But I would still keep this, as I can think of those corner cases:
if the selected item is filtered out for any reason, launching the returned page would start from a different video;
if sortName is missing in the DTO and we fall back to name, which may not match the server-normalized sort name;
duplicate/equivalent sort names can affect ordering around the selected item

If you think those cases should be ignored, then I can remove this.

app/src/main/java/org/jellyfin/androidtv/util/sdk/SdkPlaybackHelper.kt
Scorpeg: Okay, I changed it to be the way you proposed. Usually should not be more than 150 item and in worst case can start with last item for next 150, if I got it right. In any case it would be nice to have a pagination here.

PR #5577, opened 2026-05-05
Enable ASS subtitle direct play via dedicated libass module

abhirham:
**Changes**

Extracts libass integration into a new `playback/libass` module with a `LibassSubtitleController` that owns ExoPlayer wiring, subtitle view attachment, and Jellyfin fallback font registration. `VideoManager` now only activates the libass pipeline for ASS/SSA tracks delivered via embed or external streams; other subtitle formats keep the existing routing. Adds an "Enable libass" toggle in advanced playback settings (default on) and renames the preference key from `libass_enabled` to `ass_direct_play` to reflect the direct-play behavior. Refactors `MediaCodecCapabilitiesTest` behind a `DeviceProfileCapabilities` interface so `createDeviceProfile` no longer requires a `Context`, and adds unit tests covering subtitle routing and ASS/SSA profile method selection.

**Code assistance**

An LLM (Claude) was used to help draft this PR description.


nielsvanvelzen:
Closing as duplicate of #5571. As mentioned there I am not interested in your AI generated code and even if I were, 80% of the changes here are not wanted.

abhirham:
@nielsvanvelzen I understand you have your reasons for rejecting the PR. But ASS is still not supported in Android TV (google tv streamer 4k). Please use this code as reference to push out a proper fix.

Its bad practice to discard code just cause it is ai generated when it actually solves a problem FYI.

nielsvanvelzen:
The support is experimental so you need to enable it first, in addition to being for the **next version** of the app. It's not in production builds.

This code just moves stuff around to places I don't even want it.

slackerlinux85:
@nielsvanvelzen looking at what the code is doing the actual "FIX" in this PR is probably switching the ASS renderer to CANVAS.

on my google chrome-cast the current code seems to work fine when enabled but on my TCL 55P8M TV the OpenGL renderer give no subtitles at all for ASS. been meaning to make a ticket but haven't had time to properly look at it all so ill add some info here for now

was able to fix it locally by adding an option to switch between OpenGL and CANVAS that way i can choose what renderer is used i don't know what way jellyfin-tv wants to go with. have had a few crashes(in CANVAS) with karaoke subtitles in 4K. upping the buffers to 512M helped but still experienced some still. just a guess but i think the buffer fills up and tv keeps playing while it tries to empty it till everything crashes back to the home screen because it cant clear the buffer fast enough. no crash log is sent to the server unfortunately other then that it works fine

not sure why its not working with the OpenGL renderer maybe there's some OpenGL call or something not implemented/working  cant get logcat to give me any output so cant see anything i am not an android dev so i don't know much about how things work here

OpenGL version on TCL P8M for reference
`..[$] <()> adb shell dumpsys SurfaceFlinger | grep "OpenGL"
GLES: ARM, Mali-470 MP, OpenGL ES 2.0 7bdaa90
`

PR #5582, opened 2026-05-07
Fix video player not unlocking screensaver when playback ends

### LINKED ISSUE #5097: Regression 0.19.2 : the screensaver doesn't start

### Describe the bug

Since the 0.19.2
The screensaver doesn't start
I tried to change the time for trigger the screensaver, different options but the behavior stills the same

### Logs

_No response_

### Application version

0.19.2

Google Play Store

### Device information

Nvidia shield pro 2019

### Android version

Android 11

### Jellyfin server version

10.11.2

### Other sources

_No response_

nielsvanvelzen:
**Changes**

The video player uses a dynamic screensaver lock that activates only when the media is playing. However, when the video player is closed automatically because playback of the final queue item ends it would never unlock, causing the screensaver to not show up until the app is restarted.

Add an additional, explicit, unlock when the player is released to avoid this problem.

**Code assistance**

**Issues**

Fixes #5097

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/VideoManager.java
@@ -607,6 +607,7 @@ public class VideoManager {
     }
 
     private void releasePlayer() {
+        _helper.setScreensaverLock(false);
         if (mExoPlayer != null) {
             mExoPlayerView.setPlayer(null);
             mExoPlayer.release();


anultravioletaurora (approved):

thornbill (approved):

ShadowWizard11:
So unless I completely misunderstand something...  When will this be available for Android?  I am unable to get more then 0.19.8 on my google tv... the device that had this issue.

nielsvanvelzen:
This fix is part of the 0.19.9 release available for all users in the stores.

ShadowWizard11:
> This fix is part of the 0.19.9 release available for all users in the stores.

Are you sure its available to Canadian customers?  Because my Google TV has no availability beyond 0.19.8.  Or perhaps because I am looking in the google store?  Do I need to download the apk separately and side load?  Or should it be available in the play store, and if it should, any idea why it isn't for me?  Or a different store completly?

nielsvanvelzen:
The 0.19.9 release is available to everyone on [Google Play](https://play.google.com/store/apps/details?id=org.jellyfin.androidtv). The slow rollout has already ended so everyone is able to update (and about 80% of Google Play users already did). If you need help updating I'd recommend using our troubleshooting chats.

ShadowWizard11:
I managed to update to 0.19.9, however problem is not solved.  Show has been paused for over 5 hours, and screensaver didn't come on.  Can someone please advise?

PR #5588, opened 2026-05-12
Fix external subtitles not showing on external player

### LINKED ISSUE #5095: Regression -> after update from 0.18.8 to 0.19.2, external subtitles not showing on external player on firetv

### Describe the bug

Tested with Firetv 4k 3rd gen and 4th gen (Max).  VERSION 0.19.2 not showing external subtitles when using external player like MXPlayer and VLC. DIdn't have this issu with 0.18.8 . Didn't test other versions.

### Media info of the file

```shell

```

### Client logs

_No response_

### FFmpeg logs

```shell

```

### Application version

0.19.2

Amazon Appstore

### Device information

Fire TV 4k 3rd and 4th (Max) generation

### Android version

Any

### Jellyfin server version

10.11.1

### Other sources

_No response_

EqualMa:
**Changes**

As of [mpv intent spec](https://mpv-android.github.io/mpv-android/intent.html), `subs` and `subs.enable` should be `ParcelableArray of Uri`.

0.18 [respects this](https://github.com/jellyfin/jellyfin-androidtv/blob/675ae7c391841c87ce1df8ed5dc5555c59f5c612/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.java#L399). 0.19 passed [`Array<String>`](https://github.com/jellyfin/jellyfin-androidtv/blob/f76083e809504c546402aec73cbe12d40b3fb5de/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt#L140) instead.
This pr fixes this regression, by introducing another var `subtitleUrlsToUris`, in case of breaking other code.
Please let me know if you think `subtitleUrls` itself should be `Array<String>`.

0.18 passed [`subs.enable`](https://github.com/jellyfin/jellyfin-androidtv/blob/675ae7c391841c87ce1df8ed5dc5555c59f5c612/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.java#L422) but 0.19 removed it. This pr adds it back.

**Code assistance**

No AI is used.

**Issues**

Fixes #5095

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt
@@ -56,6 +56,7 @@ class ExternalPlayerActivity : FragmentActivity() {
 		private const val API_MX_SUBS = "subs"
 		private const val API_MX_SUBS_NAME = "subs.name"
 		private const val API_MX_SUBS_FILENAME = "subs.filename"
+		private const val API_MX_SUBS_ENABLE = "subs.enable"
 		private const val API_MX_RESULT_ID = "com.mxtech.intent.result.VIEW"
 		private const val API_MX_RESULT_END_BY = "end_by"
 		private const val API_MX_RESULT_END_BY_PLAYBACK_COMPLETION = "playback_completion"
@@ -149,6 +150,7 @@ class ExternalPlayerActivity : FragmentActivity() {
 				routeFormat = format,
 			)
 		}.toTypedArray()
+		var subtitleUrlsToUris = subtitleUrls.map { it.toUri() }.toTypedArray();
 		val subtitleNames = externalSubtitles.map { it.displayTitle ?: it.title.orEmpty() }.toTypedArray()
 		val subtitleLanguages = externalSubtitles.map { it.language.orEmpty() }.toTypedArray()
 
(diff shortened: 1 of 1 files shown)


nielsvanvelzen (changes requested):
The problem with this change is that some players expect an extra called "subs" that contains strings and others expect uris. We cannot provide both so we have to choose which to send, and during the last rewrite of this code the strings were chosen.

EqualMa:
We might have to maintain a dataset of which players require which specification so we can send the intent based on the player app id?

Furthermore, we could allow advanced users to extend/override the defaults by uploading a config file.

I don't know if this is too much.

nielsvanvelzen:
Part of the reason behind #5280 was so we can optimize the communication for the selected player. My idea was to create an interface that can be implemented for various players (VLC/MPV/MX/etc.) that contains a few things;

- Applicable app ids
- Function to create bundle for start
- Function to read activity result

The settings could then also show some indicator as to what functionality we support for the apps. I have not had the time to work on that yet. It would definitely help solve this specific issue  though.

nielsvanvelzen:
I've now implemented said interface in #5625 and it should fix this exact issue (the mpv implementation uses uris). Feel free to test it!

Closing this PR in the meantime as it is effectively superseded now.

PR #5604, opened 2026-05-24
Refactor: Extract MediaCodecQuery interface and AVC codec module

johnpc:
**Changes**

Extract codec query logic from `MediaCodecCapabilitiesTest` into modular, testable components.

**Motivation**

I asked on Matrix how I could help prepare the codebase for the new player (I want to enjoy new player features such as https://github.com/jellyfin/jellyfin-androidtv/pull/5527 and https://github.com/jellyfin/jellyfin-androidtv/issues/4177), and @nielsvanvelzen suggested:

> One thing that you might want to work on is the code around our device profile building. We're going to re-use the existing code for that in the new player but the code for it isn't the prettiest with that mediatest class etc. I was thinking of making it more modular (e.g. separate files for h264/h265/av1 etc.), add tests to it (with mocked system responses) and make it more configurable (e.g. allow settings to disable entire codecs or tweak them, although how much we'd expose to UI is to be determined).

This PR establishes the pattern with AVC/H.264. HEVC and AV1 modules can follow in subsequent PRs using the same interface after this is approved.

I am hopeful that this effort will help accelerate release for the new player ui and unblock a lot of feature development over time.

**Code assistance**

Claude Code was used for code review, identifying dead code (`display` property), and cleanup (removing unused methods, tightening visibility modifiers).

**Issues**

N/A

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
@@ -1,22 +1,20 @@
 package org.jellyfin.androidtv.util.profile
 
-import android.content.Context
-import android.media.MediaCodecInfo
 import android.media.MediaCodecInfo.CodecProfileLevel
 import android.media.MediaCodecList
 import android.media.MediaFormat
 import android.os.Build
 import android.util.Size
-import androidx.core.content.ContextCompat
 import androidx.media3.common.MimeTypes
-import timber.log.Timber
+import org.jellyfin.androidtv.util.profile.codec.AvcCodecCapabilities
+import org.jellyfin.androidtv.util.profile.codec.MediaCodecQuery
 
 class MediaCodecCapabilitiesTest(
-	private val context: Context,
-	private val softwareCodecsEnabled: Boolean
+	private val softwareCodecsEnabled: Boolean,
 ) {
-	private val display by lazy { ContextCompat.getDisplayOrDefault(context) }
 	private val mediaCodecList by lazy { MediaCodecList(MediaCodecList.REGULAR_CODECS) }
+	private val codecQuery by lazy { MediaCodecQuery(mediaCodecList, softwareCodecsEnabled) }
+	private val avc by lazy { AvcCodecCapabilities(codecQuery) }
 
 	// Map common Dolby Vision Profiles to their corresponding CodecProfileLevel constant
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt
```diff
@@ -6,15 +6,16 @@ import android.media.MediaCodecList
...
 import android.os.Build
 import android.util.Size
-import androidx.core.content.ContextCompat
 import androidx.media3.common.MimeTypes
-import timber.log.Timber
+import org.jellyfin.androidtv.util.profile.codec.AvcCodecCapabilities
+import org.jellyfin.androidtv.util.profile.codec.DefaultMediaCodecQuery
 
 class MediaCodecCapabilitiesTest(
 	private val context: Context,
```
nielsvanvelzen: Looks like context is no longer used

app/src/main/java/org/jellyfin/androidtv/util/profile/codec/MediaCodecQuery.kt
```diff
@@ -0,0 +1,24 @@
+package org.jellyfin.androidtv.util.profile.codec
+
+import android.util.Size
+
+/**
+ * Abstraction over Android's [android.media.MediaCodecList] for querying device codec capabilities.
+ * Enables unit testing with mocked system responses.
+ */
+interface MediaCodecQuery {
```
nielsvanvelzen: Do we need this interface? We only have a single implementation

app/src/main/java/org/jellyfin/androidtv/util/profile/codec/DefaultMediaCodecQuery.kt
```diff
@@ -0,0 +1,121 @@
...
+			}
+		}
+
+		Timber.d("Computed max resolution for %s: %dx%d", mime, maxWidth, maxHeight)
+
+		return Size(maxWidth, maxHeight)
+	}
+
+	override fun supportsMultiInstance(mime: String): Boolean {
+		if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) return false
```
nielsvanvelzen: This line can be removed now as we recently bumped the SDK to be >=23 (M)

nielsvanvelzen (changes requested):
I think we have a great start here

nielsvanvelzen (approved):

PR #5611, opened 2026-05-29
Fix next episode row showing wrong episode when episodes are missing

### LINKED ISSUE #1297: Next up episode is incorrect following a multi-part episode
### Describe the bug

1. Open episode following a multi-part episode 
a. eg: The Good Place S01E03 (Follows S01E01-E02)
2. Look at "Next Episodes" row
3. See S01E05
image
image

### Logs

```shell
n/a
```

### Application version

0.12.3 (and master)

Google Play

### Device information

Android TV (1080p) API 31 (emulator)

### Android version

Android 12

### Jellyfin server version

10.8.0-alpha3

rgalic99:
## The Problem

The "Next Episodes" row on the episode detail screen uses `startIndex` (a pagination offset) set to the episode's `indexNumber`. This only works when all episodes from 1 onward exist sequentially. When episodes are deleted or multi-part episodes shift the numbering, the offset lands on the wrong position in the list, showing an incorrect next episode.

For example: if episodes 1-12 are deleted, viewing episode 27 would show episode 39 as "next" because offset 27 in a list starting at episode 13 skips ahead.

## The Fix

Replace the generic `GetItemsRequest` with `startIndex` approach with `tvShowsApi.getEpisodes` using `startItemId`. This correctly asks the server for episodes starting from the current episode's ID, regardless of gaps in numbering.

## Changes

- `FullDetailsFragment.java`: Replace inline `ItemRowAdapter` creation with a call to `loadNextEpisodes()`
- `FullDetailsFragmentHelper.kt`: Add `loadNextEpisodes()` that uses `tvShowsApi.getEpisodes(startItemId = currentEpisodeId)` and creates a static `ItemRowAdapter` from the results

Fixes #1297

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
@@ -618,10 +618,8 @@ public class FullDetailsFragment extends Fragment implements RecordingIndicatorV
                     addItemRow(adapter, additionalPartsAdapter, 0, getString(R.string.lbl_additional_parts));
                 }
 
-                if (mBaseItem.getSeasonId() != null && mBaseItem.getIndexNumber() != null) {
-                    // query index is zero-based but episode no is not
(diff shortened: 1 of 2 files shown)


nielsvanvelzen:

rgalic99:
I understand, I saw this bug, used AI to fix it and built the APK and tested it myself. Should I open an issue and wait for an official fix or would it be a duplicate of #1297 ?
I am fine using my debug apk in the meantime

PR #5616, opened 2026-06-02
Suppress overlay between queued home videos

Scorpeg:
**Changes**
This avoids showing Next Up / overlay UI between consecutive `VIDEO` items so queued home videos can continue without interruption. Useful for this PR: https://github.com/jellyfin/jellyfin-androidtv/pull/5573

**Code assistance**
No assistance


Scorpeg:
As currently there is no way to disable this suppression and/or detect what type was video collection it plays, I think this have to be recoded before this could be useful.

Issue #4666, opened 2025-05-20
Enable AC-4 audio codec


avionics75:
### This request respects the following points:

- [x] This request is **not** already on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/labels/enhancement) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Problem description

Hi,

Would it be possible to enable AC-4 audio codec on the ATV app ?

LiveTV using this audio codec gets remuxed when using the embedded player, but when using an external player with Exoplayer it is direct streaming.

That would avoid unnecessary remux/transcode :-)

### Proposed solution

Enable AC-4 audio codec in the app.

### Alternatives considered

Use an external player for now to avoid remuxing/transcode.

Hello-World-Traveler:
Does the device support AC-4? Jellyfin Server supports AC-4  since FFmpeg 7.0

avionics75:
All my devices support AC-4, even my Android TV 8 apparently. I don't know why it get transcoded in Jellyfin ATV using the embedded player. When using an external player (for example Just Player (based on media3) or Nova Player) it is working fine on direct stream :/

Hello-World-Traveler:
https://github.com/jellyfin/jellyfin-androidtv/pull/3188 (Fix seeking issues in AC4 streams caused by not identifying decode-only) AC4 support is there but might not be in playback yet.

@nielsvanvelzen  Do we have support for AC4 in ExoPlayer?  (app/src/main/java/org/jellyfin/androidtv/util/profile/ExoPlayerProfile.kt)

avionics75:
Noted ! :-)

kinke:
I've had a quick stab at it: https://github.com/jellyfin/jellyfin-web/pull/7329

PR #5638, opened 2026-06-09
Add AC4 audio codec for supporting device profiles

JkBoyo:
Checking AC4 support and including in Device Profile

**Changes**
Current changes are setting it up so that we have AC4 as a codec that can be checked during device profile creation.
Created the function next to the other video codec checks in [MediaCodecCapabilitiesTest.kt](app/src/main/java/org/jellyfin/androidtv/util/profile/MediaCodecCapabilitiesTest.kt). Modified the codec object to include AC4 in [Codec.kt](app/src/main/java/org/jellyfin/androidtv/constant/Codec.kt).  Added the codec in [deviceProfile.kt](app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt).

I'm currently planning to do a check and filter for if the device supports the codec using the function defined in MediaCodecCapabilitiesTest.kt to determine if the codec needs filtered out kind of like you do [here](https://github.com/JkBoyo/jellyfin-androidtv/blob/e62696e33ced734e82e8ee7d94fae8ea260d316b/app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt#L105) using the settings section in the createDeviceProfileFunction.

I'm assuming it shouldn't be inside of the same block above but wasn't sure how/where you would want to implement it.

**Code assistance**
N/A

**Issues**
#4666


app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
```diff
@@ -27,6 +27,7 @@ private val supportedAudioCodecs = arrayOf(
 	Codec.Audio.AAC,
 	Codec.Audio.AAC_LATM,
 	Codec.Audio.AC3,
+	Codec.Audio.AC4,
```
nielsvanvelzen: You should only claim AC4 support here is the device actually supports it.

nielsvanvelzen (changes requested):
Normally reviews only happen after a PR goes out of draft but @BotBlake asked me "kindly" to review this while in draft.

Have you tested the "supportsAC4" function on a devices that should or shouldn't support AC4 and did they return the expected results?

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
JkBoyo: That makes sense I just wasn't sure how to add it to that array bc it seemed like it was static and would need to be filtered down. Based on the only other audio codecs I had seen anything done on what I had seen done later during the device profile for audio I wasn't sure how to handle its insertion. 

That's why I made the draft PR because I wanted to see if you wanted me to mimic the video check and set behavior for audio or do something different.

JkBoyo:
> Normally reviews only happen after a PR goes out of draft but @BotBlake asked me "kindly" to review this while in draft.
> 
> Have you tested the "supportsAC4" function on a devices that should or shouldn't support AC4 and did they return the expected results?

Most of my coding experience up to this point has been debuggerless so I'm used to trying to implement the whole thing and then test the feature and playing whack-a-mole from there 🗿🫠.

It makes sense that I should be able to do that and I will before I ask for help again next time 🫡.

JkBoyo:
Needs testing with devices that support AC4 as the CCwGTV doesn't seem to support it based on the debugger.

PR #5639, opened 2026-06-09
Fix presenter focus on old Android versions

nielsvanvelzen:
**Changes**

Older Android versions have a screwed up focus system that doesn't work nicely, these kinds of bugs only show up when testing with old (unstable) emulators.

Set focusable to presenter compose views so they correctly apply their effects.

**Code assistance**

**Issues**


BotBlake (approved):
Works for me 🗿

BotBlake:
@jkboyo we're the best testers for finding issues like this :tm:

thornbill (approved):

PR #5419, opened 2026-02-14
Feature - Added optional selecting preferred audio transcode codec and enable/disable of any audio codec.

Ranft65:
These commits provide the following user options under advanced playback preferences:

1) Optionally set preferred audio transcoding to AAC, AC3, and MP3.  

2) Optionally enable/disable any audio codec. Transcoding will occur for any disabled audio codec (if supported by the Jelly server).

Many clients have poor AAC support and only support stereo AAC.  This PR allows surround sound AAC media files to provide surround sound output on devices that do not support AAC surround sound by forcing a transcode of AAC.  This PR can do the same for the other audio codecs as well by forcing transcodes of them (if supported by the Jellyfin server).

The setting of the preferred audio codec for audio transcoding was limited to AAC, AC3, and MP3 since there were issues with the Jellyfin server audio transcoding of the other audio codecs.  Unexpected or failure of audio output may occur if disabling of the audio codecs is done such to force the Jellyfin server audio transcode to anything other than AAC, AC3, or MP3, 

This PR was inspired by the open PR https://github.com/jellyfin/jellyfin-androidtv/pull/5025 by https://github.com/andreasgerner.

**Issues**
Fixes https://github.com/jellyfin/jellyfin-androidtv/issues/2867, https://github.com/jellyfin/jellyfin-androidtv/issues/2602, https://github.com/jellyfin/jellyfin-androidtv/issues/2991, https://github.com/jellyfin/jellyfin-androidtv/issues/2432 and possibly a lot of similar issues.
Rework for https://github.com/jellyfin/jellyfin-androidtv/pull/3109.
Rework for some parts of https://github.com/jellyfin/jellyfin-androidtv/pull/3110.

PR #5642, opened 2026-06-10
Prioritize audio codecs

### LINKED ISSUE #2867: Add a fallback to AC3/EAC3 option

Ranft65:
First off, this is rework of the rejected pull request #5419.  This pull request implements a cleaner approach to the desired feature of having more control over the enabling, disabling, and prioritizing audio codec usage.

This pull request is based off the main branch.  It adds the following features:

It adds a "Bitstream AAC audio" checkbox (similar to the existing "Bitstream Dolby Digital audio" checkbox) on the "Advanced playback preferences" screen.   This allows quick enabling/disabling the use of the AAC codec which has been problematic for some when it is multi-channel.

It adds a "Select and prioritize audio codecs" button to the "Advanced playback preferences" screen.  This button opens up a selection screen that allows adding, deleting, and setting the priority order of the supported audio codecs.

Code assistance was used for an explanation of how to create a window bigger than the current settings window since my prioritization approach needed a wider window.   The explanation provided the concept of a popup that could be forced to redraw.
 
This PR was inspired by the closed PR https://github.com/jellyfin/jellyfin-androidtv/pull/5025 by https://github.com/andreasgerner.

This PR addressed the provided comment on #5025:
"We can add a new section in the settings where it is possible to configure the audio codecs (which are enabled and in what order). But it should not be a single checkbox."

Fixes https://github.com/jellyfin/jellyfin-androidtv/issues/2867, https://github.com/jellyfin/jellyfin-androidtv/issues/2602, https://github.com/jellyfin/jellyfin-androidtv/issues/2991, https://github.com/jellyfin/jellyfin-androidtv/issues/2432 and possibly a lot of similar issues.
Rework for https://github.com/jellyfin/jellyfin-androidtv/pull/3109.
Rework for some parts of https://github.com/jellyfin/jellyfin-androidtv/pull/3110.


nielsvanvelzen:

Issue #5536, opened 2026-04-17
Option to disable DTS playback


pikausp:
### This request respects the following points:

- [x] This request is **not** already on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/labels/enhancement) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Problem description

When I play DTS-HD (MA) files I end up with stereo playback. I am assuming this is caused by my soundbar supporting plain DTS, but none of the higher quality format. The media report from the device regarding supported codes lists dts, but nothing else from the DTS family.

### Proposed solution

Can we get this option also for DTS?

https://github.com/jellyfin/jellyfin-androidtv/blob/eb2459a3eb7bcad93f9c4883c09161502ac7a68d/app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt#L98

### Alternatives considered

Using Kodi with the Jellyfin plugin which plays back the file just fine. The user experience is just worse and a lot of my media have DTS-HD audio.

BlackstormFirst:
Hi,
This is a known bug in Jellyfin Android TV that's been around for several years, particularly with Samsung TVs that no longer support DTS. Jellyfin TV explicitly tells the server that it can handle this format, even though the device can't.

The simplest solution is to add an option in the menu to disable unsupported codecs, as I've already done (and suggested in another post on the same topic, but I don't have the link anymore).

I can provide my debug version which includes this fix, but I'd prefer to have it included in an official release.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/da5b8924-baf5-4fe6-9d91-80d6d5a884f6" />

DrWhovahkiin:
It would be cool if we could also just get an option on the server side to set the default settings for this. For example, setting the default for clients to show they are compatible with x audio codec or not. setting this all to off, would make it transcode everything, fixing the issue with no audio at least. 
and then the user, if they have the better audio options available, can turn it on, overriding the default.

hattmt:
I like BlackstormFirst proposition , I'm on same case but I would like to downmux to 5.1, old onkyo Hifi amplifier 

Receiver: Onkyo TX-DS989 Ver2 (7.1, THX Ultra, Dolby EX, DTS-ES,
Pro Logic IIx — no HDMI input)

[...]

PR #3929, opened 2024-08-24
Fix DTS playback - ask for transcoding when unsupported

rickysixx:
**Changes**

Check if DTS audio codec is supported by the device before adding it to the supported codec list.

Probably a similar check should be done also for other "advanced" audio codecs (e.g. Dolby TrueHD).

**Issues**

Fixes #3896, #3868

PR #5645, opened 2026-06-11
Implement auto-detection for audio passthrough and allow overrides

### LINKED ISSUE #3917: Dolby TrueHD 2.0 playback never starts
### Describe the bug

1.  Set audio output to "direct".  No other audio setting appears to matter.
2.  Play any video with Dolby TrueHD 2.0

My device does not support Dolby TrueHD playback nor passthrough.  So this should have triggered a transcode, but doesn't.  Playback doesn't start, but the UI is still available and responsive.  If I select a low max bit rate in the playback UI it will trigger a full transcode to AAC and play.  If I change the audio output to "downmix to stereo" it will trigger an audio transcode to AAC and play.

If I playback 5.1 or 7.1 TrueHD content, it will immediately transcode the audio to AAC and play.

It appears that just 2.0 TrueHD doesn't trigger a transcode when audio output is set to "direct".

### Logs

```
[2024-08-21 03:53:58.185 +00:00] [INF] [208] Jellyfin.Api.Helpers.MediaInfoHelper: User policy for "Chris". EnablePlaybackRemuxing: True EnableVideoPlaybackTranscoding: True EnableAudioPlaybackTranscoding: True
[2024-08-21 03:53:58.185 +00:00] [INF] [208] Jellyfin.Api.Helpers.MediaInfoHelper: RemoteClientBitrateLimit: 120000000, RemoteIP: "<CLIENT PUBLIC IPv6>", IsInLocalNetwork: True
[2024-08-21 03:53:58.310 +00:00] [INF] [213] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: Adding playback tracker : 722fedab350d3d62de384934adfceb34e256afd3-3a783413b0e344a0ad5622204ef4c275-3f0be8283bfdca57a7e98bc1fcf9ee56
[2024-08-21 03:53:58.310 +00:00] [INF] [213] Jellyfin.Plugin.PlaybackReporting.Data.PlaybackTracker: PlaybackTracker : Adding Start Event : 08/21/2024 03:53:58
[2024-08-21 03:53:58.310 +00:00] [INF] [213] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: Creating StartPlaybackTimer Task
[2024-08-21 03:53:58.310 +00:00] [INF] [293] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : Entered
[2024-08-21 03:53:59.084 +00:00] [INF] [213] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: Processing playback tracker : "722fedab350d3d62de384934adfceb34e256afd3-3a783413b0e344a0ad5622204ef4c275-3f0be8283bfdca57a7e98bc1fcf9ee56"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: session.RemoteEndPoint : "<CLIENT PUBLIC IPv6>"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : event_playing_id     = "3f0be8283bfdca57a7e98bc1fcf9ee56"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : event_user_id        = "3a783413b0e344a0ad5622204ef4c275"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : event_user_id_int    = 1
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : session_playing_id   = "3f0be8283bfdca57a7e98bc1fcf9ee56"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : session_user_id      = "3a783413b0e344a0ad5622204ef4c275"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : play_method          = "DirectPlay"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : e.ClientName         = "Android TV"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : e.DeviceName         = "Basement"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : ItemName             = "FLCL - s01e01 - Fooly Cooly"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : ItemId               = "3f0be8283bfdca57a7e98bc1fcf9ee56"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : ItemType             = "Episode"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : All matches, playback registered
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: Playback tracker found, adding playback info : "722fedab350d3d62de384934adfceb34e256afd3-3a783413b0e344a0ad5622204ef4c275-3f0be8283bfdca57a7e98bc1fcf9ee56"
[2024-08-21 03:54:18.310 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: Saving playback tracking activity in DB
[2024-08-21 03:54:18.321 +00:00] [INF] [214] Jellyfin.Plugin.PlaybackReporting.EventMonitorEntryPoint: StartPlaybackTimer : Exited
[2024-08-21 03:54:27.582 +00:00] [INF] [293] Emby.Server.Implementations.Session.SessionManager: Playback stopped reported by app "Android TV" "0.17.3" playing "Fooly Cooly". Stopped at "0" ms
```

### FFmpeg logs

```shell
n/a
```

### Media info of the file

```shell
s01e01
Container: mkv
Path: /media/storage2/testShows/FLCL/Season 01/s01e01.mkv
Size: 69 MB

Subtitle
Title: English - Default - Forced - PGSSUB
Language: eng
Codec: PGSSUB
AVC: No
Resolution: 1920x1080
Default: Yes
Forced: Yes
External: No

Video
Title: 1080p HEVC SDR
Codec: HEVC
AVC: No
Profile: Main 10
Level: 120
Resolution: 1920x1080
Aspect ratio: 16:9
Anamorphic: No
Interlaced: No
Framerate: 23.976025
Bitrate: 3844 kbps
Bit depth: 10 bit
Video range: SDR
Video range type: SDR
Color space: bt709
Color transfer: bt709
Color primaries: bt709
Pixel format: yuv420p10le
Ref frames: 1

Audio
Title: Stereo - Japanese - TRUEHD - Default
Language: jpn
Codec: TRUEHD
AVC: No
Layout: stereo
Channels: 2 ch
Sample rate: 48000 Hz
Bit depth: 24 bit
Default: Yes
Forced: No
External: No
```

### Application version

0.17.3

### Device information

Chromecast w/Google TV 4K

### Android version

Android 12

### Jellyfin server version

10.9.9

BlackstormFirst:
**Changes**
I've added checkboxes to enable/disable some audio codecs on demand.
This prevents playback issues on certain devices (ie. Samsung TV, Android Studio...) without changing Jellyfin TV's default behavior.
If the codec is manually disabled, the server will transcode.

In addition, the device report indicates now certain Passthrough possibilities.

**Code assistance**
Gemini Flash quick suggestion for "Compose" and codec exclusion syntaxe.

**Issues**
Fix: https://github.com/jellyfin/jellyfin-androidtv/issues/3917
Fix old PR: https://github.com/jellyfin/jellyfin-androidtv/pull/3929
Feature: https://github.com/jellyfin/jellyfin-androidtv/issues/5536

**New Checkboxes**
Screenshot_20260611_102443


app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
```diff
@@ -547,3 +569,22 @@
...
+		.setUsage(C.USAGE_MEDIA)
+		.setContentType(C.AUDIO_CONTENT_TYPE_MOVIE)
+		.build()
+	// Get audio capabilities
+	val audioCapabilities = AudioCapabilities.getCapabilities(context, audioAttributes, null)
+	// Set audio format for a passthrough audio codec 2.0 check
+	val format = Format.Builder()
+		.setSampleMimeType(mimetype)
+		.setChannelCount(Integer.bitCount(AudioFormat.CHANNEL_OUT_STEREO))
+		.setSampleRate(48000)
```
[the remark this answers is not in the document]
BlackstormFirst: resolved

ErikDB87:
Don't merge the master branch into your development branch. That'll create more commits, which would eventually pollute the master branch, if this PR would ever be accepted. (That's at least my understanding.)

Instead, 'rebase' your development branch onto the master branch. That will 'rewrite' the history of your development branch, and make it so, as if you made your commits, based (hence the name, I assume) on the latest pulled version of the master branch.

During a rebase, conflicts may arise, which you will then need to fix. After the rebase, you'll have to 'push with lease'. I suggest you try to find some more information about it. It took me some time and explanation to grasp it.

BlackstormFirst:
Hello,
Thanks for the advice; indeed, just like riding a bike, I still need training wheels to help me make real progress.

I still find the hoops you have to jump through just to contribute to a project to be incredibly cumbersome. Initially, I just wanted to report a bug. Then, to speed things up, I decided to contribute, but wrapping my head around everything—just for a small snippet of code—is a real hassle.
And on top of that, I just want to add 3 or 4 new variables to the MediaStream section, yet I find myself forced to modify the Server and API SDK projects in addition to the Android TV part. Having to learn OpenAPI, EF Core... it’s a massive headache and takes days, if not weeks, of learning just for three measly variables... the whole system is incredibly over-engineered.
It took an hour to write the code, but two weeks just to figure out how to integrate it...
Still... if that’s what it takes to fix bugs or add a new feature, I guess I just have to go along with it.

It’s worth noting that the last update was prompted by a "merge conflict," which I resolved by accepting both parts of the code. If I understand correctly, I shouldn't intervene in situations like that either, but rather let the maintainers handle it.

ErikDB87:
> It’s worth noting that the last update was prompted by a \"merge conflict,\" which I resolved by accepting both parts of the code.

Just pull the latest commits of the main branch, and then rebase your development branch onto main, and the label 'merge conflict' goes away.

In your case, you'll have to remove some commits (the ones with which you merged main into your branch). After rebasing, you'll have to 'push with lease'; otherwise `git` will complain. 

Once you feel you've grasped what should be done, I would explain the situation to Claude AI. It'll guide you through the process.

pikausp:
It would be amazing to have this. Jellyfin is unusable for my DTS-HD MA collection and I have to use Kodi.

@BlackstormFirst since you added explicit configuration for different levels of DD do you think it would make sense to do the same for DTS? Similar to what Kodi has.

For example Sonos can playback base DTS, but not higher versions.

Also a I would consider using the abbreviation DTS etc for labels as it's more widely recognized.

app/src/main/res/values-fr/strings.xml
```diff
@@ -669,6 +669,9 @@
     <string name="playback_buffer_auto">Auto</string>
     <string name="playback_buffer_large">Grand</string>
     <string name="playback_buffer_extra_large">Très grand</string>
+    <string name="lbl_bitstream_eac3">Flux audio Dolby Digital Plus</string>
```
nielsvanvelzen: Do not add translations via PR, they MUST be done via weblate only

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
```diff
@@ -265,12 +268,44 @@ fun SettingsPlaybackAdvancedScreen() {
...
+			)
+		}
+
+		item {
+			var truehdEnabled by rememberPreference(userPreferences, UserPreferences.truehdEnabled)
+
+			ListButton(
+				headingContent = { Text(stringResource(R.string.lbl_bitstream_truehd)) },
+				trailingContent = { Checkbox(checked = truehdEnabled) },
+				onClick = { truehdEnabled = !truehdEnabled }
```
nielsvanvelzen: The list of checkboxes is getting too long here now. I think it'd be best to create a subscreen for all these. Or maybe make them part of the existing media codecs screen?

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfileReport.kt
```diff
@@ -294,4 +295,14 @@ fun createDeviceProfileReport(
 		if (isS) appendItem("Device SKU") { appendValue(Build.SKU) }
 		if (isS) appendItem("Device SOC") { appendValue(Build.SOC_MODEL) }
 	}
+
+	appendSection("Audio Passthrough Capabilities") {
```
nielsvanvelzen: Definitely not the correct place to add these

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
```diff
@@ -265,12 +268,44 @@ fun SettingsPlaybackAdvancedScreen() {
...
 			ListButton(
 				headingContent = { Text(stringResource(R.string.lbl_bitstream_ac3)) },
 				trailingContent = { Checkbox(checked = ac3Enabled) },
-				onClick = { ac3Enabled = !ac3Enabled }
+				onClick = {
+					val newValue = !ac3Enabled
+					ac3Enabled = newValue
+					if (!newValue) {
+						eac3Enabled = false
+					}
```
nielsvanvelzen: ehhh this can be written more nicely

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -137,6 +137,21 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 */
 		var ac3Enabled = booleanPreference("pref_bitstream_ac3", true)
 
+		/**
+		 * Enable EAC3
+		 */
+		var eac3Enabled = booleanPreference("pref_bitstream_eac3", true)
```
nielsvanvelzen: We avoid the pref_ prefix in new preferences

app/src/main/res/values/strings.xml
```diff
@@ -664,4 +664,7 @@
     <string name="hlg">HLG</string>
     <string name="pref_software_codecs_enabled">Use software codecs</string>
     <string name="pref_software_codecs_enabled_description">Some software-based implementations may degrade performance. Disabling this option makes transcoding more likely.</string>
+    <string name="lbl_bitstream_eac3">Bitstream Dolby Digital Plus</string>
```
nielsvanvelzen: new strings shouldn't start with lbl_

nielsvanvelzen (changes requested):
Initial review pass

BlackstormFirst:
@nielsvanvelzen Hello and thanks for your review and advises, i will change/remove the code asap.

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfileReport.kt
BlackstormFirst: Hello,
I've handled all the other requests on my end, but I'm a bit unsure about this one.

Could you give me a few more details on what you'd like to see changed, please ?

This seems like the right file to report to the server the client's ability to handle these streams in stereo (I was also thinking of adding surround sound tests), in case the user wants to check and test their HDMI equipment before permanently disabling a codec.

Placing it at the end of the report seems appropriate too, though I don't have enough details to judge fully.
Could it be interfering with something else ?

Thanks in advance.

app/src/main/res/values-fr/strings.xml
BlackstormFirst: modifications from this file has been removed

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
BlackstormFirst: Moved to the MediaCodec page

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackAdvancedScreen.kt
BlackstormFirst: has been simplified

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
BlackstormFirst: pref_ prefix removed

app/src/main/res/values/strings.xml
BlackstormFirst: lbl_ prefix removed

nielsvanvelzen:
This pull request has had merge conflicts for over a week. You have to fix those first.

BlackstormFirst:
I fixed the current merge issue (twice) and re-run the app. It's OK now.
If you have some advices about the content added into the device profile report, can you please tell me what for correction. Again, thanks a lot.

image

app/src/main/java/org/jellyfin/androidtv/ui/settings/screen/playback/SettingsPlaybackCodecScreen.kt
```diff
@@ -55,6 +57,45 @@ fun SettingsPlaybackCodecScreen() {
...
+					if(!ac3Enabled) eac3Enabled = false
+				}
+			)
+		}
+
+		item {
+			ListButton(
+				headingContent = { Text(stringResource(R.string.bitstream_eac3)) },
+				trailingContent = { Checkbox(checked = eac3Enabled) },
+				onClick = { eac3Enabled = !eac3Enabled }, enabled = ac3Enabled
```
nielsvanvelzen:
```suggestion
				onClick = { eac3Enabled = !eac3Enabled },
				enabled = ac3Enabled
```

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
```diff
@@ -547,3 +568,29 @@ private fun DeviceProfileBuilder.subtitleProfile(
 	if (hls) subtitleProfile(format, SubtitleDeliveryMethod.HLS)
 	if (encode) subtitleProfile(format, SubtitleDeliveryMethod.ENCODE)
 }
+
+@OptIn(UnstableApi::class)
+fun isPassthroughAudioAvailable(context: Context, mimetype: String): Boolean {
```
nielsvanvelzen: These kinds of utility functions shouldn't be in the device profile file

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfileReport.kt
nielsvanvelzen: At the very least it should be above the app/device info where all the sections are. I would place it below the device codec decoders.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfile.kt
BlackstormFirst: I moved the function into a dedicated file and class within the codec directory; I think that if I'm moving it anyway, this makes for a cleaner setup.

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfileReport.kt
BlackstormFirst: done. moved after the device codec decoders section.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -135,7 +135,22 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		/**
 		 * Enable AC3
 		 */
-		var ac3Enabled = booleanPreference("pref_bitstream_ac3", true)
+		var ac3Enabled = booleanPreference("bitstream_ac3", true)
```
nielsvanvelzen: Existing preferences should not be renamed as that would reset their value

app/src/main/java/org/jellyfin/androidtv/util/profile/codec/HDMIAudioPassthroughCapabilities.kt
```diff
@@ -0,0 +1,38 @@
...
+			.setContentType(C.AUDIO_CONTENT_TYPE_MOVIE)
+			.build()
+		// Get audio capabilities
+		val audioCapabilities = AudioCapabilities.getCapabilities(
+			context,
+			audioAttributes,
+			null,
+			listOf(
+				AudioFormat.CHANNEL_OUT_STEREO,
+				AudioFormat.CHANNEL_OUT_5POINT1 )
```
nielsvanvelzen: This should be formatted either multi-line or single-line. Not a mix of both.

app/src/main/java/org/jellyfin/androidtv/util/profile/codec/HDMIAudioPassthroughCapabilities.kt
```diff
@@ -0,0 +1,38 @@
...
+import android.content.Context
+import android.media.AudioFormat
+import androidx.annotation.OptIn
+import androidx.media3.common.AudioAttributes
+import androidx.media3.common.C
+import androidx.media3.common.Format
+import androidx.media3.common.util.UnstableApi
+import androidx.media3.exoplayer.audio.AudioCapabilities
+
+class HDMIAudioPassthroughCapabilities {
```
nielsvanvelzen: The name implies this is for HDMI pass through only, but the implementation is about pass through in general. I'd reflect that in the name by calling is just "AudioPassthroughCapabilities".

nielsvanvelzen: This class only has a single function that does not have any other members. It could easily be an object, or what I'd probably do is move the context to a constructor property.

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfileReport.kt
```diff
@@ -164,6 +166,17 @@ fun createDeviceProfileReport(
 		}
 	}
 
+	appendDetails("HDMI Audio Passthrough Capabilities") {
```
nielsvanvelzen:
```suggestion
	appendDetails("Audio passthrough capabilities") {
```

We don't use word casing, and removing the "hdmi" part per my other comment.

nielsvanvelzen (changes requested):
What I don't really understand is why we're introducing code to detect passthrough support, but then just don't use it for the actual device profile. What is the rationale behind that choice?

BlackstormFirst:
> What I don't really understand is why we're introducing code to detect passthrough support, but then just don't use it for the actual device profile. What is the rationale behind that choice?

That was my initial setup, but...
The goal is to avoid detection errors that would result in the program making the decision on its own, rather than the user.
The equipment might return an incorrect value, be connected via optical cable, fail to report its actual capabilities, or be in a situation that hasn't been or cannot be properly tested.

So, I think the best approach is to offer users a manual override option on the client side; for advanced users, the details included in the report will help identify, if necessary, what the program perceives as the device's capabilities.

Regarding the HDMI tag, I seem to recall it being limited to that specific type of device, but upon re-reading the documentation, I can no longer find a specific reference to it. I’m going to remove those references.

app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
BlackstormFirst: prefix "pref_" restored for ac3

app/src/main/java/org/jellyfin/androidtv/util/profile/codec/HDMIAudioPassthroughCapabilities.kt
BlackstormFirst: Yes, I checked the other files in the directory, but apart from `deviceProfile.kt`, I didn't find a suitable location, so I decided to create a file dedicated to this functionality.
I made the fix by adding the context to the constructor and adjusting the calls; I hope that will be sufficient.

BlackstormFirst:
No answer to my question, so i update the code to what i suppose to be good to answer your ask.

misomarcell (approved):

nielsvanvelzen (approved):
Thanks for the PR, I've made some additional changes that you see in the commits I added. These changes actually use the detection function and allow for overrides in the settings. This is a similar design to the recently added HDR overrides. In testing on my emulator and shield devices the auto detection seemed to work as expected.

Note: opus 5.6 used for some code generation

hattmt:
Great improvements, 

Thank you for your hard work ! These are the functionalities I needed for my Onkyo TX-DS989 Ver2 amplifier ( 5.1 amplifier )

But there are a few options I'd still need:

- Transcoding target codec. Disabling a codec triggers a transcode, but the target is always AAC. Over optical, AAC can't be bitstreamed — it ends up as stereo PCM. Exposing TranscodingProfiles.AudioCodec (AAC / AC3 / E-AC3) would give 5.1 instead.

- Maximum audio channels. MaxAudioChannels already exists in the profile but has no UI. A 2 / 6 / 8 selector would let 5.1-only receivers get a proper downmix.

PR #5646, opened 2026-06-12
Refactor: Extract HEVC and AV1 codec modules

johnpc:
**Changes**

This follows up on #5604 by extracting the HEVC and AV1 codec detection logic out of MediaCodecCapabilitiesTest into their own modules. After this change, MediaCodecCapabilitiesTest is just a thin delegation facade with no logic of its own.

**Code assistance**

Claude Code was used as an assistant during implementation, code review, and test authoring.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/codec/Av1CodecCapabilities.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/codec/Av1CodecCapabilities.kt
@@ -1,0 +1,65 @@
+package org.jellyfin.androidtv.util.profile.codec
+
+import android.media.MediaCodecInfo.CodecProfileLevel
+import androidx.media3.common.MimeTypes
+import org.jellyfin.androidtv.util.AndroidVersion
+
+class Av1CodecCapabilities(
+	private val query: MediaCodecQuery,
+) {
+	companion object {
+		private const val MIME_AV1 = MimeTypes.VIDEO_AV1
+		private const val MIME_DOLBY_VISION = MimeTypes.VIDEO_DOLBY_VISION
+
+		// Fallback values from AOSP MediaCodecInfo.java for pre-Q/R devices
+		// Some devices (e.g., Fire OS) support AV1 below the official API level
+		internal const val AV1_PROFILE_MAIN10 = 0x2
+		internal const val AV1_PROFILE_MAIN10_HDR10 = 0x1000
+		internal const val AV1_PROFILE_MAIN10_HDR10_PLUS = 0x2000
+		internal const val AV1_LEVEL5 = 0x1000
+		internal const val DV_PROFILE_DVAV1_10 = 0x400
+	}
+
+	private val profileMain10: Int
+		get() = if (AndroidVersion.isAtLeastQ) CodecProfileLevel.AV1ProfileMain10 else AV1_PROFILE_MAIN10
+
+	private val profileMain10HDR10: Int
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/jellyfin/androidtv/util/profile/codec/Av1CodecCapabilities.kt
```diff
@@ -0,0 +1,66 @@
+package org.jellyfin.androidtv.util.profile.codec
+
+import android.media.MediaCodecInfo.CodecProfileLevel
+import android.os.Build
+import androidx.media3.common.MimeTypes
+
+class Av1CodecCapabilities(
+	private val query: MediaCodecQuery,
+	private val sdkInt: Int = Build.VERSION.SDK_INT,
```
nielsvanvelzen: Why do we need to set the sdkInt as a parameter? It's a constant (see the default value you're assigning).

johnpc: There's branching logic in `Av1CodecCapabilities` based on the value of sdkInt, which I wanted to stress in the unit tests. Mocking `Build.VERSION.SDK_INT` within the test file itself might be possible but there's not a pattern for it in the code, so I'd need to invent one (Robolectric? Reflection? Both feel more controversial than exposing the defaulted param)

Alternatively I might be able to remove the branching completely if it's truly dead code (and therefore have no need to test it), but my goal was to move and test the existing logic, not delete it.

What do you think? My preference personally would be to merge as-is, but happy to defer to your opinion on it.

nielsvanvelzen: I believe you can just mock the version_sdk_int with mockk for that

johnpc: I think `Build.VERSION.SDK_INT` can't be directly mocked. I tried it and the tests failed with `IllegalAccessException`.

I've added a commit to remove the field by adding a small layer of indirection via a new `DeviceSdk` internal object that I can mock. Let me know if that is better

johnpc: gentle ping for review on this

nielsvanvelzen: No need for pings, I'm just slow as I shift focus between various projects. With currently the mobile Android app being my main priority.

I understand the mocking of the Build.xxx constants doesn't work, I think this is also why our mobile app already wraps the SDK_INT. We can use the exact same approach as there:
https://github.com/jellyfin/jellyfin-android/blob/master/app/src/main/java/org/jellyfin/mobile/utils/AndroidVersion.kt

Instead of just making a facade for a single integer it has separate variables for different android versions which makes the usage side more readable. Ideally we'd adapt this pattern first (in a separate PR) and then update this one.

To make the mocking easier we'd probably still add a single wrapping getter for the SDK_INT so you can just mock that one in the tests.

johnpc: That makes sense. In that case, I've converted this PR to a draft while I wait for https://github.com/jellyfin/jellyfin-androidtv/pull/5672 which is a dependency.

johnpc: https://github.com/jellyfin/jellyfin-androidtv/pull/5672 is merged so I've rebased and switched over to using it in this branch

nielsvanvelzen (approved):

PR #5672, opened 2026-06-27
Add AndroidVersion helper for readable, mockable SDK version checks

johnpc:
**Changes**

Introduces a shared `AndroidVersion` object with named `isAtLeastX` properties, mirroring the [pattern in jellyfin-android](https://github.com/jellyfin/jellyfin-android/blob/master/app/src/main/java/org/jellyfin/mobile/utils/AndroidVersion.kt). A single `sdkInt` getter is the only place `Build.VERSION.SDK_INT` is read, so tests can mock the SDK version by mocking that one property (the static final field can't be set reflectively on JDK 17+). Existing inline `SDK_INT >= VERSION_CODES.X` comparisons across the app are migrated to use it.

This unblocks #5646, which can then be rebased to consume `AndroidVersion` and drop its `DeviceSdk` shim (see [this comment](https://github.com/jellyfin/jellyfin-androidtv/pull/5646#discussion_r3485893997)).

**Code assistance**

Claude Code was used as an assistant during implementation and test authoring.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfileReport.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfileReport.kt
@@ -251,7 +251,9 @@ fun createDeviceProfileReport(
 				appendLine()
 				appendLine("- ${HdrFormats.DOLBY_VISION.label}: ${supportedHdrTypes.contains(Display.HdrCapabilities.HDR_TYPE_DOLBY_VISION)}")
 				appendLine("- ${HdrFormats.HDR10}: ${supportedHdrTypes.contains(Display.HdrCapabilities.HDR_TYPE_HDR10)}")
-				if (AndroidVersion.isAtLeastQ) appendLine("- ${HdrFormats.HDR10_PLUS.label}: ${supportedHdrTypes.contains(Display.HdrCapabilities.HDR_TYPE_HDR10_PLUS)}")
+				if (AndroidVersion.isAtLeastQ) {
+					appendLine("- ${HdrFormats.HDR10_PLUS.label}: ${supportedHdrTypes.contains(Display.HdrCapabilities.HDR_TYPE_HDR10_PLUS)}")
+				}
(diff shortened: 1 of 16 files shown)


app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfileReport.kt
```diff
@@ -47,13 +48,13 @@ private val featureNames = setOf(
...
-private val isS = Build.VERSION.SDK_INT >= Build.VERSION_CODES.S // API 31
-private val isUpsideDownCake = Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE // API 34
-private val isBaklava = Build.VERSION.SDK_INT >= Build.VERSION_CODES.BAKLAVA // API 36
+private val isN = AndroidVersion.isAtLeastN // API 24
+private val isO = AndroidVersion.isAtLeastO // API 26
+private val isQ = AndroidVersion.isAtLeastQ // API 29
+private val isR = AndroidVersion.isAtLeastR // API 30
+private val isS = AndroidVersion.isAtLeastS // API 31
+private val isUpsideDownCake = AndroidVersion.isAtLeastUpsideDownCake // API 34
+private val isBaklava = AndroidVersion.isAtLeastBaklava // API 36
```
nielsvanvelzen: Don't think we need these variables anymore now, instead use the AndroidVersion.x getters directly.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/util/profile/deviceProfileReport.kt
johnpc: Pushed in abe060053.

nielsvanvelzen (changes requested):
I'm assuming the ones in MediaCodecQuery and MediaCodecCapabilitiesTest were intentionally omitted because of the other PR. Otherwise you missed one SDK_INT usage in copyAction.kt

johnpc:
Thanks! got the copyAction one too now :D

nielsvanvelzen (approved):

PR #5673, opened 2026-07-01
Add support for separate skip ahead and skip back lengths

kferguson42:
**Changes**
* Allow existing `skipBackLength` variables to be visible and changed in the Settings, and to be used during video seeking.

**Code assistance**
No AI agents were used to write this code.  However, cards on the table, I have no real background in software/web development; the vast majority of my programming experience is in data science.  The changes made here are just the result of me looking for places where `skipForwardLength` or a similarly named variable already existed in the codebase, but had no corresponding `skipBackLength`.

**Issues**
Partially addresses #5510

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -16,6 +16,7 @@
  - [MajMongoose](https://github.com/majmongoose)
  - [Olaren15](https://github.com/Olaren15)
  - [dtrexler](https://github.com/dtrexler)
+ - [Kyle Ferguson](https://github.com/kferguson42)
 
 # Emby Contributors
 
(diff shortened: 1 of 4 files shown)


nielsvanvelzen:
I don't think it makes sense to add a setting that isn't actually used. The current video player only uses the forward duration and we cannot make it use both forward and backward as it is not our code.

kferguson42:
> I don't think it makes sense to add a setting that isn't actually used. The current video player only uses the forward duration and we cannot make it use both forward and backward as it is not our code.

Ah, sorry, I hadn't realized that (see the aforementioned lack of experience with this type of programming).  What video player is used?  If I play a video in my (Firefox) web browser, there are settings for both skipping ahead and back, and the player seems able to handle both separately, so I guess different platforms use different players?

nielsvanvelzen:
This is not jellyfin-web, this is the Android TV app which does NOT run in Firefox.

nielsvanvelzen:
Closing as invalid (see explanations above)

PR #5683, opened 2026-07-07
Add configChanges to manifest

### LINKED ISSUE #5682: TV resolution change immediately stops playback

### Describe the bug

When you attempt to play videos that force a resolution change, the MainActivity is recreated, which stops the currently playing media immediately.

This prevents playing these videos in "Scale on TV" mode (which if you have a decent TV is where you want scaling to happen).

Work-around is to set the to `Scale: Disabled`, but the correct solution is to update the `AndroidManifest.xml` to declare it can handle `configChanges` related to layout like the generic Android client [already does](https://github.com/jellyfin/jellyfin-android/blob/master/app/src/main/AndroidManifest.xml#L51) so the application isn't torn-down and recreated just as the resolution negotiation has succeeded.

### Logs

_No response_

### Application version

v0.19.9

Google Play Store

### Device information

Google TV Streamer

### Android version

Android 14 for TV

### Jellyfin server version

10.11

### Other sources

_No response_

IDisposable:
Ensures that the application is not destroyed and recreated when screen resolution (and layout, etc.) is changed so that media that just negotiated a video resolution and rate doesn't immediately crash the playing video.

**Changes**
Add `configChanges` to _AndroidManifest.xml_ to ensure that the application isn't recreated for mere config changes.

Note: This is already in the non-TV version.

**Code assistance**
No

**Issues**
Fixes #5682

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -125,6 +125,7 @@
         <!-- Main application -->
         <activity
             android:name=".ui.browsing.MainActivity"
+            android:configChanges="density|screenSize|smallestScreenSize|screenLayout|orientation|keyboard|keyboardHidden|navigation|uiMode"
             android:launchMode="singleTask"
             android:windowSoftInputMode="adjustNothing" />
 


nielsvanvelzen (approved):

Pentaphon:
@IDisposable @nielsvanvelzen confirmed to fix the video failing to play but there is 1 minor issue. When you play 720p video with the "refresh matching > scale to TV" option enabled on a 4K TV, the player UI also scales down to 720p, making the player UI elements appear much bigger on screen. Please confirm.

nielsvanvelzen:
That's to be expected, if you play 720p content and ask your tv to switch to 720p resolution the UI will also be 720p.

Pentaphon:
> That's to be expected, if you play 720p content and ask your tv to switch to 720p resolution the UI will also be 720p.

I wouldn't really expect that, actually, I would expect the player UI to also be scaled down to 720p so the player UI elements don't appear to be much larger than they are at 1080p or 4K for the sake of UI consistency. This is the first time I've seen a player UI do that when resolution matching. It doesn't "break" anything but all the buttons, graphics, UI are huge when playing 720p video.

IDisposable:
I think a picture would be helpful to address this, but you should open a separate Issue for it.

nielsvanvelzen:
The UI rendering at 720p when using the display scaling option has always worked that way. I'm not opposed to improving it but that would be a feature request, as it is currently working as intended. Please create a (feature request-) issue and I'll see what I can do with it for the new video player.

PR #5687, opened 2026-07-09
Add Amazon Fire TV 4K (1st Gen) to modelsWithDoViHdr10PlusBug

nyanmisaka:
**Changes**
- Add **Amazon Fire TV 4K (1st Gen)** to modelsWithDoViHdr10PlusBug

```
Fire TV Stick 4K - 1st Gen (2018)	AFTMM	API Level 25 (Android 7.1)	Fire OS 6
```
https://developer.amazon.com/docs/device-specs/identify-fire-tv-devices.html

**Code assistance**

**Issues**
- https://old.reddit.com/r/jellyfin/comments/1ur7xr5/black_screen_with_audio_on_fire_tv_4k_same_file/


nielsvanvelzen (approved):
Idk how we missed that when the list was created

nyanmisaka:
> Idk how we missed that when the list was created

I suspect that many less popular Fire OS based devices are also affected by this firmware bug, but few people run Jellyfin on them. Checking for devices with the vendor name `Amazon` to add to the bug list might be a more permanent solution.

nielsvanvelzen:
Ideally we'd not have to maintain lists like this at all. The fact we added this was a last-effort, I still don't like it.

PR #5697, opened 2026-07-17
Show server splash screen on launch.

DerrickDW:
**Changes**
Makes "Enable splash screen image" display during app launch instead of only on the login page. Uses the default logo as a fallback when no server splash screen is available. Delays fallback logo rendering until session restoration completes to prevent a brief flicker.

**Code assistance**
Claude helped trace the origin of `splashscreenUrl`, `ServerRepository`, and `SessionRepository`, and assisted with Kotlin syntax cleanup and initial class structure.

**Issues**


app/src/main/java/org/jellyfin/androidtv/ui/startup/fragment/SplashFragment.kt
```diff
@@ -33,25 +42,50 @@ fun SplashScreen() {
...
+				)
+			} else if (showLogo) {
+				Image(
+					painter = painterResource(R.drawable.app_logo),
+					contentDescription = stringResource(R.string.app_name),
+					modifier = Modifier
+						.width(400.dp)
+						.fillMaxHeight()
+				)
+			}
```
nielsvanvelzen: The logo should always be visible, if the splashscreen is used the logo should be shown on top of it.

app/src/main/java/org/jellyfin/androidtv/ui/startup/fragment/SplashFragment.kt
```diff
@@ -33,25 +42,50 @@ fun SplashScreen() {
...
-					.width(400.dp)
-					.fillMaxHeight()
-			)
+			if (splashscreenUrl != null) {
+				AsyncImage(
+					model = splashscreenUrl,
+					contentDescription = stringResource(R.string.app_name),
+					modifier = Modifier.fillMaxSize(),
+					error = painterResource(R.drawable.app_logo)
+				)
```
nielsvanvelzen: We should mimic the styling the backgroundservice currently uses as right now the splash screen is extremely bright as there is no dimming.

nielsvanvelzen (changes requested):

app/src/main/java/org/jellyfin/androidtv/ui/startup/fragment/SplashFragment.kt
DerrickDW: Thanks for taking a look! The dimming suggestion is a great idea. I hadn't considered how bright a custom splash image could be in a dark room as the one I tested with had a black background, so matching the background service styling definitely makes sense. I'll get another commit cooking.

app/src/main/java/org/jellyfin/androidtv/ui/startup/fragment/SplashFragment.kt
DerrickDW: For the logo, I had one question: the current splash logo is fairly large, so when it's overlaid on a custom splash image it can cover quite a bit of the artwork. Did you have a preferred placement/size in mind, or would you like me to just overlay as is?

app/src/main/java/org/jellyfin/androidtv/ui/startup/fragment/SplashFragment.kt
nielsvanvelzen: The app is optimized for default usage, so for this specific change you'll have to test with the server generated splash screen and not a custom one to see how it will look for 90% of the users.

app/src/main/java/org/jellyfin/androidtv/ui/startup/fragment/SplashFragment.kt
nielsvanvelzen: Keep it at the same position so the logo will never shift if there is or isn't a splash image

app/src/main/java/org/jellyfin/androidtv/ui/startup/fragment/SplashFragment.kt
DerrickDW: Implemented in commit 43e9efc
Reused the existing background_filter tint and ContentScale.Crop to match the background styling.

app/src/main/java/org/jellyfin/androidtv/ui/startup/fragment/SplashFragment.kt
DerrickDW: implemented in commit c4ea033
overlaid jellyfin logo over custom splash images, keeps existing position and size, falls back to default display if no custom splash set.
tested on emulator screenshots of both tint, and logo overlay
Screenshot_2026-07-25_11-08-16
Screenshot_2026-07-25_11-13-25
Screenshot_2026-07-25_11-04-53

app/src/main/java/org/jellyfin/androidtv/ui/startup/fragment/SplashFragment.kt
```diff
@@ -20,38 +22,79 @@ import androidx.compose.ui.res.stringResource
...
+		val currentServer by serverRepository.currentServer.collectAsState()
+
+		val splashscreenUrl = currentServer
+			?.takeIf { it.splashscreenEnabled }
+			?.let { jellyfin.createApi(baseUrl = it.address).imageApi.getSplashscreenUrl() }
+
+		// Only fall back to the default logo once session restoration has
+		// actually resolved, otherwise we flash the logo before swapping
+		// to the server splashscreen once it's known.
+		val showLogo = sessionState == SessionRepositoryState.READY
```
nielsvanvelzen: I don't think we need this? The logo can just be always visible

nielsvanvelzen (changes requested):
I'm not really sold on the effect yet. When I tested this branch it showed the Jellyfin logo and then suddenly the splashscreen image popped into existence, I would say it needs a fade-in effect at the minimum.

Perhaps it would also be better if we use the `AppBackground` composable here instead so it can smoothly transition to the next screen.

app/src/main/java/org/jellyfin/androidtv/ui/startup/fragment/SplashFragment.kt
DerrickDW: Addressed both in commit 71882cc

I was able to reproduce the background pop-in on an uncached launch. I simplified the implementation by always displaying the Jellyfin logo and fading the server background in once it finishes loading. This removes the logo timing issue while keeping the startup transition smooth. I also removed the now-unnecessary sessionState logic.

I used a simple 200 ms fade-in for the background. Let me know if you'd prefer the transition to more closely match AppBackground.

While working through this, I also noticed BackgroundService.setBackground(server) already loads the server splash through the shared ImageLoader. It made me wonder whether startup should eventually reuse that path instead of loading the splash independently, or if the current approach is a better fit for the startup lifecycle.

nielsvanvelzen:
So playing around with this branch, I think using `AppBackground()` is the best way to make this work. This also gives a much smoother transition from the login screen.

```kt

@Composable
fun SplashScreen() {
	val serverRepository = koinInject<ServerRepository>()
	val backgroundService = koinInject<BackgroundService>()
	val currentServer by serverRepository.currentServer.collectAsState()

	LaunchedEffect(currentServer) {
		if (currentServer != null) backgroundService.setBackground(currentServer!!)
		else backgroundService.clearBackgrounds()
	}

	AppBackground()

	Box(
		modifier = Modifier
			.fillMaxSize(),
		contentAlignment = Alignment.Center,
	) {
		Image(
			painter = painterResource(R.drawable.app_logo),
			contentDescription = stringResource(R.string.app_name),
			modifier = Modifier
				.width(400.dp)
				.fillMaxHeight(),
		)
	}
}
```

DerrickDW:
> So playing around with this branch, I think using `AppBackground()` is the best way to make this work. This also gives a much smoother transition from the login screen.
> 
commit [4ec0dc9](https://github.com/jellyfin/jellyfin-androidtv/pull/5697/commits/4ec0dc92d0333608f29046462a9bcc53c1cb823d)
Thanks! I updated the implementation to use AppBackground() and BackgroundService as suggested. I agree that is a much smoother and less jarring transition.

One thing I noticed while testing: on a cold startup there seems to be a brief period where the default background is shown before the server splash fades in. I wonder if that's because currentServer briefly emits null, causing clearBackgrounds() before the server is restored. I haven't changed that behavior since it was in your example, but I thought I'd mention it in case it's worth investigating.

nielsvanvelzen:
The app always starts with the default background (See _currentBackground in BackgroundService), so it's expected behavior that the splash screen needs a brief moment to load in.

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks for the PR! :)

PR #5712, opened 2026-07-25
Align version handling with server and web logic

Shadowghost:
The playback logic for different versions doesn't really work as intended by the server right now.

**Changes**
* Report the played media source during playback and not just the main item
* Play the media source the server prioritizes (server does it for a reason and it is necessaary for version-aware resume
* Pass the selected version through the playback launch path (otherwise we won't play that version)
* Show and resume the selected version on the details page
* Continue playing the matching version of the next item

**Code assistance**
Opus 5 for exploration and some cross checking between the web logic and the one in atv.

**Issues**


nielsvanvelzen:
In the Jellyfin API an alternative version of a file always has it's own item id, therefor reporting the media source id it not necessary unless breaking behavior changed were introduced in the server. In which case those have not been communicated properly.

PR #5033, opened 2025-10-19
Add Bluetooth media control support (Media Session)

PraaneshSelvaraj:
## Summary
Implements MediaSession to enable Bluetooth media controls (play/pause) for Android TV.

## Changes Made
- Added MediaSession initialization in VideoManager constructor
- Connected MediaSession to ExoPlayer instance
- Added proper MediaSession cleanup in releasePlayer()

## Testing
Manually tested on Android TV with Bluetooth headset. Play/pause controls and in-ear detection working as expected.

**Test Device:**
- Device: Vu GloQLED QLED Ultra HD
- Android TV OS: 12 (Build: STT2.230929.001.080)
- Bluetooth Audio: TWS earbuds with MediaButton support

Fixes: Users were unable to control playback using Bluetooth headset buttons

## Question for Maintainers
While implementing this, I noticed inconsistent next/previous button behavior between web and Android TV for series playback:

**Web version:** Next/Previous buttons navigate between episodes
**Android TV:** Next does nothing, Previous restarts video at 0:00

Should episode navigation be added to match the web behavior? Happy to implement this in a follow-up PR if desired.

PR #5735, opened 2026-08-04
Add media session for video playback

makhnanov:
owning an active media session, never to the foreground key handling. A single
click on Bluetooth headphones (AirPods and similar) therefore did nothing while
a video was playing, even though audio playback reacts to it through its own
session in `:playback:media3:session`.

This adds `VideoMediaSession`, owned by `CustomPlaybackOverlayFragment`, which
publishes a session while the video player is active and maps its transport
controls (play/pause, stop, seek, skip, fast forward/rewind) onto
`PlaybackController`. The published playback state is kept in sync because AVRCP
devices decide from it whether a click should send play or pause. Media keys
that are handled on key up are now also consumed on key down, otherwise the
system forwards the unhandled down event to the session and a single press on a
remote would toggle playback twice.

**Code assistance**
Implemented with Claude Code (Opus 5): code base exploration, the implementation
itself and the commit message. Reviewed and manually tested by me on a Google TV
Streamer with AirPods.


nielsvanvelzen:
Duplicate of #5033

PR #5644, opened 2026-06-11
Fix for audio and subtitle auto selection from previous media

BlackstormFirst:
Changes:

- Audio verification of the previous media is no longer limited to just the language but also checks the codec.
If the same codec isn't found, then the selection reverts to language only.

- Subtitle verification is no longer limited to just the language. It now checks the Title, Codec, Tag (forced or not), and finally the language.
If the subtitles don't exactly match, it performs a fallback of Codec/Tag/Language, then if necessary another on Tag/Language, and finally a last one if needed on language only, as is currently the case for the latter.

Code assistance:
Android Studio (without AI)

Issues:
Fix: https://github.com/jellyfin/jellyfin-androidtv/issues/5636

PR #5744, opened 2026-08-08
fix(playback): restore subtitles by language + title across episodes

Tonyha7:
**Changes**
Sometimes the same sub could be on different ids in episodes. And the playback could use a different language sub. This commit fixed the issue by getting subtitle info. The entire process looks like this:
exact (language + title) -> language only -> server default

**Code assistance**
Most of the code is generated by deepseek v4 flash, I tested and made some changes.


BlackstormFirst:
There is already a similar, more comprehensive PR that was moved to draft status last week (https://github.com/jellyfin/jellyfin-androidtv/pull/5644).
Another one is currently under development with better handling and should be available soon.

Issue #4007, opened 2024-09-19
Jellyfin won't let Nvidia Shield Energy Saver (sleep timer) setting work


TorrentFiend:
### Describe the bug

Jellyfin Client on Nvidia Shield seems to ignore my setting for Energy Saver preferences.  To be clear I have my sleep timer set for 1 hour and this with the CEC features my TV has works great as a sleep timer.  It is a frequent thing that I fall asleep watching things and it is great knowing that my nvidia shields sleep mode will also turn off the soundbar and TV all at once after 1 hour of no activity.  If I can't stay awake for the rest of a movie or show I wake up a few hours later and my TV and everything turns itself off instead of everything running all night wasting electricity and hours of backlight use on the display.  Even worse this is a BIG issue on OLEDs where burn in can happen if static images are on too long.  

I have tested this on Emby, Plex, Nova Video Player, SmartTube, Netflix, Max, hulu, amazon prime.... etc etc.  They all work fine and no other app on my shield keeps my display on forever but if I use Jellyfin to watch anything and fall asleep it stays on until I eventually find it still on the next morning and manually shut it off wasting away for hours needlessly.  It will be completely idle on my Jellyfin page of whatever the last thing watched that ended was.

I have tested this several times simply by opening any page within Jellyfin, for example just opening the app and walking away....it will stay on all day and never goes to sleep like it should.  This has meant for now I am reluctantly using Emby still until this is hopefully fixed.  If I fall asleep using Emby the shields sleep timer functions normally and I can trust it won't stay on all night.

I've also tested this turning on the option under energy saver preferences to "Allow turning screen off during media playback" as a TEST ONLY and this DOES allow it to turn off as it should but I never would want to use it this way since if a file is playing from a spinning drive and the display suddenly stops and turns off mid playback I fear it may be bad on the drive to abruptly stop playback like this suddenly.  This will act as an override on this setting I guess to turn the TV off NO MATTER WHAT but applications shouldn't force users to resort to this setting right?  I don't want it to shut off if media is playing, only during the idle time after something has ended. Like if I fall asleep the last 20 minutes of a movie, the energy saver timer can take it from there and turn it off while jellyfin is open but IDLE after the credits end.  Under normal working functionality this overkill option shouldn't need to be enabled, nothing else to date has needed this on, Jellyfin shouldn't either right?

[...]

PR #5746, opened 2026-08-09
Stop playback when the display link drops

sinful1992:
## Problem

Playback continues indefinitely after the television is switched off. In my case four episodes autoplayed unattended over an hour with the TV off the whole time.

This has been reported before as #4007, which was closed as not planned. I think the reason no fix was found is that the obvious signals genuinely do not exist on this hardware, and it is worth documenting why.

## Why the obvious approaches do not work

Chromecast with Google TV 4K (`sabrina`, Amlogic), Android 14, attached to an LG panel. With the television **switched off**:

```
/sys/class/amhdmitx/amhdmitx0/hpd_state = 1
```

Hotplug detect stays asserted, so from Android's point of view nothing changes:

- `mScreenState=ON`, `mState=ON`, `committedState ON`, `mIsEnabled=true`
- EDID still readable, `deviceProductInfo{name=LG TV, ...}` still populated
- no hotplug event
- `mWakefulness=Awake` throughout
- `STREAM_MUSIC Devices: hdmi(400)` unchanged

So a `DisplayManager.DisplayListener` never fires, and there is no `ACTION_SCREEN_OFF` or `PowerManager` transition to observe.

CEC is not available either. Over three days of logs on this device:

```
dumpsys hdmi_control | grep -c '\[S\]'   ->  238
dumpsys hdmi_control | grep -c '\[R\]'   ->    0
```

238 messages sent, zero ever received, and the TV was never discovered as a CEC device. Every `<Give Device Power Status>` to logical address 0 is `(NACK, NACK)`.

## What does change

HDCP. With the television off the transmitter cannot authenticate and retries continuously:

```
E/hdcp_tx22: [TX] Capability check failed: -662.  Retrying...
E/hdcp_tx22: [ESM] SYSTEM_RESET / RESET COMPLETE / COMM RD-NAK
E/hdcp_tx22: [TX] enable low value content
```

With the television on, zero such errors. This is why DRM protected apps stop on their own and unprotected local playback does not — they are not detecting the television, they are simply losing their secure path.

`MediaDrm.getConnectedHdcpLevel()` exposes this to applications, and it tracks the sink correctly here: `6` (`HDCP_V2_3`) with the TV on, `1` (`HDCP_NONE`) with it off.

## The change

`HdcpMonitor` polls the connected HDCP level during playback and ends playback when it reports no protection.

- Polls every 5 seconds and requires two consecutive readings before acting, because changing video mode renegotiates HDCP and a single reading can be a false positive between episodes of differing frame rate.
- The flag recording that a protected level was ever seen is process wide. Per instance it reset on every fragment recreation, which meant a monitor created while the link was already down started unarmed and never acted — precisely the autoplay path this is meant to catch.
- Displays that never negotiate HDCP at all are unaffected, since the monitor only acts on a transition away from a real protected level.
- The poll loop exits as soon as playback has been ended, releasing the `MediaDrm` session without depending on view teardown, which does not run on every path out of playback.
- No-op below API 28.

## Testing

On the hardware above:

- Mid-episode, TV switched off: stopped 2.4s after the HDCP failure appeared in the kernel log.
- Across an autoplay rollover: polls continue uninterrupted through `itemComplete()` -> next up -> new fragment, then the stop fires correctly during episode 2, ~10s after power off.
- Play -> stop -> play -> stop: second stop fires correctly.
- Session release confirmed 149ms after the stop, with no further polling.
- In each case verified actually stopped, not merely logged: `dumpsys media.audio_flinger` active streams 0 and the player torn down.

I only have this one device to test on, so I would welcome checks on other hardware — particularly anything where the panel drops HPD on power off, where I would expect the existing display path to already work and this to be redundant but harmless.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/CustomPlaybackOverlayFragment.java
@@ -66,6 +66,7 @@ import org.jellyfin.androidtv.ui.presentation.MutableObjectAdapter;
 import org.jellyfin.androidtv.ui.presentation.PositionableListRowPresenter;
 import org.jellyfin.androidtv.util.CoroutineUtils;
 import org.jellyfin.androidtv.util.DateTimeExtensionsKt;
+import org.jellyfin.androidtv.util.HdcpMonitor;
 import org.jellyfin.androidtv.util.ImageHelper;
 import org.jellyfin.androidtv.util.InfoLayoutHelper;
 import org.jellyfin.androidtv.util.TextUtilsKt;
@@ -137,6 +138,8 @@ public class CustomPlaybackOverlayFragment extends Fragment implements LiveTvGui
     private final Lazy<ImageHelper> imageHelper = inject(ImageHelper.class);
 
     private final PlaybackOverlayFragmentHelper helper = new PlaybackOverlayFragmentHelper(this);
+
+    private HdcpMonitor hdcpMonitor;
 
     @Override
     public void onCreate(Bundle savedInstanceState) {
@@ -223,11 +226,18 @@ public class CustomPlaybackOverlayFragment extends Fragment implements LiveTvGui
         if (playbackController != null) {
             playbackController.init(new VideoManager(requireActivity(), view, helper), this);
         }
+
+        hdcpMonitor = CustomPlaybackOverlayFragmentHelperKt.startHdcpMonitor(this);
     }
 
     @Override
     public void onDestroyView() {
         super.onDestroyView();
+
+        if (hdcpMonitor != null) {
+            hdcpMonitor.stop();
+            hdcpMonitor = null;
+        }
 
         binding = null;
         // To fix race condition in hide timer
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:

sinful1992:
Ha, it makes sense why people leave this company,  even something that solves real issue is ignored....

nielsvanvelzen:
We're not a company, we're a group of volunteers. If there was a "real issue" you could've linked it (you didn't) and bothered to make a proper PR instead of having Claude generate some garbage in 2 seconds.

sinful1992:
There is a proper issue, playback keeps playing in background on google tv and firestick, and its detailed if you bothered to read or look into it yourself...

nielsvanvelzen:
Asking Claude to link a random, closed, issue from 2 years ago is "proper"? Lol.

sinful1992:
Issue that has not been solved,  and I solved it my end, and tested it, amd it works , as all documented. Now it doesn't play all night with tv switched off. But keep your head buried afraid of ai . People find the time and money to contribute amd you do this kind of nonsense, hopefully someone clones and rewrites the whole thing and we won't use to deal with you people for much longer..

Issue #5636, opened 2026-06-09
Incorrect subtitle and audio auto selection from previous media


BlackstormFirst:
### This issue respects the following points:

- [x] I have discussed this issue in the [Jellyfin forum](https://forum.jellyfin.org/) and/or [troubleshooting chat](https://jellyfin.org/contact) before opening this issue.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

Hello, there are several recurring problems with subtitle management and selection from one video to another.

Here's one of the problems: A video with at least two subtitles in the same language doesn't automatically select the correct subtitle, but rather the first one it finds in the previously selected language.

For example, if you have a video with:
1 - French subtitles (FULL ASS)
2 - French subtitles (FORCED ASS)
3 - French subtitles (FULL SRT)
4 - French subtitles (FORCED SRT)

If you select subtitle 4 (FORCED SRT) while playing the first episode, when you switch to the second video, subtitle 1 (FULL ASS) is always selected.

The "PlaybackController" doesn't differentiate between the two, and the user has to manually change the subtitles from one episode to the next.

The "videoQueueManager" doesn't seem to remember whether the subtitle is forced or what type it is (codecID).

Is it possible to fix this, for example, by adding this information and then using it in the "PlaybackController" to correctly load the right subtitle from one episode to the next?

Thank you in advance for this fix.

### Media info of the file

```shell
Général
Identifiant unique                       : 212762968432444715352945450597141794006 (0xA010A83E00861C2B1B329EAAC1B14CD6)
Nom complet                              : X:\Videos\Animes\Animes VF\I Was Reincarnated as the 7th Prince (2024)\I Was Reincarnated as the 7th Prince so I Can Take My Time Perfecting My Magical Ability - S01E02 - 1080p - x264 - BDRip - MULTi.mkv
Format                                   : Matroska
Format, Version                          : Version 4
Taille du fichier                        : 1,33 Gio
Durée                                    : 23 min 51 s
Type de débit global                     : Variable
Débit global                             : 7 954 kb/s
[...]

Issue #5407, opened 2026-02-11
Playback crash due to invalid subtitle index (IndexOutOfBoundsException)


tomatenbruder:
### This issue respects the following points:

- [x] I have discussed this issue in the [Jellyfin forum](https://forum.jellyfin.org/) and/or [troubleshooting chat](https://jellyfin.org/contact) before opening this issue.
- [x] This issue is **not** already reported on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/issues?q=is%3Aopen+is%3Aissue) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).
- [x] This report addresses only a single issue; If you encounter multiple issues, kindly create separate reports for each one.

### Describe the bug

The Jellyfin Android TV app crashes immediately when starting playback of a movie.
The crash happens consistently during playback initialization, right after the
PlaybackInfo request, before video playback actually starts.

According to the stack trace, the crash is caused by an IndexOutOfBoundsException
in `StreamInfo.getSubtitleDeliveryMethod()`.

[upload_Jellyfin.Android.TV_0.19.7_20260211111711_14f20f93443c43b0b4ec45ffd3560185.log](https://github.com/user-attachments/files/25232385/upload_Jellyfin.Android.TV_0.19.7_20260211111711_14f20f93443c43b0b4ec45ffd3560185.log)

### Media info of the file

```shell

```

### FFmpeg logs

```shell

```

### Application version

0.19.7

### Where did you install the app from?

Google Play Store

### Device information

FireTV 4k Max

### Android version

Android 11

### Jellyfin server version

10.11.6

stosumarte:
Having the same issue.

Is there any workaround currently?
Also, I'm able to provide logs if needed.

nielsvanvelzen:
At the very least share the crash logs in combination with the media info for the file that crashes the player. I'm not a wizard, I need information to work with 🧙‍♂️.

stosumarte:
@nielsvanvelzen I'm attaching the logs here!

[upload_Jellyfin Android TV_0.19.7_20260217205055_39d8110ad3da4dbc87ebba83ce08f83b.log](https://github.com/user-attachments/files/25395278/upload_Jellyfin.Android.TV_0.19.7_20260217205055_39d8110ad3da4dbc87ebba83ce08f83b.log)

Media info here:

```
Audio
[...]

PR #5724, opened 2026-07-29
fix appcrash when call updateBurningSubs

BlackstormFirst:
Fix appcrash IndexOutOfBoundsException when number of streams differs from the current video and the previous one and targeted index doesn't exist.

**Changes**
Get the default streams index in place of the previous stream index.

**Code assistance**
None

**Issues**
Fix issue: https://github.com/jellyfin/jellyfin-androidtv/issues/5407

Issue #5656, opened 2026-06-17
Add Impaired/Commentary/AudioDescription indicators management from previous media


BlackstormFirst:
### This request respects the following points:

- [x] This request is **not** already on [GitHub](https://github.com/jellyfin/jellyfin-androidtv/labels/enhancement) _(I've searched it)_.
- [x] I agree to follow Jellyfin's [Code of Conduct](https://jellyfin.org/docs/general/community-standards.html#code-of-conduct).

### Problem description

Hi,
I'm currently working on and already have a private branch to add SDH management (in addition of my previous PR https://github.com/jellyfin/jellyfin-androidtv/pull/5644) but it missing constants to the Jellyfin Server and SDK to manage others indicators.

I will try to add them with 2 new PR in each project, but i think it will take some time to validate them.
If a maintainer can add them before.. i will be very happy ;=)

It will solve at least 5 or 6 asks for features in the discussion for features in Jellyfin server forum, for the Jellyfin Android TV app.

Audio:

<img width="758" height="270" alt="Image" src="https://github.com/user-attachments/assets/71259323-52fd-4d4e-968c-da418e9f3884" />

Sub:

<img width="752" height="268" alt="Image" src="https://github.com/user-attachments/assets/de2e70cc-3c69-4fa2-b7c1-132c7176765a" />

Ffprobe:

<img width="307" height="408" alt="Image" src="https://github.com/user-attachments/assets/59cc7b5c-b99b-489e-9c06-2c2367b61bb9" />

### Proposed solution

Add at least 3 indicators in Jellyfin Server and SDK projects:
- Visual Impaired
- Commentary (comment)
- Audio description

### Alternatives considered

I will send my own PR on others projects

BlackstormFirst:
Currently, I've managed to implement the constants variables on the server side and in the SDK, but I'm still having trouble updating the columns in the database.

If anyone knows the best way to modify the database using the new EF Core system (directly in the code? via the migration system? using a migration XML file? another method?), I would appreciate some documentation on the subject to save time.

Also, I've modified the SDK's openapi.json file locally and generated the .kt files using the generator, but I don't quite understand how to submit a pull request for this modification. Do I only submit the JSON file? Or do I need to include all the modified files? What would be the cleanest way to do this?

[...]

PR #5748, opened 2026-08-10
fix audio and subtitle playback from previous media

BlackstormFirst:
- Fix audio and subtitle management from previous media.
- Fix appcrash with NullPointException on updateburningsub call (in a few instances where the number of indexes differs)
- Clear videoqueue when endPlayback is triggered

This is a first working draft for the update of my old PR (https://github.com/jellyfin/jellyfin-androidtv/pull/5644 & https://github.com/jellyfin/jellyfin-androidtv/pull/5724).

When you start a new media item, the client requests index information from the server.
You can change the audio and subtitles while playback is in progress.
When playing the next or previous episode, the same information is reused to provide the new indices; otherwise, the system tests various fallback options to determine the best solution.
Upon fully stopping playback (returning to the menu), the information from the last session is cleared, allowing the remote server to provide its own indices when the next piece of content is played.

**Changes**
The main change involves the addition of two functions used to return the best index depending on the situation.
The other modifications relate to the integration of these functions and the mechanism.

**Code assistance**
Android Studio (No AI).

**Issues**
Fix issue: https://github.com/jellyfin/jellyfin-androidtv/issues/5407
Fix issue: https://github.com/jellyfin/jellyfin-androidtv/issues/5636
Feature: https://github.com/jellyfin/jellyfin-androidtv/issues/5656

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/compat/StreamInfo.java
+++ b/app/src/main/java/org/jellyfin/androidtv/data/compat/StreamInfo.java
@@ -72,7 +72,7 @@ public class StreamInfo {
 
     public final org.jellyfin.sdk.model.api.SubtitleDeliveryMethod getSubtitleDeliveryMethod() {
         Integer subtitleStreamIndex = MediaSource.getDefaultSubtitleStreamIndex();
-        if (subtitleStreamIndex == null || subtitleStreamIndex == -1) return SubtitleDeliveryMethod.DROP;
+        if (subtitleStreamIndex == null || subtitleStreamIndex == -1 || MediaSource.getMediaStreams() == null) return SubtitleDeliveryMethod.DROP;
         return MediaSource.getMediaStreams().get(subtitleStreamIndex).getDeliveryMethod();
     }
 
@@ -96,7 +96,7 @@ public class StreamInfo {
 
     public final ArrayList<org.jellyfin.sdk.model.api.MediaStream> getSelectableStreams(MediaStreamType type) {
         ArrayList<org.jellyfin.sdk.model.api.MediaStream> list = new ArrayList<org.jellyfin.sdk.model.api.MediaStream>();
-
+        if (MediaSource.getMediaStreams() == null) return list;
         for (org.jellyfin.sdk.model.api.MediaStream stream : getMediaSource().getMediaStreams()) {
             if (type == stream.getType()) {
                 list.add(stream);
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackController.java
@@ -50,6 +50,7 @@ import java.time.Instant;
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
 import java.util.List;
+import java.util.Objects;
 
 import kotlin.Lazy;
 import timber.log.Timber;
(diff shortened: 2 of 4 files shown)


BlackstormFirst:
As for the code itself, it is not yet fully optimized, properly structured, or finalized. There may also be sections or fallbacks remaining that are never triggered.

I would appreciate feedback on the targeted features, specifically if you encounter issues with track selection or experience a crash with this version that wouldn't have occurred with the stable version (including directplay and transcode).

My tests are limited to Android Studio and Nvidia Shield TV Pro 2019.

BotBlake:
Hi there @BlackstormFirst,

Why did you close your already existing PRs #5644 and #5724?

Both of those were already on our review board and maintainers had started reviewing them. Closing them and replacing them with a new, significantly larger PR effectively throws away that review work and makes the changes much harder to follow.

I'm also confused by the preview URL you provided. We cannot reasonably review or test arbitrary builds distributed through an external file-sharing service. Testing and previews should be reproducible through GitHub, CI, and the changes in the PR itself.

Just skimming through the diff, this is also way out of scope for the issues it is trying to address.

This PR changes 481 lines across core playback code and combines multiple different fixes and features into one large rewrite. That is significantly more complex than necessary for the described changes and makes it very difficult to review the individual fixes for correctness or regressions.

Especially considering that #5724 already had review feedback about a possible regression, I don't think replacing the smaller PRs with an even broader rewrite is the right direction here.

Please keep the individual changes focused and scoped to the issues they are intended to solve. If the previous PRs need additional changes, those should generally be addressed there instead of replacing reviewed changes with an increasingly larger PR.

Kind regards

BlackstormFirst:
Hello, @BotBlake 

This was done for good reasons, even though I understand your arguments.
As indicated to the reviewers in the follow-up to the other two PRs:
One of the other PRs solves the issue but causes a regression even when combined with the changes from the second PR (which has since been largely superseded by this one).

Indeed, certain mechanisms related to index loading were revised to account for new elements and avoid duplication, fix a conceptual error, at the same time, some warnings regarding potential null pointers were addressed. I admit this approach is debatable and complicates the analysis of changes, but it was sometimes necessary within my implementation logic to stay as close as possible to the existing codebase.

Given the regressions introduced by PR https://github.com/jellyfin/jellyfin-androidtv/pull/5724, it made sense to provide a better solution, one that incorporates the benefits of the new features while also removing the release of erroneous or incomplete elements.

I should also point out that this PR was in "Draft" mode; it wasn't intended for review (as Niels had already noted) except by interested parties, though I appreciate you taking the time to explain your reasoning.

I also understand why an external link might be problematic, so I removed it. It would have been nice to discuss it, or simply have it removed from the conversation, given that the APK couldn't be uploaded (even when compressed), unlike what I’ve seen in other threads (though perhaps those linked to a different GitHub repository...).

Finally, it is a shame that this PR isn't being accepted in its entirety, especially on the day it would have been finalized, because, even as a draft, I believe it is already quite functional.

It will be difficult to propose certain fixes, let alone improvements, without touching, sometimes deeply, on existing mechanisms.

So, if I understand correctly... what you're recommending now, just when I’d almost finished standardizing and stabilize a working fix, is to break it all down again... in the right order... with minimal changes...
Well... that’s going to be a tall order...

BotBlake:
I am on my way to bed so the response is going to fall a bit short.

But it comes down to: 
If you get a PR review, you will have to Adress it inside that PR instead of closing the old one and opening an even bigger new one.

What I am asking for might be "a tall order", however this was entirely self caused. You have never been asked to close your original PRs, just to Adress the individual PR reviews.

It does not matter if this PR is a draft or not, it is already fundamentally going into the wrong direction, so I closed it.

Have a good day
Kind regards

PR #5759, opened 2026-08-16
Build release assets before publishing

nielsvanvelzen:
**Changes**

Update the publish workflow to allow building the release assets (apk/aab/version.txt) for draft releases. With this approach we can upload the AAB to Google/Amazon ahead of the release so the GitHub release can be delayed until the app stores have approved it. Giving a better experience for users that now often have to wait for days (if not weeks) for a release to become available.

**Code assistance**

Codex was heavily used to plan this change and we went over multiple variants to come to this solution.

**Issues**

### DIFF
--- a/.github/workflows/app-publish.yaml
+++ b/.github/workflows/app-publish.yaml
@@ -1,37 +1,55 @@
 name: App / Publish
 
 on:
-  push:
-    tags:
-      - v*
+  release:
+    types:
+      - created
+      - edited
+      - published
 
 jobs:
   publish:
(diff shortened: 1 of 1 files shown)


thornbill (approved):

nielsvanvelzen:
Reverted in c995d5aef8156d3e1da6c91b8ab1ebf0e08c5e76 (accidentally pushed to master instead of creating PR 🫠) - the change did not work as expected. Might revisit it in the future.

PR #5774, opened 2026-08-24
Add relative skip mode for d-pad seeking

### LINKED ISSUE #1185: Use left/right buttons to seek backward/forward

- When media is playing without any controls/menu on the screen, pressing left/right should seek backward/forward respectively
- When controls/menu is up and the focus is on the scrubber, then pressing left/right should seek backward/forward respectively. I don't think we can currently skip backward/forward when media is paused, so this would apply when media is playing and focus is on the scrubber.

Note: This is for FireTv remote. It has actual buttons to skip backward/forward. I'm referring to left/right buttons next to the middle/main button.

cwernert:
### Changes

**1: "Skip mode" setting**

- Within Settings > Playback > Advanced, a new "Skip mode" setting is available.
- Setting has two options: "Absolute" and "Relative", defaulting to the status quo "absolute" behaviour.
- When set to "absolute", the existing behaviour of stepping the seek bar cursor and confirming the seek is preserved (ie. users need to select an absolute timestamp on the seek bar and confirm it).
- When set to "relative", pressing the right or left buttons during playback (while the control overlay UI is hidden) skips immediately forward or backward (respectively) from the current timestamp and does not reveal the control overlay UI or require a confirmation press.
- The amount of time skipped is user-configurable via `skipBackLength` and `skipForwardLength`.
- Relative skipping uses the same pattern as the dedicated FF/RW buttons (including press accumulation).

**Note:** only the legacy player consumes this setting as the rewrite player's seekbar already seeks relatively when focused.

**2: Exposes skipBackLength slider**

- Previously the `skipBackLength` preference was not exposed in the UI, but now gets a slider just like the `skipForwardLength` preference.
- These sliders determine the skip amounts for the dedicated FF/RW buttons (as before), and now also for the left/right buttons when "Skip mode" is set to "Relative".

### Code Assistance

This PR has been implemented with Claude Code including: code generation, codebase exploration, and quality control and self-review. All changes have also been reviewed, understood, and tested by me on my Sony Bravia 8 (direct play and transcode).

### Issues

Fixes #1185
Related to #3033

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -12,6 +12,7 @@ import org.jellyfin.androidtv.preference.constant.ClockBehavior
 import org.jellyfin.androidtv.preference.constant.HEVCLevel
 import org.jellyfin.androidtv.preference.constant.NextUpBehavior
 import org.jellyfin.androidtv.preference.constant.RefreshRateSwitchingBehavior
+import org.jellyfin.androidtv.preference.constant.SkipMode
 import org.jellyfin.androidtv.preference.constant.StillWatchingBehavior
 import org.jellyfin.androidtv.preference.constant.WatchedIndicatorBehavior
 import org.jellyfin.androidtv.preference.constant.ZoomMode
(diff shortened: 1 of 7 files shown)


Pentaphon:
Honestly, the "relative" setting should become the default. It makes just as much sense on a remote as a touchscreen.

PR #5777, opened 2026-08-26
Fix NextUp row skipping episodes when season has a numbering gap

xanderphillips:
## Bug

On the series-detail page, the "Next Up" row (and the series-level Play button, which reads from the same source) silently skips unwatched episodes for shows whose current season has a numbering gap before its first available episode.

Example: a season with only episodes 10–30 present (1–9 missing/not yet acquired) showed `E10, E20, E21, ..., E30` in the row — silently dropping `E11`–`E19`.

## Root cause

`retrieveNextUpItems()` in `ItemRowAdapterHelper.kt` expanded a single `/Shows/NextUp` result into a full row via:

```kotlin
api.itemsApi.getItems(parentId = firstNextUp.seasonId, startIndex = firstNextUp.indexNumber)
```

`startIndex` is a pagination offset into the sorted result set, not a filter on the episode's `IndexNumber` field. The two only coincide when a season's episodes start at index 1 with no gaps.

## Fix

Switch to `tvShowsApi.getEpisodes(seriesId=..., startItemId=firstNextUp.id, isMissing=false, ...)`, which resolves by item ID (inclusive) instead of positional offset. This pattern already exists elsewhere in the codebase (`SdkPlaybackHelper.kt`) for the same "this episode and everything after it" purpose.

## Testing

Added a regression test reproducing the numbering-gap scenario. Verified the fix on-device.


nielsvanvelzen:

PR #5780, opened 2026-08-27
Add user-controlled HDR support reporting

mrworf:
**Changes**
Gives the user a menu option under Settings > Playback > Advanced > Media Codecs that lets them configure HDR to be automatically detected, always on, or always off. The reason is that NVIDIA Shield (and possibly other devices) may report HDR capability based on codecs rather than on the display device. This allows for easy override of this reporting. However, it will not lie about the codec capability. If the device lacks codec support, enabling it will not magically provide HDR support.

**Code assistance**
OpenAI Codex wrote the entire change using a custom skill designed for this purpose. It was based on my instructions and created a plan that I reviewed. The resulting debug and later release was verified on an NVIDIA Shield 2019. The final code was manually reviewed to avoid unnecessary dependencies or changes.

**Issues**


nielsvanvelzen:

PR #5781, opened 2026-08-27
Fix back navigation exiting the app when opened into a deep screen

### LINKED ISSUE #5010: Pressing back from an episode takes user to app launcher

### Describe the bug

1. Open episode page
2. Go to app launcher and launch another app
3. Come back to jellyfin, you should get back to the previously opened episode 
4. Pressing "Back" on the remote makes the app close, not back to the home screen

### Logs

_No response_

### Application version

0.19.0 beta 6

Google Play Store

### Device information

Chromecast 4K last generation 

### Android version

Android 14

### Jellyfin server version

10.10.7

### Other sources

_No response_

webcultist:
**Changes**

Pressing back closed the app instead of navigating to the home screen when the app was opened directly into a deeper screen, such as from a launcher tile, a deep link or a search result.

`NavigationRepositoryImpl.reset()` cleared `fragmentHistory` and navigated to the destination without pushing it onto the stack, while the view layer (`DestinationFragmentView`) does push it onto its own stack. The two stacks therefore diverged by exactly one level. With an empty history `canGoBack` was false, so the `BackHandler` in `AppNavigationHost` stayed disabled, the key press reached the activity and finished it.

`fragmentHistory` now also holds the visible destination, matching the view. `canGoBack` no longer asks whether the stack is empty but whether the current destination is the default one, and `goBack()` navigates to the default destination when there is no history left but the app is not on it. A separate `hasBackStack` was added because `MainActivity` used `canGoBack` to discard stale navigation state on a fresh activity, which under the new meaning would have sent every deep link straight back to home.

Six unit tests cover the deep entry point, unwinding levels above it, replace, and the unchanged behaviour at the root.

**Code assistance**

Substantial. The root cause was located by driving a physical device over ADB and comparing `dumpsys activity` output between navigation steps, which is what showed that the UI survives backgrounding while the navigation stack does not. An LLM was used for reading the navigation code, forming the hypothesis, writing the patch and the tests, and drafting this description. Every claim above is backed either by an observation recorded in https://github.com/jellyfin/jellyfin-androidtv/issues/5010#issuecomment-5421048326 or by the code itself, and the reasoning is laid out so it can be checked rather than taken on trust.

One limitation stated plainly: this passes test, lint and build in CI, but it has **not** been verified on a device yet. The box used for the original measurements went offline before the patch existed. If a maintainer would rather hold this until a device test is posted, that is fair, and I will add one.

**Issues**

Fixes #5010

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/MainActivity.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/MainActivity.kt
@@ -54,7 +54,9 @@ class MainActivity : FragmentActivity() {
 				else window.clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
 			}.launchIn(lifecycleScope)
 
-		if (savedInstanceState == null && navigationRepository.canGoBack) navigationRepository.reset(clearHistory = true)
+		// Drop leftover navigation from a previous run, but keep a single destination the app was
+		// opened into directly, such as a launcher tile or a deep link.
+		if (savedInstanceState == null && navigationRepository.hasBackStack) navigationRepository.reset(clearHistory = true)
 
 		navigationRepository.currentAction
 			.flowWithLifecycle(lifecycle, Lifecycle.State.STARTED)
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:

Issue #1232, opened 2021-10-27
Favourites on Home Screen


o0oKeegs:
### Describe the feature you'd like

Would it be possible to collate all Favourites from different folders (Movies, TV, Documentaries) to be displayed in a tab on the home screen rather than having to go into each Folder and select the Heart icon at the top? Cheers.

fmchris10:
The favourites function does seem a missing part of the jigsaw.

Ddgdhfjb:
Anything new on that topic?

pierrehusson23:
Taking a shot at this. #5668/#5669 got closed for using legacy leanback instead of Compose.

Approach: same pattern as the Search screen — Compose `MainToolbar` hosting a `RowsSupportFragment` via `AndroidFragment` interop, querying `IsFavoriteOrLikes` across item types, grouped by category. Doesn't depend on the full Compose grid from #5146.

Have a working local prototype, will open a draft PR once cleaned up.

Jaakurin:
Is anyone else willing to take a stab at this much-desired feature? The official client has been lacking a favorites page for a very long time and #5786 was closed for using AI. 🥺

PR #5786, opened 2026-08-29
Add Favorites screen to Android TV client

pierrehusson23:
**Changes**
Adds a dedicated Favorites screen, accessible via a new "Favorites" button in the main toolbar. Lists favorited items across all media types (movies, series, episodes, music, playlists, photos, collections...), grouped by category.

Follows the same Compose-shell + `AndroidFragment<RowsSupportFragment>` interop pattern used by the existing Search screen, rather than legacy leanback views (the reason #5668/#5669 were previously closed). Doesn't depend on the full Compose grid work tracked in #5146.

**Code assistance**
Built with substantial assistance from Claude Code: it explored the codebase to identify the Search screen as the closest reference pattern, generated the new files (repository/view model/delegate/fragment) based on that pattern, helped debug a leanback focus-navigation bug by porting the existing workaround from `SearchFragment`, and helped with local environment setup and manual testing on an emulator against a real server. This is my first open-source contribution and first time working with this stack.

**Issues**
Ref #1232

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/favorites/FavoritesFragmentDelegate.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/favorites/FavoritesFragmentDelegate.kt
@@ -1,0 +1,58 @@
+package org.jellyfin.androidtv.ui.favorites
+
+import android.content.Context
+import androidx.leanback.widget.HeaderItem
+import androidx.leanback.widget.ListRow
+import androidx.leanback.widget.OnItemViewClickedListener
+import androidx.leanback.widget.OnItemViewSelectedListener
+import androidx.leanback.widget.Row
+import org.jellyfin.androidtv.constant.QueryType
+import org.jellyfin.androidtv.data.service.BackgroundService
+import org.jellyfin.androidtv.ui.itemhandling.BaseRowItem
+import org.jellyfin.androidtv.ui.itemhandling.ItemLauncher
+import org.jellyfin.androidtv.ui.itemhandling.ItemRowAdapter
+import org.jellyfin.androidtv.ui.presentation.CardPresenter
+import org.jellyfin.androidtv.ui.presentation.CustomListRowPresenter
+import org.jellyfin.androidtv.ui.presentation.MutableObjectAdapter
+
+class FavoritesFragmentDelegate(
+	private val context: Context,
+	private val backgroundService: BackgroundService,
+	private val itemLauncher: ItemLauncher,
+) {
+	val rowsAdapter = MutableObjectAdapter<Row>(CustomListRowPresenter())
+
+	fun showResults(favoritesResultGroups: Collection<FavoritesResultGroup>) {
+		rowsAdapter.clear()
+		val adapters = mutableListOf<ItemRowAdapter>()
+		for ((labelRes, baseItems) in favoritesResultGroups) {
+			val adapter = ItemRowAdapter(
+				context,
+				baseItems.toList(),
+				CardPresenter(),
+				rowsAdapter,
+				QueryType.Search
+			).apply {
+				setRow(ListRow(HeaderItem(context.getString(labelRes)), this))
+			}
+			adapters.add(adapter)
+		}
+		for (adapter in adapters) adapter.Retrieve()
+	}
+
+	val onItemViewClickedListener = OnItemViewClickedListener { _, item, _, row ->
+		if (item !is BaseRowItem) return@OnItemViewClickedListener
+		row as ListRow
+		val adapter = row.adapter as ItemRowAdapter
+		itemLauncher.launch(item as BaseRowItem?, adapter, context)
+	}
+
+	val onItemViewSelectedListener = OnItemViewSelectedListener { _, item, _, _ ->
+		val baseItem = item?.let { (item as BaseRowItem).baseItem }
+		if (baseItem != null) {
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/jellyfin/androidtv/ui/favorites/FavoritesRepository.kt
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	override suspend fun getFavorites(
+		itemTypes: Collection<BaseItemKind>,
+	): Result<List<BaseItemDto>> = try {
+		val request = GetItemsRequest(
+			limit = QUERY_LIMIT,
+			imageTypeLimit = 1,
+			includeItemTypes = itemTypes,
+			filters = setOf(ItemFilter.IS_FAVORITE_OR_LIKES),
```
nielsvanvelzen:
```suggestion
			filters = setOf(ItemFilter.IS_FAVORITE),
```

Likes are deprecated, we shouldn't use them

app/src/main/java/org/jellyfin/androidtv/ui/favorites/FavoritesViewModel.kt
```diff
@@ -0,0 +1,48 @@
...
+
+	private val _favoritesResultsFlow = MutableStateFlow<Collection<FavoritesResultGroup>>(emptyList())
+	val favoritesResultsFlow = _favoritesResultsFlow.asStateFlow()
+
+	fun load() {
+		viewModelScope.launch {
+			_favoritesResultsFlow.value = groups.map { (stringRes, itemKinds) ->
+				async {
+					val result = favoritesRepository.getFavorites(itemKinds)
+					val items = result.getOrNull().orEmpty()
```
nielsvanvelzen: If you're never gonna use the Result just return an empty list directly

app/src/main/java/org/jellyfin/androidtv/ui/favorites/FavoritesViewModel.kt
```diff
@@ -0,0 +1,48 @@
...
+			R.string.lbl_songs to setOf(BaseItemKind.AUDIO),
+			R.string.photo_albums to setOf(BaseItemKind.PHOTO_ALBUM),
+			R.string.photos to setOf(BaseItemKind.PHOTO),
+			R.string.lbl_collections to setOf(BaseItemKind.BOX_SET),
+			R.string.lbl_people to setOf(BaseItemKind.PERSON),
+		)
+	}
+
+	private val _favoritesResultsFlow = MutableStateFlow<Collection<FavoritesResultGroup>>(emptyList())
+	val favoritesResultsFlow = _favoritesResultsFlow.asStateFlow()
```
nielsvanvelzen: I would argue SearchResultGroup could be re-used here (maybe rename it to something more generic)

app/src/main/java/org/jellyfin/androidtv/ui/favorites/FavoritesFragmentDelegate.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.jellyfin.androidtv.constant.QueryType
+import org.jellyfin.androidtv.data.service.BackgroundService
+import org.jellyfin.androidtv.ui.itemhandling.BaseRowItem
+import org.jellyfin.androidtv.ui.itemhandling.ItemLauncher
+import org.jellyfin.androidtv.ui.itemhandling.ItemRowAdapter
+import org.jellyfin.androidtv.ui.presentation.CardPresenter
+import org.jellyfin.androidtv.ui.presentation.CustomListRowPresenter
+import org.jellyfin.androidtv.ui.presentation.MutableObjectAdapter
+
+class FavoritesFragmentDelegate(
```
nielsvanvelzen, line 19: This seems like a copy of the way search is implemented, which is not up to current standards. The viewmodel is responsible for most of what is happening here

nielsvanvelzen (changes requested):
Initial (brief) code review. Aside from that I have not tested all different groups yet to see if they render properly. What I did notice is that all items are showing the heart icon because they're favorited. we should not do that as the entire page is just favorited items

pierrehusson23:
Pushed fixes for the concrete points:
- Switched to `ItemFilter.IS_FAVORITE`.
- `FavoritesRepository`/`FavoritesViewModel` no longer wrap results in `Result`, just return an empty list on failure.
- Merged `SearchResultGroup`/`FavoritesResultGroup` into a single `LabeledItemGroup` in `ui.itemhandling`.
- Favorite heart indicator is now hidden on cards in the Favorites screen.

Also found and fixed an actual copy-paste bug: the delegate was building `ItemRowAdapter` with `QueryType.Search`, which doesn't apply here (there's a dedicated static-items constructor for a pre-fetched list, now used instead).

On the broader point about the delegate being "not up to current standards" compared to search: I looked around and couldn't find any screen in the codebase that renders these item rows without leanback (`RowsSupportFragment`/`ItemRowAdapter`/a `*FragmentDelegate`) — Home and Search both still use this same structure. Could you clarify what you'd like to see instead here? Happy to move more logic into the ViewModel if you can point to an example, but I don't want to invent a new pattern for this screen alone.

nielsvanvelzen:
Unfortunately we won't be accepting it as it violates our [LLM/"AI" Development Policy](https://jellyfin.org/docs/general/contributing/llm-policies).

PR #5343, opened 2026-01-18
Recreate LegacyImageCardView with compose

nielsvanvelzen:
**Changes**

- Add ItemPreview
  - A composable that displays an ItemCard with additional metadata below (currently title & subtitle)
- Add ItemCard
  - A composable that displays a card for an item with optional overlay
- Add ItemCardBaseItemOverlay
  - A composable that displays an overlay on ItemCard for a baseitem
  - Show live progress bar for currently playing media (audio)
  - Favorite state on top-left
  - Recording state on top-left
  - Watched state on top-right
  - Progress on bottom
- Rewrite CardPresenter to use ItemPreview/ItemCard
  - Includes poetry to adapt existing BaseRowItem mess to the new design
- Communicate JellyfinImage via ChapterItemInfo
  - Add `BaseItemDto.chapterImages` helper
- Update BaseRowItem to return JellyfinImage instead of image URL
- Replace UserViewCardPresenter with CardPresenter
- Remove old (now unused) functions from ImageHelper
- Remove old (now unused) LegacyImageCardView implementation
- Remove old (now unused) functions from BaseRowItem

**Screenshots**

The design is currently very similar to the previous design, with some smaller changes (removed banner edges, removed rating, slight spacing/color tweaks). To see the difference open both of the following images in a new tab and switch between the tabs.

| New | Old |
| --- | --- |
|  |  |

**Code assistance**

**Issues**

PR #5789, opened 2026-08-31
Improve D-pad navigation performance

brandonp2412:
**Changes**

Improve D-pad navigation responsiveness across the home and browse screens by reducing work performed on focus changes and row transitions.

- Retain and pre-layout home rows so moving vertically does not synchronously create/measure a full row of Compose cards during the D-pad frame.
- Keep card focus state local to the focus-dependent overlay/marquee content instead of recomposing the full card/image subtree.
- Only subscribe to live playback progress for the focused card; unfocused cards use their static server-provided progress.
- Render card artwork directly with Compose/Coil while preserving BlurHash placeholders and crossfade behavior, avoiding the previous Compose/AndroidView image bridge.
- Debounce expensive backdrop/metadata updates until focus has settled while leaving pagination immediate.
- Restore the adaptive BrowseGrid chunk size instead of overriding it with a fixed value.
- Size backdrop image requests to the display rather than decoding unconstrained images.

**Performance testing**

Tested on a physical Google TV Streamer at 1080p/60 Hz using repeated D-pad navigation and Android `gfxinfo`/atrace measurements.

For real home-screen up/down navigation, the original implementation measured 7.38% janky frames with p95 77 ms and p99 129 ms. After the row retention/pre-layout changes, a warm trace measured 2.87% janky frames with p95 38 ms and p99 48 ms. The largest `RV OnLayout` slice dropped from 90.6 ms to 7.6 ms.

A separate media-card workload exercising actual `BaseItemDto` cards, unique network images, BlurHash placeholders, overlays and playback progress produced the following five-pair mixed-navigation averages:

| Metric | Baseline | Improved |
| --- | ---: | ---: |
| Janky frames | 10.44% | 3.58% |
| p90 | 46.4 ms | 29.6 ms |
| p95 | 58.6 ms | 35.2 ms |
| p99 | 98.6 ms | 57.0 ms |
| Deadline misses | 138.8 | 48.0 |

Focus-state isolation was also traced independently. Compared with the row-cache-only version, `Recomposer:recompose` time fell from 678 ms to 473 ms over the same navigation workload, with the worst recomposition falling from 22.1 ms to 11.5 ms. Total animation work fell from 1434 ms to 1001 ms and traversal work from 2719 ms to 2119 ms.

The final behavior was also manually tested on the same Google TV Streamer with normal Jellyfin navigation.

**Validation**

- `./gradlew test lint :app:assembleRelease` passes locally.
- `git diff --check` passes.
- Local `./gradlew detekt` currently aborts with `26.0.2.1`; an untouched checkout of current upstream `master` fails identically in the same environment, so this is not introduced by these changes.

**Related work**

I checked existing open and historical PRs before opening this. #5543 and #5544 improve home startup/API loading, #4927 and #5719 address focus correctness/restoration, #5146 is a separate Compose BrowseGrid rewrite, and #5343 introduced the Compose card presenter. I could not find an existing PR using the row caching/offscreen layout or focus-recomposition approach in this change.

**Code assistance**

ChatGPT was used to inspect the code paths, create and refine the implementation, build benchmark harnesses, drive repeated tests on the physical device, and analyze Android frame traces. Changes were iteratively retained or reverted based on measured results; several initially proposed optimizations were discarded after benchmarks showed regressions.

**Issues**

None.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/data/service/BackgroundService.kt
@@ -108,9 +108,13 @@ class BackgroundService(
 		// Cancel current loading job
 		loadBackgroundsJob?.cancel()
 		loadBackgroundsJob = scope.launch(Dispatchers.IO) {
+			val displayMetrics = context.resources.displayMetrics
 			_backgrounds = backdropUrls.mapNotNull { url ->
 				imageLoader.execute(
-					request = ImageRequest.Builder(context).data(url).build()
+					request = ImageRequest.Builder(context)
+						.data(url)
+						.size(displayMetrics.widthPixels, displayMetrics.heightPixels)
+						.build()
 				).image?.toBitmap()?.asImageBitmap()
 			}
 
--- a/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseFolderFragment.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseFolderFragment.kt
@@ -12,6 +12,8 @@ import androidx.leanback.widget.RowPresenter
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.lifecycleScope
 import androidx.lifecycle.repeatOnLifecycle
+import kotlinx.coroutines.Job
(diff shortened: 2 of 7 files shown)


nielsvanvelzen:

PR #5796, opened 2026-09-04
Add documentary zoom and pan feature for photo playback

hraabis:
**Changes**
Add documentary style photo presentation feature. During photo presentations randomly zoom and pan photos, or zoom and scroll panorama photos. If the presentation delay is less than 8 seconds, use simple animations of zoom in, zoom out, or pan from zoom 1  to zoom 2. If presentation delay is 8 seconds or greater, use 2 phase animations to zoom in or zoom out, if the aspect ratio is a panorama zoom and scroll the photo in a 3 phase animation. 
 
In new file AsyncImageViewHelper.kt, add AsyncImageView extension function photoAnimateWithZoomAndPan. In PhotoPlayerScreen.kt provide parameters to PhotoPlayerContent, which provides parameters to AsyncImage, which calls the AsyncImageView animateWithZoomAndPan function. In ItemRepository.kt add Width and Height to itemFields to ensure those values are retrieved from the server. In SettingsPlaybackPhotoPlayerScreen.kt add checkbox control to enable animate photos. 

**Code assistance**

No code assistance used in creation of this pull request.

**Issues**

Adds new feature not identified by an issue.

**Screenshot**
Screenshot_20260908-144812

**Video**
[https://drive.google.com/file/d/1DeMcAV5YiVSJys1BtIBpSTcz0T2BoL7T/view?usp=sharing](https://drive.google.com/file/d/1DeMcAV5YiVSJys1BtIBpSTcz0T2BoL7T/view?usp=sharing)


nielsvanvelzen (changes requested):
We'll first need to change the PhotoPlayerContent composable to rely on a compose-based render flow instead of the current AsyncImage that is backed by an AndroidView. That way the animations can be fully done via compose.

Additionally this has too many settings for a single feature. Just the checkbox is all we should have.

hraabis:
> We'll first need to change the PhotoPlayerContent composable to rely on a compose-based render flow instead of the current AsyncImage that is backed by an AndroidView. That way the animations can be fully done via compose.

Being new to Android development, can you give clarification on how to implement it? Changing render flows may need to be its own pull request. 

I see the Composable keyword used in files: PhotoPlayerScreen, PhotoPlayerContent, and AsyncImage. It is not used at AsyncImageView created by AsyncImage with AndroidView's factory function. 

Would the required change be implementing AsyncImageView as Composable and calling it directly in AsyncImage, instead of using the AndroidView factory function to create AsyncImageView?

> Additionally this has too many settings for a single feature. Just the checkbox is all we should have.

 The refactoring of this code links the panning limit to the random zoom selected to prevent a small zoom and large pan from showing lots of blank space on screen.

The effect strength can be hard coded to remove the 2 slider settings for Pan and Zoom.

hraabis:
> Additionally this has too many settings for a single feature. Just the checkbox is all we should have.

Commit [37e7e45](https://github.com/jellyfin/jellyfin-androidtv/pull/5796/commits/37e7e458ff403366309a79987b57a2c8a4f3fa27) Removes UserPreferences for pan and zoom strength and removes sliders from settings page. Pull request description updated with new screenshot and video of feature at this time.

nielsvanvelzen:
> > We'll first need to change the PhotoPlayerContent composable to rely on a compose-based render flow instead of the current AsyncImage that is backed by an AndroidView. That way the animations can be fully done via compose.
> 
> Being new to Android development, can you give clarification on how to implement it? Changing render flows may need to be its own pull request. 
>
> I see the Composable keyword used in files: PhotoPlayerScreen, PhotoPlayerContent, and AsyncImage. It is not used at AsyncImageView created by AsyncImage with AndroidView's factory function. 
> 
> Would the required change be implementing AsyncImageView as Composable and calling it directly in AsyncImage, instead of using the AndroidView factory function to create AsyncImageView?

The app is being migrated to the newer compose UI framework. Compose has excellent support for animations, so if we want to introduce animation to the photo player it will have to use compose first to avoid the hacky code you have now to do the zoom/pan.
The refactor for that should be in a separate PR from this enhancement though.

PR #5799, opened 2026-09-07
Add next up date cutoff setting

### LINKED ISSUE #2406: Hide shows from next up if they haven't been watched in X days

leonida92:
**Changes**
Adds a user preference under Home settings to filter the home Next Up row using the server's built-in `nextUpDateCutoff` parameter. Users can select from 7, 14, 30, 60, 90, 180, or 365 days, or disable the cutoff.
**Code assistance**
Code drafted and adapted to upstream Compose settings architecture with assistance from an LLM coding assistant; verified with Detekt and unit tests.
**Issues**
Fixes #2406 #4772

image

image

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -196,6 +196,12 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		var seriesThumbnailsEnabled = booleanPreference("pref_enable_series_thumbnails", true)
 
 		/**
+		 * Maximum days since last watch to show an item in next up.
+		 * Set to 0 to disable (no cutoff).
+		 */
+		var maxDaysInNextUp = intPreference("pref_max_days_in_next_up", 0)
+
+		/**
 		 * Subtitles foreground color
 		 */
 		var subtitlesBackgroundColor = longPreference("subtitles_background_color", 0x00FFFFFF)
--- a/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentHelper.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentHelper.kt
@@ -5,6 +5,7 @@ import org.jellyfin.androidtv.R
 import org.jellyfin.androidtv.auth.repository.UserRepository
 import org.jellyfin.androidtv.constant.ChangeTriggerType
 import org.jellyfin.androidtv.data.repository.ItemRepository
+import org.jellyfin.androidtv.preference.UserPreferences
 import org.jellyfin.androidtv.ui.browsing.BrowseRowDef
 import org.jellyfin.sdk.model.api.BaseItemDto
 import org.jellyfin.sdk.model.api.BaseItemKind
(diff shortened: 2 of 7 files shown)


app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
```diff
@@ -195,6 +195,12 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		 */
 		var seriesThumbnailsEnabled = booleanPreference("pref_enable_series_thumbnails", true)
 
+		/**
+		 * Maximum days since last watch to show an item in next up.
+		 * Set to 0 to disable (no cutoff).
+		 */
+		var maxDaysInNextUp = intPreference("pref_max_days_in_next_up", 0)
```
nielsvanvelzen:
```suggestion
		var homeNextUpMaxDays = intPreference("home_next_up_max_days", 0)
```

app/src/main/java/org/jellyfin/androidtv/ui/home/HomeFragmentHelper.kt
```diff
@@ -54,11 +57,19 @@ class HomeFragmentHelper(
 	}
 
 	fun loadNextUp(): HomeFragmentRow {
+		val maxDays = userPreferences[UserPreferences.maxDaysInNextUp]
+		val nextUpDateCutoff = if (maxDays > 0) {
+			LocalDateTime.now().minusDays(maxDays.toLong())
+		} else {
+			null
+		}
```
nielsvanvelzen: could be written something like
```suggestion
		val nextUpDateCutoff = maxDays.takeIf { it > 0 }?.let { LocalDateTime.now().minusDays(it.toLong()) }
```

app/src/main/res/values/strings.xml
```diff
@@ -431,6 +431,8 @@
     <string name="home_section_livetv">Live TV</string>
     <string name="home_section_none">None</string>
     <string name="home_section_i">Home section %1$d</string>
+    <string name="pref_max_days_in_next_up">Next up date cutoff</string>
+    <string name="pref_max_days_in_next_up_disabled">Disabled</string>
```
nielsvanvelzen: new strings should not be prefixed with `pref_`. I'd also rename them to something like `home_next_up_max_days` & `home_next_up_max_days_disabled`

nielsvanvelzen (changes requested):

leonida92:
Thanks for the review and suggestions @nielsvanvelzen!
Updated string resource names to drop the pref_ prefix, updated the preference property name to homeNextUpMaxDays, simplified the date calculation with takeIf, and resolved the Detekt warnings.

nielsvanvelzen (approved):

nielsvanvelzen:
Thanks for the PR!

PR #5800, opened 2026-09-07
feat(playback): add skip button duration preference

leonida92:
**Changes**
Adds a user preference under Playback > Media segments to configure the auto-hide duration of the skip button. Users can select from 3, 5, 8, 10, 15, 20, or 30 seconds, or disable auto-hiding (until segment ends).

**Code assistance**
Code drafted and adapted to upstream Compose settings architecture with assistance from an LLM coding assistant; verified with Detekt and unit tests.

image

skip_duration_options_screenshot

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/preference/UserPreferences.kt
@@ -262,6 +262,12 @@ class UserPreferences(context: Context) : SharedPreferenceStore(
 		)
 
 		/**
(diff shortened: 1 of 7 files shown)


johnpc:
I am in favor of this. I attempted to add a similar change (https://github.com/jellyfin/jellyfin-androidtv/pull/4519) a _year and half_ ago and it was closed as "invalid" since there was a playback rewrite going on from 2021 (https://github.com/jellyfin/jellyfin-androidtv/issues/1057).

My understanding is that the playback rewrite is now stalled and only available in dev builds. So I don't see why this small improvement can't be included in the meantime.

PR #5801, opened 2026-09-08
Fix Live TV guide crash when channel loading fails

andrew749:
**Changes**

When a Live TV channel-list request fails before any channels have been cached, `loadAllChannels()` still invokes its completion callback, but `getAllChannels()` returns null. Both the standalone guide and playback guide then call `isEmpty()` on that value and can crash with a `NullPointerException`. Return an empty list from the shared getter when no cache exists so the guides use their existing empty-state handling; a previously loaded channel list remains available after a failed refresh.

Validation: reviewed all three getter call sites and passed `git diff --check`. Attempted `./gradlew test` with Java 21, but Gradle stopped before running tests because the Android SDK is unavailable. Device regression testing has not been performed.

**Code assistance**

OpenAI Codex traced the failure path, authored the patch, and prepared this description.

**Issues**

No linked issue. Addresses a channel-list timeout followed by a null dereference in `LiveTvGuideFragment.load()`.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/LiveProgramDetailPopup.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/LiveProgramDetailPopup.java
@@ -257,9 +257,11 @@ public class LiveProgramDetailPopup {
 
     public android.widget.ImageButton createFavoriteButton() {
         BaseItemDto channel = TvManager.getChannel(TvManager.getAllChannelsIndex(mProgram.getChannelId()));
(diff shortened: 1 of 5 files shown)


nielsvanvelzen:
There's also the getChannel method that could cause an NPE. I guess that one would need fixing as well?

andrew749:
> There's also the getChannel method that could cause an NPE. I guess that one would need fixing as well?

good call, updated!

nielsvanvelzen (approved):

PR #5802, opened 2026-09-08
Merge cast & crew cards for people with multiple crew roles

shbernal:
**Changes**

The server sends one `People` entry per credit, so a person credited for several crew jobs on the same item gets one card each — a director who also produced appears twice in the cast & crew row, same photo, different subtitle. This merges the director, writer and producer credits belonging to the same person into a single card listing every role, e.g. "Director / Producer". Actors are deliberately left alone, so someone playing two characters still gets a card per character.

It mirrors what jellyfin-web has done since https://github.com/jellyfin/jellyfin-web/pull/8209 (shipped in 12.0), keeping the same merged kinds — director, writer, producer, which is also what the server keeps in `TmdbUtils.WantedCrewKinds` — and the same `" / "` separator.

Merging lives in a new `Collection<BaseItemPerson>.mergeCrewRoles()` extension, applied to the three cast & crew rows in `FullDetailsFragment` (movie, trailer, series). Episode guest stars are untouched, as guest stars are never merged. Unit tests cover the merge, the actor exemption, ordering, repeated jobs and missing job names.

One thing worth a maintainer's opinion: with three or more roles the subtitle truncates at the card width, e.g. Edgar Wright on *Scott Pilgrim vs. the World* renders as "Director / Screenpl…". Two roles fit comfortably. I kept the `" / "` join to stay consistent with the web client, but I'm happy to change the separator or fall back to the person kinds if you'd prefer.

**Code assistance**

Written with Claude Code using the Opus 5 model. It audited which clients were missing the behavior, wrote the extension, the call-site changes and the unit tests, ran the build, unit tests and detekt, and verified the result on an Android TV emulator (Android 16, x86_64) against a Jellyfin 12.0 server. I reviewed the change myself.

I only reproduced the duplicate cards on Android TV, but the same gap exists in Swiftfin, jellyfin-roku and jellyfin-vue, so I'm opening the equivalent PR against each of those as well rather than leaving the fix partial across clients.

**Issues**

N/A — no existing issue. Follows up on https://github.com/jellyfin/jellyfin-web/pull/8209.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/itemdetail/FullDetailsFragment.java
@@ -78,6 +78,7 @@ import org.jellyfin.androidtv.util.Utils;
 import org.jellyfin.androidtv.util.apiclient.BaseItemUtils;
 import org.jellyfin.androidtv.util.apiclient.Response;
 import org.jellyfin.androidtv.util.sdk.BaseItemExtensionsKt;
+import org.jellyfin.androidtv.util.sdk.BaseItemPersonExtensionsKt;
 import org.jellyfin.androidtv.util.sdk.TrailerUtils;
 import org.jellyfin.androidtv.util.sdk.compat.JavaCompat;
 import org.jellyfin.sdk.model.api.BaseItemDto;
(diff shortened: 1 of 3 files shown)


nielsvanvelzen (changes requested):
This is not something the client should be responsible for. The server needs to make API changes to communicate multiple roles for a single person before we'll adapt this change.

PR #5817, opened 2026-09-12
Explicitly request the negotiated HLS segment container

mvletter:
The device profile declares two HLS transcoding profiles split by audio codec: MPEG-TS for AAC/AC3/EAC3, and fMP4 for codecs MPEG-TS can't carry, such as DTS (see `deviceProfile.kt`). The server negotiates which one applies and reports it back as `MediaSourceInfo.TranscodingContainer`, but doesn't apply that same choice when building the HLS master playlist/segment URL it returns in `MediaSourceInfo.TranscodingUrl`. Without an explicit `segmentContainer` query parameter, that endpoint falls back to its own `.ts` default regardless of what was negotiated. In practice this meant DTS sources requiring any HLS remux got silently transcoded to AAC instead of getting the passthrough their own declared profile allows.

**Container alone is not sufficient.** Without an explicit `audioCodec` parameter either, the server infers a target audio codec from the segment request's own file extension (`.mp4` -> `aac`), which is meaningless for a video segment (the extension there is the container, not the audio format) and forces a transcode even once the container is correct — verified live, `segmentContainer=mp4` alone still produced AAC. So this PR sends both:

- `segmentContainer`, matching `MediaSourceInfo.TranscodingContainer` (`PlaybackManager.kt`, `segmentContainerQueryParameters()`).
- `audioCodec`, the ts-incompatible subset of `hlsFmp4AudioCodecs` for the fMP4 case (elevated from `private` to `internal` in `deviceProfile.kt` so both places share one source of truth), or `hlsMpegTsAudioCodecs` as-is for the ts case. Only the ts-incompatible subset is sent for fMP4: that's exactly why fMP4 was negotiated in the first place, and the server's own `audioCodec` parameter has a hard 40-character limit (regex-enforced) that a longer list could exceed — verified live for the previous, longer version of this list, rejected with a 400.

**This PR and the companion server-side PR are both required together**, not independently: this app-side fix has no effect unless the server (jellyfin/jellyfin#18018) stops defaulting the HLS segment container to `.ts` regardless of audio codec; and the server fix has no effect for this app unless it explicitly declares `segmentContainer`/`audioCodec`, since the raw HLS segment endpoints never consult the device-profile negotiation the way `PlaybackInfo` does.

## Update after testing this live on real Chromecast with Google TV + Denon AVR hardware

The segmentContainer + audioCodec forwarding fix above is still right and still needed. But testing surfaced a problem with one entry that was already in `hlsFmp4AudioCodecs` before this PR touched it: `TRUEHD`.

Playback runs fine and the AVR shows "Dolby Atmos" on screen, but there's no actual sound. The audio bytes are correct (checked with ffprobe and a manual decode), so this isn't a muxing bug in Jellyfin. It's ExoPlayer/Media3: `FragmentedMp4Extractor` doesn't use the `TrueHdSampleRechunker` that the plain `Mp4Extractor` uses, and without it TrueHD in a fragmented MP4 throws `AudioSink$UnexpectedDiscontinuityException`. That's an open, unfixed issue on Google's side: [androidx/media#1519](https://github.com/androidx/media/issues/1519). A Media3 maintainer already said they're unlikely to work on it soon. The non-fragmented version of the same file plays fine per that issue, so this is about fragmentation (which HLS requires), not TrueHD itself.

So this PR now removes `TRUEHD` from `hlsFmp4AudioCodecs`. Without it, TrueHD sources fall back to AAC, which works. With it, they'd go silent on real hardware, which is worse than what we started with.

DTS stays in the list. We found no report of the same issue for DTS, but we also haven't verified it actually works here, flagging that rather than assuming either way.

## Test plan

- Added unit tests (`PlaybackManagerSegmentContainerTests`) showing red before this change (capability doesn't compile) and green after.
- Full `:app:testDebugUnitTest` (95 tests) passes with no regressions.
- `:app:assembleDebug` succeeds and produces a working APK.
- Confirmed live on the actual hardware: fMP4+TrueHD produces byte-correct output (ffprobe-verified) but no audible sound. Matches androidx/media#1519 exactly.

### DIFF
--- a/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackManager.kt
+++ b/app/src/main/java/org/jellyfin/androidtv/ui/playback/PlaybackManager.kt
@@ -5,10 +5,13 @@ import androidx.lifecycle.lifecycleScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import org.jellyfin.androidtv.constant.Codec
 import org.jellyfin.androidtv.data.compat.PlaybackException
 import org.jellyfin.androidtv.data.compat.StreamInfo
 import org.jellyfin.androidtv.data.compat.VideoOptions
 import org.jellyfin.androidtv.util.apiclient.Response
+import org.jellyfin.androidtv.util.profile.hlsFmp4AudioCodecs
+import org.jellyfin.androidtv.util.profile.hlsMpegTsAudioCodecs
 import org.jellyfin.sdk.api.client.ApiClient
 import org.jellyfin.sdk.api.client.extensions.hlsSegmentApi
 import org.jellyfin.sdk.api.client.extensions.mediaInfoApi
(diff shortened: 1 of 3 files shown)


nielsvanvelzen:
We won't be accepting this contribution as it violates our [LLM/"AI" Development Policy](https://jellyfin.org/docs/general/contributing/llm-policies).

mvletter:
I did write most of the code, I did not write the PR text. I did spend a lot of time writing the new PR's but am using AI to get my typos out (dyslexia). Sorry for the extra work I caused.

PR #5818, opened 2026-09-13
Don't declare TrueHD under the fMP4 HLS profile

mvletter:
The fMP4 HLS profile lists TrueHD, but ExoPlayer cannot play TrueHD in a fragmented MP4. The fragmented extractor doesn't group TrueHD samples the way the non-fragmented one does, so the audio clock runs sixteen times too fast and the sink throws a discontinuity. That's androidx/media#1519, open since July 2024.

Today the declaration is harmless, because the server falls back to an MPEG-TS segment container that can't carry TrueHD either and quietly re-encodes to AAC instead. jellyfin/jellyfin#18018 fixes that fallback, and the moment it lands this line turns a silent downgrade into no audio at all. I ran into that myself while debugging something else on my own server: picture fine, amplifier lit up with Dolby Atmos on the display, and no sound at all.

So take it out until Media3 ships the fix. I've submitted that as androidx/media#3411; once it's in a release this line can come back.


nielsvanvelzen:
We won't be accepting this contribution as it violates our [LLM/"AI" Development Policy](https://jellyfin.org/docs/general/contributing/llm-policies).

mvletter:
Niels, stop saying this is AI haha. I wrote a longer post on what I'm trying to do here:https://markvletter.com/blog/dolby-atmos-truehd-samsung-tv/

I'm not an AI and I've been trying to get TrueHD to work (and failed) but did find a lot of real issues in the code and more importantly I got decent surround sound out of the client instead of stereo from a 7.1 file send from the Jellyfin server. It runs perfectly at home now but it's such a waste to not fix the bugs and improve the code.

mvletter:
This is where I'm at and I would love you advise because I'm afraid to submit any more PRs haha. https://github.com/jellyfin/jellyfin-androidtv/issues/5819

nielsvanvelzen:
Your other contributions use clearly AI generated descriptions, and since you're omitting our (required) pull request templates and not disclosing any AI usage here I can only assume this pull request is also AI generated.

I already linked our policy on LLM usage but in short; It's not disallowed to use LLM but you MUST follow our templates, disclose the LLM usage and own the changes (e.g. you are responsible for it).
If you can follow these guidelines and update the PR you can reopen/recreate it.

mvletter:
https://github.com/jellyfin/jellyfin-androidtv/pull/5821

Issue #5819, opened 2026-09-13
Device profile offers surround codecs the device can't deliver, which ends up as stereo


mvletter:
I spent a weekend trying to get TrueHD from my server to my amplifier, which turned out to be impossible on my hardware, and on the way there I found something worth reporting. Mind you this is one home setup, and the Jellyfin code is new to me as is this field.

The setup: a Chromecast with Google TV in a Samsung TV, with a Denon behind it over eARC. Playing a TrueHD 7.1 file, the app asks for direct play, the app decodes it locally, and the only active Android output at that moment is stereo PCM.

I would like it to be more than PCM.

Why is it stereo?

* `deviceProfile.kt` hands over `supportedAudioCodecs` as a static list.
* Video goes through `MediaCodecCapabilitiesTest`, audio isn't checked against the device at all.
* My Chromecast advertises a direct 5.1 PCM profile and eight channels on a chain that delivers two.

What I've been running at home is a profile that asks what the device can really pass through and stops offering the rest, but only when it passes the compressed formats and no lossless ones, because that seems to mean there's a TV in the middle. Here that turns the fake surround into real 5.1 the amplifier decodes itself. But again, that's just in this one living room.

I have this working on a branch with tests, but I don't want to send a PR for a design decision that is this big.

* Do you want this at all, or is the current behaviour deliberate for a reason I'm not seeing?
* Should this be automatic, a setting or both? (My worry with a setting is that the people who have this problem don't know it exists.)
* How far should the capability probe go? Testing per codec, per sample rate and per channel count takes about 110 ms on my Chromecast, and the profile is rebuilt on every `getPlaybackInfo` call in `JellyfinMediaStreamResolver`, so that's 110 ms on every playback start, and again whenever the stream is re-requested. That feels heavy, and if it was my code I would cache it and invalidate on audio route changes through `AudioCapabilitiesReceiver`. But that requires more code.

Let me know what would be useful. I'm ok with doing a WIP in code and somebody else who knows the code better finalizing it.

mvletter:
Since Android 13 there is AudioManager.getDirectProfilesForAttributes which seems applicable and would solve most of this. But I would still do a backwards compatible cache for older Android versions.

mvletter:
I don't think I would build the cache. All you need to know is how many channels it can carry, so that's 5 calls instead of 210.

mvletter:
[...]

PR #5822, opened 2026-09-14
Offer only the surround audio the device can deliver

mvletter:
**Changes**

### What happens now

The device profile offers every surround codec unconditionally. The server sees TrueHD in the list, hands over the original track, and the app decodes it to PCM. That is fine as long as everything behind the device carries multichannel PCM, and on a streaming box plugged into a television it does not: the television downmixes the PCM it is given, while it forwards a bitstream untouched. A 7.1 track arrives at the amplifier as stereo.

### What the client can tell us

Android can say which formats the current output accepts as a bitstream. media3 already asks that question for its own playback path, but the device profile never uses the answer, so the server keeps being told about codecs this chain cannot deliver.

### What this changes

The profile now asks which surround formats pass through and leaves out what does not, so the server transcodes to something that survives the trip instead of sending a track that gets flattened. The transcode targets are ordered so the first one is a codec that arrives intact, AC-3 ahead of E-AC-3 because ffmpeg's E-AC-3 encoder is the weaker of the two.

This only kicks in when compressed formats pass and lossless does not. Lossless is left alone.

### What I do not know

The profile sees the client, not the chain behind it. Passthrough tells you what the current output accepts and nothing about what a television further along does to multichannel PCM, so the rule above is an assumption, not a measurement. It is a good automatic default, but the honest shape for this is a user setting with automatic as its default value. I would rather ask than decide that for you: is that something you would want, and would you want it in this PR or a follow-up?

### Before and after

Measured on a Chromecast with Google TV into a television into a Denon amplifier.

Before: the app asks for TrueHD, the only active Android output is stereo PCM, and the amplifier reports PCM with its own upmixer filling the other channels.

After: the server transcodes to AC-3 5.1, the box opens an `AUDIO_FORMAT_AC3` output on six channels, and the amplifier decodes it itself and reports a 5.1 signal at 48 kHz.

**Code assistance**

Used Claude Code to find this in the code and Codex to review the findings, since the codebase and the field are new to me. The debugging and measuring I did myself.

**Issues**

#5819


mvletter:
https://github.com/jellyfin/jellyfin-androidtv/pull/5822 found this while working on it. Very small one. 

The research and reasoning behind it can be found here: https://github.com/jellyfin/jellyfin-androidtv/issues/5819

mvletter:
I was pondering on this a little more. I think this should be a user setting. The reason is: you can't always "see" what is best.

Take a Chromecast wired straight into the receiver. Passthrough says no TrueHD, but that no is send by the Chromecast, not the receiver, which takes TrueHD fine. So we drop it and the server sends AC-3 at 640, where leaving it would have given the receiver multichannel LPCM.

Issue #3230, opened 2023-12-15
Display Timeout Never Occurs


rocket4321:
### Describe the bug

Previously to jellyfin-androidtv.0.16 release, the app display would timeout after a long period of no user interaction. This allowed a TV to auto-turn off when no user input was received and the app discontinued any output video.

However, with 0.16, the display will never alter and will remain on indefinitely with no screensaver option enabled. With the screensaver, random images from the library are displayed, but the display still remains operational indefinitely.

Restore or allow a user option for display timeout to restore the display timeout functionality after no user interaction for an extended time period.

### Logs

N/A

### Application version

0.16.0

### Where did you install the app from?

Google Play

### Device information

NVidia Sheild

### Android version

Current

### Jellyfin server version

latest

nielsvanvelzen:
It is not possible to have the screen dim automatically when using the in-app screensaver so you need to turn it off in the app preferences.

rocket4321:
Yes, the screen does not timeout with the scrrensaver off (or on) in 0.16

nielsvanvelzen:
Does that also happen in other apps? Have you tried re-opening the Jellyfin app?

rocket4321:
I reboot the device, and now a timeout occured after about 15 min. I don't know why it never triggered all last night. Thanks so much for support, time, and all your excellent work. Closing issue.

rocket4321:
This issue still occurs, and the display often will not auto dim on my device. If I exit the jellyfin app, the display will properly timeout. Otherwise, after playing a selected video in jellyfin, the display will remain on indefinitely and never turn off... unless I exit the jellyfin app.

rocket4321:
Resolved in 16.1

PR #5823, opened 2026-09-15
Add option to let the device sleep after the in-app screensaver

MrGcGamer:
While the in-app screensaver is enabled, the app sets `FLAG_KEEP_SCREEN_ON` for as long as it is in the foreground. That means the device's display timeout and sleep timer never fire, even hours after the last input. This PR adds a setting under Screensaver, "Allow device to sleep after", with the values Immediately, 5 min, 15 min, 30 min, 1 h, 2 h, 4 h and Never. Once the screensaver has been visible for that long, the app stops holding the screen on and the system's own timeout takes over. The default is Never, which is the current behaviour.

It does not make the screensaver dim or hide itself: the screensaver stays on screen until the system turns the display off. Any interaction resets the timer, as it does today. Playback locks still keep the screen on (`(inAppEnabled && !sleepAllowed) || locks > 0`).

**Implementation notes:**

The timer starts when the screensaver appears. It is a second `delay` inside the existing screensaver timer job, so the `timer?.cancel()` already in `notifyInteraction` also cancels it. The only new state is a `sleepAllowed` flag, which resets in the same place.

Adds kotlinx-coroutines-test as a test dependency and a new InteractionTrackerViewModelTests. The tests use virtual time to cover Never, Immediately, the timeout running out, the screensaver staying visible afterwards, and an interaction taking the screen back.
Tested with ./gradlew detekt lint test assembleDebug. This has also been tested on device and works fine.

**Code assistance**

I wrote this with Claude Code Opus 5.0. It helped read the code, design the change, write the implementation and tests, and draft the commit messages. I reviewed all the changes.

**Related Issues**

Related to #4007 and #3230, both closed. Each reported that the device's sleep or display timeout never fires while the in-app screensaver is enabled.


nielsvanvelzen (commented):
The FLAG_KEEP_SCREEN_ON is required to avoid the regular system screensaver from showing. This is a known limitation of the in-app screensaver with no work around. We can not in any way override the screensaver and keep auto-sleep working at the same time. If you do want this behavior you should disable the in-app screensaver and change your system screensaver to use the Jellyfin one.

This change is a huge hack that will just cause weird behavior. In addition it violates our LLM policy. Closing as invalid.

MrGcGamer:
alright.. I was looking for LLM policy, but I couldn't find it

nielsvanvelzen:
[LLM/"AI" Development Policy](https://jellyfin.org/docs/general/contributing/llm-policies).

Issue #5398, opened 2026-02-07
Playback fails at runtime on some Android TV decoders (AAC channelCount=0), should fall back to transcode


nalet:
### Describe the bug

On some Android TV devices, playback can fail at runtime due to device-specific decoder/init issues even when the server playback plan considers the stream compatible for direct play/direct stream.

On my Sony BRAVIA 4K VH2 (Android 12), some AAC tracks cause Media3/ExoPlayer to try to initialize the platform AAC decoder with a `Format` that effectively reports `channelCount=0` (see `...[0, 48000]` in the `Format` dump). The decoder (`c2.android.aac.decoder`) then fails, and Jellyfin Android TV shows the toast: `Player error encountered. Will re-try...`.

### Environment

Device properties:
- `ro.product.manufacturer`: Sony
- `ro.product.brand`: Sony
- `ro.product.model`: BRAVIA 4K VH2
- `ro.product.device`: BRAVIA_VH2
- `ro.product.name`: BRAVIA_VH2_EU
- `ro.build.version.release`: 12
- `ro.build.version.sdk`: 31
- `ro.build.version.security_patch`: 2025-09-01
- `ro.build.fingerprint`: `Sony/BRAVIA_VH2_EU/BRAVIA_VH2:12/STT2.230505.001.S88/681561:user/release-keys`

Jellyfin Android TV:
- Built from upstream `master` (local build from source; versionName fallback: `0.0.0-dev.1`)

### Steps to reproduce

1. Use a media item that the server marks compatible for direct play/direct stream and contains AAC-LC audio (mime `audio/mp4a-latm`, codec `mp4a.40.2`).
2. Play it on the device above.
3. Observe the toast and a playback error.

### Logs

Minimal excerpt (logcat):

```text
CCodec: allocate(c2.android.aac.decoder)
MediaCodecAudioRenderer: Audio codec error
ExoPlaybackException: MediaCodecAudioRenderer error ... audio/mp4a-latm ... [0, 48000]
MediaCodecDecoderException: Decoder failed: c2.android.aac.decoder
```

### Expected behavior

If playback fails at runtime (decoder/init errors that the server cannot predict), the client should automatically re-request a more conservative server playback plan (e.g. force audio transcoding, then full transcode if needed), rather than giving up or requiring users to toggle settings.

MAnECiaC:
**Affected Device:** TCL 55Q7C (TCL Smart TV Pro)  
[...]

PR #5827, opened 2026-09-16
Recover playback instead of failing when a video decoder crashes mid-stream

HaraldNordgren:
This started as investigating a Jellyfin resume failure on a Philips Android TV (MediaTek chipset). After ruling out file corruption and ffmpeg bugs (a full frame-by-frame decode and remux with both stock ffmpeg and Jellyfin's bundled ffmpeg build handled the file perfectly), I connected to the TV over adb and captured logcat during a live reproduction. The actual fault: `OMX.MTK.VIDEO.DECODER.HEVC`, a MediaTek hardware decoder, throws a vendor-internal error after successfully initializing, partway through decoding. That's not fixable here: it's closed-source vendor firmware with no realistic OTA path.

`setEnableDecoderFallback` only retries with another decoder when the first fails to *initialize*, not when it starts fine and then crashes mid-stream, so `PlaybackController`'s existing retry just blindly retries the same decoder up to 3 times before giving up with an error toast.

I built a fix that excludes the failing decoder and recovers locally, and verified it on the same Philips/MediaTek TV: after the hardware decoder fails, playback now recovers within ~1s via the device's software HEVC decoder (`c2.android.hevc.decoder`), with no error toast and no restart.

**Changes**

This tracks decoders that fail at runtime and excludes them from future selection via a custom `MediaCodecSelector`, so the next attempt picks a different decoder, recovering locally (re-`prepare()` + seek back) instead of through the app's full error/restart path. Exclusions clear per queue item. Logic lives in a small `FailedVideoDecoderTracker`, with unit tests.

**Code assistance**

Developed with Claude Code (Sonnet 5, extra high reasoning):
- Investigated the failure and root-caused it via on-device adb/logcat capture during a live reproduction.
- Read Media3's source to verify the `MediaCodecSelector`/renderer-selection mechanics this fix relies on.
- Wrote the implementation and unit tests.
- Built the app locally, sideloaded a debug APK onto the same physical TV via adb, and confirmed the recovery behavior with real device logs, including catching and fixing an initial version that didn't actually engage (found by adding temporary diagnostic logging and re-testing).

I reviewed and directed the work throughout.

**Issues**

None filed for this specific case, related to the unresolved #5398.


HaraldNordgren:
Ping @nielsvanvelzen @thornbill, this unbroke video playback on my TV.

nielsvanvelzen (changes requested):
While the concept here is interesting I find it too risky to add this in our current playback code that is in maintenance mode. Dealing with bad decoders should also not be a responsibility of our app, this should be done upstream in media3. As long as the library tells us it can play something, it should do so.

HaraldNordgren:
@nielsvanvelzen That sounds reasonable (even if it's a bit sad because the upstream fix will probably take a long time). In the meantime I'll run my own debug version of Jellyfin when there are playback issues.

Created issue here: https://github.com/androidx/media/issues/3421

HaraldNordgren:
And media3 PR here: https://github.com/androidx/media/pull/3422

PR #5828, opened 2026-09-16
Add Seasons and Series navigation rows to season browse page

johnpc:
**Changes**

Adds two navigation rows to the season browse page (`GenericFolderFragment`) so users can move around a series without backing all the way out: a **Seasons** row listing the other seasons of the same series (excluding the current one), and a **Series** row linking back to the parent series detail page. Both rows only appear when browsing a `SEASON` and are automatically hidden when empty (e.g. the Seasons row does not show for a single-season show).

Demo. Left is before (current `master`), right is after this change:

Season navigation demo

**Code assistance**

Code assistance (Claude) was used to implement the change.

**Issues**

N/A


nielsvanvelzen (changes requested):
Please fill in the pull request template.

johnpc:
my apologies, updated it.

nielsvanvelzen (approved):

PR #5602, opened 2026-05-23
Add VIMU external subtitles support

dieselw0w:
**Changes**
Add external subtitles support for VIMU Player

**Code assistance**
No AI used

**Issues**
Fixes #3343

PR #5835, opened 2026-09-19
Fix external player intent extras and result handling

RealGreenDragon:
**Changes**

Follow-up on #5625. I went through the new `ExternalPlayerApi` implementations against the intent documentation of each player and found a few extras and result values that are no longer sent or read, including two that came from #5556 and #5602.

Fixes:

* **MX Player** – read the `end_by` result extra again, as added in #5556. The documentation states that the `position` and `duration` result extras *"will not exist if playback is completed"*, so `end_by` is the only way to tell a completed item from an interrupted one. Without it, an item watched to the end is reported with no position and the next item is not played. `end_by` is `"user"` or `"playback_completion"`, so this also gives a definite completed/not-completed value instead of falling back to the 90% heuristic.
* **MX Player** – add `com.mxtech.videoplayer.pro` to the package list. The documentation lists the Pro edition next to the free one. Without it, MX Pro resolves to `DefaultExternalPlayerApi`, which does not send `return_result`, so no result is returned at all and nothing is reported back to the server.
* **MX Player** – send `secure_uri` again, as it was before #5625. The video url contains the access token and this extra keeps it out of the player property dialog. Also send the documented `filename` extra, which helps the player match subtitles for remote media; it is already collected in `ExternalPlayData` but was unused.
* **MPV** – send the title as `title`. `media-title` is the name of the mpv property; the documented intent extra is `title`.
* **Vimu** – send `forceresume` as `false` again, as it was before #5625. The documentation does not state the default value when the extra is absent, so setting it explicitly keeps the position we pass in `startfrom` authoritative.
* Sort external subtitles with the default one first. `compareBy { it.isDefault }` sorts `false` before `true`, so the default subtitle ended up last. Vimu and VLC accept a single subtitle and take `first()`, so they received a non-default one whenever more than one external subtitle exists. This predates #5625 but it also affects the Vimu subtitle support added in #5602.

Additions:

* **MX Player** and **MPV** – select the default subtitles through the documented `subs.enable` extra. MPV documents it as a subset of `subs`, so it is only sent together with `subs`. Happy to drop this from the PR if you would rather keep it to the fixes.

Documentation used: [MX Player](https://mx.j2inter.com/api), [MPV](https://mpv-android.github.io/mpv-android/intent.html), [Vimu](https://www.vimu.tv/player-api), [VLC](https://wiki.videolan.org/Android_Player_Intents). VLC needs no change: it does not document a completion signal, so the 90% heuristic still applies there.

Tested on app `v0.20.0-beta.2` installed in `Fire TV Stick 4K Max 2nd gen (Fire OS 8)`, with MX Player (free edition) and MPV: watching an item to the end, leaving at about half, and resuming. For `subs.enable` I used a movie with two external subtitles, one flagged as default in the file name (`Film.it.default.srt`): with the current beta no subtitle is selected at start, with this change both players start with the default one. Vimu and MX Player Pro are paid apps and I could not test them, so those changes follow the documentation only.

**Code assistance**

An LLM was used to cross-check each implementation against the player API documentation and to review the syntax of the changes.

**Issues**

None open for these.


app/src/main/java/org/jellyfin/androidtv/ui/playback/external/VimuExternalPlayerApi.kt
```diff
@@ -30,6 +31,7 @@ class VimuExternalPlayerApi : ExternalPlayerApi {
 	override fun populateIntent(intent: Intent, data: ExternalPlayData) {
 		intent.putExtra(EXTRA_TITLE, data.title)
 		intent.putExtra(EXTRA_POSITION, data.position.inWholeMilliseconds.toInt())
+		intent.putExtra(EXTRA_RESUME, false)
```
nielsvanvelzen: This is already false by default, so not needed

app/src/main/java/org/jellyfin/androidtv/ui/playback/external/MxExternalPlayerApi.kt
```diff
@@ -14,15 +14,22 @@ class MxExternalPlayerApi : ExternalPlayerApi {
 	companion object {
 		val PACKAGE_NAMES = arrayOf(
 			"com.mxtech.videoplayer.ad",
+			"com.mxtech.videoplayer.pro",
```
nielsvanvelzen: This version of the app has been removed over 2 years ago. no need to add it here.

app/src/main/java/org/jellyfin/androidtv/ui/playback/external/MxExternalPlayerApi.kt
```diff
@@ -31,20 +38,35 @@ class MxExternalPlayerApi : ExternalPlayerApi {
...
+				intent.putExtra(EXTRA_SUBS_ENABLE, defaultSubtitles.map { it.url }.toTypedArray())
+			}
 		}
 	}
 
 	override fun parseResult(result: ActivityResult): ExternalPlayResult = when (result.resultCode) {
 		Activity.RESULT_OK -> {
-			// Try reading end position
-			val position = result.data?.getIntExtra(RESULT_EXTRA_POSITION, -1)?.takeIf { it >= 0 }?.toLong()?.milliseconds
+			val data = result.data
```
nielsvanvelzen: No need to assign this here

nielsvanvelzen (changes requested):

RealGreenDragon:
Thanks, I fixed all points, tell me if I need to squash

nielsvanvelzen (approved):

